PR #13, opened 2018-11-21
Open Model Zoo for OpenVINO 2018 R4

openvino-pushbot:


kchechil:
Ok, old models must be cleaned by a separate PR.

PR #23, opened 2018-12-20
Added face recognition demo

zhiltsov-max:
Added interactive face recognition demo and its documentation. Implemented in Python. Tested on Windows 10, Ubuntu 16.04, OpenVINO R4, Python 2.7, Python 3.5, Python 3.6, in Docker. Supports both camera and video file input.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 # [OpenVINO™ Toolkit](https://01.org/openvinotoolkit) - Open Model Zoo repository
-[![Stable release](https://img.shields.io/badge/version-2018.R4-green.svg)](https://github.com/opencv/open_model_zoo/releases/tag/2018_R4)
+[![Stable release](https://img.shields.io/badge/version-2018.R5-green.svg)](https://github.com/opencv/open_model_zoo/releases/tag/2018_R5)
 [![Apache License Version 2.0](https://img.shields.io/badge/license-Apache_2.0-green.svg)](LICENSE)
 
 This repository includes optimized deep learning models and a set of demos to expedite development of high-performance deep learning inference applications. Use these free pre-trained models instead of training your own models to speed-up the development and production deployment process.


demos/python_demos/face_recognition_demo.py
```diff
@@ -0,0 +1,1037 @@
...
+            face_identifier.start_async(image, rois, landmarks)
+            descriptor = face_identifier.get_descriptors()[0]
+            self.database.append(self.Identity(label, descriptor))
+
+    def match_faces(self, descriptors):
+        database = self
+        distances = np.empty((len(descriptors), len(database)))
+        for i, desc in enumerate(descriptors):
+            for j, identity in enumerate(database):
+                distances[i][j] = cosine(desc, identity.descriptor) * 0.5
```
sovrasov: Why do you need to multiply distance by  `0.5`?

demos/python_demos/face_recognition_demo.py
```diff
@@ -0,0 +1,1037 @@
...
+
+        # Find best assignments, prevent repeats (assuming faces can not repeat)
+        _, assignments = linear_sum_assignment(distances)
+        matches = []
+        for i in range(len(descriptors)):
+            id = assignments[i]
+            distance = distances[i, id]
```
sovrasov: When for some `i` distance is close to 1, the `i`th face is unlikely to be a person from the database.
IMHO here should be thresholding: something like if distance > 0.8 -> person is unknown.

demos/python_demos/face_recognition_demo.md
```diff
@@ -0,0 +1,179 @@
...
+results use images of square shapes. Image file name is used as a person name
+during the visualization.
+Use the following name convention: `person_N_name.png` or `person_N_name.jpg`.
+
+### Installation and dependencies
+
+The demo depends on:
+- OpenVINO library (R4)
+- Python (any of 2.7+ or 3.5+, which is supported by OpenVINO)
+- NumPy
```
grib0ed0v: @zhiltsov-max please specify particular versions of **NumPy**, **SciPy** and **OpenCV** and add them into `requirements.txt` file as well.

grib0ed0v:
@zhiltsov-max please add appropriate demo info into `open_model_zoo/blob/face-recognition-demo/demos/Readme.md` file.

demos/python_demos/face_recognition_demo.md
```diff
@@ -0,0 +1,179 @@
+# Interactive Face Recognition Demo
+
+This example demonstrates an approach to create interactive applications
+for video processing. It shows the basic architecture for building model
+pipelines supporting model placement on different devices and simultaneous
+parallel or sequential execution using OpenVINO library in Python.
+In particular, this demo uses 4 models to build a pipeline able to detect
+faces on videos, detect face direction and its keypoints (aka "landmarks"),
+and recognize persons using provided face database (gallery). The corresponding
+pretrained models can be bound in this repository:
```
grib0ed0v: I assume misprint here **found** instead of **bound**

demos/python_demos/face_recognition_demo.py
```diff
@@ -0,0 +1,1037 @@
...
+        super(FaceDetector, self).__init__(model)
+
+        assert len(model.inputs) == 1, "Expected 1 input blob"
+        assert len(model.outputs) == 1, "Expected 1 output blob"
+        self.input_blob = next(iter(model.inputs))
+        self.output_blob = next(iter(model.outputs))
+        self.input_shape = model.inputs[self.input_blob].shape
+        self.output_shape = model.outputs[self.output_blob].shape
+
+        assert np.array_equal([1, 3, 300, 300], self.input_shape), \
```
grib0ed0v: I'm not sure that here we should check input shapes against **hardcoded** values. Generally speaking, this sample can be executed with `intel_models/face-detection-adas-0001` model, which has **384x672** input resolution.

demos/python_demos/face_recognition_demo.py
```diff
@@ -0,0 +1,1037 @@
...
+        else:
+            raise Exception("Wrong face images database path. Expected path to" \
+                            "a directory containing '%s' files, got '%s'" % \
+                            (self.IMAGE_EXTENSIONS, path))
+
+        self.database = []
+        for path in paths:
+            label = osp.splitext(osp.basename(path))[0]
+            image = cv2.imread(path)
+
```
grib0ed0v: It's better to assume that as an input you have **_uncropped_**  image with face, so, first you need to run face detector, second  - landmarks regressor and then face identifier in order to obtain correct embedding for the face in gallery.

demos/python_demos/face_recognition_demo.py
```diff
@@ -0,0 +1,1037 @@
...
+            n, c, h, w = face_identifier.get_input_shape()
+            image = cv2.cvtColor(image, cv2.COLOR_BGR2RGB)
```
grib0ed0v: @zhiltsov-max I don't think that this conversion is needed for model inference

grib0ed0v:
@zhiltsov-max please wait until PR #24 will be merged and make the appropriate changes, there is no `face-reidentification-retail-0071` model anymore.

demos/python_demos/face_recognition_demo.py
zhiltsov-max: By default `cosine` outputs values in the range `[0; 2]`, so multiplication is to shrink the output to the more familiar range `[0; 1]`.

``` python
>>> from scipy.spatial.distance import cosine
>>> cosine([1, 1], [-1, -1])
2.0
>>> cosine([1, 1], [1, 1])
0.0
```

demos/python_demos/face_recognition_demo.py
grib0ed0v: I'm not saying that this should hold for other models. I'm saying that this is True for `face-detector` and it's better to add `intel_models/face-detection-adas-0001` as 1 more choice for face detector.

zhiltsov-max:
@grib0ed0v, should we remove the model numbers from the demo description? It may be hard to update all the docs in future when models change, but the model compatibility may be broken at some day.

grib0ed0v:
cc @snosov1 

@zhiltsov-max right now, the best option is to fix version of OpenVINO (R5) and models in this release with their numbers in names to avoid user confusion.

> @grib0ed0v, should we remove the model numbers from the demo description? It may be hard to update all the docs in future when models change, but the model compatibility may be broken at some day.

asmorkalov:
Jenkins please retry a build

snosov1:
@grib0ed0v Alexey, R5 content was merged. Is this PR ok to merge as well or something should be fixed?

grib0ed0v:
@snosov1 cool! After quick look I see that comment https://github.com/opencv/open_model_zoo/pull/23#issuecomment-449355038 is still unresolved.

grib0ed0v:
@zhiltsov-max I've tested it against our implementation of FR demo with the same models & gallery. It seems that implementation of `FaceIdentifier` or `FacesDatabase` contain bug which leads to instability at identity predictions.

asmorkalov:
Jenkins please retry a build

zhiltsov-max:
@grib0ed0v, updated.

demos/python_demos/face_recognition_demo/face_recognition_demo.py
```diff
@@ -0,0 +1,1071 @@
...
+            tuple(roi.position), tuple(roi.position + roi.size),
+            (0, 220, 0), 2)
+
+        # Draw identity label
+        text_scale = 0.5
+        font = cv2.FONT_HERSHEY_SIMPLEX
+        text_size = cv2.getTextSize("H1", font, text_scale, 1)
+        line_height = np.array([0, text_size[0][1]])
+        self.draw_text_with_background(frame,
+            '%s %.2f%%' % (label, 100.0 * (1 - identity.distance)),
```
grib0ed0v: In case of *Unknown* class, expression `100.0 * (1 - identity.distance)` no more represents confidence degree, so it's better not to show it to the user.

zhiltsov-max: The goal was to display model confidences to the user so that he could adjust the threshold (`-t_id`) used for person identification. Thus, this feature is mostly for debugging. May be it should be displayed only in `--verbose` mode?

grib0ed0v:
@asmorkalov what about CI status?

asmorkalov:
Jenkins please retry a build

grib0ed0v:
@snosov1 @zhiltsov-max accuracy issue is gone & CI is green now, so LGTM from my side.

asmorkalov:
@snosov1 @zhiltsov-max Unfortunatelly we still have connectivity issues on CI machines and working on it with IT. Please use the same magic phrase to restart pre-commit build if you suspect network issues.

snosov1:
There's a hell lot of presumably unrelated changes in this PR - license headers, dependencies and new targets in CMake, etc. Where do they come from? 

Maxim, maybe you forgot to rebase to the latest release?

asmorkalov:
Jenkins please retry a build

PR #33, opened 2019-02-12
Fix pointer error in face emotion detection demo

zimka:
It looks for me that there is a bug in outputIdxPos calculation (line 616).
Because emotionsValues is a linear array with [face0emotion0, face0emotion1, ... face2emotion0... faceNemotion4], shift between face0emotion0 and faceNemotion0 is N*5, not N.
Here is output before and after fix checked on some random [image](http://www.pmamed.com/wp-content/uploads/2016/10/COPD_04_17_2018.jpg) of three happy people.
Because of index shift they get wrong final emotions.
 
[face_emotions_after_fix.txt](https://github.com/opencv/open_model_zoo/files/2856588/face_emotions_after_fix.txt)
[face_emotions_before_fix.txt](https://github.com/opencv/open_model_zoo/files/2856589/face_emotions_before_fix.txt)

### DIFF
--- a/demos/interactive_face_detection_demo/detectors.cpp
+++ b/demos/interactive_face_detection_demo/detectors.cpp
@@ -613,7 +613,7 @@ std::string EmotionsDetection::operator[] (int idx) const {
     }
 
     auto emotionsValues = emotionsBlob->buffer().as<float *>();
-    auto outputIdxPos = emotionsValues + idx;
+    auto outputIdxPos = emotionsValues + idx * numOfChannels;
 
     /* identify an index of the most probable emotion in output array
            for idx image to return appropriate emotion name */


snosov1:
Hey, @zimka ! 

Thanks for the find! To me, it seems like you're on point!

@vmatveic, care to merge?

snosov1:
Jenkins please retry a build

vmatveic:
Hi,

We will check it.

Thanks!

demos/interactive_face_detection_demo/detectors.cpp
```diff
@@ -613,10 +613,15 @@ std::string EmotionsDetection::operator[] (int idx) const {
     }
 
     auto emotionsValues = emotionsBlob->buffer().as<float *>();
-    auto outputIdxPos = emotionsValues + idx;
+    auto outputIdxPos = emotionsValues + idx*numOfChannels;
```
vmatveic: Hi @zimka , 

Thanks a lot for the solution and your contribution.
Actually, we have already found it and fixed in the internal working tree. Moreover, this demo is completely redesigned in this internal release. So, next release will overwrite this part of source code and fix this problem.

But I can merge it anyway to have your solution working until our release is published.

Just one thing is missed. Unfortunately we don't have completely working CI for open source yet. And code style check was missed here. So, please put additional spaces around `*` symbol.

demos/interactive_face_detection_demo/detectors.cpp
```diff
@@ -613,10 +613,15 @@ std::string EmotionsDetection::operator[] (int idx) const {
...
 
     /* identify an index of the most probable emotion in output array
            for idx image to return appropriate emotion name */
+    std::string emotionsValuesRepr = "";
+    for (auto emValPtr=outputIdxPos; emValPtr<outputIdxPos + emotionsVecSize; emValPtr++){
+        emotionsValuesRepr +=  "(" + emotionsVec[emValPtr-outputIdxPos] + ":" + std::to_string(*emValPtr) + ");";
+    }
+    slog::info << "Face emotions:" << emotionsValuesRepr << slog::endl;
```
vmatveic: Also, I would ask to remove this source code because tracking of resulting emotions is already supported by FLAGS_r option enabled (it means raw output enabled). Please double check. If this flag is enabled all detection/recognition results are printed to the console.

demos/interactive_face_detection_demo/detectors.cpp
zimka: Hi @vmatveic, it is a very minor bug of course, but just in my opinion it still is better to fix it, than not, until next release is available.

demos/interactive_face_detection_demo/detectors.cpp
zimka: It looks for me that only the max score emotion is logged with -r flag. There are examples of demo execution in the files attached to the PR ([face_emotions_after_fix.txt](https://github.com/opencv/open_model_zoo/files/2856588/face_emotions_after_fix.txt)) with -r flag, I haven't noticed the same logged values. Am I missing something?
If these logs are still unnecessary for any reason, I can remove them;  I have added them just because I could not find this info in logs myself.
Please specify once again if it is better to remove them anyway, thanks.

demos/interactive_face_detection_demo/detectors.cpp
vmatveic: Hi,

Actually it's not so minor because impacts a user by wrong results ;) Sure, I will merge, please just update it with required spaces.

demos/interactive_face_detection_demo/detectors.cpp
vmatveic: It should be placed under _r flag at least because in case of long video input passed we will get overwritten console output. Moreover, what's the reason to track all of these values in real time? which benefit it gives for a user? I think the necessary value is only max prob emotion which we show currently. Other values are smaller and not so interesting for a user, isn't it? I still think it should be removed and you can add such debug information to own application any time when needed. But official version doesn't need it I think... There are many places where we could use detailed tracing but we avoid it to not overload samples/demos by such long debug info.

zimka: Sounds very reasonable, I have removed this part.

demos/interactive_face_detection_demo/detectors.cpp
zimka: Fixed the spaces, please check it.

vmatveic:
Hi @zimka ,
The solution is ok now. But currently CI is failed. Let us check it locally (just to repeat, our public CI process is not perfect yet). I will merge it nearest time! Thanks a lot!
@asmorkalov , could you please check it? I'm going to merge the solution bu current CI status is failed. Can I merge anyway? The solution is OK!

snosov1:
Jenkins please retry a build

snosov1:
Jenkins please retry a build

vmatveic:
@snosov1 , please merge this PR as discussed locally.
@zimka thanks a lot for contribution.

snosov1:
@vmatveic please, resolve all discussions and I'll hit the button

Issue #31, opened 2019-02-06
text-detection-0001 output shape is different from paper and description


banderlog:
Hi, I am a bit confused with the output shape of text-detection-0001.

As stated in [paper](https://arxiv.org/pdf/1801.01315.pdf):
> As shown in Fig. 3, the whole model has two separate headers, one for text/non-text prediction, and the other for link prediction. Softmax is used in both, so their outputs have 1\*2=2 and 8\*2=16 channels, respectively.

As stated in [model description](https://github.com/opencv/open_model_zoo/blob/2018/intel_models/text-detection-0001/description/text-detection-0001.md):
>The net outputs two blobs. Refer to PixelNet for details.
>    [1x2x192x320] - logits related to text/no-text classification for each pixel.
>    [1x8x192x320] - logits related to linkage between pixels and their neighbors.

Shapes of "logits related to linkage between pixels and their neighbors" from description and paper are different.

And model's outputs shape is `(1, 16, 192, 320)` in single object, which is a third output shape (not described in description or paper). What does it mean? Please help to understand.

snosov1:
@Ilya-Krylov can you comment?

Ilya-Krylov:
@banderlog Thank you for your finding. Actually this is a mistake. There should be 
> [1x2x192x320] - logits related to text/no-text classification for each pixel.
> [1x**16**x192x320] - logits related to linkage between pixels and their neighbors.

If so it corresponds to description in the paper. Output blob format is [BATCH_SIZE, CHANNELS_NUMBER, H, W].

banderlog:
Thank you for fast answer, I'm glad that I was able to help and that the second output's shape now the same in the all 3 sources. But I still have a problem: text-detection-0001 output is only *one* tensor.

Please, look on my example below:

```python
import cv2

img = cv2.imread('test.jpg')

td = cv2.dnn.readNet('./text-detection-0001.xml','./text-detection-0001.bin')
blob = cv2.dnn.blobFromImage(img, 1, (768, 1280))
td.setInput(blob)
a, b = td.forward()
>>> ValueError: not enough values to unpack (expected 2, got 1)
```
And if I'll check an output's shape:

```python
a = td.forward()
a.shape
>>> (1, 16, 192, 320)
```
As far as I understand, I still need logits related to text/no-text classification for each pixel with shape [1x2x192x320]. Could you comment it please, maybe you need some additional info, like `cv2.getBuildInformation()` output? 

[...]

PR #37, opened 2019-02-18
Fix original NN name and output tensor dimensions.

banderlog:
According to original paper, the model name is PixelLink, not PixelNet. And second tensor output dimensions are [1x16x192x320], according to original paper and #31


snosov1:
@Ilya-Krylov good to merge?

Ilya-Krylov:
LGTM

PR #39, opened 2019-02-21
Open Model Zoo documentation update

openvino-pushbot:
Models version: 82d08b07bfdeda432fdcd622d3eaf8b986d81921

Tools version: efc786978d44d438d1cbad74c55cb6288d8fe72f


kchechil:
@asmorkalov why do we still merge to 2018 branch?

kchechil:
Another comment is that introduction of text-detection-0002 should be synced with pushing to download.01.org - there is no such a model yet.

PR #51, opened 2019-03-19
model_downloader: use yaml.safe_load to prevent configs from executing arbitrary code

IRDonch:


IRDonch:
This will also test how well the `--num_attempts` option works, as I've just enabled it in our CI.

PR #59, opened 2019-03-21
U8 input for Human Pose Estimation Demo

dkurt:
With uint8 input and Myriad NCS2 device we can achieve better efficiency in case of USB 2.0:

FPS @ 456x256:

| USB | before | with PR |
|---|---|---|
|  2.0 | 3.78 | **4.02** (x1.06) |
|  3.0 | 4.71 | 4.7 |

### DIFF
--- a/demos/human_pose_estimation_demo/include/human_pose_estimator.hpp
+++ b/demos/human_pose_estimation_demo/include/human_pose_estimator.hpp
@@ -24,7 +24,7 @@ public:
     ~HumanPoseEstimator();
 
 private:
-    void preprocess(const cv::Mat& image, float* buffer) const;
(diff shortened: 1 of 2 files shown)


dkurt:
@Daniil-Osokin, May I ask you to review it?

kchechil:
@dkurt default recommended option for NCS2 is to use USB3.0, I see that there is no perf.regression for that, what was the number of iterations/nireq used for that test?

dkurt:
@kchechil, it's just an observation from the demo. It prints number of FPS on the screen. So I used it just to point that using U8 improves overall FPS if user has USB2.0

dkurt:
@eaidova, @kchechil, @Daniil-Osokin, Can merge it?

PR #65, opened 2019-03-27
[DEMO] Added detection logger and event-based metric script for Smart Classroom

evgeny-izutov:


evgeny-izutov:
@snosov1 @LeonidBeynenson @sovrasov @grib0ed0v

PR #69, opened 2019-04-01
add face-detection-retail-0044 to list of topologies

elenash:


elenash:
@snosov1 @atsyplen Please take a look

snosov1:
@atsyplen is it ok from your standpoint? Is it ok that the CI is red?

atsyplen:
@snosov1 It's ok .

PR #78, opened 2019-04-05
model_downloader: Tolerate cache access errors

IRDonch:
If we fail to read the cache, we can still download the needed file directly.
If we fail to update the cache, that doesn't affect the output of
the downloader, so we can ignore it.

These changes make the downloader more resilient to network connectivity issues
(in case when the cache is on a remote FS) and enable its usage with read-only
cache directories.


kchechil:
Jenkins please retry a build

kchechil:
Jenkins please retry a build

PR #80, opened 2019-04-09
[DEMO] Added parameters to control action smoothing in Smart Classroom demo

evgeny-izutov:


demos/smart_classroom_demo/main.cpp
```diff
@@ -607,7 +607,7 @@ int main(int argc, char* argv[]) {
         tracker_action_params.forget_delay = 150;
         tracker_action_params.affinity_thr = 0.9f;
         tracker_action_params.averaging_window_size_for_rects = 5;
-        tracker_action_params.averaging_window_size_for_labels = actions_type == TOP_K ? 5 : 1;
+        tracker_action_params.averaging_window_size_for_labels = actions_type == TOP_K ? FLAGS_ss_t : FLAGS_ss_d;
```
snosov1: Can't we use a single parameter? I.e. if it's set - it's set unconditionally here. By default - the "old" logic works.

snosov1: I see that. So, if nothing is set - I suggest to leave the current logic. But if it's set - just use the provided value (assuming that a person running the application understands what they're doing).

evgeny-izutov: Caught. Fixed.

PR #83, opened 2019-04-12
Publish 2019 r1

vmatveic:


Wovchena:
@vmatveic This pull request was merged without a respect to CI results. Now every pull request based on this source will be red too.

vmatveic:
@Wovchena , CI for this release is not completely adapted yet (it used 2018 R5 InferenceEngine to build). The same time this content was verified in our internal GitLab. So, to fix the problem we need to adapt CI here to use latest Inference Engine from 2019R1. Let's discuss in details internally. If it's really stopping you now then let's revert this commit.
adding of @asmorkalov and @moslex to the thread.

PR #84, opened 2019-04-15
 model_downloader: make postprocessing more generic 

IRDonch:
Instead of hard-coding specific types of patches, allow defining per-file
regex replacements. This decouples the downloader from specifics of data
formats, making it more flexible.

Also, add some improvements to the postprocessing process:

* check that the file we're supposed to post-process exists and is located
  inside the output directory;

* check that at least one replacement actually took place (because if not,
  that indicates an error in the config);

* back up the original file, so that the user can examine the difference.

----

This is the first half of a series of changes that decouple the downloader from framework specifics. The second half will make downloading more generic, as well.

### DIFF
--- a/model_downloader/downloader.py
+++ b/model_downloader/downloader.py
@@ -253,31 +253,36 @@ def get_confirm_token(response):
             return value
     return None
 
-def delete_param(model):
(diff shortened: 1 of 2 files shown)


IRDonch:
CC @kchechil

PR #90, opened 2019-04-18
model_downloader: make the specification of files to download more generic

IRDonch:
This frees us from having to contort the meanings of "model" and "weights" to support various frameworks and makes it possible to support more than two files per model.

This change also makes the downloader keep the downloaded archive and all of its contents for models downloaded as archives. This is possible now, because the config now contains hashes for archives rather than their contents, and therefore the archives can be retrieved from a cache (when using one).

----

This is the second half of the series of changes started in #84. It completely decouples the downloader from framework specifics (in fact, the `framework` field in the config is now unused).

### DIFF
--- a/model_downloader/downloader.py
+++ b/model_downloader/downloader.py
@@ -17,7 +17,6 @@
 """
 
 import argparse
-import collections
 import fnmatch
 import hashlib
 import re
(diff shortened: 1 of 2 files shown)


model_downloader/downloader.py
```diff
@@ -33,16 +33,8 @@
 
 from pathlib import Path
 
-Framework = collections.namedtuple('Framework', ['model_extension', 'weights_extension'])
 MemberRequest = collections.namedtuple('MemberRequest', ['path', 'destination', 'expected_hash'])
```
zmaslova: not used

zmaslova: also `import collections`

IRDonch: Good catch. Fixed.

IRDonch:
Jenkins please retry a build

zmaslova:
Jenkins please retry a build

PR #96, opened 2019-04-22
 model_downloader: make model_optimizer_args more machine-readable

IRDonch:
* Make it a list rather than a string. This makes it possible to use the arguments regardless of what the system shell is. It's also a lot more human-readable.

* Replace the angle bracket placeholder syntax with shell-like variable substitutions. This will enable using Python's `string.Template` to substitute specific paths.

-----

This is a prelude to adding a tool that will automatically run Model Optimizer for each model using these arguments.


zmaslova:
Looks good

PR #123, opened 2019-05-20
added models for R2'19

atsyplen:


kchechil:
@atsyplen please change the header to "for R2'19"
cc @jkamelin 
why 3 flavors of densenet and 1 facenet are not in this PR?

kchechil:
also, please make sure that license file is properly updated.

PR #126, opened 2019-05-24
added Accuracy Checker

eaidova:


tools/accuracy-checker/.gitattibutes
```diff
@@ -0,0 +1,9 @@
+data/annotations/* filter=lfs diff=lfs merge=lfs -text
+data/test_models/*.caffemodel filter=lfs diff=lfs merge=lfs -text
+data/test_models/*.params filter=lfs diff=lfs merge=lfs -text
+data/test_models/*.bin filter=lfs diff=lfs merge=lfs -text
+data/test_models/*.pb filter=lfs diff=lfs merge=lfs -text
+data/test_data/*.jpg filter=lfs diff=lfs merge=lfs -text
+*.caffemodel filter=lfs diff=lfs merge=lfs -text
+*.params filter=lfs diff=lfs merge=lfs -text
+*.pb filter=lfs diff=lfs merge=lfs -text
```
IRDonch: Have you actually uploaded these files via LFS? It [doesn't look like it](https://github.com/opencv/open_model_zoo/blob/15033f66f38e68a72f87770b9bcf78bcb6abb05e/tools/accuracy-checker/data/test_models/samplenet.pb).

Frankly, I think we should reconsider using LFS for all these files to begin with. They're under 2 MB in total, and they change pretty infrequently, so maybe we should just commit them directly to Git.

eaidova: So..., what should I do? Reupload using LFS or without it?
Yes, you are right, the total size of all files ~ 1MB and I do not plan to change existing set of files (except probably IR, when its version will be deprecated), but it can be only extended by ONNX and PyTorch realization of a model for unit testing.

alalek: GitHub users have no access to discussed LFS storage.
So,
- "LFS"-related lines should be removed from the PR.
- any LFS-ed files should be added into .git repository directly.

eaidova: Fixed

IRDonch: > GitHub users have no access to discussed LFS storage.

GitHub has an LFS store of its own.

That said, my vote is for no LFS as well, since I don't think it's worth it to introduce the complexity of LFS for such (relatively) small files.

IRDonch:
Jenkins: ok to test

PR #132, opened 2019-05-30
model_downloader: merge different precision variants of the same model

IRDonch:
This eliminates a lot of information duplication within the config file,
as the different variants are the same in everything but the files.

### DIFF
--- a/model_downloader/common.py
+++ b/model_downloader/common.py
@@ -124,10 +124,7 @@ class TopologyFile:
 
     @classmethod
     def deserialize(cls, file):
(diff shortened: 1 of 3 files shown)


IRDonch:
CC @kchechil @eizamaliev

I was going to combine this change with a more thorough refactoring of the config file, but then I realized that I don't have to, so I submitted it early.

IRDonch:
Jenkins please retry a build

kchechil:
@zmaslova @snosov1 I suspect there are no comments so merging this.

PR #134, opened 2019-05-31
Remove "dilation" from the model downloader config file

IRDonch:
The server hosting the .caffemodel file for it appears to be defunct, so we can no longer support it.


IRDonch:
CC @yssemaev

yssemaev:
got it

PR #144, opened 2019-06-07
Rename "accuracy-checker" to "accuracy_checker"

IRDonch:
Every other directory (except for the model directories) uses underscores to separate words, so this one should too.


IRDonch:
CC @eaidova

PR #149, opened 2019-06-13
Add a model converter option to override the output directory

IRDonch:
Also, make a couple of related changes:

* Rename `--download_root` into `--download_dir` for consistency with the other options. This script hasn't been a part of a release yet, so we don't need to retain the old name.

* Place converted models into a subdirectory corresponding to the precision (always "FP32" at the moment). This makes the directory structure more consistent between models that are converted and ones that are initially in the IR format, and will enable us to support alternate conversions in the future.

### DIFF
--- a/tools/downloader/README.md
+++ b/tools/downloader/README.md
@@ -81,15 +81,21 @@ The basic usage is to run the script like this:
 ```
 
 This will convert all models into the Inference Engine IR format. Models that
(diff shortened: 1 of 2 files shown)


atsyplen:
Jenkins please retry a build

atsyplen:
LGTM

PR #159, opened 2019-06-17
Clean up the dotfiles in the root of the repository

IRDonch:
They seem to be copies of the ones in the dldt repository, and, for the most part, don't apply to this project.


.gitignore
```diff
@@ -1,343 +1 @@
...
-
-# from Model Optimizer repo
-.idea
-.project
-.cproject
-.pydevproject
-.settings
-/bin/
-/gen/
-__pycache__
```
alalek: It makes sense to keep `__pycache__` for Python 3

IRDonch: `*.pyc` is sufficient. `__pycache__` directories contain only `.pyc` files, and Git doesn't notice directories that only contain ignored files.

PR #160, opened 2019-06-17
update demos from dldt

Wovchena:


Wovchena:
@IRDonch it is required to set up newer dldt on Jenkins for develop branch

PR #164, opened 2019-06-18
fix mo args for mtcnn models

zmaslova:


kchechil:
Jenkins please retry a build

PR #168, opened 2019-06-18
Accuracy Checker: added base config set

eaidova:


eaidova:
> There are some models here that aren't actually defined in the model downloader config, but I assume that those'll be added later.

Yes, it is replacement for EOL in this release. 
I add only configs range delivered with previous relase with EOL replacement. In this release should be added several new models config for which we also can share, but I think to add their configs is unnecessary or will be added later (the main goal is to align content with previous release).

PR #169, opened 2019-06-19
Accuracy Checker fixed namespace compatibility

eaidova:


eaidova:
@IRDonch, please merge. It brokes calibration precommits.

IRDonch:
Could you explain what this does? It looks a bit strange.

eaidova:
calibration uses ac from namespace openvino.tools.accuracy_checker
annotations were generated using accuracy checker from just accuracy_checker namespace so simple check types gives always False in comparision classes from different namespaces. The easiest way to fix it is compare by class name (or regenerate annotations for avoiding this problem)

I think it, of cause, can have more beautiful solution or even to align to use only one namespace, but it just hot fix.

PR #170, opened 2019-06-19
Add a requirements file for Accuracy Checker CI

IRDonch:
I used pip-compile to create it, although I manually removed `opencv-python` from the resulting file, since I would prefer to use the OpenCV bindings included in OpenVINO.


IRDonch:
CC @eaidova

Test build using this branch: http://134.191.240.124/job/omz/job/precommit/job/accuracy_checker_ubuntu16/171/console

PR #173, opened 2019-06-20
security_barrier_camera_demo: fix static analysis issues

Wovchena:


demos/security_barrier_camera_demo/grid_mat.hpp
```diff
@@ -22,6 +22,9 @@ class GridMat {
             maxWidth = std::max(maxWidth, static_cast<size_t>(sizes[i].width));
             maxHeight = std::max(maxHeight, static_cast<size_t>(sizes[i].height));
         }
+        if (0 == maxWidth || 0 == maxHeight) {
```
IRDonch: I'm not sure this is the correct check. Shouldn't we make sure that _none_ of the input sizes are zero?

demos/security_barrier_camera_demo/main.cpp
```diff
@@ -336,8 +336,11 @@ void Drawer::process() {
         out << "Mean overall time per all inputs: " << std::fixed << std::setprecision(2) << std::setw(6);
         const auto t1 = std::chrono::steady_clock::now();
         uint64_t frameCounter = context.frameCounter;
-        const ms meanOverallTimePerAllInputs = std::chrono::duration_cast<ms>((t1 - context.t0)
-                                               * context.readersContext.inputChannels.size()) / frameCounter;
+        ms meanOverallTimePerAllInputs = std::chrono::duration_cast<ms>((t1 - context.t0)
+                                               * context.readersContext.inputChannels.size());
+        if (0 != frameCounter) {
```
IRDonch: This seems strange. Is it actually possible for this code to be executed when `frameCounter` is 0?

Wovchena: starting and immediately stopping can result in zero frame processing. For example if VideoCapture will claim there are no frames in a file.

demos/security_barrier_camera_demo/grid_mat.hpp
Wovchena: Having one of inputs with zero resolution can be strange, but it still can be processed by GridMat. While all zero inputs will result in division by zero

demos/security_barrier_camera_demo/main.cpp
Wovchena: Well, I am not sure I can reproduce this case, because providing empty file results in "Cannot open" error, but I am still not sure that it is not impossible.

IRDonch: Well, OK. But isn't it misleading to print the time per frame (as well as FPS) when there haven't been any frames? Maybe a better fix would be to place the whole block from the declaration of `meanOverallTimePerAllInputs` to the FPS output into the `if`.

Wovchena: You are right. Fixed

PR #179, opened 2019-06-24
demos: bug fixes

Wovchena:


demos/interactive_face_detection_demo/main.cpp
```diff
@@ -169,9 +169,6 @@ int main(int argc, char *argv[]) {
         // --------------------------- 3. Doing inference -----------------------------------------------------
         // Starting inference & calculating performance
         slog::info << "Start inference " << slog::endl;
-        if (!FLAGS_no_show) {
-            std::cout << "Press any key to stop" << std::endl;
-        }
```
IRDonch: Could you explain this? It's not mentioned in your commit messages.

Wovchena: There is a similar message a few lines below.
```
        std::cout << "To close the application, press 'CTRL+C' here";
        if (!FLAGS_no_show) {
            std::cout << " or switch to the output window and press any key";
        }
```

PR #181, opened 2019-06-24
demos: align build scripts with dldt

Wovchena:


kchechil:
@IRDonch This is a fix for Windows demo build that you mentioned during stand-up. Could you please update Windows CI so that we check demo builds? Also please review and I'll merge.

demos/build_demos.sh
```diff
@@ -27,10 +27,25 @@ trap 'error ${LINENO}' ERR
...
 
-if [[ -z "${InferenceEngine_DIR}" ]]; then
-    printf "\nInferenceEngine_DIR environment variable is not set. Please set it to build the demo applications. \n"
+printf "\nSetting environment variables for building demos...\n"
+
+if [ -z "$INTEL_OPENVINO_DIR" ]; then
+    if [ -e "$DEMOS_PATH/../../bin/setupvars.sh" ]; then
+        setvars_path="$DEMOS_PATH/../../bin/setupvars.sh"
+    elif [ -e "$DEMOS_PATH/../../../bin/setupvars.sh" ]; then
+        setvars_path="$DEMOS_PATH/../../../bin/setupvars.sh"
```
IRDonch: Why do you try two paths here? Aren't the demos always going to be installed in the same directory (relative to the install root)?

Wovchena: As I understand, the reason is the following: there is a sim link and you can get two different paths: `openvino/deplyment_toolkit/inference_engine` and `openvino/inference_engine` pointing to the same folder

IRDonch: OK, good point.

PR #186, opened 2019-06-25
[Accuracy Checker] remove convert_annotation.py and mentions about it

eaidova:


tools/accuracy_checker/setup.py
```diff
@@ -83,7 +83,11 @@ def find_version(*path):
     version=version,
     long_description=long_description,
     packages=find_packages(),
-    entry_points={"console_scripts": ["accuracy_check=accuracy_checker.main:main", "convert_annotation=accuracy_checker.annotation_converters.convert:main"]},
+    entry_points={
+        "console_scripts": [
+            "accuracy_check=accuracy_checker.main:main",
+            "convert_annotation=accuracy_checker.annotation_converters.convert:main"
```
IRDonch: I would suggest a trailing comma here.

eaidova: Added

PR #188, opened 2019-06-25
[Accuracy Checker]: fix namespace compatibility bug

eaidova:
the problem is the same as in https://github.com/opencv/open_model_zoo/pull/169, but this condition checked very rarely in usage dldt tools with pregenerated annotation pickles (happed only on several datasets from private part)

### DIFF
--- a/tools/accuracy_checker/accuracy_checker/metrics/metric.py
+++ b/tools/accuracy_checker/accuracy_checker/metrics/metric.py
@@ -20,7 +20,8 @@ from ..utils import is_single_metric_source, get_supported_representations
 from ..presenters import BasePresenter
 from ..config import ConfigValidator, NumberField, StringField
 from ..dependency import ClassProvider
-from ..utils import zipped_transform, get_parameter_value_from_config
+from ..utils import zipped_transform, get_parameter_value_from_config, contains_any
+
 
 class Metric(ClassProvider):
     """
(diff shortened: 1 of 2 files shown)


IRDonch:
Jenkins: ok to test

tools/accuracy_checker/accuracy_checker/metrics/metric.py
```diff
@@ -129,8 +130,15 @@ def _update_state(self, fn, state_key, default_factory=None):
...
-                return representation
+            def is_container(representation):
+                if isinstance(representation, ContainerRepresentation):
+                    return True
+                representation_parents = type(representation).__bases__
+                representation_parents_names = [parent.__name__ for parent in representation_parents]
+
+                return contains_any(representation_parents_names, (ContainerRepresentation.__name__, ))
+
+            if not is_container(representation):
```
IRDonch: Incomplete statement here.

eaidova: Fixed, thanks

tools/accuracy_checker/accuracy_checker/metrics/metric.py
```diff
@@ -129,7 +130,15 @@ def _update_state(self, fn, state_key, default_factory=None):
 
     def _resolve_representation_containers(self, annotation, prediction):
         def get_resolve_subject(representation, source=None):
-            if not isinstance(representation, ContainerRepresentation):
+            def is_container(representation):
+                if isinstance(representation, ContainerRepresentation):
+                    return True
+                representation_parents = type(representation).__bases__
```
IRDonch: I would suggest using `__mro__` instead of `__bases__`, because:

1. It includes indirect base classes, which better mirrors the original `isinstance` check.
2. It includes the class itself, which will make the preceding `if` unnecessary.

eaidova: Completely agree, thanks

IRDonch: You didn't remove the redundant `if`, though.

eaidova: fixed

PR #189, opened 2019-06-26
update parameters

zmaslova:
But we don't have in our local repo instance-segmentation-security-0049 and instance-segmentation-security-0033. Documentation should be updated


kchechil:
@zmaslova as far as i know, these models should be replaced. cc @yssemaev

PR #196, opened 2019-07-01
Accuracy Checker: fix getting input data shape for scalar data

eaidova:
usually, input data for inference in accuracy checker is ndarray. But there is model NCF where input is one scalar value. This fix handles this issue.

P.S. priority is very small, the model is not in OMZ and even has not been added to public validation yet, due to another issue related to dataset conversion.

### DIFF
--- a/tools/accuracy_checker/accuracy_checker/utils.py
+++ b/tools/accuracy_checker/accuracy_checker/utils.py
@@ -379,7 +379,8 @@ def set_image_metadata(annotation, images):
     if not isinstance(data, list):
         data = [data]
     for image in data:
-        image_sizes.append(image.shape)
+        data_shape = np.shape(image) if not np.isscalar(image) else 1
+        image_sizes.append(data_shape)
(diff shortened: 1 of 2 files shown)


vladimir-dudnik:
Jenkins please retry a build

vladimir-dudnik:
@eaidova there are some fails on tests

eaidova:
@vladimir-dudnik, @IRDonch  fails happend due to update pytest version.
My tests were written for pytest 4.0.0 and I need time to update them for new mock objects syntaxes.
As fast work around I can suggest to fix pytest version in setup.py

vladimir-dudnik:
Jenkins please retry a build

IRDonch:
> As fast work around I can suggest to fix pytest version in setup.py

Fine by me. In fact, I think it's a good idea to pin the test dependencies, to ensure reproducible results.

IRDonch:
Jenkins: ok to test

PR #197, opened 2019-07-01
demos: remove unused headers from common

Wovchena:


Wovchena:
@IRDonch we need newer dldt to pass builds :)

Wovchena:
@IRDonch we need newer dldt to pass builds :)

IRDonch:
If these headers change from release to release, maybe we shouldn't be including them in the first place. How about including `<inference_engine.h>` instead? That ought to work on all versions.

Wovchena:
I am agree. Moreover, it appears that common does not need most of its IE includes. Updated

PR #209, opened 2019-07-03
accuracy checker: made workaround flag switched

eaidova:
in accuracy valiation enabled a workaround for models which can not be reshaped in runtime, until bug in dldt not fixed.
Calibration tool can not get per layer statistic if model reloaded without output specification.
This PR should help calibration turn off network reloading in reshaping (they do not use such validation approach for networks)


eaidova:
@vladimir-dudnik, @IRDonch could you please take a look

IRDonch:
Jenkins: ok to test

PR #218, opened 2019-07-08
demos: fix build_demos.sh

Wovchena:


IRDonch:
What does this fix? It looks to me like this shouldn't make any difference.

Wovchena:
It does not make any difference in appearance, but it makes use of variable that seems to be introduced just for that

IRDonch:
It it doesn't actually fix anything, I'm going to hold it until the next release. We're in code freeze, after all.

vmatveic:
> It it doesn't actually fix anything, I'm going to hold it until the next release. We're in code freeze, after all.

@IRDonch @Wovchena 

I think we may include this fix to this Release:

1. There is no any risk to include it (no functional updates which may affect users negatively)
2. It improves usability of this script and decreases any risks if, for example, a customer will take this script and modify output directory store in `build_dir` var himself. But in this case without this solution merged he will get wrong output message with old path (because of hard-coded `omz_demos_build` directory in the log).

So, I don't see any real objections to include this fix to the release.

Thanks.

IRDonch:
> It improves usability of this script

How?

> and decreases any risks if, for example, a customer will take this script and modify output directory store in build_dir var himself.

This seems like a rather far-fetched scenario. These files are installed read-only. And even if it does happen, the consequences are purely cosmetic.

> So, I don't see any real objections to include this fix to the release.

That's not how this works. :-) There must be a reason to include it, not a lack of objections.

Wovchena:
I am OK with holding.

vmatveic:
Ok, I agree. Let's wait.

PR #223, opened 2019-07-10
Add link to Open Visual Cloud

snosov1:


vladimir-dudnik:
Jenkins please retry a build

IRDonch:
@snosov1 What's the relation of this Open Visual Cloud to OMZ?

snosov1:
Well, I guess, it's part of the same ecosystem. I don't have a better answer.

The authors were asking how they can fit or contribute to the OMZ. My proposal was to simply mention them in the Readme in an "external resources" section. There's also a couple of other links that aren't tightly related, but I think are worthwhile having in such a section. Namely,

- https://github.com/incluit/OpenVino-For-SmartCity
- https://github.com/incluit/OpenVino-Driver-Behaviour

IRDonch:
OK, I looked at all these links and I would say that their relation to OMZ is that they offer usage examples of OMZ models (beyond what we already do with our demos). Perhaps the section title should reflect that?

snosov1:
I've added the links and modified the title.

PR #236, opened 2019-07-16
Accuracy Checker: fix cpu_extension search for Mac OS

eaidova:
@IRDonch, please take a look. 
P.S. Internal validation pipeline passed.

### DIFF
--- a/tools/accuracy_checker/accuracy_checker/launcher/dlsdk_launcher.py
+++ b/tools/accuracy_checker/accuracy_checker/launcher/dlsdk_launcher.py
@@ -343,29 +343,20 @@ class DLSDKLauncher(Launcher):
                 'for {} platform. Please, set cpu extension library manually.'.format(system_name)
             )
 
-        def get_cpu_extensions_list(file_format, base_name, selection_mode='',):
-            if not selection_mode:
-                default_cpu_extension = file_format.format(base_name)
-                extension_list = list(extensions_path.glob(default_cpu_extension))
+        if not selection_mode:
+            default_cpu_extension = file_format.format('cpu_extension')
+            extension_list = list(extensions_path.glob(default_cpu_extension))
 
-                if extension_list:
-                    return extension_list
+            if extension_list:
+                return extension_list[0]
 
-                cpu_info_flags = get_cpu_info()['flags']
-                supported_flags = ['avx512', 'avx2', 'sse4']
-                for flag in supported_flags:
-                    if flag in cpu_info_flags:
-                        selection_mode = flag
-                        break
-
-            extension_list = list(extensions_path.glob(file_format.format('{}_{}'.format(base_name, selection_mode))))
-
-            return extension_list
-
-        extension_list = get_cpu_extensions_list(file_format, 'cpu_extension', selection_mode)
-
-        if not extension_list and system_name == 'Darwin':
-            extension_list = get_cpu_extensions_list(file_format, 'cpu_extensiond', selection_mode)
+            cpu_info_flags = get_cpu_info()['flags']
+            supported_flags = ['avx512', 'avx2', 'sse4']
+            for flag in supported_flags:
+                if flag in cpu_info_flags:
+                    selection_mode = flag
+                    break
+        extension_list = list(extensions_path.glob(file_format.format('cpu_extension_{}'.format(selection_mode))))
 
         if not extension_list:
             raise ConfigError('suitable CPU extension lib not found in {}'.format(extensions_path))
--- a/tools/accuracy_checker/accuracy_checker/launcher/dlsdk_launcher.py
+++ b/tools/accuracy_checker/accuracy_checker/launcher/dlsdk_launcher.py
@@ -331,37 +331,48 @@ class DLSDKLauncher(Launcher):
             return cpu_extensions
         extensions_path = cpu_extensions.parent
         os_specific_formats = {
+            'Darwin': 'lib{}.dylib',
+            'Linux': 'lib{}.so',
             'Windows': '{}.dll',
-            'Darwin': 'lib{}d.dylib',
-            'Linux': 'lib{}.so'
         }
         system_name = platform.system()
         file_format = os_specific_formats.get(system_name)
         if not file_format:
             raise ConfigError(
-                'Accuracy Checker can not automatically found cpu extensions library '
+                'Accuracy Checker can not automatically find cpu extensions library '
                 'for {} platform. Please, set cpu extension library manually.'.format(system_name)
             )
 
-        if not selection_mode:
-            default_cpu_extension = file_format.format('cpu_extension')
-            extension_list = list(extensions_path.glob(default_cpu_extension))
+        def get_cpu_extensions_list(file_format, base_name, selection_mode='',):
+            if not selection_mode:
+                default_cpu_extension = file_format.format(base_name)
+                extension_list = list(extensions_path.glob(default_cpu_extension))
 
-            if extension_list:
-                return extension_list[0]
+                if extension_list:
+                    return extension_list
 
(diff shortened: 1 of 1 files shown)


tools/accuracy_checker/accuracy_checker/launcher/dlsdk_launcher.py
```diff
@@ -330,7 +330,19 @@ def get_cpu_extension(cpu_extensions, selection_mode):
         if cpu_extensions_name != 'AUTO':
             return cpu_extensions
         extensions_path = cpu_extensions.parent
-        file_format = '{}.dll' if platform.system() == 'Windows' else 'lib{}.so'
+        os_specific_formats = {
+            'Windows': '{}.dll',
+            'Darwin': 'lib{}d.dylib',
```
IRDonch: This should be `lib{}.dylib`. The `d` version is the debug build, I believe.

Also, I would suggest to list these in alphabetical order and to add a trailing comma.

tools/accuracy_checker/accuracy_checker/launcher/dlsdk_launcher.py
```diff
@@ -330,7 +330,19 @@ def get_cpu_extension(cpu_extensions, selection_mode):
...
+            'Linux': 'lib{}.so'
+        }
+        system_name = platform.system()
+        file_format = os_specific_formats.get(system_name)
+        if not file_format:
+            raise ConfigError(
+                'Accuracy Checker can not automatically found cpu extensions library '
```
IRDonch: find*

eaidova: fixed

tools/accuracy_checker/accuracy_checker/launcher/dlsdk_launcher.py
eaidova: > The d version is the debug build, I believe.
Most likely. I also add additional search loop if a release lib is not found.

IRDonch: > Most likely. I also add additional search loop if a release lib is not found.

Why would it not be found?

eaidova: Possibly, if a user builds IE from sources in debug mode, he can not have release lib (but I do not sure about it...)

IRDonch: Fair enough. However, macOS support should go into R2, while I would prefer to postpone the debug mode support until R3, to minimize the potential for regressions. So let's stick with just enabling macOS in this PR, OK?

IRDonch: (FWIW, I don't think searching for either version of the library makes sense anyway. The extension library should match the main IE library.)

eaidova: Ok, I remove debug.

tools/accuracy_checker/accuracy_checker/launcher/dlsdk_launcher.py
```diff
@@ -330,27 +330,50 @@ def get_cpu_extension(cpu_extensions, selection_mode):
...
-
-            cpu_info_flags = get_cpu_info()['flags']
-            supported_flags = ['avx512', 'avx2', 'sse4']
-            for flag in supported_flags:
-                selection_mode = flag
-                if selection_mode in cpu_info_flags:
-                    break
-        extension_list = list(extensions_path.glob(file_format.format('cpu_extension_{}'.format(selection_mode))))
+        os_specific_formats = {
+            'Darwin': 'lib{}.dylib',
```
alalek: BTW, macOS loadable modules (plugins) [should have `.so` extension](https://gitlab.kitware.com/cmake/cmake/blob/v3.14.1/Modules/Platform/Darwin.cmake#L34-38) (internal bug has been filled for DLDT), so it make sense to add support for multiple variants.

eaidova: What should have higher priority `.so` or `dylib`?

eaidova: ok, I add multiple variants

IRDonch: I don't think it matters. There shouldn't ever be both.

tools/accuracy_checker/accuracy_checker/launcher/dlsdk_launcher.py
```diff
@@ -326,31 +326,59 @@ def _prepare_bitstream_firmware(self, config):
...
+                default_cpu_extension = file_format.format(base_name)
+                extension_list = list(extensions_path.glob(default_cpu_extension))
+
+                if extension_list:
+                    return extension_list
+
+                cpu_info_flags = get_cpu_info()['flags']
+                supported_flags = ['avx512', 'avx2', 'sse4']
+                for flag in supported_flags:
+                    if flag in cpu_info_flags:
```
IRDonch: Hold on. `py-cpuinfo` doesn't actually support a flag named `sse4`. It has `sse4_1` and `sse4_2`. So this change is going to break extension library selection on CPUs that don't support AVX2.

There could probably be a better fix for this, but for now, let's just revert to the original logic here. I don't want any regressions.

tools/accuracy_checker/accuracy_checker/launcher/dlsdk_launcher.py
```diff
@@ -326,31 +326,59 @@ def _prepare_bitstream_firmware(self, config):
...
+        if not file_formats:
+            raise ConfigError(
+                'Accuracy Checker can not automatically find cpu extensions library '
+                'for {} platform. Please, set cpu extension library manually.'.format(system_name)
+            )
 
-            if extension_list:
-                return extension_list[0]
+        if not isinstance(file_formats, (list, tuple)):
+            file_formats = [file_formats]
```
IRDonch: This feels like a workaround for a self-inflicted problem. Just make every value in `os_specific_formats` a tuple.

tools/accuracy_checker/accuracy_checker/launcher/dlsdk_launcher.py
```diff
@@ -326,31 +326,59 @@ def _prepare_bitstream_firmware(self, config):
 
     @staticmethod
     def get_cpu_extension(cpu_extensions, selection_mode):
+        def get_cpu_extensions_list(file_format, base_name, selection_mode='',):
```
IRDonch: The default value is never used here, so it's a bit pointless.

tools/accuracy_checker/accuracy_checker/launcher/dlsdk_launcher.py
eaidova: fixed

tools/accuracy_checker/accuracy_checker/launcher/dlsdk_launcher.py
eaidova: fixed

tools/accuracy_checker/accuracy_checker/launcher/dlsdk_launcher.py
eaidova: I add a mapping from cpu_info flags to library suffixes, however, I do not know, has it ever worked on platforms without sse4_1 & sse4_2

eaidova: and also return previously used logic too (for sse4 as default suffix)

PR #237, opened 2019-07-16
accuracy checker fix postprocessing for instance segmentation

eaidova:
previous variant does not assume not aspect ratio resize and it gives wrong results for instance-segmentation-security-0050


kchechil:
cc @IRDonch - looks like it is required for R2.

vladimir-dudnik:
looks good to me

PR #241, opened 2019-07-17
demos: print HDDL as available

Wovchena:
Currently Core::GetAvailableDevices() will not print HDDL as available in any case. To avoid regression in the list of demos' supported devices (shown in help messages) we will just print HDDL as available as it was in R1.

### DIFF
--- a/demos/common/samples/common.hpp
+++ b/demos/common/samples/common.hpp
@@ -1114,14 +1114,11 @@ inline std::size_t getTensorBatch(const InferenceEngine::TensorDesc& desc) {
 inline void showAvailableDevices() {
     InferenceEngine::Core ie;
     std::vector<std::string> devices = ie.GetAvailableDevices();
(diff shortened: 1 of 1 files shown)


Wovchena:
CC @vmatveic

kchechil:
how we can say that HDDL is in the list of available devices if it can be even not physically available???

vmatveic:
> how we can say that HDDL is in the list of available devices if it can be even not physically available???

As far as I understand it's some kind of a recommendation that the particular demo MAY be executed on this device. Yes, you are right, the idea is to show only operable devices. And in worst case the user will just get an error message that the device is not available. But this is some kind of workaround for R2 which will be fixed for R3. @Wovchena please correct me if I'm wrong.

kchechil:
Jenkins please retry a build

kchechil:
Merging this only as a temporary workaround.

PR #246, opened 2019-07-18
Update CTPN input parameters

eaidova:
According to the source code provided with frozen graph, model waits for BGR image as input. Image resized with aspect ratio, minimal size is 600, maximal size is 1000.

It means that we have incorrect arguments for model optimizer and docs for this model.


kchechil:
cc @jkamelin

PR #254, opened 2019-07-24
fix potential crash in crossroad_camera_demo

antonrom23:
Added checks  in per-layer performance statistics print option


kchechil:
to be fixed after R2.

IRDonch:
Jenkins: ok to test

PR #262, opened 2019-07-29
Added parameter WITHOUT_EXTENSIONS

ilyachur:
Added a new parameter WITHOUT_EXTENSIONS in order to be able to build demos without dependency on CpuExtensions.


ilyachur:
Yes we are moving cpu_extensions into CPU plugin. I guess today or tomorrow the related changes will be merged to the private master branch.

WITHOUT_EXTENSIONS is necessary only for previous releases. Samples will be updated together with the Inference Engine  I guess.

demos/CMakeLists.txt
```diff
@@ -154,6 +154,9 @@ macro(ie_add_sample)
             add_definitions(-DUSE_OPENCV)
         endif()
     endif()
+    if(WITHOUT_EXTENSIONS)
```
IRDonch: There's no need for this to be an option. If there is an `IE::ie_cpu_extension` target, then use it. If there isn't, then don't.

demos/crossroad_camera_demo/main.cpp
```diff
@@ -26,7 +26,9 @@
 #include <samples/slog.hpp>
 #include <samples/ocv_common.hpp>
 #include "crossroad_camera_demo.hpp"
+#ifndef WITHOUT_EXTENSIONS
```
IRDonch: This is a double negative. I think it would be better to write it as `#ifdef WITH_EXTENSIONS`.

demos/CMakeLists.txt
ilyachur: Fixed

demos/crossroad_camera_demo/main.cpp
ilyachur: Fixed

IRDonch:
Jenkins: ok to test

PR #267, opened 2019-08-01
Move intel_models to models/intel

IRDonch:
This creates a single hierarchy of all model documentation. It's also the first step towards the splitting of Model Downloader config files into per-model configs.


IRDonch:
Jenkins please retry a build

kchechil:
My only suggestion would be to double check that OMZ public documentation is properly generated. cc @azaytsev1980

PR #269, opened 2019-08-02
Update README.md for cloning branch

jqueguiner:


vladimir-dudnik:
Jenknis please retry a build

kchechil:
cc @yssemaev @moslex @IRDonch 
@jqueguiner just FYI - the most up-to-date changes that are compliant with OpenVINO R2 are on master branch. Do you have a real need to use 2019 branch? Actually there was an intention to deprecate it.

IRDonch:
@jqueguiner Thanks for the contribution, but I'm afraid I'm going to decline for a couple of reasons:

1. The 2019 branch is closed. The most relevant branch to an average user is `master`, which always contains content compatible with the most recent OpenVINO toolkit release. And `master` is already the default branch, so it doesn't require any special flags to clone.

2. Your addition is not very specific to Open Model Zoo. It essentially just tells the reader how to use Git. I don't think that's within the scope of the README.

PR #272, opened 2019-08-03
Interactive Open Model Zoo as OpenCV module

dkurt:
```python
import cv2 as cv

from cv2.open_model_zoo.topologies import mobilenet_ssd
from cv2.open_model_zoo import DnnDetectionModel

frame = cv.imread('example.jpg')

ssd = mobilenet_ssd()
net = DnnDetectionModel(ssd)

classIds, confidences, boxes = net.detect(frame, confThreshold=0.5)
for box in boxes:
    cv.rectangle(frame, box, (0, 255, 0))

cv.imshow('out', frame)
cv.waitKey()
```
:snake: :wolf: :lion: :hear_no_evil: :elephant: :penguin: :octopus: :shark: :mouse: :camel: :frog: :ant: :leopard: :fox_face: :fish: :smirk_cat: :beetle: :whale: :koala: :spider: :dragon: :dove: :owl: :rabbit: :chicken: :crab: :bee:

**WIP**

- [x] Add inference. There are multiple options:
  ~~1. Each topology is derived from OpenCV's `dnn::Model`. Using downloaded models `IENetwork` can be initialized separately.~~
  2. Each topology is just meta class which can be casted to `dnn::Model` or `IENetwork`
- [x] Hash verification: do not download if hash is correct and verify hash after downloading (show warning in case of mismatch)
- [x] Cross platform cache dir (using `OPENCV_OPEN_MODEL_ZOO_CACHE_DIR` environment variable to change default caching location)
- [x] Archives management: tar.gz
- [x] Model Optimizer conversion
- [x] Added aliases for OpenVINO models (by highest version)
- [x] Archives management: zip
- [x] Downloading info: size and progress
- [x] Downloading and inference tests.
- [x] OpenCV DNN: input size
- [x] Documentation at help(topology): description, license etc.
- [ ] Replace pyyaml to own YAML implementation due it's not default Python package (is required only for build)
- [x] Errors handling (Downloading, Model Optimizer)
- [x] Remove archives after extract
- [x] Let user add and override Model Optimizer arguments
- [ ] Keep layout similar to the downloader

**related**: https://github.com/opencv/opencv/issues/14730

## Pipelines

example with text recognition demo API:
```python
import cv2 as cv
import numpy as np

from cv2.open_model_zoo import TextRecognitionPipeline

frame = cv.imread('text.jpg')

p = TextRecognitionPipeline()

rects, texts, confs = p.process(frame)

for rect, text in zip(rects, texts):
    vertices = cv.boxPoints(rect)

    for j in range(4):
        p1 = (vertices[j][0], vertices[j][1])
        p2 = (vertices[(j + 1) % 4][0], vertices[(j + 1) % 4][1])
        cv.line(frame, p1, p2, (0, 255, 0), 1);

    x = np.min(vertices[:,0])
    y = np.min(vertices[:,1])
    cv.putText(frame, text, (x, y), cv.FONT_HERSHEY_SIMPLEX, 1.0, (0, 255, 0), thickness=2)

cv.imshow('frame', frame)
cv.waitKey()
```
res

```python
import cv2 as cv
from cv2.open_model_zoo import HumanPoseEstimation

p = HumanPoseEstimation()

frame = cv.imread('example.png')
poses = p.process(frame)

p.render(frame, poses)

cv.imshow('res', frame)
cv.waitKey()
```
out


snosov1:
- Is it really necessary to have such a long prefix? `cv.open_model_zoo_TextRecognitionPipeline(det, rec)`
- Do you plan to add the example into the demos folder?

dkurt:
> Is it really necessary to have such a long prefix? cv.open_model_zoo_TextRecognitionPipeline(det, rec)

@snosov1, For now it's a kind of bug in OpenCV's wrappers (related feature request https://github.com/opencv/opencv/issues/14730). For now there is a workaround for it to define a method with proper name:

```cpp
class TextRecognitionPipelineImpl
{
    //...
};

Ptr<TextRecognitionPipelineImpl> TextRecognitionPipeline() { return new TextRecognitionPipelineImpl() }
```

corresponding Python code:
```python
from cv2.open_model_zoo import TextRecognitionPipeline
p = TextRecognitionPipeline()
```

> Do you plan to add the example into the demos folder?
It'd be nice. In example, add more Python demos for existing C++ ones. The only thing we have to decide is a code duplication: move C++ code from samples to the module and replace it to corresponding APIs.

snosov1:
Not sure I understand the answers, Dmitry =)

> For now it's a kind of bug in OpenCV's wrappers 

Are you saying that it will look "normal" after the merge of the respective PR?

> add more Python demos for existing C++ ones. 

That's a different story. For starters we should have at least one to showcase this type of usage.

dkurt:
@snosov1, 

```python
from cv2.open_model_zoo import TextRecognitionPipeline
p = TextRecognitionPipeline()
```
would be available :) I'll provide the code which do it in future commits.

jrosebr1:
@dkurt This is a really amazing PR, incredible work! If you need help putting together examples of how to use it let me know and I'll publish some content on the PyImageSearch blog 😄

dkurt:
Adrian, thank you! That would be great!

jrosebr1:
Absolutely! I'll keep an eye on this PR and once it's merged I'll put something together 😄 

I was looking at the `TextRecognitionPipeline` code -- I assume it's using EAST to localize text in the image, but what about the actual OCR component. What model is being used there?

dkurt:
@jrosebr1, `text-detection-0004`: https://github.com/opencv/open_model_zoo/blob/master/intel_models/text-detection-0004/description/text-detection-0004.md.

I think that next step is to enable these pipelines to with with different kind of models (it could be as EAST text detection as Intel's models).

IRDonch:
Without going into a detailed review, I have a few high level comments:

1. I'm confused as to how this is supposed to be used. OMZ is included into the OpenVINO toolkit as source code, while OpenCV is included as binaries. Thus an OpenVINO tookit user has no way to build this module, as his OpenCV is already built.

2. Half of this code is a reimplementation of Model Downloader/Converter. This is extremely fragile. The only supported way to download/convert/query models is by running the tools in `tools/downloader`. Anything else is liable to break as soon as the format of the config changes.

    This code, from what I can see, is already non-functional for models with regex-based postprocessing and PyTorch models. And it's going to be completely broken after #276, and then broken some more by another PR I have planned. And I heard rumors about a planned Caffe model which requires a new kind of post-processing, so that's going to be broken too. And that's just the changes that I can foresee.

    The bottom line is that this is unacceptable. You either need to run Model Downloader or just request that the user supply a directory with everything already downloaded. IMO, the latter option is way simpler.

dkurt:
@IRDonch, thanks for review!

> I'm confused as to how this is supposed to be used. OMZ is included into the OpenVINO toolkit as source code, while OpenCV is included as binaries. Thus an OpenVINO tookit user has no way to build this module, as his OpenCV is already built.

We can also build it as a part of OpenCV's binaries for OpenVINO. Just adding `OPENCV_EXTRA_MODULES_PATH` during build time.

> Half of this code is a reimplementation of Model Downloader/Converter. This is extremely fragile. The only supported way to download/convert/query models is by running the tools in tools/downloader. Anything else is liable to break as soon as the format of the config changes.

Python code is based on OMZ's downloader and https://github.com/opencv/opencv_extra/blob/master/testdata/dnn/download_models.py. Binaries uses only default Python packages (`urllib`, `zipfile`, `tarfile`, etc.)

> This code, from what I can see, is already non-functional for models with regex-based postprocessing and PyTorch models. And it's going to be completely broken after #276, and then broken some more by another PR I have planned. And I heard rumors about a planned Caffe model which requires a new kind of post-processing, so that's going to be broken too. And that's just the changes that I can foresee.

Most of postprocessing doesn't seem actual:
```yaml
    postprocessing:
      - $type: regex_replace
        file: mtcnn-p.prototxt
        pattern: 'dim: 12'
        replacement: 'dim: 720'
        count: 1
      - $type: regex_replace
        file: mtcnn-p.prototxt
        pattern: 'dim: 12'
        replacement: 'dim: 1280'
        count: 1
    model_optimizer_args:
      - --framework=caffe
      - --data_type=FP32
      - --input_shape=[1,3,720,1280]
      - --input=data
      - --mean_values=data[127.5,127.5,127.5]
      - --scale_values=data[128.0]
      - --output=conv4-2,prob1
      - --input_model=$dl_dir/mtcnn-p.caffemodel
      - --input_proto=$dl_dir/mtcnn-p.prototxt
```
I believe MO can ignore input dims at prototxt if `--input_shape=[1,3,720,1280]` specified (I can check it). Haven't found PyTorch models for now.

I wanted to adapt this PR after changes proposed at https://github.com/opencv/open_model_zoo/pull/276. That shouldn't be a problem.

IRDonch:
> We can also build it as a part of OpenCV's binaries for OpenVINO. Just adding OPENCV_EXTRA_MODULES_PATH during build time.

This seems impractical. It means that the OpenCV team would have to rebuild their binaries after every update to the model configuration file(s). From past experience I can say the models tend to get updated close to the time of the release due to last-minute Model Optimizer bugfixes, so the chances of OpenCV shipping with an out-of-sync module are quite high.

> Python code is based on OMZ's downloader and https://github.com/opencv/opencv_extra/blob/master/testdata/dnn/download_models.py. Binaries uses only default Python packages (urllib, zipfile, tarfile, etc.)

I can see that, but it does nothing to address my concern. Also, it's not entirely true - when you're running Model Optimizer, you depend on all of its dependencies.

> I believe MO can ignore input dims at prototxt if --input_shape=[1,3,720,1280] specified (I can check it).

That's good, but I know at least one regex-based fix that cannot be skipped - the one in `googlenet-v2`. In that case, the original `prototxt` file is actually ill-formed, and we have to patch it to make it work.

> Haven't found PyTorch models for now.

They were merged recently. You might want to rebase.

> I wanted to adapt this PR after changes proposed at #276. That shouldn't be a problem.

I'm sure you can adapt to that specific change, but the problem is that you _have_ to. Any new feature in the downloader/converter would require you (or someone else) to adapt the module to it. That's not sustainable.

dkurt:
@IRDonch, Thanks!

I agree that adding new module is a kind of overhead. There is another one option is to add it to regular OpenCV distribution. However the name of the package will be a bit different:

```python
from cv2.dnn.open_model_zoo import ...
```

Or

```python
from cv2.dnn.zoo import ...
```

This way we can use OMZ's topologies as optional ones and refactored https://github.com/opencv/opencv_extra/blob/master/testdata/dnn/download_models.py as a base.

We can start with it and check it's stability first.

vladimir-dudnik:
@IRDonch 
> I'm sure you can adapt to that specific change, but the problem is that you have to. Any new feature in the downloader/converter would require you (or someone else) to adapt the module to it. That's not sustainable.

Can we rely on downloader/converter API to minimize needs of changes in case of any new features in them, is not it?

dkurt:
@vladimir-dudnik, Proposed changes use pure Python to download files by URL. The module depends only on `list_topologies.yml` (or set of them after https://github.com/opencv/open_model_zoo/pull/276) and source code of some demos (for now it's `text_recognition` and `human_pose_estimation`).

IRDonch:
> Can we rely on downloader/converter API to minimize needs of changes in case of any new features in them?

That's what I'm suggesting...

dkurt:
@IRDonch, I'd like to propose the following solution: the package is still generated by OMZ structure (`intel/` and `public/` subfolders). For every model it's name is preserved in any way. In case of success - YAML is parser as well.

Then, at `download()` we will use `downloader.py` as a first priority tool for downloading (using the name of model we have). In case of failed `download.py` command (in example, when OpenCV is compiled with OMZ without OpenVINO and script has not been found) - try to download by existing code.

dkurt:
On the other hand - there is a lightweight solution with just Python module which can only download and convert models without demos.

I'll propose a separate PR with it.

ocv_module/open_model_zoo/README.md
```diff
@@ -0,0 +1,108 @@
...
+img = cv.imread('example.jpg')
+
+classId, confidence = net.classify(img)
+```
+
+## Algorithms
+
+Some of networks may have pretty complicated pre- or post- processing procedures.
+Another models can be combined to solve interesting problems. For these kind of
+topologies you can use ready Algorithms. In example, to recognize text:
```
mmphego, line 82:
```suggestion
Some networks may have pretty complicated pre- or post-processing procedures.
Other models can be combined to solve interesting problems. For this kind of
topologies you can use ready Algorithms. This example shows how to recognize text:
```

ocv_module/open_model_zoo/gen.py
```diff
@@ -0,0 +1,164 @@
...
+
+    fileURL, fileSHA, fileName = getSource(files[0])
+    if fileName.endswith('.tar.gz') or fileName.endswith('.zip'):
+        config['archive_url'], config['archive_sha256'], config['archive_name'] = fileURL, fileSHA, fileName
+    else:
+        config['config_url'], config['config_sha256'], config['config_path'] = fileURL, fileSHA, fileName
+        if len(files) > 1:
+            config['model_url'], config['model_sha256'], config['model_path'] = getSource(files[1])
+
+    s = ', '.join(['{"%s", "%s"}' % (key, value) for key, value in config.items()])
```
mmphego, line 90: No need to cast into a list.
```suggestion
    s = ', '.join('{"%s", "%s"}' % (key, value) for key, value in config.items())
```

mmphego (commented):
Some suggestions
Overral great PR can't wait for it to be merged.

Less is more!!!

PR #286, opened 2019-08-08
pedestrian_tracker: generalize detection

Wovchena:


demos/pedestrian_tracker_demo/src/detector.cpp
```diff
@@ -99,20 +106,37 @@ ObjectDetector::ObjectDetector(
...
+    switch(inputInfo.size()) {
+        case 1: {
+            InputInfo::Ptr inputInfoFirst = inputInfo.begin()->second;
+            inputInfoFirst->setPrecision(Precision::U8);
+            inputInfoFirst->getInputData()->setLayout(Layout::NCHW);
+            input_name_ = inputInfo.begin()->first;
+            im_info_name_.clear();
+            break;
+        }
+        case 2: {
```
IRDonch: Doesn't the `case 2` code also work for `case 1`?

IRDonch:
This is to support `person-detection-retail-0002`, right? If so, please add a corresponding case to the test script.

IRDonch:
(By the way, the tests failed because the build ran on the wrong node. I fixed the pipeline to make sure that doesn't happen anymore.)

demos/pedestrian_tracker_demo/src/detector.cpp
Wovchena: Updated

Wovchena:
> is is to support `person-detection-retail-0002`, right? If so, please add a corresponding case to the test script.

Added

PR #289, opened 2019-08-09
AC: allows to use regular expression instead of full output layer name for ssd pytorch

eaidova:
From one to other MO running model can have different suffixes for output layers.
It can become a problem for inplace conversion during Accuracy Checker running.


eaidova:
Jenkins please retry a build

PR #293, opened 2019-08-13
Add tests for segmentation_demo (C++ and Python)

IRDonch:


ci/requirements-demos.txt
```diff
@@ -0,0 +1 @@
+numpy==1.17.0 ; python_version >= "3.4"
```
Wovchena: There is another [`requirements.txt`](https://github.com/opencv/open_model_zoo/blob/develop/demos/python_demos/requirements.txt) for demos. Should we use it instead of creating another one?

IRDonch: The idea with the `ci/requirements-*.txt` files is that they contain a complete set of pinned packages, to ensure that CI builds are consistent. That's different from the file you linked, which is just a list of packages.

(Also, that `requirements.txt` includes `opencv-python`, which we _don't_ want to use, although that's arguably a bug.)

IRDonch:
Jenkins please retry a build

PR #294, opened 2019-08-13
AC: generalized converter for semantic segmentation

eaidova:
general converter for semantic segmentation without bounding for a specific dataset.


tools/accuracy_checker/accuracy_checker/annotation_converters/__init__.py
```diff
@@ -61,5 +62,6 @@
     'BratsNumpyConverter',
     'Cifar10FormatConverter',
     'MNISTCSVFormatConverter',
-    'WMTConverter'
+    'WMTConverter',
+    'common_semantic_segmentation'
```
vladimir-dudnik: is there any reason to have different naming convention?

eaidova: fixed, thanks

PR #296, opened 2019-08-13
ci/requirements-*.txt: bump all versions

IRDonch:
~Except for numpy in `requirements-ac.txt`, which needs to stay at 1.16 due to an issue with Pylint.~

EDIT: Bumped numpy too.

### DIFF
--- a/ci/requirements-ac.txt
+++ b/ci/requirements-ac.txt
@@ -1,12 +1,12 @@
 joblib==0.13.2            # via scikit-learn
-nibabel==2.4.1
-numpy==1.16.0
-pillow==6.0.0
+nibabel==2.5.0
+numpy==1.17.0
+pillow==6.1.0
 py-cpuinfo==4.0.0
-pyyaml==5.1.1
-scikit-learn==0.21.2
+pyyaml==5.1.2
+scikit-learn==0.21.3
 scipy==0.19.0
 shapely==1.6.4.post2
 six==1.12.0               # via nibabel
-tqdm==4.32.1
+tqdm==4.33.0
 yamlloader==0.5.5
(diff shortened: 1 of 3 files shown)


eaidova:
> Except for numpy in requirements-ac.txt, which needs to stay at 1.16 due to an issue with Pylint

Pylint issue: https://github.com/PyCQA/pylint/issues/2865
looks like it was fixed in the last pylint version, but I need additional check

IRDonch:
> looks like it was fixed in the last pylint version, but I need additional check

Oh, that's great to know. Let me know if you confirm it.

eaidova:
@IRDonch , pylint 2.3.1(+ astroid 2.2.5)  successfuly works with numpy 1.16.1  and higher.
But it detects one new issue, which I fix in https://github.com/opencv/open_model_zoo/pull/294

IRDonch:
Alright, in that case I'll bump NumPy too after #294 is merged.

eaidova:
@IRDonch , another requirement which can not be bumped is py-cpuinfo==4.0.0 (the next release has the bug with showing CPU flags from frozen code, it is not fixed yet). But we use this lib only for cpu_extension loading. As I know, CPU extension is going to be a part of CPU plugin in R3, so probably it makes sense to remove this lib from accuracy checker dependency at all.

IRDonch:
> another requirement which can not be bumped is py-cpuinfo==4.0.0

Good thing I didn't bump it then, right? :-)

(I use `pip-compile` to build these files, so I automatically obey all requirements set by `setup.py`.)

IRDonch:
Jenkins please retry a build

PR #298, opened 2019-08-15
Enbale check availability of HDDL plugin without booting HDDL

lgynju:
Previously, when querying the availability of HDDL, it will cost too much time to boot daemon which resulting in bad experience. As a workarouond, it print "HDDL" by default regardless of the result. Now, HDDL plugin supports availability of without booting HDDL.

Signed-off-by: Lu, Gaoyong <gaoyong.lu@intel.com>

### DIFF
--- a/demos/common/samples/common.hpp
+++ b/demos/common/samples/common.hpp
@@ -1114,11 +1114,14 @@ inline std::size_t getTensorBatch(const InferenceEngine::TensorDesc& desc) {
 inline void showAvailableDevices() {
     InferenceEngine::Core ie;
     std::vector<std::string> devices = ie.GetAvailableDevices();
-
-    std::cout << std::endl;
-    std::cout << "Available target devices:";
-    for (const auto& device : devices) {
-        std::cout << "  " << device;
+    if (!devices.empty()) {
+        std::cout << std::endl;
+        std::cout << "Available target devices:";
+        for (const auto& device : devices) {
+            std::cout << "  " << device;
+        }
     }
-    std::cout << "  HDDL" << std::endl;
+    else {
+        THROW_IE_EXCEPTION << "Not available any target device to infer on";
+    }
 }


kchechil:
@lgynju please re-target your PR to opencv:develop. My understanding is that you aim to deliver this support along with OpenVINO R3 which corresponds to develop branch of OMZ.

demos/common/samples/common.hpp
```diff
@@ -1114,11 +1114,14 @@ inline std::size_t getTensorBatch(const InferenceEngine::TensorDesc& desc) {
...
-        std::cout << "  " << device;
+    if (!devices.empty()) {
+        std::cout << std::endl;
+        std::cout << "Available target devices:";
+        for (const auto& device : devices) {
+            std::cout << "  " << device;
+        }
+    }
+    else {
+        THROW_IE_EXCEPTION << "Not available any target device to infer on";
```
IRDonch: How could this be possible? I'd assume that at least CPU is always available.

Wovchena: I suggest replacing this message with `"Could not find a device available to infer on"`.

lgynju: Yeah, it supposed to be, I just revert a walkaround commit of Ilya.

Wovchena: As I remember, @ilya-lavrenov said there is a config in IE that allows to disable plugins.

IRDonch: Alright.

IRDonch:
@lgynju Could you address @Wovchena's comment?

demos/common/samples/common.hpp
lgynju: yeah, as I have seen the PR https://github.com/opencv/open_model_zoo/pull/331, At least it won't print hard code `HDDL` no matter whether it has or not. And I am OK with the exception message. By the way, I think this PR is an option, you can opt to close it if not that necessary.

IRDonch: This PR still has value, since it adds the error message. Are you going to update it?

IRDonch:
Closing due to inactivity. Feel free to address the comments and reopen.

PR #125, opened 2019-05-23
added insightface face recognition models

elenash:

PR #299, opened 2019-08-15
Accuracy Checker: configs for insightface face recognition models

eaidova:
According to #125, added configs for accuracy checker. For both MxNet and IR (after model conversion)

cc: @snosov1, @kchechil, @eizamaliev, @elenash


snosov1:
Thinking out loud - does it make sense to make it part of @elenash's PR? (for demonstration purposes)

eaidova:
@snosov1 , Yes. of course, but #125 should be rebased before. I made these changes only for demonstrating which structure configs should have, I could not add them to @elenash's PR, because it does not contain `tools` and `accuracy_checker` at all.

eaidova:
no needs, configs become the part of #125

PR #304, opened 2019-08-15
Dynamic Open Model Zoo as Python module

dkurt:
**WIP**

an alternative solution for https://github.com/opencv/open_model_zoo/pull/272 with limited usage (no demos and no C++ but much more simpler).

- [ ] Decide location of `topologies.py` file to make it possible:
  ```python
  from openvino.open_model_zoo.topologies import mobilenet_ssd
  ```
- [x] Relative paths: to models and to `downloader.py` and `converter.py` scripts
- [ ] Add an ability to append / override MO arguments (in example, to vary precision)
- [x] Aliases for DLDT networks versions (like in https://github.com/opencv/open_model_zoo/pull/272)
- [x] Tests
- [x] OpenCV conversion (with mean and scales setting)
- [ ] Try to emit `requests` package dependency (non default)

## Current state:

* Documentation

```python
from topologies import mobilenet_ssd
help(mobilenet_ssd)
```
```
Help on class mobilenet_ssd in module topologies:

class mobilenet_ssd(Topology)
 |      The `mobilenet-ssd` model is a Single-Shot multibox Detection (SSD) network intended to perfo
rm object detection. This model is implemented using the Caffe\* framework. For details about this mo
del, check out the repository <https://github.com/chuanqi305/MobileNet-SSD>.
 |  The model input is a blob that consists of a single image of 1x3x300x300 in BGR order, also like 
the `densenet-121` model. The BGR mean values need to be subtracted as follows: [127.5, 127.5, 127.5]
 before passing the image blob into the network. In addition, values must be divided by 0.007843.
 |  The model output is a typical vector containing the tracked object data, as previously described.
 |  
 |      License: https://raw.githubusercontent.com/chuanqi305/MobileNet-SSD/master/LICENSE
 |  
 |      Attributes
 |      ----------
 |          config : str
 |              Absolute path to text file with network configuration.
 |          model : str
 |              Absolute path to weights file of network.
 |          framework : str
 |              Name of the framework in which format network is represented.
 |  
 |  Method resolution order:
 |      mobilenet_ssd
 |      Topology
 |      builtins.object
 |  
 |  Methods defined here:
 |  
 |  __init__(self, precision='FP32')
 |      Initialize self.  See help(type(self)) for accurate signature.
 |  
 |  ----------------------------------------------------------------------
 |  Methods inherited from Topology:
 |  
 |  getIENetwork(self)
 |      Creates openvino.inference_engine.IENetwork instance.
 |      Model Optimizer is lauched to create OpenVINO Intermediate Representation (IR).
 |  
 |  getIR(self, precision='FP32')
 |      Creates OpenVINO Intermediate Representation (IR) network format.
 |      
 |      Returns
 |      -------
 |          xmlPath
 |              Path to generated or downloaded .xml file
 |          binPath
 |              Path to generated or downloaded .bin file
 |  
 |  getOCVModel(self, useIR=True)
 |      Creates OpenCV's cv::dnn::Model from origin network. Depends on network type,
 |      returns cv.dnn_DetectionModel, cv.dnn_ClassificationModel, cv.dnn_SegmentationModel
 |      or cv.dnn_Model if not specified.
 |      
 |      Preprocessing parameters are set.
 |  
 |  ----------------------------------------------------------------------
 |  Data descriptors inherited from Topology:
 |  
 |  __dict__
 |      dictionary for instance variables (if defined)
 |  
 |  __weakref__
 |      list of weak references to the object (if defined)
```

* Downloading
```python
from topologies import mobilenet_ssd
ssd = mobilenet_ssd()
```
```
###############|| Downloading models ||###############

========= Downloading /home/dkurtaev/open_model_zoo/tools/downloader/models/public/mobilenet-ssd/mobilenet-ssd.prototxt
... 100%, 28 KB, 47565 KB/s, 0 seconds passed

========= Downloading /home/dkurtaev/open_model_zoo/tools/downloader/models/public/mobilenet-ssd/mobilenet-ssd.caffemodel
... 100%, 22605 KB, 6388 KB/s, 3 seconds passed


###############|| Post processing ||###############
```

### Usage example
* With OpenCV
```python
import cv2 as cv

from topologies import mobilenet_ssd

ssd = mobilenet_ssd()
net = ssd.getOCVModel()

frame = cv.imread('example.jpg')

classIds, confidences, boxes = net.detect(frame, confThreshold=0.5)
for box in boxes:
    cv.rectangle(frame, box, (0, 255, 0))

cv.imshow('out', frame)
cv.waitKey()
```
out

* With Inference Engine
```python
from topologies import mobilenet_ssd
from openvino.inference_engine import IENetwork

ssd = mobilenet_ssd()

xmlPath, binPath = ssd.getIR()
net = IENetwork(xmlPath, binPath)
```


dkurt:
@IRDonch, @vladimir-dudnik, Please review this PR. This is Python only tool to download models using `downloader.py` and convert them to IRs by `converter.py`.

Open question is location of this script in OpenVINO distribution. I want to see something like
```python
import openvino.open_model_zoo.topologies.public
import openvino.open_model_zoo.topologies.intel
```

tools/downloader/topologies.py
```diff
@@ -0,0 +1,233 @@
...
+import types
+
+import common
+
+if 'INTEL_OPENVINO_DIR' in os.environ:  # As a part of OpenVINO
+    _loc = os.path.join(os.environ['INTEL_OPENVINO_DIR'], 'deployment_tools', 'open_model_zoo')
+    _modelsDir = os.path.join(_loc, 'models')
+    _downloader = os.path.join(_loc, 'tools', 'downloader', 'downloader.py')
+else:  # Standalone
+    _loc = os.path.dirname(__file__)
```
IRDonch: Use `pathlib`, please.

tools/downloader/topologies.py
```diff
@@ -0,0 +1,233 @@
...
+    _modelsDir = os.path.join(_loc, '..', '..', 'models')
```
IRDonch: The variable naming convention is `snake_case`.

tools/downloader/topologies.py
```diff
@@ -0,0 +1,233 @@
...
+_precisionMarker = '$precision'
+_cache = os.path.abspath(os.getenv('OPENCV_OPEN_MODEL_ZOO_CACHE_DIR', 'models'))
+
+# Parse topologies.
+for _moduleName in os.listdir(_modelsDir):
+    _modelsSubDir = os.path.join(_modelsDir, _moduleName)
+    if not os.path.isdir(_modelsSubDir):
+        continue
+
+    _module = types.ModuleType(_moduleName)
```
IRDonch: I am unconvinced that each folder has to become a module, or that folders should be reflected in the API at all. The folders exist mainly to make development easier. Each model has a globally unique name, so the folder names are not required for disambiguation.

tools/downloader/topologies.py
```diff
@@ -0,0 +1,233 @@
...
+    sys.modules['topologies.' + _moduleName] = _module
+    globals()[_moduleName] = _module
+    _module.__dict__['sys'] = sys
+    _module.__dict__['os'] = os
+
+    # List of models which have versions (mostly DLDT models). For such kind of models
+    # with the highest version we will add short names. In example, for
```
IRDonch: This logic is flawed for a few reasons:

* They are not versions, but identifiers. A bigger number doesn't necessarily mean that the model is newer or better.
* Models with different identifiers might have different interfaces, so automatically choosing the biggest identifier might break compatibility with existing code if a new model with a higher identifier is added.

tools/downloader/topologies.py
```diff
@@ -0,0 +1,233 @@
...
+    # "face_detection_retail_0004" and "face_detection_retail_0005" method with name
+    # "face_detection_retail" returns result of "face_detection_retail_0005".
+    _versionedNames = {}
+
+    for _topologyName in os.listdir(_modelsSubDir):
+        _config = os.path.join(_modelsSubDir, _topologyName, 'model.yml')
```
IRDonch: This still duplicates the config parsing logic from `common.py`. You should reuse that instead, or use the info dumper.

tools/downloader/topologies.py
```diff
@@ -0,0 +1,233 @@
...
+                _configPath = _moArgs['--input_proto'].replace('$dl_dir/', '')
+
+            if _configPath:
+                _configPath = os.path.realpath(os.path.join(_downloadDir, _configPath))
+            if _modelPath:
+                _modelPath = os.path.realpath(os.path.join(_downloadDir, _modelPath))
+
+
+            exec("""
+class {className}:
```
IRDonch: Does each model really need to be its own class? I'd think that they could all be instances of the same class.

tools/downloader/topologies.py
```diff
@@ -0,0 +1,233 @@
+import sys
+import os
+import re
+import yaml
+import types
+
+import common
+
+if 'INTEL_OPENVINO_DIR' in os.environ:  # As a part of OpenVINO
```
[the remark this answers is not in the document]
dkurt: @IRDonch, The thing is that I want to do it in the following way:
```python
import openvino.open_model_zoo.topologies
```
but not
```python
import openvino.tools.open_model_zoo.downloader.topologies
```

tools/downloader/topologies.py
dkurt: It's a kind of open question. Do we want to have `intel` prefixes in users' apps or not to highlight which models are Intel trained and which are public.

tools/downloader/topologies.py
dkurt: I see, thank you!

tools/downloader/topologies.py
dkurt: That's a good point. I'll check how docstring will look like in this way.

tools/downloader/topologies.py
dkurt: Removed submodules

tools/downloader/topologies.py
dkurt: `common.load_topologies` reused

tools/downloader/topologies.py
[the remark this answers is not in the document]
IRDonch: I don't see how that's relevant, but you've removed this code anyway, so I guess it doesn't matter. :-)

tools/downloader/topologies.py
```diff
@@ -0,0 +1,191 @@
...
+    _files = _topology.files
+    _download_dir = _cache / _topology.subdirectory
+    _mo_args = {}
+    if _topology.mo_args:
+        for _arg in _topology.mo_args:
+            _tokens = _arg.split('=')
+            _mo_args[_tokens[0]] = _tokens[1] if len(_tokens) == 2 else None
+
+    # Resolve weights and text files names.
+    assert(len(_files) > 0)
```
IRDonch: I wouldn't recommend asserting this. We already have some PyTorch models where the model definition is pulled implicitly via a dependency on `torchvision`; I don't think it's beyond the realm of possibility for a future model to have its weights also embedded in some dependency.

Plus, nothing seems to depend on this assertion anyway.

tools/downloader/topologies.py
```diff
@@ -0,0 +1,191 @@
...
+
+    if len(_files) > 2:
```
IRDonch: This condition is definitely wrong:

1. There are some DLDT format models that have only 2 files (the `INT1` ones).
2. While I don't think there are currently any models in any other format that have more than 2 files, there's no reason why there couldn't be in the future. For example, TensorFlow models can comprise at least 3 files (metagraph, checkpoint, pipeline config), and there's no telling what future formats could be like.

tools/downloader/topologies.py
```diff
@@ -0,0 +1,191 @@
...
+        _paths = []
+        for i in range(2):
+            # Get only basename and add precision marker
+            _path = str(_files[i].name)
+            _pos = _path.find('/')
+            if _path[:_pos] in common.KNOWN_PRECISIONS:
+                _path = _precision_marker + _path[_pos:]
+            _paths.append(_path)
+
+        _config_path, _model_path = _paths
```
IRDonch: This relies on the assumption that the first two files listed are an `.xml` file and a `.bin` file (in that order), which is too unreliable. What you should depend on is the fact that the models will always be saved to `<precision>/<name>.xml`/`.bin`. Which, admittedly, is still an assumption, but I think I'm going to add logic to `common.py` to enforce this soon, at which point it'll be a contract.

dkurt: @IRDonch, It's useful to pass parameters in correct order. We have a method in OpenCV with is independent to network files order. It can can determine origin framework and config/model paths depends on extensions.

I can add some similar logic here.

tools/downloader/topologies.py
dkurt: Sure. Removed an assertion.

tools/downloader/topologies.py
dkurt: Thanks! Removed this condition and replaced onto `_topology.framework == 'dldt'`.

tools/downloader/topologies.py
dkurt: Added reversed paths in case of swapped `.xml` and `.bin` files. Thanks!

tools/downloader/topologies.py
dkurt: Created a base class. Thanks!

dkurt:
Is there any chance that we can add this module to 2019R3?

IRDonch:
> Is there any chance that we can add this module to 2019R3?

Honestly, the chances are pretty slim. I still need to review a lot of the code, plus there needs to be documentation and (preferably) CI integration.

tools/downloader/downloader.py
```diff
@@ -225,6 +235,11 @@ def main():
             output.mkdir(parents=True, exist_ok=True)
 
             for model_file in model.files:
+                # Download only files with specific precision.
+                prefix = str(model_file.name)[:str(model_file.name).find('/')]
+                if prefix in common.KNOWN_PRECISIONS and not prefix in requested_precisions:
+                    continue
```
IRDonch: The precisions option for the downloader is clearly a separate feature, so I'm going to request that you submit it as a separate PR. You'll also need to document this option in the README.

tools/downloader/downloader.py
```diff
@@ -225,6 +235,11 @@ def main():
             output.mkdir(parents=True, exist_ok=True)
 
             for model_file in model.files:
+                # Download only files with specific precision.
+                prefix = str(model_file.name)[:str(model_file.name).find('/')]
```
IRDonch: The logic that decides which file belongs to which precision should be located in `common`, and in fact there essentially already _is_ such logic there. You just need to change the data that `common.load_topologies` returns to include this information.

On the downloader level, the logic should be more high level: "if this file belongs to a precision, and the user requested this precision, then download it".

tools/downloader/common.py
```diff
@@ -228,6 +228,7 @@ def apply(self, output_dir):
         print('========= Unpacking {} ========='.format(postproc_file))
 
         shutil.unpack_archive(str(postproc_file), str(output_dir), self.format)
+        postproc_file.unlink()  # Remove an archive
```
IRDonch: This isn't a bad idea, but don't sneak it into an unrelated PR. Submit it separately.

tools/downloader/downloader.py
dkurt: @IRDonch, sure, thank you!

tools/downloader/common.py
dkurt: Opened https://github.com/opencv/open_model_zoo/pull/376

tools/downloader/downloader.py
dkurt: Opened a separate PR https://github.com/opencv/open_model_zoo/pull/384

tools/downloader/topologies.py
```diff
@@ -0,0 +1,196 @@
...
+        assert(len(_topology.files) >= 2), 'Expected to have at least two files in IR format'
+        _paths = []
+        # DLDT topologies usually come with multiple precision configuration files.
+        for i in range(2):
+            # Get only basename and add precision marker
+            _path = str(_topology.files[i].name)
+            _pos = _path.find('/')
+            if _path[:_pos] in common.KNOWN_PRECISIONS:
+                _path = _precision_marker + _path[_pos:]
+            _paths.append(_path)
```
IRDonch: This logic is redundant. For `dldt` models it's guaranteed that the files are `<precision>/<name>.xml` and `<precision>/<name>.bin`.

tools/downloader/topologies.py
```diff
@@ -0,0 +1,196 @@
...
+        _config_path, _model_path = _paths if _paths[0].endswith('.xml') else _paths[::-1]
+    else:
+        # Get paths from Model Optimizer arguments
+        if '--input_model' in _mo_args:
+            _model_path = _mo_args['--input_model'].replace('$dl_dir/', '')
+            _config_path = ''  # If there is a text file - we will set it next
+
+        for key in ['--input_proto', '--input_symbol']:
+            if key in _mo_args:
+                _config_path = _mo_args[key].replace('$dl_dir/', '')
```
IRDonch: This is very hacky, and I don't like it.

* There might be variables other than `dl_dir`.
* There might be different variable syntax (`string.Template`, which we use, supports `${var}` as well as `$var`).
* This logic will fail for models that have to be converted to ONNX first.
* It will also fail for TensorFlow checkpoint-based models.

In general, the whole practice of parsing Model Optimizer arguments is flawed, because you have no guarantee that any arguments you know about now are sufficient to fully describe any future model.

I don't think the model configs currently contain enough information to enable safe automated usage of the models in source formats. I think it might be useful to provide this information, but that is a feature in its own right. For now, though, I think you should remove any such usage.

tools/downloader/topologies.py
```diff
@@ -0,0 +1,196 @@
...
+
+    _model_path = _download_dir / _model_path
+    if _config_path:
+        _config_path = _download_dir / _config_path
+
+
+    exec("""
+class {className}(Topology):
```
IRDonch: I cannot respond to the original thread for this for some reason, so I'll open a new one.

It's even clearer now that the only difference between the topology classes is how they are initialized. So why do they need to be different classes? Having them be objects of the same class seems like a natural fit.

tools/downloader/topologies.py
dkurt: It's used to replace it after to requested precision.

tools/downloader/topologies.py
```diff
@@ -0,0 +1,196 @@
...
+
+        if self.config and not os.path.exists(self.config) or \
+           self.model and not os.path.exists(self.model):
+            sys.argv = ['', '--name=' + self.name, '--precisions=' + precision,
+                        '--output_dir=' + str(_cache).replace('\\', '\\\\')]
+            from downloader import main
+            main()
+
+
+    def getIR(self, precision='FP32'):
```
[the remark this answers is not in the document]
dkurt: done

tools/downloader/topologies.py
```diff
@@ -0,0 +1,196 @@
...
+
+        Returns
+        -------
+            xmlPath
+                Path to generated or downloaded .xml file
+            binPath
+                Path to generated or downloaded .bin file
+        '''
+        if self.framework == 'dldt':
+            return self.config, self.model
```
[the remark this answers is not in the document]
dkurt: nice catch! thanks

tools/downloader/topologies.py
dkurt: replaced

tools/downloader/topologies.py
dkurt: replaced to functions. If we create objects they will start downloading.

tools/downloader/topologies.py
dkurt: Replaced to string's substitution as it's implemented in https://github.com/opencv/open_model_zoo/blob/develop/tools/downloader/converter.py

tools/downloader/topologies.py
```diff
@@ -0,0 +1,196 @@
...
+        self.config = config.replace(_precision_marker, precision)
+        self.model = model.replace(_precision_marker, precision)
+        self.framework = framework
+
+        if self.config and not os.path.exists(self.config) or \
+           self.model and not os.path.exists(self.model):
+            sys.argv = ['', '--name=' + self.name, '--precisions=' + precision,
+                        '--output_dir=' + str(_cache).replace('\\', '\\\\')]
+            from downloader import main
+            main()
```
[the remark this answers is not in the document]
dkurt: That's a good point. Thanks! Replaced to direct main invocation (with parser arguments as parameters).

tools/downloader/topologies.py
```diff
@@ -0,0 +1,196 @@
...
+        self.mo_args = mo_args
+        self.download_dir = download_dir
+        self.config = config.replace(_precision_marker, precision)
+        self.model = model.replace(_precision_marker, precision)
+        self.framework = framework
+
+        if self.config and not os.path.exists(self.config) or \
+           self.model and not os.path.exists(self.model):
+            sys.argv = ['', '--name=' + self.name, '--precisions=' + precision,
+                        '--output_dir=' + str(_cache).replace('\\', '\\\\')]
```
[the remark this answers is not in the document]
dkurt: For Windows paths.

tools/downloader/topologies.py
```diff
@@ -0,0 +1,196 @@
...
+        self.name = name
+        self.task_type = task_type
+        self.mo_args = mo_args
+        self.download_dir = download_dir
+        self.config = config.replace(_precision_marker, precision)
+        self.model = model.replace(_precision_marker, precision)
+        self.framework = framework
+
+        if self.config and not os.path.exists(self.config) or \
+           self.model and not os.path.exists(self.model):
```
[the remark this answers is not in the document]
dkurt: The problem of cache is size. In fact it takes x2 time more memory to store models (cache + used files). That's why I wanted to store models in the working directory by default so user can easily find it and reuse in a different way or check that they are correct.

tools/downloader/topologies.py
```diff
@@ -0,0 +1,196 @@
...
+import yaml
+import types
+import argparse
+
+from pathlib import Path
+
+import common
+
+_precision_marker = '$precision'
+_cache = Path(os.getenv('OPENCV_OPEN_MODEL_ZOO_CACHE_DIR', Path('.').resolve() / 'models'))
```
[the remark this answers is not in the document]
dkurt: User-defined directory can be used (`OPENCV_OPEN_MODEL_ZOO_CACHE_DIR` environment variable).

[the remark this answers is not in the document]
ngaloppo: There should be a reasonable fallback folder in case `OPENCV_OPEN_MODEL_ZOO_CACHE_DIR` is not defined. Could be created by `tempfile.mkdtemp()`. 

Why is the `OPENCV_OPEN_MODEL_ZOO_CACHE_DIR` prefixed with `OPENCV`? This pull request is for an OpenVINO module. I would just drop the prefix, or prefix with `OPENVINO`.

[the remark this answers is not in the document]
dkurt: > There should be a reasonable fallback folder in case OPENCV_OPEN_MODEL_ZOO_CACHE_DIR is not defined. Could be created by tempfile.mkdtemp().

There is a fallback already - folder named `models` in the current working directory.

> Why is the OPENCV_OPEN_MODEL_ZOO_CACHE_DIR prefixed with OPENCV? This pull request is for an OpenVINO module. I would just drop the prefix, or prefix with OPENVINO.

Renamed it to just `OPEN_MODEL_ZOO_CACHE_DIR`. `OPENVINO_OPEN_MODEL_ZOO_CACHE_DIR` is to long :)

dkurt:
Removed tests for now. We can enable them after merge of this PR.

dkurt:
@IRDonch, @vladimir-dudnik, @ngaloppo, may I ask you to review this patch one more time? If there are no critical issues which we cannot fix further without breaking compatibility, can we merge it now? Tests and integration questions can be solved separately due they depend on different repositories.

tools/downloader/topologies.py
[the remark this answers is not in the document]
IRDonch: > The problem of cache is size. In fact it takes x2 time more memory to store models (cache + used files).

Yeah. That's the price of robustness.

> That's why I wanted to store models in the working directory by default so user can easily find it and reuse in a different way or check that they are correct.

This doesn't make sense. If the user wants to use our models in a different way, they can run the downloader themselves. The purpose of this API, as far as I can tell, is to let the user use our models _without_ paying attention to the model files. And for that it has to be robust.

tools/downloader/topologies.py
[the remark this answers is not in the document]
IRDonch: Please elaborate.

tools/downloader/topologies.py
IRDonch: I'm not sure what you replaced, but it's still redundant. You don't need a loop or a `_paths` list. You know what the names of the files are.

tools/downloader/topologies.py
IRDonch: But this doesn't address my point at all... You just made your ad-hoc logic a little bit more ad-hoc.

tools/downloader/topologies.py
IRDonch: > If we create objects they will start downloading.

They won't if you don't implement them that way. There's no reason why the object has to start downloading the model files as soon as it's created.

tools/downloader/topologies.py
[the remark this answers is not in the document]
IRDonch: Great, this addresses 1 point out of 3. What about the rest of them?

Seriously, the downloader is not a library. It is a program. Use it as a program.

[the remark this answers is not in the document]
ngaloppo: I agree with @IRDonch on this one. The current solution feels hacky, and not robust. I believe that there is a need for a library that loads model zoo models, aside from having a __program__ that downloads models, but let me elaborate on that in a separate thread. It might be that we need a more significant refactor to achieve the two goals (downloader program & library). Let's figure that out once we have figured out what exactly we want out of the library - so let me get feedback on that one first :)

ngaloppo:
First, let me comment on the value of a library that can load models from the model zoo, _in addition to a program to do so_. In particular, this is useful for interactive experimentation notebooks (e.g. Jupyter), so that you don't need to go through shell execution to call `downloader.py` (and potentially have to set up the shell to have the program dependencies set up). The requirements for the program and the library are different. The program will download and store serialized versions of the models to the filesystem, immediately. The behavior and features of the library is different - so let's align on those first. I have taken some inspiration from `torch.utils.model_zoo` and `tf.hub`. 

### Features

I don't think the current approach (which mimics `torchvision`) is maintainable. It requires creating classes for each new model in the model zoo. So please allow me to suggest a slight difference to the approach. 

#### Minimal features

- Get available models in [open\_model\_zoo](https://github.com/opencv/open_model_zoo), return a list of model names
- Load a model by model name (e.g. `'se-resnext-50'`), return an instance of `IENetwork`

That's it. All else is transparent to the user. Users should not have to care about where or how models are stored/cached. This approach implies that the model is downloaded immediately (if not already cached), unless we make significant changes to how IENetwork works, I **think**. The user can convert to serialized version by using `IENetwork.serialize()`.

#### Optional features

- Load a model by URL pointing to a .tgz file (containing any supported serialized version)
- Get references to the cached serialized (OpenVINO) IR of the model.
- Some way to get a `cv::dnn::Model` out of this library. Is there a way to convert `IENetwork` instances to `cv::dnn::Model` instances? Alternatively, if we implement the feature above, we could go through the `cv::dnn::readNet()` path. What 

#### API

These are the minimal API's I'd like to see.  Let's iterate over the API, and fine-tune the behavior here.

- `def get_model_names() -> List[str]`
- `def get_model(name: str, progress: bool = True, cache_location: Optional[str]) -> openvino.IENetwork`

Basic usage: 

```
import openvino.open_model_zoo as omz

name = 'resnet-50'
if name in omz.get_model_names():
	model = omz.load_model(name)
...
```

#### Notes

**Download only vs. load model**: I could be convinced to make it even simpler: 

```
import openvino.open_model_zoo as omz
name = 'resnet-50'
xml, bin = omz.download_model(name)
model = IENetwork.from_ir(xml, bin)
```

The API spec would be `download_model(name:str, progress: bool = True, cache_location: Optional[str]) -> Tuple[str, str]` returning the paths of the cached `.xml,.bin` files. 

This also solves the problem of wanting a `cv::dnn::Model`, and not a `ie_api.IENetwork`.

**Cached location**: I like the approach in `torch.utils.model_zoo`:

> If the object is already present in model_dir, it’s deserialized and returned. The default value of model_dir is $TORCH_HOME/checkpoints where environment variable $TORCH_HOME defaults to $XDG_CACHE_HOME/torch. $XDG_CACHE_HOME follows the X Design Group specification of the Linux filesytem layout, with a default value ~/.cache if not set.

Replace `$TORCH_HOME` by `$OPEN_MODEL_ZOO_HOME`.

IRDonch:
I'm certainly all for starting with a minimal viable API.

I'd like to suggest a modification, though. I think the usage should look like this:

```python
import openvino.open_model_zoo as omz

with omz.Zoo() as zoo:
    name = 'resnet-50'
    assert name in zoo.model_names:
	model = zoo.load_model(name)
```

The loading of information about the modules would be encapsulated in `Zoo.__init__`, so that the importing of the module by itself doesn't trigger reading of config files / launching programs. And `Zoo` is a context manager, so that it can remove any temporary files it creates by the end of the program.

IRDonch:
> First, let me comment on the value of a library that can load models from the model zoo, in addition to a program to do so. In particular, this is useful for interactive experimentation notebooks (e.g. Jupyter), so that you don't need to go through shell execution to call downloader.py (and potentially have to set up the shell to have the program dependencies set up).

FWIW, this could potentially work for models that are published in the IR format, but for other models you have to run Model Optimizer (and potentially other scripts) to convert them, so the ability to execute external programs is required anyway.

ngaloppo:
> FWIW, this could potentially work for models that are published in the IR format, but for other models you have to run Model Optimizer (and potentially other scripts) to convert them, so the ability to execute external programs is required anyway.

I think that support for IR format models only is acceptable for the minimum viable feature set of the first version. What's the percentage of OMZ models that has IR format published?

ngaloppo:
> I'm certainly all for starting with a minimal viable API.
> 
> I'd like to suggest a modification, though. I think the usage should look like this:
> 
> ```python
> import openvino.open_model_zoo as omz
> 
> with omz.Zoo() as zoo:
>     name = 'resnet-50'
>     assert name in zoo.model_names:
> 	model = zoo.load_model(name)
> ```
> 
> The loading of information about the modules would be encapsulated in `Zoo.__init__`, so that the importing of the module by itself doesn't trigger reading of config files / launching programs. And `Zoo` is a context manager, so that it can remove any temporary files it creates by the end of the program.

That's acceptable syntax to me. However, I wouldn't want it to remove files that can (and should) be cached. The library should be intelligent enough to check OMZ whether its cache is stale, in case it needs an update. This brings up a good topic though: we may need some kind of versioning of models, for reproducibility.

dkurt:
see https://github.com/openvinotoolkit/open_model_zoo/pull/1352

PR #311, opened 2019-08-19
tools/downloader/converter.py: add support for parallel conversion

IRDonch:


tools/downloader/converter.py
```diff
@@ -130,18 +164,28 @@ def main():
...
+                if not prefixed_subprocess(stdout_prefix, mo_cmd):
+                    return False
+
+            prefixed_printf(stdout_prefix, '')
 
-                if subprocess.run(mo_cmd).returncode != 0:
-                    failed_topologies.add(top.name)
+        return True
+
+    if args.jobs == 1:
```
zmaslova: I suggest to run in one thread for dry-run also

IRDonch: Excellent point; will do.

IRDonch: Done.

PR #316, opened 2019-08-20
AC: supported CVAT annotation format for attributes recogntion task

eaidova:
this converter will cover emotion-recognition and vehicle-attributes-recognition for pre-commit CI


vladimir-dudnik:
minor but still, may we change names from CVATForAttributes to CVATAttributesConvertion or something like this to be similar with other convertors?

eaidova:
ok, renamed

PR #317, opened 2019-08-20
AC:  supproted CVAT annotation format for Object Detection task 

eaidova:
this converter will cover all object detection tasks in pre-commit CI


eaidova:
> May be use CVATObjectDetectionConverter ?

Ok, renamed

PR #326, opened 2019-08-22
[DEMOS] minimum required CMake version for demos is updated to 2.8.12

vmatveic:


vmatveic:
@Wovchena @IRDonch this change is already implemented and tested for the original openVINO samples. Needs to be extended for the demos too.

Wovchena:
@vmatveic, what is a reason for the change?

vmatveic:
> @vmatveic, what is a reason for the change?

@Wovchena , forgot to specify it. The idea is that original openVINO CMake infrastructure was updated (in scope of upcoming release) with new Interface Compile Options propagated to all the users of the Inference Engine library. But these flags are only supported starting from 2.8.12.

Also, this change is already doublechecked from the perspective of the minimum supported CMake version (now it's 2.8.12 which is default CMake version for oldest support CentOS version).

Thanks.

P.S.: due to cmake_minimum_required is called with 2.8.12 version some CMake policies are automatically applied and additional small change in CMake is required. I think we can merge it after pipeline is passed.

PR #331, opened 2019-08-22
Revert hddl workaround

AndrewBakalinIntel:


Wovchena:
We have a similar PR: https://github.com/opencv/open_model_zoo/pull/298 and I think it covers more cases.

IRDonch:
@Wovchena That is true, but that one has stalled and I think this change is valuable by itself.

IRDonch:
Jenkins: test this please

PR #345, opened 2019-08-27
Model Downloader: add a machine-readable output format

IRDonch:


IRDonch:
Don't merge yet; I just noticed some things that need to be fixed.

IRDonch:
False alarm, but I used it to change the name of the option. Ready for review now.

tools/downloader/downloader.py
```diff
@@ -43,53 +43,46 @@ def process_download(chunk_iterable, size, file):
...
             return True
-        except requests.exceptions.ConnectionError as e:
-            print("Error Connecting:", e)
-        except requests.exceptions.Timeout as e:
-            print("Timeout Error:", e)
-        except requests.exceptions.TooManyRedirects as e:
-            print("Redirects Error: requests exceeds maximum number of redirects", e)
-        except (requests.exceptions.RequestException, ssl.SSLError) as e:
-            print(e)
+        except (requests.exceptions.RequestException, ssl.SSLError):
```
zmaslova: why do you remove all other exceptions?

IRDonch: The other exceptions are subclasses of `RequestException`, and their exception messages contain enough detail to make special handling for them unnecessary.

PR #354, opened 2019-08-30
demos/tests: add smart_classroom

Wovchena:


demos/multichannel_demo/models.lst
```diff
@@ -1,3 +1,5 @@
 # This file can be used with the --list option of the model downloader.
 face-detection-retail-????
+face-detection-adas-????
+face-detection-adas-binary-????
```
IRDonch: Keep these in sorted order, please.

demos/tests/image_sequences.py
```diff
@@ -135,4 +135,17 @@
         image_net_arg('00037128'),
         image_net_arg('00048316'),
     ],
+
+    '500x375x3': [
```
IRDonch: Is this just a sequence of arbitrary same-sized images? I don't think we're going to get good code coverage with a sequence like that - you can't reidentify faces if there aren't any faces.

demos/multichannel_demo/models.lst
Wovchena: Updated

demos/tests/image_sequences.py
Wovchena: Updated images

vladimir-dudnik:
looks good to me

PR #359, opened 2019-09-02
Fix some issues for NCF

dliang0406:


eaidova:
Jenkins ok to test

tools/accuracy_checker/accuracy_checker/metrics/hit_ratio.py
```diff
@@ -62,11 +62,13 @@ def evaluate(self, annotations, predictions):
         for user in range(self.users_num):
             map_item_score = {}
             for j in range(iter_num):
-                item = self.pred_per_user[user][j][0]
-                score = self.pred_per_user[user][j][1]
-                map_item_score[item] = score
+                if len(self.pred_per_user[user]) != 0:
```
eaidova: hmm...
if we check that user in mapping and have predictions, could we move this check into the first cycle (where we enumerate users)?
just:

```python
for user in range(self.users_num):
     if not self.pred_per_user[user]:
          continue
     map_item_score = {}
     for j in range(iter_num):
           item = self.pred_per_user[user][j][0]
           score = self.pred_per_user[user][j][1]
           map_item_score[item] = score

eaidova: Also this place has pylint warning:
accuracy_checker/metrics/hit_ratio.py:65: [C1801(len-as-condition), BaseRecommenderMetric.evaluate] Do not use `len(SEQUENCE)` to determine if a sequence is empty

please fix it

dliang0406: Thank you @eaidova !

PR #363, opened 2019-09-02
WIP: Conversion script from PyTorch to ONNX updated

eizamaliev:
- Added support of model's creation parameters
- Fixed conversion error for some models

### DIFF
--- a/tools/downloader/pytorch_to_onnx.py
+++ b/tools/downloader/pytorch_to_onnx.py
@@ -21,6 +21,14 @@ def positive_int_arg(values):
             sys.exit('Invalid value for input argument: {!r}, a positive integer is expected'.format(value))
     return result
 
+def model_parameters(parameters):
+    if not parameters:
+        return None
+    model_params = dict((param, value) for param, value in (element.split('=') for element in parameters.split(',')))
+    for param in model_params.keys():
+        model_params[param] = eval(model_params[param])
+    return model_params
+
 
 def parse_args():
     """Parse input arguments"""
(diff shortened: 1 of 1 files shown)


tools/downloader/pytorch_to_onnx.py
```diff
@@ -21,6 +21,14 @@ def positive_int_arg(values):
             sys.exit('Invalid value for input argument: {!r}, a positive integer is expected'.format(value))
     return result
 
+def model_parameters(parameters):
+    if not parameters:
+        return None
+    model_params = dict((param, value) for param, value in (element.split('=') for element in parameters.split(',')))
+    for param in model_params.keys():
+        model_params[param] = eval(model_params[param])
```
IRDonch: This seems too ad-hoc to me. If you're going to use `eval` anyway, why not just take the entire constructor call as a script argument and `eval` the whole thing at once?

tools/downloader/pytorch_to_onnx.py
```diff
@@ -95,6 +104,7 @@ def convert_to_onnx(model, input_shape, output_file, input_names, output_names):
     output_file.parent.mkdir(parents=True, exist_ok=True)
     model.eval()
     dummy_input = torch.randn(input_shape)
+    model(dummy_input)
```
IRDonch: What is the purpose of this?

eizamaliev: This fixes some conversion errors, when some layer's shape isn't defined explicitly by constant (_for some models_).

tools/downloader/pytorch_to_onnx.py
eizamaliev: I've already tried your suggestion, but got some problems. This is most suitable solution for this moment. I can seek the way to improve it, if you insist.

IRDonch: I'd like to at least know what problems you ran into. Maybe we could come up with a better solution together.

> This is most suitable solution for this moment.

I should note that so far I don't even see the problem that this is a solution for. :-)

tools/downloader/pytorch_to_onnx.py
IRDonch: OK, but could you add a comment explaining this? Otherwise, it just looks like debugging code that was left in.

eizamaliev:
Added in [PR 442](https://github.com/opencv/open_model_zoo/pull/442)

PR #376, opened 2019-09-05
Remove archives after unpacking

dkurt:


tools/downloader/common.py
```diff
@@ -286,6 +286,7 @@ def apply(self, reporter, output_dir):
         reporter.print_section_heading('Unpacking {}', postproc_file)
 
         shutil.unpack_archive(str(postproc_file), str(output_dir), self.format)
+        postproc_file.unlink()  # Remove an archive
```
IRDonch: ITYM "remove _the_ archive"

dkurt: Sure. Replaced

PR #384, opened 2019-09-07
--precisions flag for downloader.py

dkurt:
resolves https://github.com/opencv/open_model_zoo/issues/308

### DIFF
--- a/tools/downloader/README.md
+++ b/tools/downloader/README.md
@@ -36,7 +36,7 @@ If you using models from PyTorch framework, you will also need to use intermedia
 conversion to ONNX format. To use automatic conversion install additional dependencies:
 
 ```sh
-python3 -mpip install --user -r ./requirements-pytorch.in 
+python3 -mpip install --user -r ./requirements-pytorch.in
 ```
 
 Model downloader usage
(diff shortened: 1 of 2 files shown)


tools/downloader/README.md
```diff
@@ -59,6 +59,13 @@ option:
 The `--all` option can be replaced with other filter options to download only
 a subset of models. See the "Shared options" section.
 
+You may use `--precisions` flag to specify comma separated precisions of weights
+to be downloaded.
+
+```sh
+./downloader.py --name face-detection-retail-0004 --precision FP16,INT8
```
grib0ed0v: @dkurt `--precisions`, otherwise it's not working

dkurt: @grib0ed0v, nice catch, thank you!

tools/downloader/downloader.py
```diff
@@ -191,6 +191,8 @@ def main():
         help='download only models whose names match at least one of the patterns in the specified file')
     parser.add_argument('--all',  action='store_true', help='download all available models')
     parser.add_argument('--print_all', action='store_true', help='print all available models')
+    parser.add_argument('--precisions', metavar='PREC[,PREC...]',
```
eaidova: why not `nargs`?  it allows setting list of arguments using space as delimiter. 
https://docs.python.org/3/library/argparse.html#nargs

dkurt: To align it with `converter.py`:

https://github.com/opencv/open_model_zoo/blob/c3b4b2b9845db2ca15e03df9804b6f98d7a3c211/tools/downloader/converter.py#L98-L99

kchechil:
Jenkins please retry a build

PR #394, opened 2019-09-10
AC: fixed axis for per point normed error metric calculation

eaidova:


eaidova:
Jenkins please retry a build

PR #395, opened 2019-09-10
multi_camera_multi_person_tracking: remove person-detection-retail-0002

Wovchena:
The demo supports only single input nets, while person-detection-retail-0002 has two inputs.


Wovchena:
@sovrasov, are you OK with change?

sovrasov:
Yes, it's OK

PR #402, opened 2019-09-11
MTMC: bugfixes

sovrasov:
Fixed:
-  IOU calculation
- reid similarity aggregation
- track-detection threshold value


vladimir-dudnik:
Jenkins please retry a build

demos/python_demos/multi_camera_multi_person_tracking/mc_tracker/sct.py
```diff
@@ -317,15 +317,15 @@ def _compute_detections_assignment_cost(self, active_tracks_idx, detections, fea
...
+                    reid_sim = max(reid_sim_avg, reid_sim_curr, reid_sim_clust)
                 else:
                     reid_sim = 0.5
                 affinity_matrix[j, i] = iou * reid_sim
         return 1 - affinity_matrix
 
     @staticmethod
     def _area(box):
-        return max((box[2] - box[0]) * (box[3] - box[1]), 0)
+        return max(box[2] - box[0], 0) * max(box[3] - box[1], 0)
```
Wovchena: Shouldn't it be `abs((box[2] - box[0]) * (box[3] - box[1]))`?

sovrasov: No. This method should return 0 in case of invalid box. Previously it returned 0 only if height or width was <= 0, but case then both h and w are negative wasn't cached since h*w > 0. Now the method considers h and w separately.

kchechil:
Jenkins please retry a build

PR #409, opened 2019-09-12
fix input for deeplabv3 in docs add config for accuracy validation

eaidova:


eizamaliev:
I think it is important to point to demo/sample can work incorrectly, since they expect `BGR` input.

eaidova:
@eizamaliev, a note about demos added, please take a look again.

PR #423, opened 2019-09-18
AC: remove usage deprecated scipy image api

eaidova:
removed `scipy.misc.imread` and `scipy.misc.imresize` usage. Changes allow deleting restriction on supported scipy version

### DIFF
--- a/tools/accuracy_checker/accuracy_checker/data_readers/data_reader.py
+++ b/tools/accuracy_checker/accuracy_checker/data_readers/data_reader.py
@@ -20,7 +20,6 @@ from collections import OrderedDict, namedtuple
 import re
 import cv2
 from PIL import Image
-import scipy.misc
 import numpy as np
 import nibabel as nib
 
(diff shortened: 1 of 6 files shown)


tools/accuracy_checker/accuracy_checker/data_readers/data_reader.py
```diff
@@ -180,8 +179,163 @@ def read(self, data_id):
...
+        ts = type(size)
+        if np.issubdtype(ts, np.signedinteger):
+            percent = size / 100.0
+            size = tuple((np.array(im.size) * percent).astype(int))
+        elif np.issubdtype(type(size), np.floating):
+            size = tuple((np.array(im.size) * size).astype(int))
+        else:
+            size = (size[1], size[0])
+        func = {'nearest': 0, 'lanczos': 1, 'bilinear': 2, 'bicubic': 3, 'cubic': 3}
+        imnew = im.resize(size, resample=func[interp])
```
IRDonch: I think this can be simplified a lot, given that `size` is always a tuple, `interp` is always `'nearest'` and `mode` is always `None`.

tools/accuracy_checker/accuracy_checker/data_readers/data_reader.py
```diff
@@ -180,8 +179,163 @@ def read(self, data_id):
...
+            raise ValueError("Invalid array shape for mode.")
+        if mode in ['RGBA', 'CMYK'] and numch != 4:
+            raise ValueError("Invalid array shape for mode.")
+
+        # Here we know data and mode is correct
+        image = Image.frombytes(mode, shape, strdata)
+        return image
+
+    @staticmethod
+    def _to_image(arr, high=255, low=0, cmin=None, cmax=None, pal=None, mode=None, channel_axis=None):
```
IRDonch: None of these keyword parameters ever get set to anything but default values, so they can all be removed.

tools/accuracy_checker/accuracy_checker/data_readers/data_reader.py
```diff
@@ -180,8 +179,163 @@ def read(self, data_id):
...
+        if mode == 'I':
+            data32 = data.astype(np.uint32)
+            image = Image.frombytes(mode, shape, data32.tostring())
+        else:
+            raise ValueError("Mode is unknown or incompatible with input array shape.")
+
+        return image
+
+    @staticmethod
+    def _process_3d(data, shape, mode, channel_axis, high, low, cmin, cmax):
```
IRDonch: I don't think it's possible to reach this function, since it's only reachable through `imresize` and you only call `imresize` to do a 2-dimensional resize.

eaidova: Under 3d here meant image which has height, width, channels. I am not sure that it is unreachable, Pascal VOC annotation is 3-channels images before encoding, so if `resize_segmentation_mask` calls before encoding, it will process 3d image. I can not guarantee that it is not used in such way.

eaidova:
Jenkins please retry a build

tools/accuracy_checker/accuracy_checker/data_readers/data_reader.py
IRDonch: I see, I was just confused about what 3D meant.

tools/accuracy_checker/accuracy_checker/data_readers/data_reader.py
```diff
@@ -180,8 +179,94 @@ def read(self, data_id):
...
+        return image
+
+    @staticmethod
+    def _process_3d(data, shape):
+        # if here then 3-d array with a 3 or a 4 in the shape length.
+        # Check for 3 in datacube shape --- 'RGB' or 'YCbCr'
+        ca = np.flatnonzero(np.asarray(shape) == 3) if 3 in shape else np.flatnonzero(np.asarray(shape) == 4)
+        if not np.size(ca):
+            raise ValueError("Could not find channel dimension.")
+        ca = ca[0]
```
IRDonch: This channel dimension searching logic seems really dodgy to me. Surely in our case the channel dimension is always the same?

tools/accuracy_checker/accuracy_checker/data_readers/data_reader.py
```diff
@@ -180,8 +179,94 @@ def read(self, data_id):
...
+    def imresize(arr, size):
+        im = ScipyImageReader._to_image(arr)
+        size = (size[1], size[0])
+        imnew = im.resize(size, resample=0)
+
+        return ScipyImageReader._from_image(imnew)
+
     def read(self, data_id):
-        return np.array(scipy.misc.imread(str(get_path(self.data_source / data_id))))
+        return ScipyImageReader._imread(str(get_path(self.data_source / data_id)))
```
IRDonch: I think you can just merge `_imread` into here. It's the only use of it, so not much point in having it as a separate function.

tools/accuracy_checker/accuracy_checker/data_readers/data_reader.py
```diff
@@ -180,8 +179,94 @@ def read(self, data_id):
...
+
+        return (bytedata.clip(0, 255) + 0.5).astype(np.uint8)
+
+    @staticmethod
+    def imresize(arr, size):
+        im = ScipyImageReader._to_image(arr)
+        size = (size[1], size[0])
+        imnew = im.resize(size, resample=0)
+
+        return ScipyImageReader._from_image(imnew)
```
IRDonch: Given that `_to_image` can't possibly return a `P` format image, the call to `_from_image` here can be replaced with just a call to `np.array`. At which point all `imresize`-related code will become independent from all `imread`-related code, so you can localize `imresize` and its dependencies to the file where it's actually used.

tools/accuracy_checker/accuracy_checker/data_readers/data_reader.py
```diff
@@ -180,8 +179,94 @@ def read(self, data_id):
 class ScipyImageReader(BaseReader):
     __provider__ = 'scipy_imread'
```
IRDonch: I think a note should be made in the documentation that this doesn't literally use SciPy anymore.

tools/accuracy_checker/accuracy_checker/data_readers/data_reader.py
```diff
@@ -180,8 +179,94 @@ def read(self, data_id):
...
+    @staticmethod
+    def _to_image(arr):
+        data = np.asarray(arr)
+        if np.iscomplexobj(data):
+            raise ValueError("Cannot convert a complex-valued array.")
+        shape = list(data.shape)
+        valid = len(shape) == 2 or ((len(shape) == 3) and ((3 in shape) or (4 in shape)))
+        if not valid:
+            raise ValueError("'arr' does not have a suitable array shape for any mode.")
+        if len(shape) == 2:
```
IRDonch: There is some duplication of logic here. I think rewriting this part as:

```
if len(shape) == 2:
    # process 2D
elif len(shape) == 3 and (...):
    # process 3D
else:
    raise ValueError(...)
```

would be more elegant.

tools/accuracy_checker/accuracy_checker/postprocessor/resize_segmentation_mask.py
```diff
@@ -79,3 +80,52 @@ def _(entry, height, width):
...
+    def _process_2d(self, data, shape):
+        shape = (shape[1], shape[0])  # columns show up first
+        bytedata = self._bytescale(data)
+        image = Image.frombytes('L', shape, bytedata.tostring())
+
+        return image
+
+    def _process_3d(self, data, shape):
+        # if here then 3-d array with a 3 or a 4 in the shape length.
+        # Check for 3 in datacube shape --- 'RGB' or 'YCbCr'
```
IRDonch: This comment refers to code that doesn't exist anymore.

tools/accuracy_checker/accuracy_checker/postprocessor/resize_segmentation_mask.py
```diff
@@ -79,3 +80,52 @@ def _(entry, height, width):
...
         return annotation, prediction
+
+    def to_image(self, arr):
+        data = np.asarray(arr)
+        if np.iscomplexobj(data):
+            raise ValueError("Cannot convert a complex-valued array.")
+        shape = list(data.shape)
+        if len(shape) == 2:
+            return self._process_2d(data, shape)
+        if len(shape) == 3 and ((3 in shape) or (4 in shape)):
```
IRDonch: Since we now know which dimension is the channel dimension, `((3 in shape) or (4 in shape))` should be replaced by `shape[2] in {3, 4}`.

tools/accuracy_checker/accuracy_checker/postprocessor/resize_segmentation_mask.py
```diff
@@ -79,3 +80,52 @@ def _(entry, height, width):
...
+        cscale = cmax - cmin
+        if cscale == 0:
+            cscale = 1
+
+        scale = float(255) / cscale
+        bytedata = (data - cmin) * scale
+
+        return (bytedata.clip(0, 255) + 0.5).astype(np.uint8)
+
+    def resize(self, mask, width, height):
```
IRDonch: Shouldn't this and `to_image` be private (i.e. underscore-prefixed)?

PR #445, opened 2019-09-24
Latency metrics

wizard2006:
Purpose: 
	When performing module optimization and quantization, in addtional to the accuracy, another major measurement of inference performance is latency, or how long it takes to complete a prediction. Shorter latency ensures good user experience. 
	Latency will be used in our quantizations alogorithom. We here proposal to add add the latency as a basic metrics as well as the accuracy. It can be got during the infrerence of predition, and the cost is not high.
	In the poposal patch, 2 kinds of latency metrics, minimum and average latency, are provided. Minimum is for the shortest one of all inference requests. The average one is the mean value of the request that filtered out noises, which is commom during warning up period.
	
	The patch should have no impact on the cases that latency metric is not interested.
	
	and example for latency metric setting in mobilenet_v2.yaml

        metrics:
          - name: accuracy@top1
            type: accuracy
            top_k: 1

          - name: accuracy@top5
            type: accuracy
            top_k: 5
	
          - name: latency@average_ms
            type: latency

          - name: latency@minimum_ms
            type: latency

### DIFF
--- a/tools/accuracy_checker/accuracy_checker/evaluators/quantization_model_evaluator.py
+++ b/tools/accuracy_checker/accuracy_checker/evaluators/quantization_model_evaluator.py
@@ -316,7 +316,7 @@ class ModelEvaluator:
         self._predictions = []
         self._metrics_results = []
         self._ir_latency = []
-        if self.dataset.reset()
+        if self.dataset
             self.dataset.reset()
 
     def release(self):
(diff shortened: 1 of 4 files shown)


vladimir-dudnik:
@wizard2006 have you noticed there are some merge conflicts which need to be resolved first. Seems you need to rebase your branch to the latest state of repo.

eaidova:
@wizard2006, I sent you my recommendation via email, please take them into consideration.

PR #463, opened 2019-09-27
Added simple command to build samples and updated command line args for multichannel fd

mahinlma:
updated command line args with respect to web cam configuration 
set -nc 0, when no web cams configured.

### DIFF
--- a/demos/README.md
+++ b/demos/README.md
@@ -144,7 +144,13 @@ cd build
 ```sh
 make
 ```
-
+or  
+Use this simple command which takes default build type `Release`:
+```sh
+mkdir build & cd build  
+cmake ..  
+make
+```
 For the release configuration, the demo application binaries are in `<path_to_build_directory>/intel64/Release/`;
 for the debug configuration — in `<path_to_build_directory>/intel64/Debug/`.
 
--- a/demos/README.md
+++ b/demos/README.md
@@ -147,7 +147,7 @@ make
 or  
 Use this simple command which takes default build type `Release`:
 ```sh
-mkdir build & cd build  
+mkdir build && cd build  
 cmake ..  
 make
 ```
--- a/demos/multichannel_demo/fd/README.md
+++ b/demos/multichannel_demo/fd/README.md
@@ -63,7 +63,7 @@ For example, to run the demo with the pre-trained face detection model on FPGA w
 To run the demo using two recorded video files, use the following command:
 ```sh
 ./multi-channel-face-detection-demo -m face-detection-retail-0004.xml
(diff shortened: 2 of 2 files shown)


mahinlma:
In CMakeList, for this `CMAKE_BUILD_TYPE STREQUAL "" ` condition, CMAKE_BUILD_TYPE setted as Release.
So added simple command to build samples:
cd to sample directory - 
```
mkdir build && cd build
cmake .. 
make
```

vladimir-dudnik:
Jenkins please retry a build

demos/README.md
```diff
@@ -144,7 +144,13 @@ cd build
...
 make
 ```
-
+or  
+Use this simple command which takes default build type `Release`:
+```sh
+mkdir build && cd build  
+cmake ..  
+make
+```
```
IRDonch: I don't quite understand the purpose of this addition. These are exactly the same commands as above, except without the `-DCMAKE_BUILD_TYPE=Release` argument.

demos/multichannel_demo/fd/README.md
```diff
@@ -100,12 +100,13 @@ You can use `-i` option to connect all the three web cameras:
 
 Alternatively, you can just set `-nc 3`, which simplifies application usage.
 
-If your cameras are connected to PC with indexes gap (for example, `0,1,3`), use the `-i` parameter.
+If your cameras are connected to PC with indexes gap (for example, `0,1,2`), use the `-i` parameter.
```
IRDonch: This is an incorrect change - the sequence 0,1,2 doesn't have a gap in it.

demos/multichannel_demo/fd/README.md
```diff
@@ -100,12 +100,13 @@ You can use `-i` option to connect all the three web cameras:
...
 
 IP сameras support:
 ```
 -i rtsp://camera_address_1/ rtsp://camera_address_2/
 ```
+ note: set `-nc 0`, when web cam not configured/connected.
```
IRDonch: But `-nc` already defaults to zero. Why set it explicitly?

demos/README.md
mahinlma: for this CMAKE_BUILD_TYPE STREQUAL "" condition, CMAKE_BUILD_TYPE setted as Release In CMakeList. so we don't require to add `-DCMAKE_BUILD_TYPE=Release` explicitly for release Configuration

IRDonch: True, but does that warrant creating a duplicate set of instructions? It's not that difficult to add `-DCMAKE_BUILD_TYPE=Release`, even if it is redundant.

PR #464, opened 2019-09-30
Conversion to ONNX  is unified

eizamaliev:
Implementation of [this](https://github.com/opencv/open_model_zoo/pull/392#discussion_r329005443) idea.

### DIFF
--- a/tools/downloader/common.py
+++ b/tools/downloader/common.py
@@ -292,7 +292,7 @@ Postproc.types['unpack_archive'] = PostprocUnpackArchive
 
 class Model:
     def __init__(self, name, subdirectory, files, postprocessing, mo_args, framework,
-            description, license_url, precisions, task_type, pytorch_to_onnx_args):
+                 description, license_url, precisions, task_type, conversion_to_onnx_args):
         self.name = name
         self.subdirectory = subdirectory
         self.files = files
@@ -303,7 +303,7 @@ class Model:
         self.license_url = license_url
         self.precisions = precisions
         self.task_type = task_type
-        self.pytorch_to_onnx_args = pytorch_to_onnx_args
+        self.conversion_to_onnx_args = conversion_to_onnx_args
 
     @classmethod
     def deserialize(cls, model, name, subdirectory):
@@ -328,10 +328,10 @@ class Model:
                 with deserialization_context('"postprocessing" #{}'.format(i)):
                     postprocessing.append(Postproc.deserialize(postproc))
 
-            pytorch_to_onnx_args = None
-            if model.get('pytorch_to_onnx', None):
-                pytorch_to_onnx_args = [validate_string('"pytorch_to_onnx" #{}'.format(i), arg)
-                                        for i, arg in enumerate(model['pytorch_to_onnx'])]
+            conversion_to_onnx_args = None
+            if model.get('conversion_to_onnx_args', None):
+                conversion_to_onnx_args = [validate_string('"conversion_to_onnx_args" #{}'.format(i), arg)
+                                           for i, arg in enumerate(model['conversion_to_onnx_args'])]
 
             framework = validate_string_enum('"framework"', model['framework'], KNOWN_FRAMEWORKS)
 
@@ -372,7 +372,7 @@ class Model:
             task_type = validate_string_enum('"task_type"', model['task_type'], KNOWN_TASK_TYPES)
 
             return cls(name, subdirectory, files, postprocessing, mo_args, framework,
-                description, license_url, precisions, task_type, pytorch_to_onnx_args)
+                description, license_url, precisions, task_type, conversion_to_onnx_args)
 
 def load_models(args):
     models = []
--- a/tools/downloader/common.py
+++ b/tools/downloader/common.py
@@ -30,6 +30,7 @@ DOWNLOAD_TIMEOUT = 5 * 60
 
 # make sure to update the documentation if you modify these
 KNOWN_FRAMEWORKS = {'caffe', 'dldt', 'mxnet', 'pytorch', 'tf'}
+KNOWN_ONNX_SUPPORT = {'pytorch'}
 KNOWN_PRECISIONS = {'FP16', 'FP32', 'INT1', 'INT8'}
 KNOWN_TASK_TYPES = {
     'action_recognition',
@@ -328,12 +329,14 @@ class Model:
                 with deserialization_context('"postprocessing" #{}'.format(i)):
                     postprocessing.append(Postproc.deserialize(postproc))
 
+            framework = validate_string_enum('"framework"', model['framework'], KNOWN_FRAMEWORKS)
+
             conversion_to_onnx_args = None
             if model.get('conversion_to_onnx_args', None):
+                if framework not in KNOWN_ONNX_SUPPORT:
+                    raise DeserializationError('Conversion to ONNX not supported for "{}" framework!'.format(framework))
                 conversion_to_onnx_args = [validate_string('"conversion_to_onnx_args" #{}'.format(i), arg)
                                            for i, arg in enumerate(model['conversion_to_onnx_args'])]
-
-            framework = validate_string_enum('"framework"', model['framework'], KNOWN_FRAMEWORKS)
 
             if 'model_optimizer_args' in model:
                 mo_args = [validate_string('"model_optimizer_args" #{}'.format(i), arg)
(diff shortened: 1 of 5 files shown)


tools/downloader/converter.py
```diff
@@ -56,19 +56,25 @@ def prefixed_subprocess(prefix, args):
...
-                                                            dl_dir=args.download_dir / model.subdirectory)
-                            for arg in model.pytorch_to_onnx_args]
-    cmd = [str(args.python), str(pytorch_converter), *pytorch_to_onnx_args]
+    prefixed_printf(stdout_prefix, '========= {}Converting {} to ONNX from {}',
+                    '(DRY RUN) ' if args.dry_run else '', model.name, model.framework)
+
+    conversion_to_onnx_args = [string.Template(arg).substitute(conv_dir=output_dir / model.subdirectory,
+                                                               dl_dir=args.download_dir / model.subdirectory)
+                               for arg in model.conversion_to_onnx_args]
+    if model.framework=='pytorch':
```
jkamelin: May be it would be better to add some variable like `common.KNOWN_FRAMEWORKS` for frameworks which support ONNX conversion and use it here.

IRDonch: You would still have to pick a converter depending on the framework, so it wouldn't help much.

eaidova: you can use dictionary for selection, where key is framework and function for call converter realisation like:
```python
onnx_converters = {
'pytorch': pytorch_to_onnx,
'caffe2': caffe2_to_onnx
}
converter = onnx_converters.get(model.framework)
if converter is None:
  # handle error of unknown converter
```
}

tools/downloader/converter.py
```diff
@@ -56,19 +56,28 @@ def prefixed_subprocess(prefix, args):
...
+    prefixed_printf(stdout_prefix, '========= {}Converting {} to ONNX from {}',
+                    '(DRY RUN) ' if args.dry_run else '', model.name, model.framework)
+
+    conversion_to_onnx_args = [string.Template(arg).substitute(conv_dir=output_dir / model.subdirectory,
+                                                               dl_dir=args.download_dir / model.subdirectory)
+                               for arg in model.conversion_to_onnx_args]
+    converter = converters.get(model.framework)
+    if converter is None:
+        prefixed_printf(stdout_prefix, 'Conversion to ONNX not supported for {} framework', model.framework)
+        return False
```
IRDonch: If we get here, it means that the model config is invalid. I think we should catch this error earlier; specifically, in `Model.deserialize`.

tools/downloader/converter.py
```diff
@@ -56,19 +56,28 @@ def prefixed_subprocess(prefix, args):
...
-    pytorch_converter = Path(__file__).absolute().parent / 'pytorch_to_onnx.py'
-    prefixed_printf(stdout_prefix, '========= {}Converting {} to ONNX',
-        '(DRY RUN) ' if args.dry_run else '', model.name)
-
-    pytorch_to_onnx_args = [string.Template(arg).substitute(conv_dir=output_dir / model.subdirectory,
-                                                            dl_dir=args.download_dir / model.subdirectory)
-                            for arg in model.pytorch_to_onnx_args]
-    cmd = [str(args.python), str(pytorch_converter), *pytorch_to_onnx_args]
+    converters = {
+        'pytorch': Path(__file__).absolute().parent / 'pytorch_to_onnx.py'
```
IRDonch: Given that we already know from #392 that we'll need multiple arguments for conversion from Caffe2, I think you should preemptively support that here (by making the values in this dictionary lists of strings).

Also, please add a trailing comma.

tools/downloader/converter.py
```diff
@@ -56,19 +56,28 @@ def prefixed_subprocess(prefix, args):
...
+    }
+    prefixed_printf(stdout_prefix, '========= {}Converting {} to ONNX from {}',
```
IRDonch: IMO, the "from" part shouldn't be added for a couple of reasons:

1) It's inconsistent with the "Converting to IR" message.
2) It's going to look a bit weird, because "ONNX" is a human-readable framework name, while `model.framework` is an identifier.

tools/downloader/converter.py
eaidova: it seems to me, that all conversion scripts will be located in the same directory.
Possibly it makes since to store only related path here (`pytorch_to_onnx.py`) and add prefix (`Path(__file__).absolute().parent /`) only in time cmd args preparation

eizamaliev: That is not true - all conversions will be executed by running actual script, e.g `script.py <params>`. This will standardize conversion parameters list.

tools/downloader/converter.py
eizamaliev: Ok, but should we specify original framework in log someway?

IRDonch: I don't think it's necessary. It'll be implicitly identified by the conversion command, anyway.

tools/downloader/converter.py
eizamaliev: Good point, thanks!

tools/downloader/converter.py
IRDonch: > That is not true - all conversions will be executed by running actual script, e.g script.py <params>.

But in the case of Caffe2, the "script" is actually `-c <python command>`. That's two arguments.

jkamelin: Since we are unifying conversion, I think some conversion parameters should be the same e.g. `input-names`, `input-shape`, while others are framework-specific. This means that we need to have some `script.py` for Caffe2 to parse arguments, to pass them to conversion module.

IRDonch: OK, I'm going to drop this, because I don't quite understand your plan and we can always change it later if need be.

I am going to insist on the trailing comma, though. :-)

tools/downloader/common.py
```diff
@@ -328,13 +329,15 @@ def deserialize(cls, model, name, subdirectory):
...
-            if model.get('pytorch_to_onnx', None):
-                pytorch_to_onnx_args = [validate_string('"pytorch_to_onnx" #{}'.format(i), arg)
-                                        for i, arg in enumerate(model['pytorch_to_onnx'])]
-
             framework = validate_string_enum('"framework"', model['framework'], KNOWN_FRAMEWORKS)
 
+            conversion_to_onnx_args = None
+            if model.get('conversion_to_onnx_args', None):
+                if framework not in KNOWN_ONNX_SUPPORT:
+                    raise DeserializationError('Conversion to ONNX not supported for "{}" framework!'.format(framework))
```
IRDonch: Please remove the exclamation mark, it goes against the error message convention.

tools/downloader/converter.py
```diff
@@ -56,19 +56,26 @@ def prefixed_subprocess(prefix, args):
...
-        '(DRY RUN) ' if args.dry_run else '', model.name)
-
-    pytorch_to_onnx_args = [string.Template(arg).substitute(conv_dir=output_dir / model.subdirectory,
-                                                            dl_dir=args.download_dir / model.subdirectory)
-                            for arg in model.pytorch_to_onnx_args]
-    cmd = [str(args.python), str(pytorch_converter), *pytorch_to_onnx_args]
+    script_dir = Path(__file__).absolute().parent
+    converters = {
+        'pytorch': 'pytorch_to_onnx.py',
+    }
```
IRDonch: I don't think we need a separate dictionary with converters. You can replace `KNOWN_ONNX_SUPPORT` with `ONNX_CONVERTERS`, which can serve both as a set of supported frameworks and a converter dictionary.

tools/downloader/converter.py
```diff
@@ -56,19 +56,26 @@ def prefixed_subprocess(prefix, args):
...
+    prefixed_printf(stdout_prefix, '========= {}Converting {} to ONNX ',
```
IRDonch: Please remove the trailing space from the message.

tools/downloader/common.py
```diff
@@ -30,6 +30,7 @@
 
 # make sure to update the documentation if you modify these
 KNOWN_FRAMEWORKS = {'caffe', 'dldt', 'mxnet', 'pytorch', 'tf'}
+KNOWN_ONNX_SUPPORT = {'pytorch'}
```
IRDonch: The comment above doesn't apply to this variable, so please move it out of this block.

eizamaliev: Disagree. If you add new framework, which is supported via conversion to ONNX, you also need to update documentation, for example contribution guide (in future).

PR #466, opened 2019-09-30
pedestrian_tracker: remove VideoCapture wrapper

### LINKED ISSUE #401: Error while using demo scripts
Hi,
I downloaded .xml and .bin files for pedestrian_tracker_demo.
When I tried the scripts i got the following error.
`./pedestrian_tracker_demo -i 0 \ -m_det person-detection-retail-0013.xml \ -m_reid person-reidentification-retail-0031.xml \ -d_det CPU
InferenceEngine: 
	API version ............ 1.6
	Build .................. custom_releases/2019/R1.1_28dfbfdd28954c4dfd2f94403dd8dfc1f411038b
[ INFO ] Parsing input parameters
[ ERROR ] Parameter -m_det is not set
` 

Can anyone help me with this..

Thanks in advance.

Wovchena:
Closes #401


Wovchena:
CC @LeonidBeynenson

demos/pedestrian_tracker_demo/main.cpp
```diff
@@ -158,35 +152,39 @@ int main_work(int argc, char **argv) {
...
-        int frame_idx = pair.second;
-
-        if (frame.empty()) break;
-
-        PT_CHECK(frame_idx >= first_frame);
-
-        if ( (last_frame >= 0) && (frame_idx > last_frame) ) {
-            std::cout << "Frame " << frame_idx << " is greater than last_frame = "
-                << last_frame << " -- break";
+    for (uint32_t frame_idx = FLAGS_first; 0 == FLAGS_last || frame_idx <= FLAGS_last; ++frame_idx) {
```
IRDonch: Considering that 0 is a valid frame index, I think it would be better if only negative `FLAGS_last` values meant "process all frames".

Wovchena: I think we should keep `FLAGS_first` and `FLAGS_last` of the same type just for consistency and clarity. So I can 
1) revert both flags back to `int32`

or

2. reformulate `FLAGS_last` as `The frame index of video sequence to stop at.` And `FLAGS_first >= FLAGS_last` will mean process nothing (instead of `FLAGS_first > FLAGS_last`), and `FLAGS_last=0` makes the demo stop immediately. 

What do you think?

IRDonch: I think option 1 makes the most sense.

Wovchena: updated

PR #480, opened 2019-10-03
Bump the version for the next patch release

IRDonch:


IRDonch:
Note that this is going to the `release` branch (which I just created).

kchechil:
@vladimir-dudnik could you please take a look?

PR #482, opened 2019-10-03
[CMAKE] Align demos CMakeLists.txt according to latest DLDT structure

vmatveic:
Additionally, support of IE_MAIN_SOURCE_DIR  was cleaned up from the demos
because they can only be built externally.

### DIFF
--- a/demos/CMakeLists.txt
+++ b/demos/CMakeLists.txt
@@ -22,31 +22,18 @@ if (NOT(BIN_FOLDER))
     set (BIN_FOLDER ${ARCH})
 endif()
 
-if (NOT(IE_MAIN_SOURCE_DIR))
-    # in case if samples are built out of IE repo
-    set (IE_MAIN_SAMPLES_DIR ${CMAKE_CURRENT_BINARY_DIR})
-else()
-    # in case if samples are built from IE repo
-    set (IE_MAIN_SAMPLES_DIR ${IE_MAIN_SOURCE_DIR})
-endif()
-
 if(NOT(UNIX))
-    set (CMAKE_LIBRARY_OUTPUT_DIRECTORY ${IE_MAIN_SAMPLES_DIR}/${BIN_FOLDER})
-    set (CMAKE_LIBRARY_PATH ${IE_MAIN_SAMPLES_DIR}/${BIN_FOLDER})
-    set (CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${IE_MAIN_SAMPLES_DIR}/${BIN_FOLDER})
-    set (CMAKE_COMPILE_PDB_OUTPUT_DIRECTORY ${IE_MAIN_SAMPLES_DIR}/${BIN_FOLDER})
-    set (CMAKE_PDB_OUTPUT_DIRECTORY ${IE_MAIN_SAMPLES_DIR}/${BIN_FOLDER})
-    set (CMAKE_RUNTIME_OUTPUT_DIRECTORY ${IE_MAIN_SAMPLES_DIR}/${BIN_FOLDER})
-    set (LIBRARY_OUTPUT_DIRECTORY ${IE_MAIN_SAMPLES_DIR}/${BIN_FOLDER})
-    set (LIBRARY_OUTPUT_PATH ${LIBRARY_OUTPUT_DIRECTORY}) # compatibility issue: linux uses LIBRARY_OUTPUT_PATH, windows uses LIBRARY_OUTPUT_DIRECTORY
+    set (CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/${BIN_FOLDER})
+    set (CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/${BIN_FOLDER})
+    set (CMAKE_COMPILE_PDB_OUTPUT_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/${BIN_FOLDER})
+    set (CMAKE_PDB_OUTPUT_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/${BIN_FOLDER})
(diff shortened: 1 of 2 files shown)


demos/CMakeLists.txt
```diff
@@ -63,7 +50,7 @@ if (WIN32)
         set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /WX") #treating warnings as errors
     endif ()
 
-    if (${CMAKE_CXX_COMPILER_ID} STREQUAL MSVC)
+    if (CMAKE_CXX_COMPILER_ID MATCHES MSVC)
```
IRDonch: Why use `MATCHES` here? The ID should be exactly `MSVC`.

demos/CMakeLists.txt
```diff
@@ -178,15 +169,12 @@ macro(ie_add_sample)
...
-    if(TARGET IE::ie_cpu_extension)
-        target_link_libraries(${IE_SAMPLE_NAME} PRIVATE IE::ie_cpu_extension)
-    endif()
-
     target_link_libraries(${IE_SAMPLE_NAME} PRIVATE ${OpenCV_LIBRARIES} ${InferenceEngine_LIBRARIES}
                                                     ${IE_SAMPLE_DEPENDENCIES} gflags)
 
     if(UNIX)
-        target_link_libraries(${IE_SAMPLE_NAME} PRIVATE pthread)
+        find_package(Threads REQUIRED)
```
IRDonch: The `find_package` call ought to be done once per CMake run, so it should be moved outside of the macro.

demos/CMakeLists.txt
```diff
@@ -178,15 +169,12 @@ macro(ie_add_sample)
...
     target_include_directories(${IE_SAMPLE_NAME} PRIVATE "${CMAKE_CURRENT_SOURCE_DIR}/../common")
 
-    if(TARGET IE::ie_cpu_extension)
-        target_link_libraries(${IE_SAMPLE_NAME} PRIVATE IE::ie_cpu_extension)
-    endif()
-
     target_link_libraries(${IE_SAMPLE_NAME} PRIVATE ${OpenCV_LIBRARIES} ${InferenceEngine_LIBRARIES}
                                                     ${IE_SAMPLE_DEPENDENCIES} gflags)
 
     if(UNIX)
```
IRDonch: If you're going to use the `Threads` module, then this check is no longer necessary.

vmatveic:
@IRDonch , thanks for review. I'm adding @ilya-lavrenov to comment. Ilya, should we update it accordingly in both repos?

demos/CMakeLists.txt
```diff
@@ -150,15 +144,9 @@ macro(ie_add_sample)
...
             message(WARNING "OPENCV is disabled or not found, " ${IE_SAMPLE_NAME} " skipped")
             return()
-        else()
-            add_definitions(-DUSE_OPENCV)
         endif()
     endif()
 
-    if(TARGET IE::ie_cpu_extension)
-        add_definitions(-DWITH_EXTENSIONS)
-    endif()
```
IRDonch: It's too early to remove this, as our CI currently uses IE from OpenVINO 2019 R3. It can be removed after the next release (or at least after the next CI update).

demos/CMakeLists.txt
vmatveic: @ilya-lavrenov , could you please clarify why you used `MATCHES` here? According to the https://cmake.org/cmake/help/v3.0/command/if.html link the MATCHES keyword is used to check regexp validity. But `STREQUAL` is used to verify lexicographical equality. Do you use it here intentionally?

demos/CMakeLists.txt
vmatveic: Agree, will fix

demos/CMakeLists.txt
vmatveic: will discuss and fix

demos/CMakeLists.txt
vmatveic: When do you plan next CI update? If it's planned for a later time then I will roll back it. Please note that this PR should be merged asap because otherwise you may get the build failed (if OpenCV_DIR is not specified explicitly). In last DLDT solution it's implemented that OpenCV_DIR is propagated by `find_package(InferenceEngine)` call to use exactly the same OpenCV both for the DLDT and Demos build.

demos/CMakeLists.txt
IRDonch: I think I know why, actually. `MSVC` is also the name of a CMake variable, so `CMAKE_CXX_COMPILER_ID STREQUAL MSVC` compares the values of the `CMAKE_CXX_COMPILER_ID` and `MSVC` variables, rather than comparing the value of `CMAKE_CXX_COMPILER_ID` to the string "MSVC".

There are better ways to fix this, though:

1. Enable policy `CMP0054` and use `if(CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")`, or...
2. Just use `if(MSVC)` instead.

demos/CMakeLists.txt
IRDonch: > When do you plan next CI update?

I haven't planned it yet. I usually do it when there's a cause, e.g. a new model that can't be converted by the old version of MO.

Either way, I don't think we need to rush this particular change. I specifically requested this code to be written in this way so that it would work regardless of whether the CPU extension library exists.

> Please note that this PR should be merged asap because otherwise you may get the build failed (if OpenCV_DIR is not specified explicitly).

In our CI we _do_ specify `OpenCV_DIR` explicitly, though.

demos/README.md
```diff
@@ -218,8 +225,8 @@ run the `setupvars` script to set all necessary environment variables:
 ```sh
 <INSTALL_DIR>\bin\setupvars.bat
 ```
-If you use your own Inference Engine and OpenCV binaries to build the demos please make sure you have added
-to the `PATH` environment variable.
+If you use your own Inference Engine build with disabled RPATH please make sure you have added
```
IRDonch: This change doesn't make sense. There is no RPATH on Windows.

vladimir-dudnik:
Jenkins please retry a build

IRDonch:
@vmatveic Ping? Are you going to update this PR?

IRDonch:
Closing due to inactivity. Feel free to reopen once if you update your branch.

PR #484, opened 2019-10-03
demos/tests/cases.py: avoid an out-of-bounds access in smart_classroom_demo

IRDonch:
This demo assumes by default that the action recognition model recognizes 3 actions, but person-detection-raisinghand-recognition-0001 only recognizes 2. This causes the demo to go out of bounds when retrieving the action probabilities.

We could fix it by adding a `-student_ac` option to the test case, but a better fix is to use the top-k mode (via the `-a_top` option) for this model, since:

a) in top-k mode the demo already defaults to 2 actions; and
b) this gives us better code coverage.

In top-k mode the other 3 models aren't used, so don't add cases with them.

### DIFF
--- a/demos/tests/cases.py
+++ b/demos/tests/cases.py
@@ -187,12 +187,16 @@ NATIVE_DEMOS = [
             '-i': ImagePatternArg('smart-classroom-demo'),
             '-m_fd': ModelArg('face-detection-adas-0001')}),
         device_cases('-d_act', '-d_fd', '-d_lm', '-d_reid'),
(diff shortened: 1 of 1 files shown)


IRDonch:
Even though test cases aren't included in releases, I'm targeting this to `release` so that if we do submit any other demo fixes to `release`, CI builds won't fail. I'll merge it to `develop` after it's merged to `release`.

IRDonch:
> Should the images be updated to trigger Tracker usage?

That would certainly be useful, but it is of secondary importance. Right now the goal is to fix CI.

PR #497, opened 2019-10-09
[Downloader] Add lost reporter to try_update_cache

zmaslova:


IRDonch:
Thanks. Could you retarget this to `release`? I think it's important (and non-intrusive) enough to include in the next minor release.

PR #511, opened 2019-10-11
tools/downloader/converter.py: serialize output of concurrent jobs

IRDonch:
Currently, when we are in multithreaded mode, we prefix each output line with the name of the thread to disambiguate it. This makes the output unambiguous, but also very ugly and difficult to read.

So remove that logic, and instead make it so that the output of the converter in multithreaded mode is exactly like it is in single-threaded mode. Do it by buffering each job's output in an queue, and emptying each such queue into stdout in turn.

The downside of this approach is that we have to keep in memory the output of all jobs whose turn hasn't come yet, and that output can potentially be unbounded. However, it shouldn't be a problem in practice, since the amount of output produced by the conversion tools we run (and the converter itself) is fairly small. It's probably also less efficient due to the inter-thread communication, however, I failed to find any actual performance degradation in my experiments.

You might wonder if there's any point in keeping the single-threaded mode if the multi-threaded mode now produces the exact same output. The reason I'm keeping it is that the single-threaded mode still works a bit better due to not redirecting the subprocesses' stdout and stderr:

* the subprocess's stderr output will go to the converter's stderr, while in multithreaded mode it will go to the converter's stdout;

* if the converter's stdout is a terminal, so will be the subprocess's stdout, and so the subprocess will be able to use terminal features, like color output (although this is only hypothetical for now, since none of the tools we run have any terminal-specific logic).

### DIFF
--- a/tools/downloader/converter.py
+++ b/tools/downloader/converter.py
@@ -18,6 +18,7 @@ import argparse
 import concurrent.futures
 import os
 import platform
+import queue
 import re
 import shlex
 import string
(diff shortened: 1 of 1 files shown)


IRDonch:
CC @maksimsazhin

tools/downloader/converter.py
```diff
@@ -29,6 +29,13 @@
 
 import common
 
+class JobContext:
+    def printf(self, format, *args, flush=False):
+        raise NotImplementedError
+
+    def subprocess(self, args):
+        raise NotImplementedError
+
```
eizamaliev: Group all `JobContext`-like classes together

tools/downloader/converter.py
```diff
@@ -40,35 +47,77 @@ def quote_windows(arg):
...
+    def subprocess(self, args):
         return subprocess.run(args).returncode == 0
 
-    with subprocess.Popen(args, stdout=subprocess.PIPE, stderr=subprocess.STDOUT,
-            universal_newlines=True) as p:
-        for line in p.stdout:
-            sys.stdout.write(prefix + ': ' + line)
-        return p.wait() == 0
 
+class BufferingRunner(JobContext):
```
eizamaliev: May be `MultiThreadedContext` more suitable? Or another name, mentioning about multi-threading?

tools/downloader/converter.py
IRDonch: Done.

tools/downloader/converter.py
IRDonch: I ended up renaming both concrete context classes to more descriptive names.

IRDonch:
I uploaded a new version, in which I simplified the thread synchronization by using `queue.Queue`.

PR #519, opened 2019-10-16
Fix error thrown from SetBatch in VPUs

aungthetnaing:
VPU calls base implementation which is to throw error for SetBatch.

https://github.com/opencv/open_model_zoo/issues/505

Signed-off-by: Aung Naing <aung.t.naing@intel.com>

### DIFF
--- a/demos/pedestrian_tracker_demo/main.cpp
+++ b/demos/pedestrian_tracker_demo/main.cpp
@@ -50,7 +50,18 @@ CreatePedestrianTracker(const std::string& reid_model,
 
     if (!reid_model.empty() && !reid_weights.empty()) {
         CnnConfig reid_config(reid_model, reid_weights);
-        reid_config.max_batch_size = 16;
+        reid_config.max_batch_size = 16;   // defaulting to 16
+
+        try {
+            if (ie.GetConfig(deviceName, CONFIG_KEY(DYN_BATCH_ENABLED)).as<std::string>() != PluginConfigParams::YES) {
+                reid_config.max_batch_size = 1;
+                std::cerr << "Dynamic batch is not supported for " << deviceName << ". Fall back to batch 1." << std::endl;
+            }
+        }
+        catch(const InferenceEngine::details::InferenceEngineException& e) {
+            reid_config.max_batch_size = 1;
+            std::cerr << e.what() << " for " << deviceName << ". Fall back to batch 1." << std::endl;
+        }
 
         std::shared_ptr<IImageDescriptor> descriptor_strong =
             std::make_shared<DescriptorIE>(reid_config, ie, deviceName);
--- a/demos/pedestrian_tracker_demo/src/cnn.cpp
+++ b/demos/pedestrian_tracker_demo/src/cnn.cpp
@@ -77,7 +77,9 @@ void CnnBase::InferBatch(
             matU8ToBlob<uint8_t>(frames[batch_i + b], input_blob_, b);
         }
 
-        infer_request_.SetBatch(current_batch_size);
+        if (1 != batch_size) {
+            infer_request_.SetBatch(current_batch_size);
+        }
         infer_request_.Infer();
 
         fetch_results(outputs_, current_batch_size);


vladimir-dudnik:
Jenkins please retry a build

aungthetnaing:
jtestme

demos/pedestrian_tracker_demo/src/cnn.cpp
```diff
@@ -77,7 +77,9 @@ void CnnBase::InferBatch(
             matU8ToBlob<uint8_t>(frames[batch_i + b], input_blob_, b);
         }
 
-        infer_request_.SetBatch(current_batch_size);
+        if ((deviceName_.find("MYRIAD") == std::string::npos) && (deviceName_.find("HDDL") == std::string::npos)) {
+             infer_request_.SetBatch(current_batch_size);    // std::cout << "No dynamic batch support for " << deviceName_ << ". Falling back to batch 1" << std::endl << std::endl;
```
Wovchena: I suggest removing that `// std::cout << "No dynamic batch support for " << deviceName_ << ". Falling back to batch 1" << std::endl << std::endl;`: put it after `reid_config.max_batch_size = 1;` and slightly clarify the message:

`std::cerr << "Dynamic batch is not supported for " << deviceName_ << ". Fall back to batch 1\n";` no second std::endl

aungthetnaing: The issue is that it will print many times and could pollute log.

aungthetnaing: Used lambda to print once.

Wovchena: Why don't you want to put this print after `reid_config.max_batch_size = 1;` as I said in the first comment? That would eliminate the need in your trick and warn a user in the beginning, but not at the moment when reid is actually triggered (which can happen at a random moment of the video).

Also it seems there is a clearer way to check in `main()` if dynamic batch supported by checking `ie.GetConfig(device, CONFIG_KEY(DYN_BATCH_ENABLED))`. https://github.com/opencv/open_model_zoo/blob/9ef91c63f968208b7a85caaed39249c93ae3050b/demos/smart_classroom_demo/main.cpp#L424

And as for `CnnDLSDKBase::InferBatch`'s check for device, it probably can be replaced by just a check that `input` blob `batch_size` matches `current_batch_size`. And if not, call `infer_request_.SetBatch(current_batch_size);`.

aungthetnaing: Thanks for cleaner way (future proof way when VPU started supporting Batching). I'm not very clear on second point.I need the condition just because calling SetBatch throws exception.

demos/pedestrian_tracker_demo/src/cnn.cpp
```diff
@@ -77,7 +77,9 @@ void CnnBase::InferBatch(
             matU8ToBlob<uint8_t>(frames[batch_i + b], input_blob_, b);
         }
 
-        infer_request_.SetBatch(current_batch_size);
+        if (ie_.GetConfig(deviceName_, CONFIG_KEY(DYN_BATCH_ENABLED)).as<std::string>() == PluginConfigParams::YES) {
```
Wovchena:
```suggestion
        if (1 != batch_size) {
```
`GetConfig` will throw exceptions, so it doesn't really differ from calling `infer_request_.SetBatch(current_batch_size);` unconditionally and catching exceptions. But exception catch can lead to a performance degradation. That's why I suggest ` if (1 != batch_size)`

demos/pedestrian_tracker_demo/src/cnn.cpp
```diff
@@ -77,7 +77,9 @@ void CnnBase::InferBatch(
...
+             infer_request_.SetBatch(current_batch_size);
+	}
```
Wovchena: Misaligned bracket

demos/pedestrian_tracker_demo/src/cnn.cpp
Wovchena: `Core::GetConfig` can throw an exception so you need not only check the value, but also catch `InferenceEngine::details::InferenceEngineException`.

As for the second point, I realized now, it won't work. There is slightly another solution, see my suggestion.

vladimir-dudnik:
Jenkins please retry a build

demos/pedestrian_tracker_demo/main.cpp
```diff
@@ -50,7 +50,11 @@ CreatePedestrianTracker(const std::string& reid_model,
 
     if (!reid_model.empty() && !reid_weights.empty()) {
         CnnConfig reid_config(reid_model, reid_weights);
-        reid_config.max_batch_size = 16;
+        reid_config.max_batch_size = 16;   // defaulting to 16
+        if ((deviceName.find("MYRIAD") != std::string::npos) || (deviceName.find("HDDL") != std::string::npos)) {
```
Wovchena: I think here you should use that approach with `GetConfig`, it just needs to add catching of `InferenceEngine::details::InferenceEngineException`.

vladimir-dudnik:
Jenkins please retry a build

demos/pedestrian_tracker_demo/main.cpp
aungthetnaing: Changed to try.. catch..

vladimir-dudnik:
Jenkins please retry a build

demos/pedestrian_tracker_demo/main.cpp
Wovchena: This block of code is misaligned now. The rest looks fine.

vladimir-dudnik:
Jenkins please retry a build

PR #540, opened 2019-10-23
Fix annotation reading issue for bert

dliang0406:
'NoneType' object is referenced otherwise.

Issue tracked by https://github.com/opencv/open_model_zoo/issues/539.

### DIFF
--- a/tools/accuracy_checker/accuracy_checker/evaluators/quantization_model_evaluator.py
+++ b/tools/accuracy_checker/accuracy_checker/evaluators/quantization_model_evaluator.py
@@ -333,18 +333,21 @@ def create_dataset_attributes(config, tag):
     data_reader_config = dataset_config.get('reader', 'opencv_imread')
     data_source = dataset_config.get('data_source')
 
-    if isinstance(data_reader_config, str):
-        data_reader = BaseReader.provide(data_reader_config, data_source)
-    elif isinstance(data_reader_config, dict):
-        data_reader = BaseReader.provide(data_reader_config['type'], data_source, data_reader_config)
-    else:
-        raise ConfigError('reader should be dict or string')
     annotation_reader = None
     dataset_meta = {}
     metric_dispatcher = None
+
     if contains_any(dataset_config, ['annotation', 'annotation_conversion']):
         annotation_reader = Dataset(dataset_config)
         dataset_meta = annotation_reader.metadata
+
+    if isinstance(data_reader_config, str):
+        data_reader = BaseReader.provide(data_reader_config, data_source)
+    elif isinstance(data_reader_config, dict) and annotation_reader:
+        data_reader = BaseReader.provide(data_reader_config['type'], data_source, data_reader_config, annotation_reader.annotation)
+    else:
+        raise ConfigError('reader should be dict or string')
+
     dataset = DatasetWrapper(data_reader, annotation_reader)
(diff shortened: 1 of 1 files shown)


vladimir-dudnik:
Jenkins please retry a build

tools/accuracy_checker/accuracy_checker/evaluators/quantization_model_evaluator.py
```diff
@@ -333,18 +333,22 @@ def create_dataset_attributes(config, tag):
...
     dataset_meta = {}
     metric_dispatcher = None
+
     if contains_any(dataset_config, ['annotation', 'annotation_conversion']):
         annotation_reader = Dataset(dataset_config)
         dataset_meta = annotation_reader.metadata
+
+    if isinstance(data_reader_config, str):
+        data_reader = BaseReader.provide(data_reader_config, data_source)
+    elif isinstance(data_reader_config, dict) and annotation_reader:
```
eaidova: hmm.. dictionary used not only for bert on our side.
I think you need to check reader type in your case and that annotation reader provided inside `elif` statement and throw appropriate error if for this reader type annotation_reader is None

eaidova: I mean, something like that:
```python
    elif isinstance(data_reader_config, dict):
       data_reader_type = data_reader_config['type']
       if data_reader_type == `annotation_features_extractor`:
           if not annotation_reader:
               raise ConfigError('annotation required')
           data_reader = BaseReader.provide(data_reader_config['type'], data_source, data_reader_config, annotation_reader.annotation)
       else:
           data_reader = BaseReader.provide(data_reader_config['type'], data_source, data_reader_config)
             
```

eaidova: But better define list of constants in data_reader.py , it will include reader providers which require annotation and check that reader not in this list instead direct string comparison

eaidova: @dliang0406, can I suggest my version of bugfix instead of your in separate PR?

dliang0406: Hi @eaidova I think your bugfix is better since it treats "annotation_features_extractor" specifically and leave other readers untouched. But I don't get your point of  separated PR...
If I submit a new patch in this PR, you can squash the three commits together at your side, right?

eaidova: I mean that, this my patch is pretty simple for avoiding extra changes in your fix, but I want made more common approach for such situation which also requires changes in other files: `data_reader.py`, `model_evaluator.py`. It is the reason why I want submit another patch which with more general solution.

dliang0406: Hi @eaidova , as I 'm not so familiar with data_reader.py, I think I can wait for your real fix and use my patch as a WA locally.

IRDonch:
@eaidova I see that the linked issue is closed. Does that mean that this PR is no longer relevant?

eaidova:
@IRDonch, yes, I merged another fix for this problem. This Pr can be closed.

PR #543, opened 2019-10-23
AC: fix annotation feature reader

eaidova:


eaidova:
@dliang0406, this is my solution for https://github.com/opencv/open_model_zoo/issues/539
It is also handle another potential problem. When we use subset of dataset with random order, the behaviour of reader for BERT is not correct due to it reads in usual order.

dliang0406:
@eaidova your solution has been verified OK with BERT. Thanks a lot!

PR #547, opened 2019-10-23
Collection of execution time for demos added

elozgacx:
Collection of execution time for demos added

### DIFF
--- a/demos/tests/run_tests.py
+++ b/demos/tests/run_tests.py
@@ -30,6 +30,9 @@ import shutil
 import subprocess
 import sys
 import tempfile
+import re
+import timeit
+import csv
 
 from pathlib import Path
 
@@ -51,6 +54,17 @@ def parse_args():
     parser.add_argument('--mo', type=Path, metavar='MO.PY',
         help='Model Optimizer entry point script')
     return parser.parse_args()
+
+def parce_result(demo_name, device, pipline, execution_time):
+    if device == "":
+        device = "CPU"
+    csv_path = Path("demo_execution_time_report.csv")
+    first_time = not csv_path.exists()
+    with csv_path.open('a+', newline='') as csvfile:
+        testwriter = csv.writer(csvfile)
+        if first_time:
+            testwriter.writerow(["DemoName", "Device", "ModelsInPipline", "ExecutionTime"])
+        testwriter.writerow([demo_name, device, ", ".join(pipline), execution_time])
 
 def main():
     args = parse_args()
@@ -143,13 +157,29 @@ def main():
                     ' '.join(shlex.quote(str(arg)) for arg in case_args))
                 print(flush=True)
 
+                execution_time = -1
+                device = ""
+                pipline = []
+                for key , value in test_case.options.items():
+                    match = re.search(r'-d.*', key)
+                    if match and device == "":
+                        device = value
+                    match = re.search(r'-m.*', key)
+                    if match:
+                        model_path = option_to_args(key, value)[1]
+                        model = re.search(r'.*/(models/.*).xml', model_path)
+                        pipline.append(model.group(1) if model else option_to_args(key, value)[1])
+                pipline.sort()
                 try:
+                    start_time = timeit.default_timer()
                     subprocess.check_output(fixed_args + case_args,
                         stderr=subprocess.STDOUT, universal_newlines=True)
+                    execution_time = timeit.default_timer() - start_time
                 except subprocess.CalledProcessError as e:
                     print(e.output)
                     print('Exit code:', e.returncode)
                     num_failures += 1
+                parce_result(demo.full_name, device, pipline, execution_time)
 
         print()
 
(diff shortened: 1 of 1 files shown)


elozgacx:
@IRDonch could you please review.

vladimir-dudnik:
Jenkins please retry a build

demos/tests/run_tests.py
```diff
@@ -52,6 +55,17 @@ def parse_args():
         help='Model Optimizer entry point script')
     return parser.parse_args()
 
+def parce_result(demo_name, device, pipline, execution_time):
```
vladimir-dudnik: pls, fix typo: pipline should be pipeline

demos/tests/run_tests.py
```diff
@@ -143,13 +157,29 @@ def option_to_args(key, value):
                     ' '.join(shlex.quote(str(arg)) for arg in case_args))
                 print(flush=True)
 
+                execution_time = -1
+                device = ""
+                pipline = []
+                for key , value in test_case.options.items():
```
vladimir-dudnik: why comma is separated by spaces from both sides? better to write 'key, value'

demos/tests/run_tests.py
```diff
@@ -143,13 +157,29 @@ def option_to_args(key, value):
...
+                    match = re.search(r'-m.*', key)
+                    if match:
+                        model_path = option_to_args(key, value)[1]
+                        model = re.search(r'.*/(models/.*).xml', model_path)
+                        pipline.append(model.group(1) if model else option_to_args(key, value)[1])
+                pipline.sort()
                 try:
+                    start_time = timeit.default_timer()
                     subprocess.check_output(fixed_args + case_args,
                         stderr=subprocess.STDOUT, universal_newlines=True)
```
vladimir-dudnik: is there any system when '=' sign separated by spaces and when it is not?

demos/tests/run_tests.py
elozgacx: Done.

demos/tests/run_tests.py
zmaslova: `parce` -> `parse`

demos/tests/run_tests.py
vladimir-dudnik: so, @elozgacx, it would be better to add spaces around '=' for stderr=subprocess calls (there are few cases in file)

vladimir-dudnik:
Jenkins please retry a build

demos/tests/run_tests.py
```diff
@@ -143,13 +157,29 @@ def option_to_args(key, value):
                     ' '.join(shlex.quote(str(arg)) for arg in case_args))
                 print(flush=True)
 
+                execution_time = -1
+                device = ""
+                pipeline = []
+                for key, value in test_case.options.items():
+                    match = re.search(r'-d.*', key)
+                    if match and device == "":
+                        device = value
```
IRDonch: You try to get a single device for each test case, which is IMO misguided. While all current test cases do run all models on the same device, the demos support running different networks on different devices, and I can see us adding test cases that cover that possibility. So it would be more prudent to record (model, device) pairs instead.

demos/tests/run_tests.py
```diff
@@ -143,13 +157,29 @@ def option_to_args(key, value):
...
+                    match = re.search(r'-m.*', key)
+                    if match:
+                        model_path = option_to_args(key, value)[1]
+                        model = re.search(r'.*/(models/.*).xml', model_path)
+                        pipeline.append(model.group(1) if model else option_to_args(key, value)[1])
```
IRDonch: This is overly complicated. For a model option, the value should be a `ModelArg`. You can get the model name directly from its attributes.

demos/tests/run_tests.py
```diff
@@ -143,13 +157,29 @@ def option_to_args(key, value):
...
                 try:
+                    start_time = timeit.default_timer()
                     subprocess.check_output(fixed_args + case_args,
                         stderr=subprocess.STDOUT, universal_newlines=True)
+                    execution_time = timeit.default_timer() - start_time
                 except subprocess.CalledProcessError as e:
                     print(e.output)
                     print('Exit code:', e.returncode)
                     num_failures += 1
+                parse_result(demo.full_name, device, pipeline, execution_time)
```
IRDonch: This will record a nonsensical time if the demo fails to run. IMO, you shouldn't record results for unsuccessful cases.

demos/tests/run_tests.py
```diff
@@ -52,6 +55,17 @@ def parse_args():
         help='Model Optimizer entry point script')
     return parser.parse_args()
 
+def parse_result(demo_name, device, pipeline, execution_time):
+    if device == "":
+        device = "CPU"
+    csv_path = Path("demo_execution_time_report.csv")
```
IRDonch: The report file path should be settable via a command line option.

PR #548, opened 2019-10-24
AC: change getting threads number way

eaidova:


IRDonch:
What's the reason for this change?

eaidova:
@IRDonch , I can not understand real reason, but usage multithreding influence on numpy random behaviur :) as the result generated subset is different even used the same seed
Please ignore it sometime until I investigate it

IRDonch:
Curious. Do you have a reproducer for this?

eaidova:
> Curious. Do you have a reproducer for this?
It reproduced on post training tool sanity tests.

eaidova:
@IRDonch, as I say I am wrong, more likely it is internal PTCT problem

PR #549, opened 2019-10-24
Change print interval in model evaluator

LeonidBeynenson:
Add possibility to change print_interval for progress reporter in ModelEvaluator

### DIFF
--- a/tools/accuracy_checker/accuracy_checker/evaluators/quantization_model_evaluator.py
+++ b/tools/accuracy_checker/accuracy_checker/evaluators/quantization_model_evaluator.py
@@ -105,7 +105,11 @@ class ModelEvaluator:
         _set_number_infer_requests(nreq)
 
         if check_progress:
-            progress_reporter = ProgressReporter.provide('print', self.dataset.size)
+            pr_kwargs = {}
+            if isinstance(check_progress, int) and not isinstance(check_progress, bool):
+                pr_kwargs = {"print_interval": check_progress}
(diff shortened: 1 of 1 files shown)


LeonidBeynenson:
2 @eaidova : Ekaterina, could you review this small PR?   
I thought it may be useful.

vladimir-dudnik:
Jenkins please retry a build

tools/accuracy_checker/accuracy_checker/evaluators/quantization_model_evaluator.py
```diff
@@ -105,7 +105,11 @@ def _set_number_infer_requests(nreq):
         _set_number_infer_requests(nreq)
 
         if check_progress:
-            progress_reporter = ProgressReporter.provide('print', self.dataset.size)
+            pr_kwargs = {}
```
eaidova: please move progress reporter generation in the function
Possibly it also make sense to rename this parameter to `progress_interval` and made default value is 0. it helps to avoid complicated logic and multiple meaning for argument.

eaidova: But possibly no, it can requires some changes on your side...
or you can stay this bool flag as is and set additional parameter in this method kwargs for progress interval if it is necessary

eaidova:
@LeonidBeynenson , my redesigned feature was merged to current `develop` branch, so I close your PR.
Thank you for your improvement!

PR #561, opened 2019-10-28
AC: change progress interval in quantization model evaluator

eaidova:


eaidova:
@LeonidBeynenson, I made a small redesign of your solution with print interval from #549 (preserve your initial logic, without changing higher lyouts)
If you do not have any comments, I am going to merge it.

LeonidBeynenson:
2 @eaidova : Ekaterina, I agree with you that after the redesign it looks better.

PR #574, opened 2019-10-30
demos/monitors: add cpu, mem

Wovchena:


demos/common/monitors/cpu_monitor.cpp
```diff
@@ -0,0 +1,274 @@
...
+        status = PdhAddCounter(query, szCounterName, 0, &counter);
+        if (ERROR_SUCCESS != status)
+        {
+            throw std::runtime_error("PdhAddCounter() failed");
+        }
+
+        status = PdhSetCounterScaleFactor(counter, -2); // scale counter to [0, 1]
+        if (ERROR_SUCCESS != status)
+        {
+            throw std::runtime_error("PdhSetCounterScaleFactor() failed");
```
IRDonch: `system_error` would be a better fit here, as it would allow you to include the error code in the exception data.

demos/common/monitors/cpu_monitor.cpp
```diff
@@ -0,0 +1,274 @@
...
+    if (ERROR_SUCCESS != status)
+    {
+        throw std::runtime_error("PdhCollectQueryData() failed");
+    }
+}
+
+void CpuMonitor::closeQuery() {
+    PDH_STATUS status = PdhCloseQuery(query);
+    if (ERROR_SUCCESS != status) {
+        throw std::runtime_error("PdhCloseQuery() failed");
```
IRDonch: The only way we can get here is if `query` is an invalid handle, which shouldn't ever happen. So this should be a `logic_error`.

demos/common/monitors/cpu_monitor.cpp
```diff
@@ -0,0 +1,274 @@
...
+    }
+}
+
+void CpuMonitor::enableHistory(std::size_t historySize) {
+    if (0 == historySize) {
+        disableHistory();
+    }
+    else {
+        this->historySize = historySize;
+        historyEnabled = true;
```
IRDonch: Isn't `historyEnabled` basically equivalent to `historySize != 0`? Why keep it as a separate variable?

demos/common/monitors/cpu_monitor.cpp
```diff
@@ -0,0 +1,274 @@
...
+    lastEnabled = false;
+    if (!historyEnabled) {
+        closeQuery();
+    }
+}
+
+void CpuMonitor::collectData() {
+    // store CPU Utulization in range [0, 1]
+    PDH_STATUS status;
+    PDH_FMT_COUNTERVALUE DisplayValue;
```
IRDonch:
```suggestion
    PDH_FMT_COUNTERVALUE displayValue;
```

demos/common/monitors/cpu_monitor.cpp
```diff
@@ -0,0 +1,274 @@
...
+    }
+
+    std::vector<std::pair<double, double>> idleNonIdleCpuStat;
+    for (std::size_t i = 0; i < coreTimeCounters.size(); i++) {
+        status = PdhGetFormattedCounterValue(coreTimeCounters[i], PDH_FMT_DOUBLE, NULL, &DisplayValue);
+        if (ERROR_SUCCESS != status) {
+            throw std::runtime_error("PdhGetFormattedCounterValue() failed");
+        }
+
+        idleNonIdleCpuStat.emplace_back(DisplayValue.doubleValue, 1 - DisplayValue.doubleValue);
```
IRDonch: This doesn't quite make sense to me. You're filling `idleNonIdleCpuStat` with pairs, but then you only ever access the first item of those pairs. Why use pairs, then?

Also, you probably should check `DisplayValue.CStatus` before using `DisplayValue.doubleValue`.

demos/common/monitors/cpu_monitor.cpp
```diff
@@ -0,0 +1,274 @@
...
+#include <regex>
+#include <utility>
+#include <fstream>
+
+namespace {
+std::size_t getNCores() {
+    std::ifstream cpuinfo("/proc/cpuinfo");
+    return std::count(std::istream_iterator<std::string>(cpuinfo),
+        std::istream_iterator<std::string>(),
+        std::string("processor"));
```
IRDonch: Why not `sysconf(_SC_NPROCESSORS_CONF)`?

demos/common/monitors/cpu_monitor.cpp
```diff
@@ -0,0 +1,274 @@
...
+
+        status = PdhSetCounterScaleFactor(counter, -2); // scale counter to [0, 1]
+        if (ERROR_SUCCESS != status)
+        {
+            throw std::runtime_error("PdhSetCounterScaleFactor() failed");
+        }
+
+        coreTimeCounters.push_back(counter);
+    }
+    status = PdhCollectQueryData(query);
```
IRDonch: Why do this if you don't read the counters afterwards?

demos/common/monitors/cpu_monitor.cpp
```diff
@@ -0,0 +1,274 @@
...
+        std::string("processor"));
+}
+
+std::vector<std::pair<unsigned long, unsigned long>> getIdleNonIdleCpuStat(std::size_t nCores) {
+    std::vector<std::pair<unsigned long, unsigned long>> idleNonIdleCpuStat(nCores);
+    std::ifstream proc_stat("/proc/stat");
+    std::string line;
+    std::smatch match;
+    std::getline(proc_stat, line);
+    while (proc_stat.good())
```
IRDonch: A simpler way to iterate over lines is `while (std::getline(proc_stat, line))`.

demos/common/monitors/cpu_monitor.cpp
```diff
@@ -0,0 +1,274 @@
...
+    {
+        std::regex core_jiffies("^cpu(\\d+)\\s+"
```
IRDonch: The regex ought to be moved outside of the loop, to avoid recompiling it every time.

demos/common/monitors/cpu_monitor.cpp
```diff
@@ -0,0 +1,274 @@
...
+            "(\\d+)\\s+" // user
+            "(\\d+)\\s+" // nice
+            "(\\d+)\\s+" // system
+            "(\\d+)\\s+" // idle
+            "(\\d+)\\s+" // iowait
+            "(\\d+)\\s+" // irq
+            "(\\d+)\\s+" // softirq
+            "(\\d+)\\s+" // steal
+            "(\\d+)\\s+" // guest
+            "(\\d+)$");  // guest_nice
```
IRDonch: I wouldn't recommend anchoring the regex at the end, since Linux might add new columns in the future.

demos/common/monitors/cpu_monitor.cpp
```diff
@@ -0,0 +1,274 @@
...
+            idleNonIdleCpuStat[core_id].first = idleInfo;
+            idleNonIdleCpuStat[core_id].second = non_idleinfo;
+        }
+        std::getline(proc_stat, line);
+    }
+    return idleNonIdleCpuStat;
+}
+}
+
+CpuMonitor::CpuMonitor() : nCores{getNCores()}, lastEnabled{false}, historyEnabled{false}, samplesNumber{0} {};
```
IRDonch: There's a lot of duplication here with the Windows version of the code. You should move the common code out of the `#if`.

demos/common/monitors/cpu_monitor.cpp
```diff
@@ -0,0 +1,274 @@
...
+            "(\\d+)\\s+" // iowait
+            "(\\d+)\\s+" // irq
+            "(\\d+)\\s+" // softirq
+            "(\\d+)\\s+" // steal
+            "(\\d+)\\s+" // guest
+            "(\\d+)$");  // guest_nice
+        if (std::regex_match(line, match, core_jiffies))
+        {
+            unsigned long idleInfo = stoul(match[5]) + stoul(match[6]),
+                non_idleinfo = stoul(match[2])
```
IRDonch: Sudden naming convention change.

demos/common/monitors/cpu_monitor.cpp
```diff
@@ -0,0 +1,274 @@
...
+                    + stoul(match[3])
+                    + stoul(match[4])
+                    + stoul(match[7])
+                    + stoul(match[8])
+                    + stoul(match[9]), // it doesn't handle overflows and overflows of /proc/stat values
```
IRDonch: Overflows and overflows?

demos/common/monitors/cpu_monitor.cpp
```diff
@@ -0,0 +1,274 @@
...
+}
+
+std::deque<std::vector<double>> CpuMonitor::getLastHistory() const {
+    return cpuLoadHistory;
+}
+
+std::vector<double> CpuMonitor::getMeanCpuLoad() const {
+    return meanCpuLoad;
+}
+#else
```
IRDonch: The `#else` suggests that the following code will work on any non-Windows systems, but actually it's quite Linux-specific. You need to change this to `#elif __linux__` and add an actual `#else` with a stub implementation.

demos/common/monitors/cpu_monitor.cpp
```diff
@@ -0,0 +1,274 @@
...
+    for (std::size_t i = 0; i < idleNonIdleCpuStat.size(); ++i) {
+        unsigned long idleDiff = idleNonIdleCpuStat[i].first - prevIdleNonIdleCpuStat[i].first;
+        unsigned long nonIdleDiff = idleNonIdleCpuStat[i].second - prevIdleNonIdleCpuStat[i].second;
+        cpuLoad[i] = static_cast<double>(nonIdleDiff) / (idleDiff + nonIdleDiff);
+    }
+    prevIdleNonIdleCpuStat = std::move(idleNonIdleCpuStat);
+
+    if (meanCpuLoad.empty()) {
+        meanCpuLoad = cpuLoad;
+        samplesNumber = 1;
```
IRDonch: Instead of this, it would be simpler to initialize `meanCpuLoad` to a vector of zeroes.

demos/common/monitors/memory_monitor.cpp
```diff
@@ -0,0 +1,258 @@
+// Copyright (C) 2019 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+//
+
+#include "memory_monitor.h"
+
+#ifdef WIN32
+#include <algorithm>
+
+#define PSAPI_VERSION 1 // for psapi
```
IRDonch: Given that we certainly don't support versions of Windows older than 7, why not define it to 2? Then you'll no longer have to link to `psapi.dll`.

demos/common/monitors/memory_monitor.cpp
```diff
@@ -0,0 +1,258 @@
...
+    }
+    return static_cast<double>(performanceInformation.PhysicalTotal * performanceInformation.PageSize)
+        / (1024 * 1024 * 1024);
+}
+double getSwapTotal() {
+    PERFORMANCE_INFORMATION performanceInformation;
+    if (!GetPerformanceInfo(&performanceInformation, sizeof(performanceInformation))) {
+        throw std::runtime_error("GetPerformanceInfo() failed");
+    }
+    return static_cast<double>(performanceInformation.CommitLimit * performanceInformation.PageSize)
```
IRDonch: This is questionable. The commit limit is not the size of the swap file; it is the amount of memory that can be committed, which comprises both the size of the page file and the size of physical memory (minus the amount of memory that can't be paged out).

demos/common/monitors/memory_monitor.cpp
```diff
@@ -0,0 +1,258 @@
...
+            } else if ("SwapFree" == match[1]) {
+                availableSwap = stod(match[2]) / (1024 * 1024);
+            }
+        }
+        std::getline(meminfo, line);
+    }
+    if (0 == availableSwap) {
+        throw std::runtime_error("Can't get available swap memory");
+    }
+    availableMem = static_cast<double>(memfree + buffers + cached + sReclaimable - shmem) / (1024 * 1024);
```
IRDonch: Why not just use `MemAvailable`?

demos/common/monitors/memory_monitor.h
```diff
@@ -0,0 +1,32 @@
...
+    void enable(std::size_t  historySize);
+    void disable();
+    void collectData();
+    std::deque<std::pair<double, double>> getLastHistory() const;
+    double getMeanMem() const;
+    double getMeanSwap() const;
+    double getMaxMem() const;
+    double getMaxSwap() const;
+
+    const double memTotal, swapTotal; // in GiB
```
IRDonch: I don't think these can be treated as constants, as the former can change if the system has hotpluggable memory, and the latter can change if swap files/partitions are added/removed/resized.

demos/common/monitors/memory_monitor.cpp
```diff
@@ -0,0 +1,258 @@
...
+    double usedSwap = static_cast<double>(performanceInformation.CommitTotal * performanceInformation.PageSize)
+        / (1024 * 1024 * 1024);
+
+    if (0 == samplesNumber) {
+        maxMem = meanMem = usedMem;
+        maxSwap = meanSwap = usedSwap;
+        samplesNumber = 1;
+    } else {
+        meanMem = (meanMem * samplesNumber + usedMem) / (samplesNumber + 1);
+        meanSwap = (meanSwap * samplesNumber + usedSwap) / (samplesNumber + 1);
```
IRDonch: It seems a bit pointless (and potentially precision-destroying) to keep multiplying and dividing by the number of samples. Why not keep the sum and only divide it by the number of samples when you need the mean?

demos/common/monitors/cpu_monitor.cpp
Wovchena: Docs say there are some counters that require an extra query before reporting data. But I didn't check if `\Processor(%u)\% Processor Time` is this counter.

demos/common/monitors/cpu_monitor.cpp
Wovchena: I'm not sure this is the only case (how do you know?). May be something happens and fails internally. So should it be `system_error` too?

demos/common/monitors/memory_monitor.cpp
Wovchena: It follows `htop`'s implementation. I'm also not sure this is a good option because of comment in `man`:
> An estimate of how much memory is available for starting new applications, without swapping.

Wovchena:
> One thing I noticed with both the CPU and memory monitors is that they collect the overall system stats. Is that really what we want? Why not the stats of the current process?

The first reason is that one could build a complex pipeline out of the demo, like a `gstreamer` process preparing data and transferring it to a device which a demo reads.
The second one is that collecting CPU load for a single process will rise a question like why we can't  utilize more

demos/common/monitors/cpu_monitor.cpp
Wovchena: `historySize == 1` can mean that history is enabled with length 1, but it also can mean that only last value is enabled.

demos/common/monitors/cpu_monitor.cpp
Wovchena: This will us to now, that `/proc/stat` is different.

demos/common/monitors/cpu_monitor.cpp
IRDonch: > how do you know?

There's a list (of one entry) of possible errors in the [documentation](https://docs.microsoft.com/en-us/windows/win32/api/pdh/nf-pdh-pdhclosequery).

> May be something happens and fails internally.

If something fails that shouldn't have, that _is_ a logic error. :-)

demos/common/monitors/cpu_monitor.cpp
IRDonch: What's the difference?

demos/common/monitors/cpu_monitor.cpp
IRDonch: Could you show me where it says that?

demos/common/monitors/cpu_monitor.cpp
IRDonch: Why would we want to know?

Wovchena: To update regex.

demos/common/monitors/cpu_monitor.cpp
Wovchena: https://docs.microsoft.com/en-us/windows/win32/perfctrs/collecting-performance-data

This page also shows that there are zeroes for `\Processor(%u)\% Processor Time` for the first call, so this counter requires two calls.

demos/common/monitors/memory_monitor.cpp
IRDonch: > It follows htop's implementation.

Are you sure that implementation was not written before [MemAvailable was implemented](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=34e431b0ae398fc54ea69ff85ec700722c9da773)?

> > An estimate of how much memory is available for starting new applications, without swapping.

Isn't that exactly what we want?

demos/common/monitors/cpu_monitor.cpp
Wovchena: Having `historySize == 1` (which is != 0) still can mean that history is disabled.

demos/common/monitors/cpu_monitor.cpp
Wovchena: OK

demos/common/monitors/memory_monitor.cpp
Wovchena: > Are you sure that implementation was not written before MemAvailable was implemented?

Looks that later: https://github.com/hishamhm/htop/commit/26422af6088129b99864b75ed4ce5395aa578f2b#diff-6a85ca50e4ed143060c0849894d89370R564, but it could be done for compatibility.

> Isn't that exactly what we want?

This note about "starting new applications" can mean that available memory for acquiring by existing applications and available memory for starting new applications aren't the same numbers. I can't say which we want for sure.

demos/common/monitors/cpu_monitor.h
```diff
@@ -0,0 +1,45 @@
+// Copyright (C) 2019 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+//
+
+#pragma once
+
+#include <iostream> // TODO remove
+#include <deque>
+#include <vector>
+#ifdef WIN32
```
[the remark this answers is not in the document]
Wovchena: replaced

demos/common/monitors/cpu_monitor.cpp
Wovchena: replaced

demos/common/monitors/cpu_monitor.cpp
Wovchena: updated

demos/common/monitors/cpu_monitor.cpp
Wovchena: moved

demos/common/monitors/cpu_monitor.cpp
Wovchena: updated

demos/common/monitors/cpu_monitor.cpp
Wovchena: updated

demos/common/monitors/cpu_monitor.cpp
Wovchena: updated

demos/common/monitors/memory_monitor.h
Wovchena: updated

demos/common/monitors/memory_monitor.cpp
Wovchena: updated

demos/common/monitors/cpu_monitor.cpp
IRDonch: Yeah, but what is the actual difference in behavior between history being enabled and disabled when `historySize == 1`?

demos/common/monitors/memory_monitor.cpp
Wovchena: Why minus the amount of memory that can't be paged out?
Should it be
```
usedSwap = performanceInformation.CommitTotal - (performanceInformation.PhysicalTotal - performanceInformation.PhysicalAvailable)
```
? And it looks I don't really need `swapTotal`

demos/common/monitors/cpu_monitor.cpp
Wovchena: There is no reason to draw empty graph if `historySize == 1`, so `historySize < 2` can be treated as disabled history. Updated

demos/common/monitors/cpu_monitor.cpp
IRDonch: There are also a few other instances of such loops.

demos/common/monitors/cpu_monitor.cpp
IRDonch: That's true, I guess, but the code will be broken until we do. I think it would be better to assume that any additional fields won't affect our calculations, and just ignore them.

demos/common/monitors/memory_monitor.cpp
IRDonch: > Why minus the amount of memory that can't be paged out?

Probably not the best wording on my part. What I mean is that some physical memory is reserved by the OS and isn't included in the commit limit (or the commit charge).

> usedSwap = performanceInformation.CommitTotal - (performanceInformation.PhysicalTotal - performanceInformation.PhysicalAvailable)

This seems close, but it's not quite correct. Imagine that a file is mapped into memory for writing and some changes have been made to it but not flushed to disk yet. The pages corresponding to this file will be counted in `PhysicalTotal`, but not in `PhysicalAvailable` or `CommitTotal`, which means they will affect `usedSwap`, even though they actually have no effect on swap. Same thing for the reserved pages I mentioned above.

Wovchena: > The pages corresponding to this file will be counted in `PhysicalTotal`, but not in `PhysicalAvailable` or `CommitTotal`

https://answers.microsoft.com/en-us/windows/forum/windows_10-performance/physical-and-virtual-memory-in-windows-10/e36fb5bc-9ac8-49af-951c-e7d39b979938# claims:
> Committed (Virtual In Use) = Physical Memory in Use + Page file in Use

If the memory is counted in (PhysicalTotal - PhysicalAvailable) than it is counted in `CommitTotal`.

Reserved memory is also described there:
> Total Physical Memory = Installed - Hardware Reserved

So in memory_monitor's computations it is just not accounted anywhere.

demos/common/monitors/memory_monitor.cpp
IRDonch: > Looks that later:

The code you've shown me has actually [existed since before htop switched to Git](https://github.com/hishamhm/htop/commit/d6231bab89d634da5564491196b7c478db038505#diff-78718b9cfe8260b8487ef93a88226e44R490). Although I don't see where the available memory calculation that we are discussing is in this code.

> This note about "starting new applications" can mean that available memory for acquiring by existing applications and available memory for starting new applications aren't the same numbers.

Well... even if they aren't the same (which I find doubtful - why wouldn't they be?), isn't it still better to use the estimate explicitly provided by the kernel, rather than a formula of questionable origins?

Wovchena: Verifying the numbers by running `htop` will give (almost) the same results. I suppose it is more convenient way of verifying and `htop` is the first thing that comes in mind. At least it was the first thing I wanted to check values with when I had an initial version.

To me `MemAvailable` gives different results (diff is about 0.4GiB).

IRDonch: What are you comparing the numbers with? This meter?

изображение

Because that's calculated as [`used` - `buffers` - `cached`](https://github.com/hishamhm/htop/blob/04cc193e3c0c39ea47eb01d61a6866b32d70baea/linux/Platform.c#L203), which is not what you do either.

> I suppose it is more convenient way of verifying and htop is the first thing that comes in mind. At least it was the first thing I wanted to check values with when I had an initial version.

That's not much of an argument. I could tell you that the first thing that comes to my mind is `free` and that uses `MemAvailable`. :-)

Wovchena: I am comparing to the number reported. It is 4.4G for you.

[The trick is how you get `cachedMem`](https://github.com/hishamhm/htop/blob/04cc193e3c0c39ea47eb01d61a6866b32d70baea/linux/LinuxProcessList.c#L960). `usedMem - buffersMem - cachedMem = usedMem - buffersMem - cachedMem(procfs) - sreclaimable + shmem`

> That's not much of an argument. I could tell you that the first thing that comes to my mind is free and that uses MemAvailable. :-)

OK than. I will update.

demos/common/monitors/cpu_monitor.cpp
```diff
@@ -0,0 +1,289 @@
...
+        if (std::regex_match(line, match, coreJiffies))
+        {
+            unsigned long idleInfo = stoul(match[5]) + stoul(match[6]),
+                nonIdleInfo = stoul(match[2])
+                    + stoul(match[3])
+                    + stoul(match[4])
+                    + stoul(match[7])
+                    + stoul(match[8])
+                    + stoul(match[9]), // it doesn't handle overflow of sum and overflows of /proc/stat values
+                core_id = stoul(match[1]);
```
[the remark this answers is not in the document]
Wovchena: updated

demos/common/monitors/memory_monitor.cpp
Wovchena: updated

demos/common/monitors/cpu_monitor.cpp
Wovchena: updated

demos/common/monitors/cpu_monitor.cpp
Wovchena: updated

demos/common/monitors/cpu_monitor.cpp
```diff
@@ -0,0 +1,289 @@
...
+#include <regex>
+#include <utility>
+#include <fstream>
+
+#include <unistd.h>
+
+namespace {
+std::vector<std::pair<unsigned long, unsigned long>> getIdleNonIdleCpuStat(std::size_t nCores) {
+    std::vector<std::pair<unsigned long, unsigned long>> idleNonIdleCpuStat(nCores);
+    std::ifstream proc_stat("/proc/stat");
```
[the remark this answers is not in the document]
Wovchena: updated

demos/common/monitors/cpu_monitor.cpp
```diff
@@ -0,0 +1,287 @@
...
+}
+
+void CpuMonitor::disableHistory() {
+    historySize = 1;
+    if (!lastEnabled) {
+        closeQuery();
+    }
+}
+
+void CpuMonitor::enableLast() {
```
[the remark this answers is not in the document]
Wovchena: updated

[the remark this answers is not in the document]
Wovchena: updated

demos/common/monitors/cpu_monitor.cpp
```diff
@@ -0,0 +1,274 @@
...
+    PDH_STATUS status;
+
+    status = PdhOpenQuery(NULL, NULL, &query);
+    if (ERROR_SUCCESS != status) {
+        throw std::runtime_error("PdhOpenQuery() failed");
+    }
+
+    for (std::size_t i = 0; i < nCores; ++i)
+    {
+        TCHAR szCounterName[MAX_PATH];
```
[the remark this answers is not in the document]
Wovchena: removed

demos/common/monitors/cpu_monitor.cpp
```diff
@@ -0,0 +1,274 @@
...
+        _stprintf_s(szCounterName, sizeof(szCounterName), TEXT("\\Processor(%u)\\%% Processor Time"), i);
```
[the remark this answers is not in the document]
Wovchena: replaced

demos/common/monitors/cpu_monitor.cpp
```diff
@@ -0,0 +1,274 @@
...
+
+        PDH_HCOUNTER counter;
+
+        status = PdhAddCounter(query, szCounterName, 0, &counter);
+        if (ERROR_SUCCESS != status)
+        {
+            throw std::runtime_error("PdhAddCounter() failed");
```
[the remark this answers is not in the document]
Wovchena: updated

demos/common/monitors/cpu_monitor.cpp
Wovchena: updated

demos/common/monitors/cpu_monitor.cpp
Wovchena: updated

demos/common/monitors/cpu_monitor.cpp
Wovchena: updated

demos/common/monitors/memory_monitor.cpp
IRDonch: > If the memory is counted in (PhysicalTotal - PhysicalAvailable) than it is counted in CommitTotal.

Unfortunately, that is not true, because not all physical pages that are in use correspond to memory that is committed. I have actually written a program that creates the scenario I described above (with the unflushed changes to a mapped file) and by running that program, I got the memory stats to look like this:

```
CommitTotal: 6768248 K
CommitLimit: 10010272 K
PhysicalTotal: 8218664 K
PhysicalAvailable: 720016 K
```

… which, if you were to apply your formula, would yield negative swap usage. :-)

> Total Physical Memory = Installed - Hardware Reserved

"Hardware reserved" is not the same thing as "OS reserved". The former is regions of memory that are inaccessible because some device's registers are mapped there. The latter is regions of memory that contain critical code/data that must not ever be paged out. However, I added up the size of the page file and `PhysicalTotal` on my machine, and the result was exactly equal to `CommitLimit`, which suggests that OS-reserved RAM is included in the commit charge. So I was wrong on this point.

IRDonch: BTW, if you want to repeat my experiments, here are the programs I used: https://gist.github.com/IRDonch/789ffef15945f08ec60696772f84d0cd

demos/common/monitors/cpu_monitor.cpp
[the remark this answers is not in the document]
IRDonch: You have added a wrapper class, but you haven't really solved the problem: it's still possible for the query to remain allocated if `CpuMonitor::openQuery` fails.

demos/common/monitors/cpu_monitor.cpp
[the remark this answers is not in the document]
IRDonch: You have replaced it with an `std::string`, but given that Windows is going to convert it to Unicode internally anyway, I think you should use `std::wstring` to start with.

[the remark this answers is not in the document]
Wovchena: [`PdhAddCounter()` accepts `LPCSTR`](https://docs.microsoft.com/en-us/windows/win32/api/pdh/nf-pdh-pdhaddcountera) which is [`typedef const char* LPCSTR;`](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-dtyp/f8d4fe46-6be8-44c9-8823-615a21d17a61).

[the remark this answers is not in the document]
IRDonch: PdhAddCounter**A** accepts `LPCSTR`. [PdhAddCounter**W**](https://docs.microsoft.com/en-us/windows/win32/api/pdh/nf-pdh-pdhaddcounterw) accepts `LPCWSTR`.

demos/common/monitors/memory_monitor.cpp
Wovchena: Do you know a way to get size of used part of pageFile from C++? I can't find any.

How does it:
> I added up the size of the page file and PhysicalTotal on my machine, and the result was exactly equal to CommitLimit

proves that
>OS-reserved RAM is included in the commit charge

?

IRDonch: > Do you know a way to get size of used part of pageFile from C++?

Not exactly, but in Performance Monitor there is a counter that shows the percentage of used page file space (I don't remember the exact name; I think it was under "Page file"). It's probably accessible through the PDH API.

> How does it: [...] proves that [...]?

The OS-reserved areas of RAM are included in `PhysicalTotal`, which is easy to check with [RAMMap](https://docs.microsoft.com/en-us/sysinternals/downloads/rammap) (for example, it shows that non-paged pool (which is one of these areas) is included in the total). Since the page file size plus `PhysicalTotal` exactly equals the commit limit, it means that the entirety of `PhysicalTotal` is included in the commit limit, including the OS-reserved areas. If the OS-reserved areas were not included in the commit charge, then the commit charge could never reach the commit limit, which would be silly.

Wovchena: Thank you!
It looks that [PDH doesn't provide Physical Memory: Total](https://docs.microsoft.com/en-us/previous-versions/windows/desktop/legacy/aa965225(v%3Dvs.85)#system-memory-performance-information), so in order to get used Paging File size I will need both PDH and `GetPerformanceInfo()` assuming that system state doesn't change significantly between collectiong info from these two sources.

demos/common/monitors/cpu_monitor.cpp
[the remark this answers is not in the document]
Wovchena: Will, kind of updated.
I think it is better to close query in case of errors manually, because now I open/close extra query to override previous extra query :)

demos/common/monitors/cpu_monitor.cpp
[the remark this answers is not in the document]
Wovchena: updated

demos/common/monitors/cpu_monitor.cpp
Wovchena: updated

demos/common/monitors/memory_monitor.cpp
Wovchena: redefined

demos/common/monitors/memory_monitor.cpp
```diff
@@ -0,0 +1,271 @@
...
+    {
+        if (std::regex_match(line, match, memRegex))
+        {
+            if ("SwapTotal" == match[1]) {
+                swapTotal = stod(match[2]) / (1024 * 1024);
+            }
+        }
+    }
+    return swapTotal;
+}
```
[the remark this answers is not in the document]
Wovchena: updated

demos/common/monitors/cpu_monitor.cpp
```diff
@@ -0,0 +1,243 @@
...
+        unsigned long idleDiff = idleNonIdleCpuStat[i].first - prevIdleNonIdleCpuStat[i].first;
+        unsigned long nonIdleDiff = idleNonIdleCpuStat[i].second - prevIdleNonIdleCpuStat[i].second;
+        if (0 == idleDiff + nonIdleDiff) {
+            if (cpuLoadHistory.empty()) {
+                cpuLoad[i] = 0;
+            } else {
+                cpuLoad[i] = cpuLoadHistory.back()[i];
+            }
+        } else {
+            cpuLoad[i] = static_cast<double>(nonIdleDiff) / (idleDiff + nonIdleDiff);
```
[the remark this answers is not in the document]
Wovchena: Nice idea. Updated

demos/common/monitors/cpu_monitor.cpp
```diff
@@ -0,0 +1,235 @@
...
+
+    PDH_FMT_COUNTERVALUE displayValue;
+    std::vector<double> cpuLoad(coreTimeCounters.size());
+    for (std::size_t i = 0; i < coreTimeCounters.size(); i++) {
+        status = PdhGetFormattedCounterValue(coreTimeCounters[i], PDH_FMT_DOUBLE, NULL, &displayValue);
+        if (ERROR_SUCCESS != status) {
+            throw std::system_error(status, std::system_category(), "PdhGetFormattedCounterValue() failed");
+        }
+        if (PDH_CSTATUS_VALID_DATA != displayValue.CStatus && PDH_CSTATUS_NEW_DATA != displayValue.CStatus) {
+            throw std::system_error(status, std::system_category(), "Error in counter data");
```
IRDonch: If you're here, `status` is guaranteed to be `ERROR_SUCCESS`, so it doesn't make sense to pass it to the exception. While notionally you could use `displayValue.CStatus` instead, I don't think the CStatus values are in the same code space as Windows error codes, so it wouldn't work correctly. I think in this case you should just throw a `runtime_error`.

demos/common/monitors/cpu_monitor.cpp
```diff
@@ -0,0 +1,235 @@
...
+    std::vector<unsigned long> idleCpuStat(nCores);
+    std::ifstream procStat("/proc/stat");
+    std::string line;
+    std::smatch match;
+    std::regex coreJiffies("^cpu(\\d+)\\s+"
+        "(\\d+)\\s+"
+        "(\\d+)\\s+"
+        "(\\d+)\\s+"
+        "(\\d+)\\s+" // idle
+        "(\\d+).*"); // iowait
```
IRDonch: `.*` at the end is redundant.

Wovchena: Regex doesn't match without it.

demos/common/monitors/presenter.cpp
```diff
@@ -0,0 +1,319 @@
...
+        addRemoveMonitor(monitor);
+    }
+}
+
+Presenter::Presenter(const std::string& keys, int yPos, cv::Size graphSize, std::size_t historySize) :
+    Presenter{strKeysToMonitorSet(keys), yPos, graphSize, historySize} {}
+
+void Presenter::addRemoveMonitor(MonitorType monitor) {
+    int sampleStep = std::max(1, static_cast<int>(graphSize.width / historySize));
+    unsigned updatedHistorySize = (graphSize.width + sampleStep - 1) / sampleStep; // round up
```
IRDonch: Could it be possible for `updatedHistorySize` to be set to 1 here? I think that would break the update logic below.

demos/common/monitors/presenter.cpp
```diff
@@ -0,0 +1,319 @@
...
+    }
+}
+
+void Presenter::addRemoveMonitor(int key) {
+    switch(std::toupper(key)) {
+        case 'C': addRemoveMonitor(MonitorType::CpuAverage);
+            break;
+        case 'D': addRemoveMonitor(MonitorType::DistributionCpu);
+            break;
+        case 'M': addRemoveMonitor(MonitorType::Memory);
```
IRDonch: There is some duplication of logic with `strKeysToMonitorSet` here. I think you should create a map from key to monitor type and reuse it in both places.

demos/common/monitors/presenter.h
```diff
@@ -0,0 +1,45 @@
...
+        cv::Size graphSize = {150, 60},
+        std::size_t historySize = 20);
+    explicit Presenter(const std::string& keys,
+        int yPos = 20,
+        cv::Size graphSize = {150, 60},
+        std::size_t historySize = 20);
+    void addRemoveMonitor(MonitorType monitor);
+    void addRemoveMonitor(int key); // handles c, d, m, h keys
+    void drawGraphs(cv::Mat& frame);
+    std::map<MonitorType, std::vector<double>> getMeans() const;
```
IRDonch: Is there a reason to expose the actual mean values to the user? The only time you use this function, you pass the result directly to `meansToOstream`, so you could simplify both the interface and the implementation by merging `getMeans` and `meansToOstream` together.

demos/common/monitors/presenter.cpp
```diff
@@ -0,0 +1,319 @@
...
+            cpuMonitor.collectData();
+        }
+        if (memoryMonitor.getHistorySize() > 1) {
+            memoryMonitor.collectData();
+        }
+    }
+
+    int numberOfEnabledMonitors = (cpuMonitor.getHistorySize() > 1) + distributionCpuEnabled
+        + (memoryMonitor.getHistorySize() > 1);
+    int pannelLength = graphSize.width * numberOfEnabledMonitors
```
IRDonch:
```suggestion
    int panelWidth = graphSize.width * numberOfEnabledMonitors
```

demos/common/monitors/presenter.cpp
```diff
@@ -0,0 +1,319 @@
...
+                rectXPos += step;
+            }
+            sum /= lastHistory.back().size();
+            int yLine = graph.rows - static_cast<int>(graphRectHeight * sum);
+            cv::line(graph, cv::Point{0, yLine}, cv::Point{graph.cols, yLine}, {0, 255, 0}, 2);
+        }
+        cv::Rect border{cv::Point{graphPos, yPos + textGraphSplittingLine},
+            cv::Size{graphSize.width, graphSize.height - textGraphSplittingLine}};
+        cv::rectangle(frame, border, {0, 0, 0});
+        strStream.str("Core distr");
```
IRDonch: Distr?

demos/common/monitors/presenter.cpp
```diff
@@ -0,0 +1,319 @@
...
+    if (memoryMonitor.getHistorySize() > 1 && --numberOfEnabledMonitors >= 0) {
+        std::deque<std::pair<double, double>> lastHistory = memoryMonitor.getLastHistory();
+        cv::Mat graph = frame(cv::Rect{cv::Point{graphPos, yPos}, graphSize} & cv::Rect(0, 0, frame.cols, frame.rows));
+        graph *= 1.3;
+        int histxPos = graph.cols - 1;
+        double range = std::min(memoryMonitor.getMaxMemTotal() + memoryMonitor.getMaxSwap(),
+            (memoryMonitor.getMaxMem() + memoryMonitor.getMaxSwap()) * 1.2);
+        if (lastHistory.size() > 1) {
+            for (auto memUsageIt = lastHistory.rbegin(); memUsageIt != lastHistory.rend() - 1; ++memUsageIt) {
+                constexpr double SWAP_TRESHOLD = 10.0 / 1024; // 10 MiB
```
IRDonch:
```suggestion
                constexpr double SWAP_THRESHOLD = 10.0 / 1024; // 10 MiB
```

demos/common/monitors/cpu_monitor.cpp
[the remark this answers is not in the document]
IRDonch: Your current implementation is not correct, because if you call `closeQuery` and then destroy the object, it will attempt to close the handle again.

I suggest you make use of `unique_ptr` for this:

```c++
class CpuMonitor {
    // [...]
    std::unique_ptr<QueryWrapper> query;
};

void CpuMonitor::openQuery() {
    std::unique_ptr<QueryWrapper> newQuery(new QueryWrapper);
    // [... configure counters ...]
    query = std::move(newQuery);
}

void CpuMonitor::closeQuery() {
    coreTimeCounters.clear();
    query.reset();
}
```

As a bonus, this will allow you to move the definition of `QueryWrapper` out of the header file.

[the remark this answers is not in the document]
Wovchena: updated.
Moved `QueryWrapper` to a separate files cos both monitors now use it.

demos/common/monitors/memory_monitor.cpp
Wovchena: Updated

demos/common/monitors/cpu_monitor.cpp
Wovchena: updated

demos/common/monitors/presenter.h
```diff
@@ -0,0 +1,45 @@
...
+    explicit Presenter(std::set<MonitorType> enabledMonitors = {},
+        int yPos = 20,
+        cv::Size graphSize = {150, 60},
+        std::size_t historySize = 20);
+    explicit Presenter(const std::string& keys,
+        int yPos = 20,
+        cv::Size graphSize = {150, 60},
+        std::size_t historySize = 20);
+    void addRemoveMonitor(MonitorType monitor);
+    void addRemoveMonitor(int key); // handles c, d, m, h keys
```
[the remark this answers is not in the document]
Wovchena: I think this is the only way presenter is going to handle a key, which allows to clarify in function name how the key is handled.

demos/common/monitors/presenter.cpp
Wovchena: It stays for "load distribution over cores", but this is too long string.

demos/common/monitors/presenter.h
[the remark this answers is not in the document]
Wovchena: I think this is the only way presenter is going to handle a key, which allows to clarify in function name how the key is handled.

demos/common/monitors/cpu_monitor.cpp
Wovchena: ok, updated

demos/common/monitors/presenter.cpp
```diff
@@ -0,0 +1,319 @@
...
+    const auto memoryMean = means.find(MonitorType::Memory);
+        if (means.end() != memoryMean) {
+            if (memoryMean->second.empty()) {
+                tmpStream << "No data collected for memory ans swap usage\n";
+            } else {
+                assert (2 == memoryMean->second.size());
+                tmpStream << "Mean memory usage: " << std::fixed << std::setprecision(1) << memoryMean->second.front()
+                    << "GiB\n";
+                tmpStream << "Mean swap usage: " << std::fixed << std::setprecision(1) << memoryMean->second.back()
+                    << "GiB\n";
```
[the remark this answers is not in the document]
Wovchena: added

demos/common/monitors/presenter.cpp
Wovchena: updated

demos/common/monitors/presenter.cpp
Wovchena: updated

demos/common/monitors/presenter.h
Wovchena: updated

demos/common/monitors/cpu_monitor.cpp
Wovchena: moved

demos/common/monitors/cpu_monitor.cpp
Wovchena: added

demos/common/monitors/presenter.h
```diff
@@ -0,0 +1,47 @@
...
+    explicit Presenter(const std::string& keys,
+        int yPos = 20,
+        cv::Size graphSize = {150, 60},
+        std::size_t historySize = 20);
+    operator std::string() const;
+    void addRemoveMonitor(MonitorType monitor);
+    void addRemoveMonitor(int key); // handles c, d, m, h keys
+    void drawGraphs(cv::Mat& frame);
+
+    friend std::ostream& operator<<(std::ostream& os, const Presenter& presenter);
```
IRDonch: I don't think using the output operator is appropriate here. You're not printing the presenter, you're printing the metrics.

demos/common/monitors/presenter.cpp
IRDonch: Maybe call it "Core load", then?

demos/common/monitors/presenter.h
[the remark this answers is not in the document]
IRDonch: > I think this is the only way presenter is going to handle a key

It is not the only way already, since the presenter handles the H key by toggling _multiple_ monitors.

demos/common/monitors/cpu_monitor.cpp
IRDonch: There is more common code than that. :-) Parts of the constructor, `setHistorySize` and `collectData` are all duplicated too.

IMO, what you should do here is expand the `PerformanceCounter` class. Move everything that is OS-dependent into there, then the `CpuMonitor` class itself can become entirely OS-independent.

Wovchena: I think further emphasizing of common parts should be avoided because it introduces a number of additional switches between Windows and Linux and I find them annoying and hard for understanding.
Why should I move everything that is OS-dependent into `PerformanceCounter`? Most of `CpuMonitor`'s logic is already OS specific. Moving it into a separate class will introduce another level of nesting because `PerformanceCounter` is supposed to live only for period when the data needs to be collected and `openQuery()` manages this object creation, so I can't just move everything into `PerformanceCounter`.

demos/common/monitors/presenter.h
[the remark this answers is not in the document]
Wovchena: renamed

demos/common/monitors/presenter.cpp
Wovchena: Updated

demos/common/monitors/query_wrapper.h
```diff
@@ -0,0 +1,14 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+
+#pragma once
+
+#include <pdh.h>
+class QueryWrapper {
+public:
+    QueryWrapper();
+    ~QueryWrapper();
```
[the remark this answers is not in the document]
Wovchena: updated

demos/common/monitors/query_wrapper.h
```diff
@@ -0,0 +1,14 @@
...
+    operator PDH_HQUERY() const;
+    PDH_HQUERY query;
```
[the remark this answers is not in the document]
Wovchena: updated

demos/common/monitors/memory_monitor.cpp
```diff
@@ -0,0 +1,223 @@
...
+MemoryMonitor::~MemoryMonitor() = default;
+
+void MemoryMonitor::openQuery() {
+    std::unique_ptr<MemoryMonitor::PerformanceCounter> newPerformanceCounter{new MemoryMonitor::PerformanceCounter};
+
+    PDH_STATUS status = PdhAddCounterW(newPerformanceCounter->query, L"\\Paging File(_Total)\\% Usage", 0,
+        &newPerformanceCounter->pagingFileUsageCounter);
+    if (ERROR_SUCCESS != status)
+    {
+        throw std::system_error(status, std::system_category(), "PdhSetCounterScaleFactor() failed");
```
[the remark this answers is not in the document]
Wovchena: updated

demos/common/monitors/presenter.h
Wovchena: updated

demos/common/monitors/cpu_monitor.cpp
IRDonch: > I think further emphasizing of common parts should be avoided because it introduces a number of additional switches between Windows and Linux 

Does it? In my mind it actually _reduces_ the number of switches. Here's what I imagine it would look like:

* `CpuMonitor::openQuery` is gone (merged into the `PerformanceCounter` constructor).
* `CpuMonitor::closeQuery` is gone (unneeded).
* `CpuMonitor::CpuMonitor` is the same, except the `nCores` member initializer is gone.
* `CpuMonitor::setHistorySize` looks like this:

```c++
    if (0 == historySize && 0 != size) {
        performanceCounter.reset(new PerformanceCounter);
    } else if (0 != historySize && 0 == size) {
        performanceCounter.reset();
    }
    historySize = size;
    std::size_t newSize = std::min(size, cpuLoadHistory.size());
    cpuLoadHistory.erase(cpuLoadHistory.begin(), cpuLoadHistory.end() - newSize);
```

* `CpuMonitor::collectData` looks like this:

```c++
    auto cpuLoad = performanceCounter->query();

    for (std::size_t i = 0; i < cpuLoad.size(); ++i) {
        cpuLoadSum[i] += cpuLoad[i];
    }
    ++samplesNumber;

    cpuLoadHistory.push_back(std::move(cpuLoad));
    if (cpuLoadHistory.size() > historySize) {
        cpuLoadHistory.pop_front();
    }
```

This makes it so that the only `#if` required is around the `PerformanceCounter` class. As a bonus, it also nicely separates the collection of metrics from their aggregation.

demos/crossroad_camera_demo/crossroad_camera_demo.hpp
```diff
@@ -64,6 +64,9 @@ static const char no_show_processed_video[] = "Optional. No show processed video
 /// @brief message resizable input flag
 static const char input_resizable_message[] = "Optional. Enables resizable input with support of ROI crop & auto resize.";
 
+/// @brief Message list of monitors to show
+static const char utilization_monitors_message[] = "Optional. List of monitors to show.";
```
IRDonch:
```suggestion
static const char utilization_monitors_message[] = "Optional. List of monitors to show initially.";
```

(presumably, you can still disable them using the keys)

demos/crossroad_camera_demo/main.cpp
```diff
@@ -817,6 +823,8 @@ int main(int argc, char *argv[]) {
                 const int key = cv::waitKey(isVideo ? 1 : 0);
                 if (27 == key)  // Esc
                     break;
+                else
```
IRDonch: The `else` is redundant.

demos/interactive_face_detection_demo/main.cpp
```diff
@@ -211,10 +212,15 @@ int main(int argc, char *argv[]) {
 
         std::cout << "To close the application, press 'CTRL+C' here";
         if (!FLAGS_no_show) {
-            std::cout << " or switch to the output window and press any key";
+            std::cout << " or switch to the output window and press Q, q or Esc";
```
IRDonch:
```suggestion
            std::cout << " or switch to the output window and press Q or Esc";
```

Q and q are the same key. I don't think it makes sense to differentiate between them.

demos/object_detection_demo_ssd_async/main.cpp
```diff
@@ -375,8 +382,10 @@ int main(int argc, char *argv[]) {
             if (27 == key)  // Esc
                 break;
             if (9 == key) {  // Tab
-                isAsyncMode ^= true;
+                isAsyncMode = true;
```
IRDonch: Doesn't this mean that async mode can no longer be reset?

demos/tests/cases.py
```diff
@@ -18,6 +18,7 @@
 from args import *
 from image_sequences import IMAGE_SEQUENCES
 
+MONITORS = 'cdm'
```
IRDonch: I think you can go further: do `MONITORS = {'-u': 'cdm'}` and merge it into the test cases with `**MONITORS` .

demos/common/monitors/cpu_monitor.cpp
Wovchena: Updated.

demos/crossroad_camera_demo/crossroad_camera_demo.hpp
Wovchena: Updated.

demos/crossroad_camera_demo/main.cpp
Wovchena: Updated.

demos/interactive_face_detection_demo/main.cpp
Wovchena: Updated.

demos/object_detection_demo_ssd_async/main.cpp
Wovchena: Reverted

demos/tests/cases.py
Wovchena: Updated.

demos/common/monitors/cpu_monitor.h
```diff
@@ -0,0 +1,29 @@
...
+    CpuMonitor();
+    ~CpuMonitor();
+    void setHistorySize(std::size_t size);
+    std::size_t getHistorySize() const;
+    void collectData();
+    std::deque<std::vector<double>> getLastHistory() const;
+    std::vector<double> getMeanCpuLoad() const;
+
+private:
+    bool lastEnabled;
```
IRDonch: This is unused now.

demos/common/monitors/cpu_monitor.cpp
```diff
@@ -0,0 +1,197 @@
...
+public:
+    PerformanceCounter() : prevIdleCpuStat{getIdleCpuStat()}, prevTimePoint{std::chrono::steady_clock::now()} {}
+
+    std::vector<double> getCpuLoad() {
+        std::vector<double> cpuLoad(nCores, 0);
+        std::vector<unsigned long> idleCpuStat = getIdleCpuStat();
+        auto timePoint = std::chrono::steady_clock::now();
+        // don't update data too frequently which may result in negative values for cpuLoad.
+        // It may happen when collectData() is called just after setHistorySize().
+        if (timePoint - prevTimePoint > std::chrono::milliseconds{100}) {
```
IRDonch: If this is false, the function will return a vector of zeroes, which will skew the stats. I think you need to either return the previous load values, or return an empty vector and have downstream code handle it properly.

demos/common/monitors/cpu_monitor.h
Wovchena: removed

demos/common/monitors/cpu_monitor.cpp
Wovchena: updated

demos/common/monitors/presenter.cpp
```diff
@@ -0,0 +1,293 @@
...
+
+        cv::Rect border{cv::Point{graphPos, yPos + textGraphSplittingLine},
+            cv::Size{graphSize.width, graphSize.height - textGraphSplittingLine}};
+        cv::rectangle(frame, {border}, {0, 0, 0});
+        if (lastHistory.empty()) {
+            strStream.str("Memory");
+        } else {
+            strStream.str("");
+            strStream << std::fixed << std::setprecision(1) << lastHistory.back().first << " + "
+                << lastHistory.back().second << " GiB";
```
IRDonch: It would be useful to print something like "X GiB RAM + Y GiB swap". Otherwise it's impossible to understand what the numbers are.

Wovchena: The suggested string is too long for putting it on small windows. I would not change it.

IRDonch: Unfortunate. Oh well.

PR #575, opened 2019-10-30
demos: replace vectors with arrays where possible

IRDonch:
The impetus for this is that an internal static analysis tool gets confused by vectors initialized with initializer lists. However, it's a good change in its own right: by using arrays we avoid unnecessary dynamic allocation and, in some cases, dynamic initialization.

In one case, the vector isn't necessary at all (it shadows another identical vector), so remove it completely.

### DIFF
--- a/demos/common/samples/common.hpp
+++ b/demos/common/samples/common.hpp
@@ -48,6 +48,11 @@ class ConsoleErrorListener : public InferenceEngine::IErrorListener {
     }
 };
 
+template <typename T, std::size_t N>
+constexpr std::size_t arraySize(const T (&)[N]) noexcept {
+    return N;
+}
+
 /**
  * @brief Trims from both ends (in place)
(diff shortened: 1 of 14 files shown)


demos/common/samples/common.hpp
```diff
@@ -48,6 +48,11 @@ class ConsoleErrorListener : public InferenceEngine::IErrorListener {
     }
 };
 
+template <typename T, std::size_t N>
+inline std::size_t arraySize(const T (&)[N]) {
```
Wovchena: I suggest adding `constexpr` and `noexcept` like in https://en.cppreference.com/w/cpp/iterator/size. And than `inline` is no longer needed.

IRDonch: Alright, done. I initially thought the CentOS 7 version of GCC might be too old to support this, but I tested, and it isn't.

PR #578, opened 2019-10-31
demos: integrate the pose_extractor Python extension into the main build

IRDonch:
This simplifies the build procedure and should make it fairly easy to add tests for `human_pose_estimation_3d_demo`. I also fixed a few small issues in the extension's CMake scripts.


demos/CMakeLists.txt
```diff
@@ -6,6 +6,8 @@ cmake_minimum_required (VERSION 2.8.12)
 
 project(Demos)
 
+option(WITH_PYTHON "Whether to build extension modules for Python demos" OFF)
```
Wovchena: I suggest `ENABLE_PYTHON` to stay consistent with IE.

demos/python_demos/human_pose_estimation_3d_demo/README.md
```diff
@@ -23,15 +23,19 @@ The demo workflow is the following:
 > **NOTE**: By default, Open Model Zoo demos expect input with BGR channels order. If you trained your model to work with RGB order, you need to manually rearrange the default channels order in the demo application or reconvert your model using the Model Optimizer tool with `--reverse_input_channels` argument specified. For more information about the argument, refer to **When to Reverse Input Channels** section of [Converting a Model Using General Conversion Parameters](https://docs.openvinotoolkit.org/latest/_docs_MO_DG_prepare_model_convert_model_Converting_Model_General.html).
 ## Prerequisites
 
-Before running it is necessary to build `pose_extractor` module. Your system should have installed:
-* Python 3.5 (or above)
-* CMake 3.10 (or above)
-* C++ Compiler (g++ or MSVC)
-
-To build `pose_extractor` module, please run in command line:
-`python setup.py build_ext`
```
Wovchena: I'm not sure we should remove that approach. As @vladimir-dudnik said, we don't want make python users mess with C++ build tools.

demos/CMakeLists.txt
IRDonch: OK, done.

demos/python_demos/human_pose_estimation_3d_demo/README.md
IRDonch: It's a fair point, but:

1. I would really prefer not to maintain two build systems.
2. The `setup.py`-based system delegates to CMake anyway. So it's not like it's substantively different.

Wovchena: `setup.py` removes a need to manually building a command line to compile the lib (like `build_demos_msvc.bat`). Alternative solution could be adding `ENABLE_PYTHON` key to `build_demos*` scripts. I'm just trying to fix a concern that it would be inconvenient for python users to use `cmake` manually. On the other hand for now we can agree that this is a hybrid demo and you should be familiar with both python and C++/cmake to use it.

So the new question is should `build_demos*` forward `ENABLE_PYTHON` key.

IRDonch: Are you suggesting to add an option to the `build_demos*` scripts to set `ENABLE_PYTHON`?

Wovchena: Yes

IRDonch: OK, I've done this for `build_demos.sh`. I'll do it for the Windows script later.

demos/build_demos.sh
```diff
@@ -25,6 +25,20 @@ error() {
 }
 trap 'error ${LINENO}' ERR
 
+extra_cmake_opts=()
+
+for opt in "$@"; do
+    case "$opt" in
+    -DENABLE_PYTHON=*)
```
Wovchena: `-D ENABLE_PYTHON=*` is also a valid cmake key. But I don't see easy way to handle it, so it is probably going to be kept like this.

IRDonch: Yeah, I don't think we need to support the full CMake argument syntax here. It is our own script and our own option (which just happens to be a valid CMake option too :-)).

demos/python_demos/human_pose_estimation_3d_demo/README.md
IRDonch: > I'll do it for the Windows script later.

And done.

PR #579, opened 2019-10-31
Collection of execution time for demos added

elozgacx:
Collection of execution time for demos added

### DIFF
--- a/demos/tests/cases.py
+++ b/demos/tests/cases.py
@@ -18,17 +18,28 @@ import sys
 from args import *
 from image_sequences import IMAGE_SEQUENCES
 
-ALL_DEVICES = ['CPU', 'GPU']
-
 TestCase = collections.namedtuple('TestCase', ['options'])
 
-class NativeDemo:
-    def __init__(self, subdirectory, test_cases):
+class Demo:
+
+    def device_args(self, device_list):
+        device_args = {}
+        if len(self.device_keys)>0:
+            for device in device_list:
+                device_args[device] = []
+                for dev in self.device_keys:
+                    device_args[device].extend([dev, device])
+        return device_args
+
+class NativeDemo(Demo):
+    def __init__(self, subdirectory, device_keys, test_cases):
         self.subdirectory = subdirectory
 
         self._name = subdirectory.replace('/', '_')
 
         self.test_cases = test_cases
+
+        self.device_keys = device_keys
 
     @property
     def full_name(self):
@@ -40,13 +51,15 @@ class NativeDemo:
     def fixed_args(self, source_dir, build_dir):
         return [str(build_dir / self._name)]
 
-class PythonDemo:
-    def __init__(self, subdirectory, test_cases):
+class PythonDemo(Demo):
+    def __init__(self, subdirectory, device_keys, test_cases):
         self.subdirectory = 'python_demos/' + subdirectory
 
         self._name = subdirectory.replace('/', '_')
 
         self.test_cases = test_cases
+        
+        self.device_keys = device_keys
 
     @property
     def full_name(self):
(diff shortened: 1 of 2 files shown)


elozgacx:
Hello @IRDonch could you please review

vladimir-dudnik:
Jenkins please retry a build

demos/tests/cases.py
```diff
@@ -40,13 +51,15 @@ def models_lst_path(self, source_dir):
...
-    def __init__(self, subdirectory, test_cases):
+class PythonDemo(Demo):
+    def __init__(self, subdirectory, device_keys, test_cases):
         self.subdirectory = 'python_demos/' + subdirectory
 
         self._name = subdirectory.replace('/', '_')
 
         self.test_cases = test_cases
+        
+        self.device_keys = device_keys
```
IRDonch: These assignments should probably happen in the same order that the parameters are listed.

demos/tests/cases.py
```diff
@@ -73,23 +86,23 @@ def combine_cases(*args):
...
-        TestCase(options={'-no_show': None,
-            '-i': ImagePatternArg('person-vehicle-bike-detection-crossroad')}),
-        device_cases('-d', '-d_pa', '-d_reid'),
-        TestCase(options={'-m': ModelArg('person-vehicle-bike-detection-crossroad-0078')}),
-        single_option_cases('-m_pa', None, ModelArg('person-attributes-recognition-crossroad-0230')),
-        single_option_cases('-m_reid', None, ModelArg('person-reidentification-retail-0079')),
+    NativeDemo(subdirectory='crossroad_camera_demo',
+                device_keys=['-d', '-d_pa', '-d_reid'],
+                test_cases=combine_cases(
+                    TestCase(options={'-no_show': None,
```
IRDonch: Don't change the indentation of cases, please. You've done this correctly with the other demos.

demos/tests/run_tests.py
```diff
@@ -50,8 +53,26 @@ def parse_args():
         help='list of demos to run tests for (by default, every demo is tested)')
     parser.add_argument('--mo', type=Path, metavar='MO.PY',
         help='Model Optimizer entry point script')
+    parser.add_argument('--device_list', required=False, default="CPU, GPU",
```
IRDonch: This should be called `--devices` for consistency with `--demos`. It should also use the same syntax.

Also, `required=False` is redundant.

demos/tests/run_tests.py
```diff
@@ -50,8 +53,26 @@ def parse_args():
...
+        help='List of devices to test')
```
IRDonch: Lowercase help messages, please.

demos/tests/run_tests.py
```diff
@@ -50,8 +53,26 @@ def parse_args():
...
+    parser.add_argument('--report_file', type=Path, required=False, default="demo_execution_time_report.csv",
```
IRDonch:
```suggestion
    parser.add_argument('--report-file', type=Path,
```

I think the default should be to not produce a report at all. Tools shouldn't dump stuff into the current directory without being asked to.

elozgacx:
Hello @IRDonch , ready for review, could you please to take a look.

demos/tests/run_tests.py
```diff
@@ -50,8 +53,26 @@ def parse_args():
         help='list of demos to run tests for (by default, every demo is tested)')
     parser.add_argument('--mo', type=Path, metavar='MO.PY',
         help='Model Optimizer entry point script')
+    parser.add_argument('--devices', default="CPU, GPU",
```
Wovchena: I suggest using just `,` without a spcae as separator to follow other tools style, like https://github.com/opencv/open_model_zoo/blob/06c73b9662118efcd66f363760bf20898f0cb763/tools/downloader/downloader.py#L194

Wovchena: Actually no. This will make it harder to parse device in case of `MULTI:CPU,GPU`. So I'm fine with space.

IRDonch: I'm also find with spaces, but then I am not fine with commas. If you want to use spaces for separation, then just use spaces.

demos/tests/run_tests.py
```diff
@@ -133,23 +154,33 @@ def option_to_args(key, value):
...
-                print(flush=True)
-
-                try:
-                    subprocess.check_output(fixed_args + case_args,
-                        stderr=subprocess.STDOUT, universal_newlines=True)
-                except subprocess.CalledProcessError as e:
-                    print(e.output)
-                    print('Exit code:', e.returncode)
-                    num_failures += 1
+                pipeline = [option_to_args(key, value)[1] for key, value in test_case.options.items() if "-m" in key]
```
IRDonch: You don't need to check the option name _or_ render the value. Just check if it's a `ModelArg`, and if yes, then extract the model name.

demos/tests/run_tests.py
```diff
@@ -133,23 +154,33 @@ def option_to_args(key, value):
...
+
+                for device, dev_arg in device_args.items():
+                    case_index +=1
```
IRDonch: I don't think we should increment the index for each device, since then there's a lack of correspondence between output and the elements of `demo.test_cases`. It would be better to print the device as part of the case ID, e.g. "Test case #5/GPU:".

demos/tests/run_tests.py
```diff
@@ -133,23 +148,32 @@ def option_to_args(key, value):
 
             print('Fixed arguments:', ' '.join(map(shlex.quote, fixed_args)))
             print()
+            device_args = demo.device_args(args.devices.split(' '))
```
IRDonch: It would be better to use `split()` to allow an arbitrary number of spaces between devices.

demos/tests/run_tests.py
```diff
@@ -133,23 +148,32 @@ def option_to_args(key, value):
...
+            for case_index, test_case in enumerate(demo.test_cases):
 
-            for test_case_index, test_case in enumerate(demo.test_cases):
```
IRDonch: Why rename `test_case_index`?

demos/tests/run_tests.py
```diff
@@ -133,23 +148,32 @@ def option_to_args(key, value):
...
-                print(flush=True)
-
-                try:
-                    subprocess.check_output(fixed_args + case_args,
-                        stderr=subprocess.STDOUT, universal_newlines=True)
-                except subprocess.CalledProcessError as e:
-                    print(e.output)
-                    print('Exit code:', e.returncode)
-                    num_failures += 1
+                pipeline = [value.name for key, value in test_case.options.items() if type(value) is ModelArg]
```
IRDonch:
```suggestion
                pipeline = [value.name for key, value in test_case.options.items() if isinstance(value, ModelArg)]
```

IRDonch:
Jenkins: test this please

PR #580, opened 2019-10-31
demos/mask_rcnn_demo: prevent OOB accesses to the colors array

IRDonch:
There's no guarantee that the number of unique classes in the image is less than the number of colors, so use a remainder operation to wrap the color index.

Also, change the way that `class_color` is populated to prevent dependence on evaluation order. Evaluating `class_color[class_id]` can change the size, so the color indexes could change depending on whether the LHS or the RHS was evaluated first. Use `emplace` to set the default value instead.

### DIFF
--- a/demos/mask_rcnn_demo/main.cpp
+++ b/demos/mask_rcnn_demo/main.cpp
@@ -318,9 +318,8 @@ int main(int argc, char *argv[]) {
             int box_height = std::min(static_cast<int>(std::max(0.0f, y2 - y1)), images_cv[batch].size().height);
             auto class_id = static_cast<size_t>(box_info[1] + 1e-6f);
             if (prob > PROBABILITY_THRESHOLD) {
-                if (class_color.find(class_id) == class_color.end())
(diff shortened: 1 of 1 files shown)


IRDonch:
> output_images should be completely removed.

(Assuming you mean `class_color`.)

I could do that. However, with the current approach we are guaranteed that each class will have a distinct color, as long as the total number of classes in the input image is less than or equal to the number of colors in the array. With your proposed approach, that no longer holds. I think for now it's more important to have distinct colors rather than consistent colors, since without labels you can't tell which color is what anyway.

Wovchena:
> (Assuming you mean class_color.)

Correct

> since without labels you can't tell which color is what anyway.

If somebody will use the demo frequently he will learn the mapping from predicted results.

Both solutions have problems, so I don't think it is important which one to choose. At least the number of problems is decreased with the patch.

IRDonch:
> At least the number of problems is decreased with the patch.

Yep, that's my approach too. :-)

Issue #567, opened 2019-10-29
Two models in a neural compute stick2


aub8016:
in the multi camera multi person tracking, we need to use two models the person detection and re-identification model, but when using MYRIAD  as a device it only loads one model and it gives the following error

```
INFO:root:Opening` file video1.mp4
INFO:root:Opening file video2.mp4
INFO:root:Creating Inference Engine
INFO:root:Loading network files:
	/home/ayoub/AI/open_model_zoo/tools/downloader/intel/person-detection-retail-0013/FP16/person-detection-retail-0013.xml
	/home/ayoub/AI/open_model_zoo/tools/downloader/intel/person-detection-retail-0013/FP16/person-detection-retail-0013.bin
INFO:root:Preparing input blobs
INFO:root:Loading model to the plugin
INFO:root:Creating Inference Engine
INFO:root:Loading network files:
	/home/ayoub/AI/open_model_zoo/tools/downloader/intel/person-reidentification-retail-0079/FP16/person-reidentification-retail-0079.xml
	/home/ayoub/AI/open_model_zoo/tools/downloader/intel/person-reidentification-retail-0079/FP16/person-reidentification-retail-0079.bin
INFO:root:Preparing input blobs
INFO:root:Loading model to the plugin
Traceback (most recent call last):
  File "multi_camera_multi_person_tracking.py", line 148, in <module>
    main()
  File "multi_camera_multi_person_tracking.py", line 140, in main
    person_recognizer = VectorCNN(args.m_reid, args.device)
  File "/home/ayoub/AI/open_model_zoo/demos/python_demos/multi_camera_multi_person_tracking/utils/network_wrappers.py", line 75, in __init__
    self.net = load_ie_model(model_path, device, None, num_reqs=self.max_reqs)
  File "/home/ayoub/AI/open_model_zoo/demos/python_demos/multi_camera_multi_person_tracking/utils/ie_tools.py", line 96, in load_ie_model
    exec_net = ie.load_network(network=net, device_name=device, num_requests=num_reqs)
  File "ie_api.pyx", line 85, in openvino.inference_engine.ie_api.IECore.load_network
  File "ie_api.pyx", line 92, in openvino.inference_engine.ie_api.IECore.load_network
RuntimeError: Can not init Myriad device: `NC_ERROR
```

vladimir-dudnik:
[...]

PR #588, opened 2019-11-06
[#567]Mcmt ncs2 fix

aungthetnaing:
IE cannot be instantiated multiple times for a single NCS2 use case. We will use single instance.


vladimir-dudnik:
Jenkins please retry a build

demos/python_demos/multi_camera_multi_person_tracking/utils/ie_tools.py
```diff
@@ -59,12 +59,12 @@ def get_input_shape(self):
         return self.inputs_info[self.input_key]
 
 
-def load_ie_model(model_xml, device, plugin_dir, cpu_extension='', num_reqs=1):
+def load_ie_model(ie, model_xml, device, plugin_dir, cpu_extension='', num_reqs=1):
     """Loads a model in the Inference Engine format"""
     model_bin = os.path.splitext(model_xml)[0] + ".bin"
     # Plugin initialization for specified device and load extensions library if specified
     log.info("Creating Inference Engine")
```
IRDonch: This log message should probably be moved together with the `ie` assignment.

aungthetnaing: My understanding is that this log is applicable to lines following after that. It was noted as specific device init, which is still left in place.

IRDonch: I don't think the lines below can reasonably be described as "Creating Inference Engine", though. So keep the log entry here if you want, but then edit it to match what actually happens.

aungthetnaing: Re-phrased log in original location. Added log in ie instance creation.

PR #589, opened 2019-11-07
Ac/add single human pose estimation accuracy checking

AnastasiaaSenina:


eaidova:
Jenkins ok to test

tools/accuracy_checker/accuracy_checker/annotation_converters/ms_coco.py
```diff
@@ -289,3 +289,70 @@ def _create_representations(
...
+                keypoints_annotation.metadata['areas'] = areas
+                keypoints_annotation.metadata['rects'] = bboxes
+                keypoints_annotation.metadata['iscrowd'] = is_crowd
+                keypoints_annotation.metadata['difficult_boxes'] = difficult
+
+                keypoints_annotations.append(keypoints_annotation)
+                if progress_callback is not None and image_id & progress_interval == 0:
+                    progress_callback(image_id / num_iterations * 100)
+
+        return ConverterReturn(keypoints_annotations, {'label_map': label_map}, None)
```
eaidova:
```suggestion
        return ConverterReturn(keypoints_annotations, {'label_map': label_map}, content_errors)
```

tools/accuracy_checker/accuracy_checker/preprocessor/geometric_transformations.py
```diff
@@ -873,3 +873,78 @@ def crop_center(img, cropx, cropy, cropz):
...
+            'dst_height': NumberField(
+                value_type=int, optional=True, min_value=1, description="Height of heatmaps output."
+            )
+        })
+
+        return parameters
+
+    def configure(self):
+        self.dst_height, self.dst_width = get_size_from_config(self.config)
+        self.input_height = 384
```
eaidova: on my view `input_height`/`input_width` and possible `stride` also should be configurable

tools/accuracy_checker/accuracy_checker/adapters/pose_estimation.py
```diff
@@ -337,3 +337,66 @@ def get_poses(subset, candidate):
         scores = np.array(scores)
 
         return persons_keypoints_x, persons_keypoints_y, persons_keypoints_v, scores
+
+
+class SingleHumanPoseAdapter(Adapter):
+    __provider__ = 'single_human_pose_estimation'
+    prediction_types = (PoseEstimationPrediction, )
+
+    @classmethod
```
eaidova: you do not need to specify output layer name explicitly if you model has only one output.
adapter has field `output_blob`, which store output layer name as default.
Just remove `parameters`, `validate_config` and `configure` method and change
heatmaps = np.transpose(raw_outputs[self.keypoints_heatmap][-1], (1, 2, 0)) -> np.transpose(raw_outputs[self.output_blob][-1], (1, 2, 0))

tools/accuracy_checker/accuracy_checker/utils.py
```diff
@@ -489,3 +489,11 @@ def check_file_existence(file):
         return True
     except (FileNotFoundError, IsADirectoryError):
         return False
+
+def get_stride_from_config(config, allow_none=False):
```
eaidova: As I understand, stride is required argument for your preprocessor, right?
What about set it optional=False in parameters? It will be mean that it should be always provided (None is not allowed and it will be checked automatically)
BTW why at the end you return tuple with 2 None values?

tools/accuracy_checker/accuracy_checker/preprocessor/geometric_transformations.py
```diff
@@ -29,7 +29,8 @@
 from ..dependency import ClassProvider
 from ..logging import warning
 from ..preprocessor import Preprocessor
-from ..utils import contains_all, get_size_from_config, string_to_tuple, get_size_3d_from_config
+from ..utils import contains_all, get_size_from_config, string_to_tuple, get_size_3d_from_config,
+                    get_stride_from_config
```
eaidova: it does not work, you need to wrap multiline import statement in the brackets for treat it as single line.
something like that:
```python
from ..utils import (
contains_all, get_size_from_config, string_to_tuple, get_size_3d_from_config, get_stride_from_config
)
```

tools/accuracy_checker/configs/single-human-pose-estimation-0001.yml
```diff
@@ -0,0 +1,21 @@
+models:
+  - name: single-human-pose-estimation-0001
+
+    launchers:
+      - framework: dlsdk
+        device: CPU
+        model:   intel/single-human-pose-estimation-0001/FP32/single-human-pose-estimation.xml
+        weights: intel/single-human-pose-estimation-0001/FP32/single-human-pose-estimation.bin
```
vladimir-dudnik:
```suggestion
        weights: intel/single-human-pose-estimation-0001/FP32/single-human-pose-estimation-0001.bin
```
other intel models have consistent naming with folder

tools/accuracy_checker/accuracy_checker/preprocessor/geometric_transformations.py
```diff
@@ -873,3 +874,83 @@ def crop_center(img, cropx, cropy, cropz):
         endz = min(startz + cropz, z)
 
         return img[startz:endz, starty:endy, startx:endx, :]
+
+
+class AffineTransform(Preprocessor):
```
Daniil-Osokin: This does not look like generic affine transform. Suggest to change the name to specific one. Not sure what to suggest, but it is some kind of crop.

tools/accuracy_checker/accuracy_checker/preprocessor/geometric_transformations.py
```diff
@@ -873,3 +874,83 @@ def crop_center(img, cropx, cropy, cropz):
...
+            bbox[2] = bbox[3] * aspect_ratio
+
+        s = np.array([bbox[2] / 200., bbox[3] / 200.], np.float32)
+        scale = s * 1.25
+
+        return center, scale
+    
+    @staticmethod
+    def get_affine_transform(center, scale, output_size, key=0):
+
```
Daniil-Osokin: :)

tools/accuracy_checker/configs/single-human-pose-estimation-0001.yml
Daniil-Osokin: Will this model come from `openvino_trainng_extensions`? If so, it should go to `public` folder.

AnastasiaaSenina: Yes, you are right. Thank you.

tools/accuracy_checker/accuracy_checker/preprocessor/geometric_transformations.py
```diff
@@ -876,8 +877,8 @@ def crop_center(img, cropx, cropy, cropz):
         return img[startz:endz, starty:endy, startx:endx, :]
 
 
-class AffineTransform(Preprocessor):
-    __provider__ = 'affine_transform'
+class CropProvidedBbox(Preprocessor):
+    __provider__ = 'crop_provided_bbox'
```
eaidova: provided from what?
I am not sure about naming... if it is possible could you please give more descriptive name

tools/accuracy_checker/accuracy_checker/preprocessor/geometric_transformations.py
```diff
@@ -874,6 +876,84 @@ def crop_center(img, cropx, cropy, cropz):
...
+        cx_bbox = bbox[0] + bbox[2] * 0.5
+        cy_bbox = bbox[1] + bbox[3] * 0.5
+        center = np.array([np.float32(cx_bbox), np.float32(cy_bbox)])
+
+        if bbox[2] > aspect_ratio * bbox[3]:
+            bbox[3] = bbox[2] * 1.0 / aspect_ratio
+        elif bbox[2] < aspect_ratio * bbox[3]:
+            bbox[2] = bbox[3] * aspect_ratio
+
+        s = np.array([bbox[2] / 200., bbox[3] / 200.], np.float32)
```
eaidova: is 200  the magic constant or it has some meaning?
Can this number be changed for some cases or it is constant?

AnastasiaaSenina: it is constant.

tools/accuracy_checker/accuracy_checker/adapters/pose_estimation.py
```diff
@@ -337,3 +337,59 @@ def get_poses(subset, candidate):
...
+            y_values.append(y)
+            if score > 0.2:
+                sum_score_thr += score
+                num_kp_thr += 1
+            sum_score += score
+        if num_kp_thr != 0:
+            pose_score = sum_score_thr / num_kp_thr
+        else:
+            pose_score = sum_score / raw_outputs[self.output_blob].shape[1]
+        result.append(PoseEstimationPrediction(identifiers[0], np.array([x_values]),
```
eaidova: I understand that this model will be executed with batch > 1 scarcely, but we prefer to define processing in the cycle over list of identifiers and batch, please look on any other adapters.

eaidova:
@AnastasiaaSenina, please update also readme files for added functionality.

tools/accuracy_checker/configs/single-human-pose-estimation-0001.yml
```diff
@@ -0,0 +1,21 @@
+models:
+  - name: single-human-pose-estimation-0001
+
+    launchers:
+      - framework: dlsdk
+        device: CPU
```
eaidova: please remove device and add fp16 model also

tools/accuracy_checker/accuracy_checker/adapters/pose_estimation.py
```diff
@@ -337,3 +337,61 @@ def get_poses(subset, candidate):
...
+        outputs_batch = raw_outputs[self.output_blob]
+        for i, heatmaps in enumerate(outputs_batch):
+            heatmaps = np.transpose(heatmaps, (1, 2, 0))
+            sum_score = 0
+            sum_score_thr = 0
+            scores = []
+            x_values = []
+            y_values = []
+            num_kp_thr = 0
+            vis = [1] * raw_outputs[self.output_blob].shape[1]
```
eaidova: you already extract predictions in `outputs_batch`. Better to use it

tools/accuracy_checker/accuracy_checker/preprocessor/geometric_transformations.py
```diff
@@ -874,6 +876,84 @@ def crop_center(img, cropx, cropy, cropz):
...
+            'stride': NumberField(
+                value_type=int, optional=False,
+                description="Stride for network. It is input size of heatmaps / output size of heatmaps."
+            )
+        })
+
+        return parameters
+
+    def configure(self):
+        self.input_height, self.input_width = get_size_from_config(self.config)
```
eaidova: `get_size_from_config` assume also that `size` parameter is available. It is used when `dst_height` and `dst_width` are equal. Please add it for alignment with other preprocessors.

tools/accuracy_checker/accuracy_checker/preprocessor/geometric_transformations.py
```diff
@@ -874,6 +876,84 @@ def crop_center(img, cropx, cropy, cropz):
...
+        self.stride = get_stride_from_config(self.config)
```
eaidova: do you really need introduce new function for getting stride?
You make this parameter required, so check that it is not None will be performed automatically. It allows you use `self.get_value_from_config('stride')` for getting stride value

tools/accuracy_checker/configs/single-human-pose-estimation-0001.yml
```diff
@@ -0,0 +1,29 @@
...
+    datasets:
+      - name: ms_coco_single_keypoints
+        preprocessing:
+          - type: transformed_crop_with_auto_scale
+            dst_height: 384
+            dst_width: 288
+            stride: 8
+        metrics:
+          - name: AP
+            type: coco_orig_keypoints_precision
```
eaidova: please add `global_definitions: ../dataset_definitions.yml` at the end like in other configs e.g. [human-pose-estimation-0001](https://github.com/opencv/open_model_zoo/blob/develop/tools/accuracy_checker/configs/human-pose-estimation-0001.yml#L62)

eaidova:
@AnastasiaaSenina, changes looks good, but linter say:

 ```
accuracy_checker/preprocessor/geometric_transformations.py:1: [C0302(too-many-lines), ] Too many lines in module (1002/1000)
```
You can just remove 2 empty lines inside process for fast fix, I am going refactor this module later and devide it on several files

eaidova:
@vladimir-dudnik, do you agree with changes? Can we merge?

PR #591, opened 2019-11-07
Image retrieval demo added "--no-show" option

eizamaliev:


demos/python_demos/image_retrieval_demo/image_retrieval_demo.py
```diff
@@ -58,6 +58,8 @@ def build_argparser():
                       help="Optional. Required for CPU custom layers. Absolute path to "
                            "a shared library with the kernels implementations.", type=str,
                       default=None)
+    args.add_argument('--no_show', action='store_true',
+                      help='Optional. Do not display output.')
```
Wovchena: `CONTRIBUTING.md` claims it should be  `Optional. Do not visualize inference results.`

demos/python_demos/image_retrieval_demo/image_retrieval_demo.py
```diff
@@ -137,14 +139,15 @@ def main():
             else:
                 log.info("ROI detected, found: %s", sorted_classes[0])
 
-        key = visualize(view_frame, position,
-                        [img_retrieval.impaths[i] for i in sorted_indexes],
-                        distances[sorted_indexes] if position is not None else None,
-                        img_retrieval.input_size, np.mean(compute_embeddings_times),
-                        np.mean(search_in_gallery_times), imshow_delay=3)
+        if not args.no_show:
+            key = visualize(view_frame, position,
```
Wovchena: I suggest forwarding that key to `visualize()`. We want to disable only `imshow` and `waitkey` while this approach disables the whole function which we could test.

eizamaliev: Ok

demos/python_demos/image_retrieval_demo/image_retrieval_demo/visualizer.py
```diff
@@ -95,6 +95,8 @@ def visualize(image, target_pos, impaths, distances, input_size, compute_embeddi
 
     demo_image = add_top10_gallery_images(demo_image, impaths, distances, input_image, target_pos)
 
-    cv2.imshow('demo_image', demo_image)
+    if not no_show:
+        cv2.imshow('demo_image', demo_image)
+        return cv2.waitKey(imshow_delay) & 0xff
 
-    return cv2.waitKey(imshow_delay) & 0xff
+    return 0
```
Wovchena: It should be `-1`, like if no key had been pressed: https://www.docs.opencv.org/master/d7/dfc/group__highgui.html#ga5628525ad33f52eab17feebcfba38bd7

Wovchena: And `-1 & 0xff` is probably something different.

Wovchena: @eizamaliev, so what about this `0xff`? `-1 & 0xff` which you will get from `cv2.waitKey(imshow_delay) & 0xff` is not equal to `-1` in case of `--no_show`. We could just drop this `0xff` from the demo or return `-1 & 0xff`. I don't know which way is correct to process pressed key.

eizamaliev: In this case `-1` will be correct output for no-key-pressed. In fact, `cv2.waitKey(imshow_delay) & 0xff` is incorrect output if no key pressed. I'll fix it.

Wovchena: `cv::waitKey` returns not only -1 or 1.
Your line `return key_pressed & 0xff if key_pressed != 1 else -1` is equivalent to 
```
if key_pressed == 1: return -1
else return key_pressed & 0xff
```
So if you press the button with code `1`, the demo will receive `-1` and since pressed key is handled out of the function, I believe it's better to forward code of pressed key unchanged (the demo doesn't handle that particular key now, but I still think it is incorrect approach). Moreover for current implementation, if you press nothing and presenting is enabled, you will get `-1 & 0xff` and if you use `--no_show` it will return `-1`, while my suggestion is that `--no_show` should behave like no key is pressed.

Wovchena: > In this case `-1` will be correct output for no-key-pressed. In fact, `cv2.waitKey(imshow_delay) & 0xff` is incorrect output if no key pressed. I'll fix it.

In your implementation for no-key-pressed the return value will be `-1 & 0xff` which is not `-1`.

eizamaliev: Sorry, my mistake. It should have been *key_pressed != **-1***

eizamaliev: I not sure, that understand you, but will summarize:

Key | with `--no_show` | without `--no_show 
----- | ----- | -----
| Pressed | -1 | waitKey() & 0xFF |
| Not pressed| -1 | -1 (not waitKey() & 0xFF) |

Correct?

eizamaliev: @Wovchena do you accept this change?

Wovchena: Yes

PR #595, opened 2019-11-07
Improve usability of super_resolution_demo

AlexanderDokuchaev:


demos/super_resolution_demo/main.cpp
```diff
@@ -136,6 +136,10 @@ int main(int argc, char *argv[]) {
             int h = static_cast<int>(lrInputInfoItem->getTensorDesc().getDims()[2]);
             int c = static_cast<int>(lrInputInfoItem->getTensorDesc().getDims()[1]);
 
+            /** Convert to 3-channels image for case when image was saved with alpha channel **/
+            if (img.channels() == 4)
+                cv::cvtColor(img, img, cv::COLOR_BGRA2BGR);
```
IRDonch: Wouldn't it be better to change `IMREAD_UNCHANGED` to `IMREAD_COLOR` above? It's not only simpler, but would also work for grayscale images.

AlexanderDokuchaev: No. 'IMREAD_COLOR' always convert to RGB Image, but text-image-super-resolution model expect grayscale Image as input.

IRDonch: I see. In this case, maybe you could use `IMREAD_COLOR`/`IMREAD_GRAYSCALE`, depending on `c`?

AlexanderDokuchaev: Maybe

AlexanderDokuchaev: @IRDonch  is it possible add testcase for grayscale input image?

IRDonch: Not at the moment, I'm afraid. I don't think we have any in our test data.

demos/super_resolution_demo/main.cpp
```diff
@@ -124,18 +124,18 @@ int main(int argc, char *argv[]) {
...
-                continue;
-            }
-
             /** Get size of low resolution input **/
             auto lrInputInfoItem = inputInfo[lrInputBlobName];
             int w = static_cast<int>(lrInputInfoItem->getTensorDesc().getDims()[3]);
             int h = static_cast<int>(lrInputInfoItem->getTensorDesc().getDims()[2]);
             int c = static_cast<int>(lrInputInfoItem->getTensorDesc().getDims()[1]);
 
+            cv::Mat img = cv::imread(i, c == 1 ? cv::IMREAD_UNCHANGED : cv::IMREAD_COLOR);
```
IRDonch:
```suggestion
            cv::Mat img = cv::imread(i, c == 1 ? cv::IMREAD_GRAYSCALE : cv::IMREAD_COLOR);
```

PR #601, opened 2019-11-08
demos: prevent several potential integer overflows

IRDonch:


demos/common/samples/common.hpp
```diff
@@ -412,60 +412,40 @@ static UNUSED bool writeOutputBmp(std::string name, unsigned char *data, size_t
 * @param width - width of the rectangle
 * @param rectangles - vector points for the rectangle, should be 4x compared to num classes
 * @param classes - vector of classes
-* @param thickness - thickness of a line (in pixels) to be used for bounding boxes
 */
-static UNUSED void addRectangles(unsigned char *data, size_t height, size_t width, std::vector<int> rectangles, std::vector<int> classes, int thickness = 1) {
+static UNUSED void addRectangles(unsigned char *data, size_t height, size_t width, std::vector<int> rectangles, std::vector<int> classes) {
```
Wovchena: May be `object_detection_demo_faster_rcnn` could just use OpenCV? And the need in this function would be eliminated. :)

demos/object_detection_demo_faster_rcnn/detectionoutput.h
```diff
@@ -107,7 +107,7 @@ class DetectionOutputPostProcessor {
                     {Precision::I32, indices_size, {indices_size, {0, 1, 2}}});
             _indices->allocate();
 
-            SizeVector detections_size{static_cast<size_t>(_num * _num_classes)};
+            SizeVector detections_size{static_cast<size_t>(_num) * static_cast<size_t>(_num_classes)};
```
Wovchena: It looks `_num` is never used as `int`. Except for `_num = static_cast<int>(conf_size);` but `conf_size` is `uint32_t`. `_num` should be defined as `size_t`.

IRDonch: Good catch. `_num` is actually entirely redundant, so I just removed it.

demos/common/samples/common.hpp
IRDonch: Fair point. I'll see if it's not too difficult to port this demo to OpenCV.

IRDonch: Ported it in #606. This change no longer applies, so I removed it.

PR #605, opened 2019-11-11
Update multi camera multi person tracker

DmitriySidnev:
Here is an update for multi camera multi person tracker. Were added the next features:
* support of instance segmentation network
* support of orientation classification network
* metrics evaluation
* history visualization
* visual analyzer
* bug fixes

@snosov1, @sovrasov, @AlexanderDokuchaev, please take a look.


IRDonch:
I don't suppose you could split your improvements into multiple patches? This is a lot to take in.

demos/python_demos/multi_camera_multi_person_tracking/README.md
```diff
@@ -32,16 +32,22 @@ and then for each detected object it extracts embeddings using re-identification
 To install required dependencies run
 
 ```bash
-pip3 install -r requirements.txt
+cat requirements.txt | xargs -n 1 -L 1 pip3 install
```
IRDonch: Why is this necessary?

demos/python_demos/multi_camera_multi_person_tracking/models.lst
```diff
@@ -1,3 +1,4 @@
 # This file can be used with the --list option of the model downloader.
 person-detection-retail-0013
 person-reidentification-retail-????
+instance-segmentation-security-????
```
IRDonch: Please maintain the alphabetical order.

demos/python_demos/multi_camera_multi_person_tracking/requirements.txt
```diff
@@ -1,3 +1,19 @@
...
+opencv-python==3.4.5.20
+Pillow==6.1.0
+pycocotools==2.0.0
+requests==2.22.0
+numpy==1.16.4
+scipy==1.3.0
+six==1.12.0
+tensorboard==1.8
+tqdm==4.32.2
+yacs==0.1.6
```
IRDonch: This file should not pin versions. If you have certain minimum requirements, you can use `>=`, but otherwise the demos are expected to work with any recent versions of dependencies.

demos/python_demos/multi_camera_multi_person_tracking/tools/run_evaluate.py
```diff
@@ -0,0 +1,151 @@
...
+"""
+
+import argparse
+import json
+
+import numpy as np
+import motmetrics as mm
+from lxml import etree
+from tqdm import tqdm
+import glog as log
```
IRDonch: The other demos use the built-in `logging` module for logging. Please do the same.

demos/python_demos/multi_camera_multi_person_tracking/README.md
DmitriySidnev: Some of modules in the list depend on `setuptools` or `Cython` and the first command `pip3 install -r requirements.txt` is crashed when you have clean system. You can check it with clean `virtualenv` and key `--no-cahce` to avoid using cached modules. So, the new command works without any troubles.

demos/python_demos/multi_camera_multi_person_tracking/models.lst
DmitriySidnev: Fixed.

demos/python_demos/multi_camera_multi_person_tracking/requirements.txt
DmitriySidnev: Fixed, used `>=`. I did not try early versions and it is possible that something turned out incompatible.

demos/python_demos/multi_camera_multi_person_tracking/tools/run_evaluate.py
DmitriySidnev: Fixed.

DmitriySidnev:
@IRDonch , actually it is difficult to split the PR onto several patches. Can we remain this PR as is?

demos/python_demos/multi_camera_multi_person_tracking/README.md
IRDonch: I have tested this, and it seems that the problematic package is `pycocotools`. And I know that `pycocotools` is also a problem because it fails to install on WIndows. Is there any way you can make due without it?

demos/python_demos/multi_camera_multi_person_tracking/requirements.txt
```diff
@@ -1,3 +1,18 @@
...
+opencv-python>=3.4.5.20
+Pillow>=6.1.0
+pycocotools>=2.0.0
+requests>=2.22.0
+numpy>=1.16.4
+scipy>=1.3.0
+six>=1.12.0
+tensorboard>=1.8
+tqdm>=4.32.2
+yacs>=0.1.6
```
IRDonch: This file should only list direct dependencies, so as to make it clear what the demo actually needs. Could you remove the packages that are not used directly?

demos/python_demos/multi_camera_multi_person_tracking/README.md
DmitriySidnev: Yes, you are right the problem is in `pycocotools` but it is a necessary module. At least via our command `pycocotools` is installed properly in Linux. I would like to remain this command as is and we can not remove `pycocotools`. So, with Windows, as I understand, it is a global problem and if we need this module any way.

demos/python_demos/multi_camera_multi_person_tracking/requirements.txt
DmitriySidnev: Fixed.

DmitriySidnev:
@IRDonch, I think I have fixed all comments. Could you please check it?

demos/python_demos/multi_camera_multi_person_tracking/README.md
```diff
@@ -32,16 +32,22 @@ and then for each detected object it extracts embeddings using re-identification
...
 ```
-usage: multi_camera_multi_person_tracking.py [-h] -i I [I ...] -m M_DETECTOR
+usage: multi_camera_multi_person_tracking.py [-h] -i I [I ...]
+                                             [--detections DETECTIONS [DETECTIONS ...]]
+                                             -m M_DETECTOR
                                              [--t_detector T_DETECTOR]
+                                             [--m_segmentation M_SEGMENTATION]
+                                             [--t_segmentation T_SEGMENTATION]
+                                             [--m_orientation M_ORIENTATION]
+                                             [--t_orientation T_ORIENTATION]
```
IRDonch: You need to expand the README with an explanation of these new parameters. In particular, it's necessary to note what models are accepted for the new `--m` options, as well as the format of the file(s) that `--detections` accepts.

demos/python_demos/multi_camera_multi_person_tracking/mc_tracker/mct.py
```diff
@@ -23,18 +24,27 @@ class MultiCameraTracker:
...
         self.time = 0
         self.last_global_id = 0
         self.global_ids_queue = queue.Queue()
+        assert time_window >= 1
         self.time_window = time_window  # should be greater than time window in scts
+        assert 0 <= global_match_thresh <= 1
         self.global_match_thresh = global_match_thresh
+        assert bbox_min_aspect_ratio >= 0
+        self.bbox_min_aspect_ratio = bbox_min_aspect_ratio
+        random.seed(random_seed)
```
IRDonch: This seems misleading. If `MultiCameraTracker` accepts a `random_seed` parameter, then it should only use that seed for its own calculations, not set it globally. If you want to set a seed globally, just do it in `main`. You don't need to pass it as a parameter.

demos/python_demos/multi_camera_multi_person_tracking/multi_camera_multi_person_tracking.py
```diff
@@ -81,24 +96,32 @@ def run(params, capture, detector, reid):
...
             all_detections[i] = [det[0] for det in detections]
             all_masks[i] = [det[2] for det in detections if len(det) == 3]
 
-        tracker.process(frames, all_detections, all_masks)
+        tracker.process(prev_frames, all_detections, all_masks)
         tracked_objects = tracker.get_tracked_objects()
 
-        fps = round(1 / (time.time() - start), 1)
-        vis = visualize_multicam_detections(frames, tracked_objects, fps)
-        if not params.no_show:
```
IRDonch: I think you've accidentally made it so `--no_show` has no effect. Please restore it.

demos/python_demos/multi_camera_multi_person_tracking/multi_camera_multi_person_tracking.py
```diff
@@ -107,19 +130,34 @@ def run(params, capture, detector, reid):
...
+
 
 def main():
     """Prepares data for the person recognition demo"""
     parser = argparse.ArgumentParser(description='Multi camera multi person \
                                                   tracking live demo script')
     parser.add_argument('-i', type=str, nargs='+', help='Input sources (indexes \
                         of cameras or paths to video files)', required=True)
+    parser.add_argument('--detections', type=str, nargs='+',
+                        help='Json files with detections')
```
IRDonch:
```suggestion
                        help='JSON files with detections')
```

demos/python_demos/multi_camera_multi_person_tracking/multi_camera_multi_person_tracking.py
```diff
@@ -107,19 +130,34 @@ def run(params, capture, detector, reid):
...
 
     parser.add_argument('-m', '--m_detector', type=str, required=True,
                         help='Path to the person detection model')
     parser.add_argument('--t_detector', type=float, default=0.6,
                         help='Threshold for the person detection model')
 
+    parser.add_argument('--m_segmentation', type=str, required=False,
+                        help='Path to the instance segmentation model')
```
IRDonch:
```suggestion
                        help='Path to the person instance segmentation model')
```

Seems like it should be consistent with the next option.

demos/python_demos/multi_camera_multi_person_tracking/multi_camera_multi_person_tracking.py
```diff
@@ -141,13 +179,28 @@ def main():
     log.info("Creating Inference Engine")
     ie = IECore()
 
-    person_detector = Detector(ie, args.m_detector, args.t_detector,
-                               args.device, args.cpu_extension,
-                               capture.get_num_sources())
+    if args.detections:
+        person_detector = DetectionsFromFileReader(args.detections, args.t_detector)
+    else:
+        if args.m_segmentation:
```
IRDonch: `elif args.m_segmentation:` would be more readable.

demos/python_demos/multi_camera_multi_person_tracking/multi_camera_multi_person_tracking.py
```diff
@@ -141,13 +179,28 @@ def main():
...
+            person_detector = MaskRCNN(ie, args.m_segmentation, args.t_segmentation,
+                                       args.device, args.cpu_extension,
+                                       capture.get_num_sources())
+        else:
+            person_detector = Detector(ie, args.m_detector, args.t_detector,
+                                       args.device, args.cpu_extension,
+                                       capture.get_num_sources())
```
IRDonch: Given that `--detection`, `--m_segmentation` and `--m_detector` are apparently alternatives, you should make it so that only one can be specified.

demos/python_demos/multi_camera_multi_person_tracking/requirements.txt
IRDonch: It doesn't seem to actually be fixes. E.g. I can immediately see `requests`, which is not used anywhere in the demo.

DmitriySidnev: My bad. Fixed.

demos/python_demos/multi_camera_multi_person_tracking/mc_tracker/mct.py
DmitriySidnev: `random_seed` setup moved to main function.

demos/python_demos/multi_camera_multi_person_tracking/multi_camera_multi_person_tracking.py
DmitriySidnev: Yes, thank you. Restored.

demos/python_demos/multi_camera_multi_person_tracking/multi_camera_multi_person_tracking.py
DmitriySidnev: Fixed.

demos/python_demos/multi_camera_multi_person_tracking/multi_camera_multi_person_tracking.py
DmitriySidnev: Fixed.

demos/python_demos/multi_camera_multi_person_tracking/multi_camera_multi_person_tracking.py
DmitriySidnev: Fixed.

demos/python_demos/multi_camera_multi_person_tracking/multi_camera_multi_person_tracking.py
DmitriySidnev: Fixed. Added check for number of detectors with error if more than one or nothing was specified.

demos/python_demos/multi_camera_multi_person_tracking/README.md
DmitriySidnev: Model specified via `--m_orientation` and `--t_orientation` arguments removed. Instance segmentation model was added in readme (also is present in `models.lst` file).

demos/python_demos/multi_camera_multi_person_tracking/README.md
IRDonch: > but it is a necessary module.

Is it, though?

I have tried to figure out how it is used, and it looks to me like the code where it's used is unreachable. The only place where it's used is in `utils.segm_postrocess.segm_postprocess` when `encode` is true, but nothing seems to actually call it with `encode=True`.

demos/python_demos/multi_camera_multi_person_tracking/multi_camera_multi_person_tracking.py
```diff
@@ -17,18 +17,42 @@
...
+
+
+def check_detectors(args):
+    detectors = {
+        '--m_detector': args.m_detector,
+        '--m_segmentation': args.m_segmentation,
+        '--detections': args.detections
+    }
+    det_counter = 0
+    for _, value in detectors.items():
```
Wovchena:
```suggestion
    for value in detectors.values():
```

demos/python_demos/multi_camera_multi_person_tracking/utils/network_wrappers.py
```diff
@@ -81,3 +105,215 @@ def forward(self, batch):
...
+        def __init__(self, transforms):
+            self.transforms = transforms
+
+        def __call__(self, sample):
+            for t in self.transforms:
+                sample = t(sample)
+            return sample
+
+    class Resize(object):
+        def __init__(self, max_size=None, window_size=None, size=None):
```
Wovchena: What is the reason to introduce `max_size` and `size` options? They are never used? Removing them would simplify the class.

demos/python_demos/multi_camera_multi_person_tracking/utils/network_wrappers.py
```diff
@@ -81,3 +105,215 @@ def forward(self, batch):
...
+            # Resize boxes.
+            if 'gt_boxes' in sample:
+                sample['gt_boxes'] *= [scale_x, scale_y, scale_x, scale_y]
+                sample['gt_boxes'] = np.clip(sample['gt_boxes'], 0, [w - 1, h - 1, w - 1, h - 1])
+
+            # Resize masks.
+            if 'gt_masks' in sample:
+                sample['gt_masks'] = [[np.clip(part * [scale_x, scale_y], 0, [w - 1, h - 1]) for part in obj]
+                                      for obj in sample['gt_masks']]
+
```
Wovchena: `gt_boxes` and `gt_masks` paths seem to be never used. The whole class can be much sorter:
```py
scale = min(net_w / image_size[0], net_h / image_size[1])
sample['image'] = cv2.resize(sample['image'], None, fx=scale, fy=scale)
```
Since `MaskRCNN` already stores that `net_w` and `net_h` as `w` and `h`, there is no reason to introduce `Resize` class (if `Normalize` is removed and `Compes` becomes unusd), it should be a function (if it is worth it to introduce a function containing only 2 lines and which is used only in one place).

demos/python_demos/multi_camera_multi_person_tracking/utils/network_wrappers.py
```diff
@@ -81,3 +105,215 @@ def forward(self, batch):
...
+    class Normalize(object):
+        def __init__(self, mean, std):
+            self.mean = mean
+            self.std = std
+
+        def __call__(self, sample):
+            sample['image'] = sample['image'].astype('float32').transpose(2, 0, 1)
+            sample['image'][:, :, 0] = (sample['image'][:, :, 0] - self.mean[0]) / self.std[0]
+            sample['image'][:, :, 1] = (sample['image'][:, :, 1] - self.mean[1]) / self.std[1]
+            sample['image'][:, :, 2] = (sample['image'][:, :, 2] - self.mean[2]) / self.std[2]
```
Wovchena: 1. The channel dimension now stays first because of `transpose(2, 0, 1)`.
2. Use  broadcasting:
    `sample['image'] -= self.mean[:, None, None]`
    `sample['image'] /= self.std[:, None, None]`
    Or in one line but not in place: `sample['image'] = (sample['image'] - self.mean[:, None, None]) / self.std[:, None, None]`
    assuming you will replace `self.mean = mean` with `self.mean = np.array(mean)` in `__init__(self, mean, std)`.
3. But for me your version with wrong dimensions works better than my suggested version. So I question whether this normalization is needed. I think you should just drop it. Intel models usually have means and scales already fused into the model. `instance-segmentation-security-????` are not exceptions. (Assuming `Normalize` is only used for `instance-segmentation-security-????` models)

demos/python_demos/multi_camera_multi_person_tracking/utils/network_wrappers.py
```diff
@@ -81,3 +105,215 @@ def forward(self, batch):
...
+        self.input_files = input_files
+        self.score_thresh = score_thresh
+        self.detections = []
+        for input_file in input_files:
+            log.info('Loading {}'.format(input_file))
+            with open(input_file) as f:
+                detections = json.load(f)
+                detections_dict = {}
+                for det in detections:
+                    detections_dict[det['frame_id']] = {'boxes': det['boxes'], 'scores': det['scores']}
```
Wovchena: I think the format of files for `--history_file` and `--detections` should be aligned. It will be possible to generate files by `--history_file` and provide the same files with `--detections` later.

demos/python_demos/multi_camera_multi_person_tracking/README.md
DmitriySidnev: Yes, you are right. Unused code removed, `pycocotools removed` too. Command to install python modules restored.

demos/python_demos/multi_camera_multi_person_tracking/multi_camera_multi_person_tracking.py
DmitriySidnev: `waitKey` works only in a window.

demos/python_demos/multi_camera_multi_person_tracking/multi_camera_multi_person_tracking.py
```diff
@@ -107,48 +152,81 @@ def run(params, capture, detector, reid):
...
     parser.add_argument('-d', '--device', type=str, default='CPU')
     parser.add_argument('-l', '--cpu_extension',
                         help='MKLDNN (CPU)-targeted custom layers.Absolute \
                               path to a shared library with the kernels impl.',
                              type=str, default=None)
-    parser.add_argument("--no_show", help="Optional. Don't show output", action='store_true')
 
     args = parser.parse_args()
+    if check_detectors(args) != 1:
+        return 1
```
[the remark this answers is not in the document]
DmitriySidnev: Fixed.

demos/python_demos/multi_camera_multi_person_tracking/README.md
```diff
@@ -75,20 +92,46 @@ Minimum command examples to run the demo:
...
     -d path/to/person-detection-retail-0013.xml \
-    -r path/to/person-reidentification-retail-0076.xml \
+    -r path/to/person-reidentification-retail-0103.xml \
+    --config config.py
+
+# videos with instance segmentation model
+python multi_camera_multi_person_tracking.py \
+    -i path/to/video_1.avi path/to/video_2.avi \
+    -d path/to/instance-segmentation-security-0050.xml \
+    -r path/to/person-reidentification-retail-0107.xml \
```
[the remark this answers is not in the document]
DmitriySidnev: Fixed all commands in readme.

demos/python_demos/multi_camera_multi_person_tracking/README.md
```diff
@@ -75,20 +92,46 @@ Minimum command examples to run the demo:
...
 # web-cameras
 python multi_camera_multi_person_tracking.py \
     -i 0 1 \
     -d path/to/person-detection-retail-0013.xml \
-    -r path/to/person-reidentification-retail-0076.xml \
+    -r path/to/person-reidentification-retail-0103.xml \
     --config config.py
 ```
 
+The demo can use file in JSON format with detections instead person detector.
```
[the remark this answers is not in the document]
DmitriySidnev: Fixed.

demos/python_demos/multi_camera_multi_person_tracking/README.md
```diff
@@ -75,20 +92,46 @@ Minimum command examples to run the demo:
...
+Structure of this file should be the next:
```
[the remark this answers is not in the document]
DmitriySidnev: Fixed.

demos/python_demos/multi_camera_multi_person_tracking/README.md
```diff
@@ -75,20 +92,46 @@ Minimum command examples to run the demo:
...
+```
```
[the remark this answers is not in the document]
DmitriySidnev: Fixed.

demos/python_demos/multi_camera_multi_person_tracking/utils/misc.py
```diff
@@ -41,10 +75,20 @@ def read_py_config(filename):
...
-def none_to_zero(a):
-    if a is not None:
-        return a
-    return 0
+def check_pressed_keys(key):
+    if key != 32:  # If was pressed space, use per-frame mode via pressing space key
+        key = cv.waitKey(1)
+    if key == 32:  # Pause
+        while True:
+            key = cv.waitKey(1)
```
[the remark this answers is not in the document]
DmitriySidnev: Fixed to `cv.waitKey(0)`.

demos/python_demos/multi_camera_multi_person_tracking/utils/network_wrappers.py
```diff
@@ -11,10 +11,29 @@
...
+import json
+import logging as log
+from collections import namedtuple
+from abc import ABC, abstractmethod
+
+import cv2
+import numpy as np
+
 from utils.ie_tools import load_ie_model
+from .segm_postrocess import postprocess
```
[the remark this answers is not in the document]
DmitriySidnev: Fixed.

demos/python_demos/multi_camera_multi_person_tracking/utils/segm_postrocess.py
```diff
@@ -0,0 +1,156 @@
...
+    boxes_exp = np.zeros(boxes.shape)
+    boxes_exp[:, 0] = x_c - w_half
+    boxes_exp[:, 2] = x_c + w_half
+    boxes_exp[:, 1] = y_c - h_half
+    boxes_exp[:, 3] = y_c + h_half
+
+    return boxes_exp
+
+
+def postprocess_batch(batch_ids, scores, classes, boxes, raw_cls_masks,
```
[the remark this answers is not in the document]
DmitriySidnev: Method `postprocess_batch` removed.

demos/python_demos/multi_camera_multi_person_tracking/utils/visualization.py
```diff
@@ -12,42 +12,131 @@
...
+            label = 'ID {}'.format(label) if not isinstance(label, str) else label
+            label_size, base_line = cv.getTextSize(label, cv.FONT_HERSHEY_SIMPLEX, 1, 2)
+            top = max(top, label_size[1])
+            cv.rectangle(frame, (left, top - label_size[1]), (left + label_size[0], top + base_line),
+                         (255, 255, 255), cv.FILLED)
+            cv.putText(frame, label, (left, top), cv.FONT_HERSHEY_SIMPLEX, 1, (0, 0, 0), 2)
 
-def visualize_multicam_detections(frames, all_objects, fps=''):
+
+def get_terget_size(frame_sizes, vis=None, max_window_size=(1920, 1080), stack_frames='vertical', **kwargs):
```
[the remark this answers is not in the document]
DmitriySidnev: Fixed.

demos/python_demos/multi_camera_multi_person_tracking/README.md
```diff
@@ -40,55 +41,102 @@ pip3 install -r requirements.txt
...
   --history_file HISTORY_FILE
+                        Optional. Path to file in JSON format to save results
+                        of the demo
+  --no_show             Optional. Don't show output
   -d DEVICE, --device DEVICE
   -l CPU_EXTENSION, --cpu_extension CPU_EXTENSION
                         MKLDNN (CPU)-targeted custom layers.Absolute path to a
                         shared library with the kernels impl.
-  --no_show             Optional. Don't show output
+
```
Wovchena:
```suggestion
```

demos/python_demos/multi_camera_multi_person_tracking/multi_camera_multi_person_tracking.py
```diff
@@ -50,28 +75,39 @@ def __call__(self):
...
-    config = {}
-    if len(params.config):
-        config = read_py_config(params.config)
+    frame_number = 0
+    avg_latency = AverageEstimator()
+    key = -1
+
+    if config['normalizer_config']['enabled']:
+        del config['normalizer_config']['enabled']
+        capture.add_transform(NormalizerCLAHE(**config['normalizer_config']))
```
Wovchena: That's confusing when I see `if something: delete that something`. Why not 
```py
if config['normalizer_config']['enabled']:
    capture.add_transform(NormalizerCLAHE(config['normalizer_config']['clip_limit'],
        config['normalizer_config']['tile_size']))
```

demos/python_demos/multi_camera_multi_person_tracking/tools/run_evaluate.py
```diff
@@ -0,0 +1,153 @@
...
+import argparse
+import json
+import logging as log
+
+import motmetrics as mm
+import numpy as np
+from lxml import etree
+from tqdm import tqdm
+
+from mc_tracker.sct import TrackedObj
```
Wovchena: This import works only if current directory is `open_model_zoo/demos/python_demos/multi_camera_multi_person_tracking`

demos/python_demos/multi_camera_multi_person_tracking/tools/run_evaluate.py
```diff
@@ -0,0 +1,153 @@
...
+        all_labels = [obj.label for obj in detections]
+        uniq = set(all_labels)
+        if len(all_labels) != len(uniq):
+            return True
+
+    return False
+
+
+def main():
+    """Computes MOT metrics for the person recognition demo"""
```
Wovchena: It seems the tools don't clarify where to get arguments for `--gt_files`. The tools look to be for a narrow circle of users who know what all this is about. Of course it is possible to dig into and find that this is about `motmetrics` and so on, but are users supposed to do this?

What is the difference between this tool and `run_history_visualize.py`? Both accept `--gt_files` both seem to compute the metric. Is it possible to merge the tools?

demos/python_demos/multi_camera_multi_person_tracking/tools/run_history_visualize.py
```diff
@@ -0,0 +1,249 @@
...
+    target_width, target_height = get_target_size(input_sizes, vis, max_window_size, stack_frames)
+    target_fps = min(fps)
+    return (target_width, target_height), target_fps
+
+
+def main():
+    """Prepares data for the person recognition demo"""
+    parser = argparse.ArgumentParser(description='Multi camera multi person \
+                                                  tracking visualization demo script')
+    parser.add_argument('--videos', type=str, nargs='+',
```
Wovchena: It should be `-i` to align with the demo.

demos/python_demos/multi_camera_multi_person_tracking/tools/run_history_visualize.py
```diff
@@ -0,0 +1,249 @@
...
+    if len(args.output_video):
+        video_output_size, fps = calc_output_video_params(frame_size, fps_source, args.gt_files, args.merge_outputs)
+        fourcc = cv.VideoWriter_fourcc(*'XVID')
+        output_video = cv.VideoWriter(args.output_video, fourcc, fps, video_output_size)
+        if args.gt_files and not args.merge_outputs:
+            ext = args.output_video.split('.')[-1]
+            output_path = args.output_video[:len(args.output_video) - len(ext) - 1] + '_gt.' + ext
+            output_video_gt = cv.VideoWriter(output_path, fourcc, fps, video_output_size)
+
+    # Create GT tracks if necessary
```
Wovchena:
```suggestion
    # Read GT tracks if necessary
```

demos/python_demos/multi_camera_multi_person_tracking/utils/network_wrappers.py
```diff
@@ -11,10 +11,29 @@
...
+import cv2
+import numpy as np
+
 from utils.ie_tools import load_ie_model
+from .segm_postprocess import postprocess
+
+
+class DetectorInterface(ABC):
+    @abstractmethod
+    def run_asynch(self, frames, index):
```
Wovchena: `async` instead of `asynch` is better:
1. It is a common way to abbreviate the word like that
2. This is how dldt abbreviates it

demos/python_demos/multi_camera_multi_person_tracking/README.md
```diff
@@ -40,55 +41,102 @@ pip3 install -r requirements.txt
...
+                                             [--m_segmentation M_SEGMENTATION]
+                                             [--t_segmentation T_SEGMENTATION]
                                              --m_reid M_REID
                                              [--output_video OUTPUT_VIDEO]
-                                             [--config CONFIG]
                                              [--history_file HISTORY_FILE]
-                                             [-d DEVICE] [-l CPU_EXTENSION]
+                                             [--no_show] [-d DEVICE]
+                                             [-l CPU_EXTENSION]
+
```
Wovchena:
```suggestion
```

demos/python_demos/multi_camera_multi_person_tracking/README.md
IRDonch: Alright, much better.

demos/python_demos/multi_camera_multi_person_tracking/multi_camera_multi_person_tracking.py
IRDonch: That's exactly why you should only call it if there is a window. :-)

demos/python_demos/multi_camera_multi_person_tracking/multi_camera_multi_person_tracking.py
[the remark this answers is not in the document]
IRDonch: Fixed how? I don't think you changed anything.

demos/python_demos/multi_camera_multi_person_tracking/tools/run_evaluate.py
```diff
@@ -0,0 +1,153 @@
...
+                ht_boxes.append([obj.rect[0], obj.rect[1],
+                                 obj.rect[2] - obj.rect[0],
+                                 obj.rect[3] - obj.rect[1]])
+                ht_labels.append(obj.label)
+
+            distances = mm.distances.iou_matrix(np.array(gt_boxes),
+                                                np.array(ht_boxes), max_iou=0.5)
+            accs[i].update(gt_labels, ht_labels, distances)
+
+        time += 1
```
IRDonch: This looks useless.

demos/python_demos/multi_camera_multi_person_tracking/tools/run_history_visualize.py
```diff
@@ -0,0 +1,249 @@
...
+        max_id = 0
+        for track in tracks:
+            if track['id'] > max_id:
+                max_id = track['id']
+        return max_id
+
+    output = []
+    for cam_tracks in all_tracks:
+        output.append(find_max(cam_tracks))
+    return max(output)
```
IRDonch: Isn't all of this basically `return max(track['id'] for cam_tracks in all_tracks for track in cam_tracks)`?

demos/python_demos/multi_camera_multi_person_tracking/tools/run_history_visualize.py
```diff
@@ -0,0 +1,249 @@
...
+
+
+def find_max_frame_num(tracks):
+    output = [0 for _ in tracks]
+    for i, cam_tracks in enumerate(tracks):
+        for track in cam_tracks:
+            output[i] = max(output[i], track['timestamps'][-1])
+    return min(output)
```
IRDonch: `return min(max(track['timestamps'][-1] for track in cam_tracks) for cam_tracks in tracks)`?

demos/python_demos/multi_camera_multi_person_tracking/multi_camera_multi_person_tracking.py
DmitriySidnev: OK, done.

demos/python_demos/multi_camera_multi_person_tracking/multi_camera_multi_person_tracking.py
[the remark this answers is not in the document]
DmitriySidnev: There are two lines with it. Now fixed both of them.

demos/python_demos/multi_camera_multi_person_tracking/README.md
```diff
@@ -75,20 +92,46 @@ Minimum command examples to run the demo:
...
     --config config.py
 ```
 
+The demo can use file in JSON format with detections instead person detector.
+Structure of this file should be the next:
+```
+[
+    {
+        'frame_id': 0
+        'boxes': [[x0, y0, x1, y1], [x0, y0, x1, y1], ...]  # N bounding boxes
```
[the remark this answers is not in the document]
DmitriySidnev: Fixed.

demos/python_demos/multi_camera_multi_person_tracking/multi_camera_multi_person_tracking.py
```diff
@@ -17,18 +17,42 @@
...
+        '--m_detector': args.m_detector,
+        '--m_segmentation': args.m_segmentation,
+        '--detections': args.detections
+    }
+    det_counter = 0
+    for _, value in detectors.items():
+        if value:
+            det_counter += 1
+    if det_counter == 0:
+        log.error('No detector specified, please specify one of the next parameters: '
```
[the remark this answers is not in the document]
DmitriySidnev: Fixed.

demos/python_demos/multi_camera_multi_person_tracking/multi_camera_multi_person_tracking.py
```diff
@@ -17,18 +17,42 @@
...
+                  '\'--m_detector\', \'--m_segmentation\' or \'--detections\'')
+    elif det_counter > 1:
+        non_empty_detectors = [{det: value} for det, value in detectors.items() if value]
```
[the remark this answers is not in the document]
DmitriySidnev: Fixed.

demos/python_demos/multi_camera_multi_person_tracking/multi_camera_multi_person_tracking.py
```diff
@@ -17,18 +17,42 @@
...
+        det_string = ''
+        for det in non_empty_detectors:
+            det_string += '\n\t{}'.format(det)
```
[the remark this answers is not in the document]
DmitriySidnev: Fixed.

demos/python_demos/multi_camera_multi_person_tracking/utils/analyzer.py
```diff
@@ -0,0 +1,222 @@
...
+                if use_images:
+                    label_img_clust = make_label_img(label_img_clust, track.crops[j])
+
+    summary_writer.add_embedding(embeddings_all, metadata=metadata_all,
+                                 label_img=label_img_all, global_step=step, tag='All')
+    summary_writer.add_embedding(embeddings_avg, metadata=metadata_avg,
+                                 label_img=label_img_avg, global_step=step, tag='Average')
+    summary_writer.add_embedding(embeddings_clust, metadata=metadata_clust,
+                                 label_img=label_img_clust, global_step=step, tag='Clustered')
+    log.info('Embeddings have saved successfully. To see the result use the next command: '
```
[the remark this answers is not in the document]
DmitriySidnev: Fixed.

demos/python_demos/multi_camera_multi_person_tracking/utils/analyzer.py
```diff
@@ -0,0 +1,222 @@
...
+                          name='SCT', show_online=self.plot_timeline_freq)
+
+
+def save_embeddings(scts, save_path, use_images=False, step=0):
+    def make_label_img(label_img, crop, target_size=(32, 32)):
+        img = cv2.resize(crop, target_size)  # Resize, size must be square
+        img = cv2.cvtColor(img, cv2.COLOR_BGR2RGB)  # BGR to RGB
+        img = torch.Tensor(img.transpose(2, 0, 1)) / 255  # Scale
+        label_img = img.unsqueeze(0) if label_img is None else \
+            torch.cat((label_img, img.unsqueeze(0)))
```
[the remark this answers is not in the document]
DmitriySidnev: Unfortunately, yes. `tensorboard` does not support `numpy` array. At least in version of `tensorboard` I used when implemented it.

demos/python_demos/multi_camera_multi_person_tracking/utils/misc.py
```diff
@@ -41,10 +75,20 @@ def read_py_config(filename):
     return cfg_dict
 
 
-def none_to_zero(a):
-    if a is not None:
-        return a
-    return 0
+def check_pressed_keys(key):
+    if key != 32:  # If was pressed space, use per-frame mode via pressing space key
```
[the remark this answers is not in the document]
DmitriySidnev: The comment removed.

demos/python_demos/multi_camera_multi_person_tracking/multi_camera_multi_person_tracking.py
DmitriySidnev: Fixed.

demos/python_demos/multi_camera_multi_person_tracking/utils/network_wrappers.py
DmitriySidnev: Fixed.

demos/python_demos/multi_camera_multi_person_tracking/utils/network_wrappers.py
DmitriySidnev: Cleaned.

demos/python_demos/multi_camera_multi_person_tracking/utils/network_wrappers.py
DmitriySidnev: `Normalize` removed.

demos/python_demos/multi_camera_multi_person_tracking/README.md
DmitriySidnev: Fixed.

demos/python_demos/multi_camera_multi_person_tracking/multi_camera_multi_person_tracking.py
DmitriySidnev: Fixed.

demos/python_demos/multi_camera_multi_person_tracking/tools/run_evaluate.py
DmitriySidnev: Fixed. My fix probably is not so pretty but it resolves the problem.

demos/python_demos/multi_camera_multi_person_tracking/tools/run_history_visualize.py
DmitriySidnev: Fixed.

demos/python_demos/multi_camera_multi_person_tracking/tools/run_history_visualize.py
DmitriySidnev: Fixed.

demos/python_demos/multi_camera_multi_person_tracking/utils/network_wrappers.py
DmitriySidnev: Fixed.

demos/python_demos/multi_camera_multi_person_tracking/README.md
DmitriySidnev: Fixed.

demos/python_demos/multi_camera_multi_person_tracking/README.md
```diff
@@ -40,55 +41,102 @@ pip3 install -r requirements.txt
...
+    --m_reid path/to/person-reidentification-retail-0103.xml \
     --config config.py
 ```
 
+The demo can use a JSON file with detections instead of a person detector.
+The structure of this file should be as follows:
+```json
+[
+    {
+        'frame_id': 0
```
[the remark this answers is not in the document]
DmitriySidnev: Fixed.

demos/python_demos/multi_camera_multi_person_tracking/tools/run_evaluate.py
DmitriySidnev: Removed.

demos/python_demos/multi_camera_multi_person_tracking/tools/run_history_visualize.py
DmitriySidnev: Fixed.

demos/python_demos/multi_camera_multi_person_tracking/tools/run_history_visualize.py
DmitriySidnev: Fixed.

demos/python_demos/multi_camera_multi_person_tracking/utils/network_wrappers.py
DmitriySidnev: Actually purpose of these two files is different. `--detections` has a simple format to make it possible create file with detected bounding boxes in easy way and use it instead some detector that helps execute demo faster or, for example, check quality of your person ReID-model. I suppose that this file must created by everyone in the simplest way. `--history_file` firstly consists of data obtained from all videos which were used in the demo, then the data is grouped by `IDs` of persons. We can't transfer `--history_file` to `--detections` format, `--detections` to `--history_file` is possible but I think for user it is not useful.

demos/python_demos/multi_camera_multi_person_tracking/multi_camera_multi_person_tracking.py
```diff
@@ -81,74 +118,120 @@ def run(params, capture, detector, reid):
...
+    print('')
+
     thread_body.process = False
     frames_thread.join()
 
     if len(params.history_file):
         history = tracker.get_all_tracks_history()
+        save_dir = os.path.dirname(params.history_file)
+        if not os.path.exists(save_dir):
+            os.makedirs(save_dir)
```
Wovchena: This fails for me if I run the demo with `--history_file a.txt`

demos/python_demos/multi_camera_multi_person_tracking/tools/run_evaluate.py
Wovchena: Updating `sys.path` is usually considered as a hack. I suggest moving the tools up one folder to stay at the level of the demo.

demos/python_demos/multi_camera_multi_person_tracking/utils/network_wrappers.py
```diff
@@ -81,3 +105,143 @@ def forward(self, batch):
...
+        self.frames = frames
+
+    def wait_and_grab(self):
+        return self.get_detections(self.frames)
+
+    def resize(self, sample):
+        image_height, image_width = sample['image'].shape[:2]
+        scale = min(self.h / image_height, self.w / image_width)
+        sample['image'] = cv2.resize(sample['image'], None, fx=scale, fy=scale)
+        sample['image'] = sample['image'].astype('float32').transpose(2, 0, 1)
```
Wovchena: Type casting and transposition doesn't belong to resizing. I suggest moving the line into `preprocess(self, frame)`.

demos/python_demos/multi_camera_multi_person_tracking/utils/network_wrappers.py
Wovchena: > this file must created by everyone in the simplest way

This doesn't seem to be simple. Where does the file for `--detections` come from, maybe there a script to build the file or it can be found in some dataset. Is it possible to clarify/simplify this? There is a note in Readme, but I think it is not very useful because it still requires saving detections somehow.

demos/python_demos/multi_camera_multi_person_tracking/multi_camera_multi_person_tracking.py
DmitriySidnev: Fixed.

demos/python_demos/multi_camera_multi_person_tracking/tools/run_evaluate.py
DmitriySidnev: `run_evaluate.py` evaluates metrics for multi camera multi person tracker. Here you must have annotated data to compare your results with ground truth. This demo show online IDs for every bounding box and all bounding boxes are merged to tracks according to the ID. Our algorithm merges tracks by some criterion and the final result may be different from the  showed online. So, to visualize this final result we use `run_history_visualize.py` script. Also here is an option to show your results and ground truth together to compare it visually. Yes, these two scripts have common functions but I would not like to merge them. They have different purpose although both show quality of the algorithm. What about `--gt_files`, it is possible to obtain them only via annotation, we can't provide it. We use here annotation from CVAT.

demos/python_demos/multi_camera_multi_person_tracking/utils/network_wrappers.py
DmitriySidnev: Moved code from `resize` to `preprocess`.

demos/python_demos/multi_camera_multi_person_tracking/tools/run_evaluate.py
DmitriySidnev: OK, moved to parent directory.

demos/python_demos/multi_camera_multi_person_tracking/tools/run_evaluate.py
Wovchena: Will you consider adding your clarification above, so other people at least had a chance to figure out how to use it?

demos/python_demos/multi_camera_multi_person_tracking/utils/network_wrappers.py
DmitriySidnev: OK, you won. I have added option to save detections.

Wovchena: This fails for me at `json.dump(data, outfile)`.

Wovchena: I suggest replacing `entry['scores'].append(det[1])` with `entry['scores'].append(det[1].item())`.

demos/python_demos/multi_camera_multi_person_tracking/utils/analyzer.py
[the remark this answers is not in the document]
IRDonch: Only half-fixed. :-)

demos/python_demos/multi_camera_multi_person_tracking/multi_camera_multi_person_tracking.py
```diff
@@ -17,18 +17,56 @@
...
+        '--m_segmentation': args.m_segmentation,
+        '--detections': args.detections
+    }
+    non_empty_detectors = [(det, value) for det, value in detectors.items() if value]
+    det_number = len(non_empty_detectors)
+    if det_number == 0:
+        log.error('No detector specified, please specify one of the following parameters: '
+                  '\'--m_detector\', \'--m_segmentation\' or \'--detections\'')
+    elif det_number > 1:
+        det_string = ''.join('\n\t{} = {}'.format(det[0], det[1]) for det in non_empty_detectors)
```
IRDonch:
```suggestion
        det_string = ''.join('\n\t{}={}'.format(det[0], det[1]) for det in non_empty_detectors)
```

You don't use spaces when specifying the parameters, so they should also not be used when printing them.

demos/python_demos/multi_camera_multi_person_tracking/utils/analyzer.py
```diff
@@ -0,0 +1,222 @@
...
+                    label_img_clust = make_label_img(label_img_clust, track.crops[j])
+
+    summary_writer.add_embedding(embeddings_all, metadata=metadata_all,
+                                 label_img=label_img_all, global_step=step, tag='All')
+    summary_writer.add_embedding(embeddings_avg, metadata=metadata_avg,
+                                 label_img=label_img_avg, global_step=step, tag='Average')
+    summary_writer.add_embedding(embeddings_clust, metadata=metadata_clust,
+                                 label_img=label_img_clust, global_step=step, tag='Clustered')
+    log.info('Embeddings have been saved successfully. To see the result use the next command: '
+             'tensorboard --logdir={}'.format(save_path))
```
IRDonch: The `tensorboard` script is installed by the `tensorboard` package, while this module uses `tensorboardX`. Is it really necessary to install both?

demos/python_demos/multi_camera_multi_person_tracking/utils/analyzer.py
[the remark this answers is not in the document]
IRDonch: The [documentation](https://tensorboard-pytorch.readthedocs.io/en/latest/tensorboard.html#tensorboardX.SummaryWriter.add_embedding) claims that NumPy arrays are supported. Maybe you could try with a newer version?

demos/python_demos/multi_camera_multi_person_tracking/utils/network_wrappers.py
DmitriySidnev: > This fails for me at `json.dump(data, outfile)`.

Could you please share a full command that leads to the error?

DmitriySidnev: > I suggest replacing `entry['scores'].append(det[1])` with `entry['scores'].append(det[1].item())`.

Please clarify why. `det` is always `list` or `tuple` and `det[1]` is always `float32`, method `item()` failed for `tuple`.

demos/python_demos/multi_camera_multi_person_tracking/utils/analyzer.py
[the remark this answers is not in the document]
DmitriySidnev: Yes, did not see. Fixed.

demos/python_demos/multi_camera_multi_person_tracking/tools/run_evaluate.py
DmitriySidnev: I have added description in readme for these two scripts with format of the annotation for the ground truth files.

demos/python_demos/multi_camera_multi_person_tracking/multi_camera_multi_person_tracking.py
DmitriySidnev: Fixed.

demos/python_demos/multi_camera_multi_person_tracking/utils/analyzer.py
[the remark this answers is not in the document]
DmitriySidnev: Yes, it works with version 2.0 of `tensorboardX`, but honestly I tried to use numpy arrays early. So, I removed PyTorch module.

demos/python_demos/multi_camera_multi_person_tracking/utils/analyzer.py
DmitriySidnev: `tensorboard` does not provide `SummaryWriter` class that is available in `tensorboardX`. `tensorboard` can be used with `tensorflow` or `pytorch` while `tensorboardX` is a wrapper for `tensorboard` module and is used independently.

IRDonch: So what you're saying is "yes". :-)

IRDonch:
Jenkins: test this please

IRDonch:
@DmitriySidnev For the tests to passs, you need to update `demos/python_demos/requirements.txt`, as well as `ci/requirements-demos.txt` (see `ci/update-requirements.py`).

IRDonch:
Note: please use `pip-tools<4.5.0` when updating `ci/requirements-demos.txt`, because 4.5.0 introduced an annoying change to the output format, which I'd rather avoid for now.

demos/python_demos/multi_camera_multi_person_tracking/utils/network_wrappers.py
Wovchena: There is the command:
`python3 open_model_zoo/demos/python_demos/multi_camera_multi_person_tracking/multi_camera_multi_person_tracking.py --m_reid /mnt/h/d/downloader/intel/person-reidentification-retail-0200/FP32/person-reidentification-retail-0200.xml -d CPU -i /home/vzlobin/Videos/cr/test_video_6.avi /home/vzlobin/Videos/cr/test_video_6.avi --config open_model_zoo/demos/python_demos/multi_camera_multi_person_tracking/config.py --m_segmentation /mnt/h/d/downloader/intel/instance-segmentation-security-0050/FP32/instance-segmentation-security-0050.xml --save_detections detections.txt`
It fails after pressing Esc.

Numpy's array cant be saved to json.

demos/python_demos/multi_camera_multi_person_tracking/tools/run_evaluate.py
Wovchena: Thank you

demos/python_demos/multi_camera_multi_person_tracking/utils/network_wrappers.py
DmitriySidnev: Fixed.

DmitriySidnev:
Jenkins: test this please

DmitriySidnev:
@IRDonch, I updated `python_demos/requirements.txt` like in the `ci/update-requirements.py` script. Finally, there appeared a lot of python modules with versions and comments. Is it right? Could you please check it?

IRDonch:
> Is it right?

No. You need to put _direct_ requirements in `python_demos/requirements.txt` and then _run_ the `ci/update-requirements.py` script to update `ci/requirements-demos.txt`.

DmitriySidnev:
@IRDonch, I fixed `python_demos/requirements` and `ci/requirements*` via `ci/update-requirements.py` (all requirements in `ci` directory). Could you please check it one more time?

demos/python_demos/multi_camera_multi_person_tracking/requirements.txt
```diff
@@ -1,3 +1,11 @@
-opencv-python
-numpy
-scipy
+Cython>=0.29.12
```
IRDonch: Cython doesn't seem to be used anywhere.

demos/python_demos/multi_camera_multi_person_tracking/run_evaluate.py
```diff
@@ -0,0 +1,151 @@
...
+ limitations under the License.
+"""
+
+import argparse
+import json
+import logging as log
+
+import motmetrics as mm
+import numpy as np
+from lxml import etree
```
IRDonch: Is the built-in `xml.etree.ElementTree` module insufficient here?

IRDonch:
@DmitriySidnev It mostly looks fine now, but you probably should not have committed the rest of the `ci/requirements-*.txt` files, because now it'll take absolutely forever to run CI on this. I suggest you revert the changes to the unrelated requirements files.

demos/python_demos/multi_camera_multi_person_tracking/requirements.txt
DmitriySidnev: You are right. Removed.

demos/python_demos/multi_camera_multi_person_tracking/run_evaluate.py
DmitriySidnev: Yes, it is enough. I did not now about built-in xml parser. `lxml` removed.

DmitriySidnev:
@IRDonch, all `ci/requirements*` files except `requirements-demo.txt` reverted.

demos/python_demos/requirements.txt
```diff
@@ -1,5 +1,10 @@
+matplotlib
+motmetrics
 nibabel
 numpy
 scikit-learn
 scipy
+setuptools
```
IRDonch: It doesn't make sense to put `setuptools` here, because you need `setuptools` to use a requirements file to begin with.

DmitriySidnev: Removed.

IRDonch:
Jenkins: test this please

DmitriySidnev:
Jenkins: test this please

DmitriySidnev:
@IRDonch, seems that `ci` failed with wrong path to config file. I fixed the path and made it absolute.

IRDonch:
Jenkins: test this please

PR #606, opened 2019-11-11
OpenCV-ify object_detection_demo_faster_rcnn

IRDonch:
This allows us to remove a ton of redundant code, at the cost of a couple slight disadvantages:

* the demo now requires OpenCV to build, which shouldn't be a problem,   since nearly every other demo does so as well (the only exception being segmentation_demo);

* the demo no longer accepts files in the MNIST format as input, which also shouldn't be a problem, since it doesn't make sense to run object detection on the MNIST dataset.

### DIFF
--- a/demos/object_detection_demo_faster_rcnn/main.cpp
+++ b/demos/object_detection_demo_faster_rcnn/main.cpp
@@ -10,13 +10,16 @@
 #include <vector>
 #include <limits>
 
-#include <format_reader_ptr.h>
 #include <inference_engine.hpp>
 #ifdef WITH_EXTENSIONS
 #include <ext_list.hpp>
 #endif
 
-#include <samples/common.hpp>
+#include <opencv2/core.hpp>
+#include <opencv2/imgcodecs.hpp>
+#include <opencv2/imgproc.hpp>
+
+#include <samples/ocv_common.hpp>
 #include <samples/slog.hpp>
 #include <samples/args_helper.hpp>
 #include "object_detection_demo_faster_rcnn.h"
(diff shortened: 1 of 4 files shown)


demos/object_detection_demo_faster_rcnn/main.cpp
```diff
@@ -373,8 +352,7 @@ int main(int argc, char *argv[]) {
 
         const float* detection = static_cast<PrecisionTrait<Precision::FP32>::value_type*>(output_blob->buffer());
 
-        std::vector<std::vector<int> > boxes(batchSize);
-        std::vector<std::vector<int> > classes(batchSize);
+        std::vector<std::vector<DetectedObject>> objects(batchSize);
```
Wovchena: With this line `DetectedObject` gets an excuse why it is not still deleted form common.hpp :)

But I doubt we need `DetectedObject` here. A rectangle can be drawn just after extracting its coordinates in the same loop. This approach simplifies the demo.

IRDonch: What's your problem with `DetectedObject`? :-) I think that's actually a pretty useful thing to have, since this structure reoccurs in many models.

In this case you're right, though. I simplified the demo as you suggested.

Wovchena: > I think that's actually a pretty useful thing to have, since this structure reoccurs in many models.

Yes, but each time you need a slightly different struct in each particular case. Demos that need that kind of struct usually redesign it. For example, `difficult` field in `DetectedObject` is not usually needed.

PR #622, opened 2019-11-15
AC: pipeline evaluator write result update

eaidova:


eaidova:
@vladimir-dudnik, please merge ASAP, action-recognition-* models validation broken without it.

PR #624, opened 2019-11-18
AC: support 3 dim tensors for ONNX runtime

eaidova:
make onnx runtime launcher work when batch is ommited.
it is necessary for Pytorch/ONNX mask rcnn support (e.g this [model](https://github.com/onnx/models/tree/master/vision/object_detection_segmentation/mask-rcnn))

### DIFF
--- a/tools/accuracy_checker/accuracy_checker/launcher/onnx_launcher.py
+++ b/tools/accuracy_checker/accuracy_checker/launcher/onnx_launcher.py
@@ -78,10 +78,13 @@ class ONNXLauncher(Launcher):
 
         return results
 
-    @staticmethod
-    def fit_to_input(data, layer_name, layout):
+    def fit_to_input(self, data, layer_name, layout):
+        layer_shape = self.inputs[layer_name]
         if len(np.shape(data)) == 4:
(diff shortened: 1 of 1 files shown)


tools/accuracy_checker/accuracy_checker/launcher/onnx_launcher.py
```diff
@@ -78,10 +78,13 @@ def predict(self, inputs, metadata=None, **kwargs):
...
 
-    @staticmethod
-    def fit_to_input(data, layer_name, layout):
+    def fit_to_input(self, data, layer_name, layout):
+        layer_shape = self.inputs[layer_name]
         if len(np.shape(data)) == 4:
-            return np.transpose(data, layout).astype(np.float32)
+            data = np.transpose(data, layout).astype(np.float32)
+            if len(layer_shape) == 3:
+                return data[0]
```
IRDonch: I think there should be a check (or at least an assert) that `data` has size 1 in the first dimension. Otherwise, we could silently lose data.

eaidova: added

PR #630, opened 2019-11-20
AC: fix segmentation metrics for non-existed classes

eaidova:
metrics should ignore classes where predictions do not exists. Result finalization skip only nan, inf values and treat zero as valid result while for these metrics it means that label should be ignored (icnet/unet trained on camvid do not have objects for unlabled class and it should not affect metric result)

### DIFF
--- a/tools/accuracy_checker/tests/test_segmentation_metrics.py
+++ b/tools/accuracy_checker/tests/test_segmentation_metrics.py
@@ -87,7 +87,7 @@ class TestMeanAccuracy:
         dataset = single_class_dataset()
         dispatcher = MetricsExecutor(create_config(self.name), dataset)
         dispatcher.update_metrics_on_batch(range(len(annotations)), annotations, predictions)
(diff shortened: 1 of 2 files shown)


tools/accuracy_checker/tests/test_segmentation_metrics.py
```diff
@@ -87,7 +87,7 @@ def test_one_class(self):
         dataset = single_class_dataset()
         dispatcher = MetricsExecutor(create_config(self.name), dataset)
         dispatcher.update_metrics_on_batch(range(len(annotations)), annotations, predictions)
-        expected = generate_expected_result([1.0, 0.0], self.name, dataset.labels)
+        expected = generate_expected_result([1.0], self.name, {0: 'dog'})
```
IRDonch: I think it would be clearer if you used `dataset.labels[0]` instead of `'dog'` (and similarly in the other cases).

eaidova: yes, thank you, updated

PR #631, opened 2019-11-20
LW AV: update configs for LPR models

eaidova:


eaidova:
@vladimir-dudnik, could you please merge? It affects models calibration

PR #632, opened 2019-11-20
AC: allow merge calibration config with path prefix and definitions

eaidova:
allow merge content of json config (which has relation with AC) using env vars


eaidova:
@vladimir-dudnik, if you do not have objestion, could you please merge this PR?
It helps to start integration json configs on benchmark side

PR #644, opened 2019-11-25
AC: support MO transformations config option

eaidova:
`--tf_custom_operation_config` option in MO is deprecated in this release.  (now in the latest versions DLDT, it raises deprecation warning). MO team recommend to replace it with `--transformations_config` 

P.S. Probably it also should affect Model Downloader configs for some TF models.

### DIFF
--- a/tools/accuracy_checker/accuracy_checker/launcher/dlsdk_launcher.py
+++ b/tools/accuracy_checker/accuracy_checker/launcher/dlsdk_launcher.py
@@ -200,7 +200,7 @@ class DLSDKLauncher(Launcher):
                 optional=True, is_directory=True, description="TF Custom Operation Config prefix."
             ),
             '_transformations_config_dir': PathField(
-                optional=True, is_directory=True, description="Transformation config for Model Optimizer"),
+                optional=True, is_directory=True, description="Transformation config prefix for Model Optimizer"),
             '_tf_obj_detection_api_pipeline_config_path': PathField(
                 optional=True, is_directory=False, description="TF Custom Operation Pipeline Config."),
             '_cpu_extensions_mode': StringField(optional=True, description="CPU extensions mode."),
--- a/tools/accuracy_checker/accuracy_checker/launcher/dlsdk_launcher.py
+++ b/tools/accuracy_checker/accuracy_checker/launcher/dlsdk_launcher.py
@@ -197,8 +197,10 @@ class DLSDKLauncher(Launcher):
                 optional=True, is_directory=True, description="TF Object Detection API Config."
             ),
             '_tf_custom_op_config_dir': PathField(
-                optional=True, is_directory=True, description="TF Custom Operation Config."
+                optional=True, is_directory=True, description="TF Custom Operation Config prefix."
             ),
+            '_transformations_config_dir': PathField(
+                optional=True, is_directory=False, description="Transformation config for Model Optimizer"),
             '_tf_obj_detection_api_pipeline_config_path': PathField(
                 optional=True, is_directory=False, description="TF Custom Operation Pipeline Config."),
             '_cpu_extensions_mode': StringField(optional=True, description="CPU extensions mode."),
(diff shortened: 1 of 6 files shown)


eaidova:
> pls see error message:
15:00:10 accuracy_checker/launcher/model_conversion.py:186: [C0326(bad-whitespace), ] Exactly one space required after comma
15:00:10 mo_params, framework, tf_custom_op_config_dir, transformations_config_dir, mo_path

@vladimir-dudnik, thank you, I already fix it.

tools/accuracy_checker/accuracy_checker/launcher/model_conversion.py
```diff
@@ -177,24 +184,35 @@ def exec_mo_binary(args, timeout=None, should_log_cmd=False):
...
         return mo_params
 
     if tf_custom_op_config_dir:
-        tf_custom_op_config_dir = Path(tf_custom_op_config_dir)
+        prefix_dir = Path(tf_custom_op_config_dir)
+    elif transformations_config_dir:
+        prefix_dir = Path(transformations_config_dir)
     else:
-        tf_custom_op_config_dir = Path('/').joinpath(*mo_path.parts[:-1]) / 'extensions' / 'front' / 'tf'
+        prefix_dir = Path('/').joinpath(*mo_path.parts[:-1]) / 'extensions' / 'front' / 'tf'
```
IRDonch: This is a fine default directory for TF configs, but presumably the `--transformations_config` option will be usable for frameworks other than TF, so shouldn't the default prefix for that be more generic?

tools/accuracy_checker/accuracy_checker/main.py
```diff
@@ -108,7 +108,15 @@ def build_arguments_parser():
     )
     parser.add_argument(
         '--tf_custom_op_config_dir',
-        help='path to directory with tensorflow custom operation configuration files for model optimizer',
+        help='path to directory with Model Optimizer transformation configuration files',
```
IRDonch: Are you sure you want to edit this description? I think the old description is more apt.

tools/accuracy_checker/accuracy_checker/launcher/model_conversion.py
```diff
@@ -177,24 +184,35 @@ def exec_mo_binary(args, timeout=None, should_log_cmd=False):
...
 
-    config_path = Path(config_path)
-    if not config_path.is_absolute():
-        config_path = tf_custom_op_config_dir / config_path
+    if tf_custom_op_config_path:
+        config_path = Path(tf_custom_op_config_path)
+        if not config_path.is_absolute():
+            config_path = prefix_dir / config_path
+            mo_params['tensorflow_use_custom_operations_config'] = str(get_path(config_path))
+        return mo_params
```
IRDonch: This `return` is IMO not a good thing. It means that if both the TF config and generic config are specified, the latter will be silently ignored, which is surprising behavior. AC should either pass both options to MO or complain and refuse to do anything.

tools/accuracy_checker/accuracy_checker/main.py
eaidova: returned back

tools/accuracy_checker/accuracy_checker/launcher/model_conversion.py
eaidova: options separated, I do not want be responsible on verifying MO arguments (it is their obligation)

tools/accuracy_checker/accuracy_checker/launcher/model_conversion.py
eaidova: changed

tools/accuracy_checker/accuracy_checker/launcher/dlsdk_launcher.py
```diff
@@ -197,8 +197,10 @@ def parameters(cls):
                 optional=True, is_directory=True, description="TF Object Detection API Config."
             ),
             '_tf_custom_op_config_dir': PathField(
-                optional=True, is_directory=True, description="TF Custom Operation Config."
+                optional=True, is_directory=True, description="TF Custom Operation Config prefix."
             ),
+            '_transformations_config_dir': PathField(
+                optional=True, is_directory=True, description="Transformation config for Model Optimizer"),
```
IRDonch: Shouldn't the description say it's a prefix, like it does for `_tf_custom_op_config_dir`?

eaidova: okey, added

PR #649, opened 2019-11-27
AC: support Yolo v3 with mask parameter in output

eaidova:
same yolo v3 models has parameter `mask` for selection anchors for specific output layer.
yolo v3 demo is able to handle this case correctly because it check layer parameters and process using OpenVINO specifically approach.
Unforutnately, accuracy checker detections processing is framework agnostic. It is the reason why we need to put this parameter explicitly in the config.


eaidova:
@vladimir-dudnik, could you please approve ASAP? It is extremely important for OpenVINO team

PR #650, opened 2019-11-27
AC: CenterNet support

jkamelin:
Support of [CenterNet](https://github.com/xingyizhou/CenterNet) ctdet_dlav0 trained on coco


tools/accuracy_checker/accuracy_checker/adapters/centernet.py
```diff
@@ -0,0 +1,161 @@
...
+        for i in range(dets.shape[0]):
+            dets[i, :, :2] = CTDETAdapter._transform_preds(
+                dets[i, :, 0:2], center[i], scale[i], (width, heigth))
+            dets[i, :, 2:4] = CTDETAdapter._transform_preds(
+                dets[i, :, 2:4], center[i], scale[i], (width, heigth))
+        return dets
+
+    def process(self, raw, identifiers=None, frame_meta=None):
+        result = []
+        hm = self._extract_predictions(raw, frame_meta)[self.hm]
```
eaidova: could you please extract predictions only one time and after that use already extracted?

tools/accuracy_checker/accuracy_checker/adapters/centernet.py
```diff
@@ -0,0 +1,161 @@
...
+
+class CTDETAdapter(Adapter):
+    __provider__ = 'ctdet'
+
+    @classmethod
+    def parameters(cls):
+        parameters = super().parameters()
+        parameters.update(
+            {
+                'hm': StringField(description="Object center points heatmap."),
```
eaidova: please give more descriptive name for parameters, if they mean output layer name, please add postfix `_out`

tools/accuracy_checker/accuracy_checker/adapters/centernet.py
```diff
@@ -0,0 +1,161 @@
...
+                                 xs + wh[..., 0:1] / 2,
+                                 ys + wh[..., 1:2] / 2), axis=2)
+        detections = np.concatenate((bboxes, scores, clses), axis=2)
+        detections = np.array(detections)
+        im_size = frame_meta[0].get('image_size')
+        scale = max(im_size)
+        center = np.array(im_size[:2])/2.0
+        dets = self._transform(detections, [np.flip(center)], [scale], height, width)
+        x_min, y_min, x_max, y_max, scores, classes = dets[0].transpose(1, 0)
+        result.append(DetectionPrediction(identifiers, classes, scores, x_min, y_min, x_max, y_max))
```
eaidova: identifiers, it is list of identifier, if model support only batch size 1, you need extract it
```python
identifiers[0]
```
if model can be evaluated with batch >1, output processing should be done in the cycle (one DetectionPrediction for one image in the batch)

tools/accuracy_checker/accuracy_checker/preprocessor/geometric_transformations.py
```diff
@@ -916,3 +916,94 @@ def process(self, image, annotation_meta=None):
...
+            'scale': NumberField(
+                value_type=int, optional=True, default=1,
+                description="Scale for input image"
+            )
+        })
+
+        return parameters
+
+    def configure(self):
+        self.input_height, self.input_width = get_size_from_config(self.config)
```
eaidova: get size from config also assume that `size` parameter also supported for case when dst_height and dts_width are equal

eaidova:
@jkamelin, please provide also update readme for added functionality (adapter/README.md, preprocessing/README.md)

PR #651, opened 2019-11-27
Simplify post-processing in object_detection_demo_faster_rcnn

IRDonch:
Originally I set out to remove the usage of a soon-to-be deprecated `InferenceEngine::Data` constructor (see the first commit), but then I noticed that the code that constructor is used in is rather superfluous, so I ended up refactoring the demo to remove it entirely.

See the commit messages for more information.


IRDonch:
#762 removes the deprecated function usage too, so I think I'm going to merge that instead and postpone this until the next release.

PR #652, opened 2019-11-27
pose_extractor: replace unnecessary vectors with arrays

IRDonch:
This is pretty much a port of 57fa87f5.

### DIFF
--- a/demos/python_demos/human_pose_estimation_3d_demo/pose_extractor/src/peak.cpp
+++ b/demos/python_demos/human_pose_estimation_3d_demo/pose_extractor/src/peak.cpp
@@ -112,11 +112,11 @@ std::vector<HumanPose> groupPeaksToPoses(const std::vector<std::vector<Peak> >& 
                                          const float foundMidPointsRatioThreshold,
                                          const int minJointsNumber,
                                          const float minSubsetScore) {
-    const std::vector<std::pair<int, int> > limbIdsHeatmap = {
+    static const std::pair<int, int> limbIdsHeatmap[] = {
         {2, 3}, {2, 6}, {3, 4}, {4, 5}, {6, 7}, {7, 8}, {2, 9}, {9, 10}, {10, 11}, {2, 12}, {12, 13}, {13, 14},
         {2, 1}, {1, 15}, {15, 17}, {1, 16}, {16, 18}, {3, 17}, {6, 18}
     };
-    const std::vector<std::pair<int, int> > limbIdsPaf = {
+    static const std::pair<int, int> limbIdsPaf[] = {
         {31, 32}, {39, 40}, {33, 34}, {35, 36}, {41, 42}, {43, 44}, {19, 20}, {21, 22}, {23, 24}, {25, 26},
         {27, 28}, {29, 30}, {47, 48}, {49, 50}, {53, 54}, {51, 52}, {55, 56}, {37, 38}, {45, 46}
     };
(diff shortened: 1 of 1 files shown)


demos/python_demos/human_pose_estimation_3d_demo/pose_extractor/CMakeLists.txt
```diff
@@ -18,8 +18,13 @@ add_library(${target_name} MODULE wrapper.cpp
...
                                   src/peak.hpp src/peak.cpp)
-target_include_directories(${target_name} PRIVATE src/ ${PYTHON_INCLUDE_DIRS} ${NUMPY_INCLUDE_DIR} ${OpenCV_INCLUDE_DIRS})
-target_link_libraries(${target_name} ${PYTHON_LIBRARIES} ${OpenCV_LIBS})
+target_include_directories(${target_name} PRIVATE
+    src "${Demos_SOURCE_DIR}/common"
+    ${PYTHON_INCLUDE_DIRS} ${NUMPY_INCLUDE_DIR} ${OpenCV_INCLUDE_DIRS})
+target_link_libraries(${target_name} PRIVATE
+    # we don't actually need Inference Engine, but the common headers depend on IE headers
+    ${InferenceEngine_LIBRARIES}
+    ${PYTHON_LIBRARIES} ${OpenCV_LIBS})
```
eizamaliev: Why did you increase dependencies for very insignificant changes? I dont see any sense here.

IRDonch: What would you rather have me do? `arraySize` is in `common`; I don't want to duplicate it. I don't see how a dependency on `common` is a bad thing, given that every other demo relies on it too.

eizamaliev: What about `${InferenceEngine_LIBRARIES}`? I understand that 
> we don't actually need Inference Engine, but the common headers depend on IE headers

but I don't see any impact of `arraySize` and its advantages over `vector.size()`.

IRDonch: > What about `${InferenceEngine_LIBRARIES}`?

The `pose_extractor` library doesn't actually become dependent on `inference_engine`, because it references no symbols from it. Although even if it did, I don't think it would be a problem, given that the demo (indirectly) depends on the IE library anyway.

> but I don't see any impact of `arraySize` and its advantages over `vector.size()`.

The advantage is that it works for arrays. :-) If you're asking why I want to use arrays here, it's because that allows us to avoid dynamic allocation and initialization each time the function is called.

IRDonch: @eizamaliev Ping?

eizamaliev: As I've already said, I dont see any reasonable impact of it which compensate increased dependency.

IRDonch: Which dependency do you object to?

eizamaliev: What wrong with `static const std::vector<std::pair<int, int> >`?
It can solve problem with 
>  dynamic allocation and initialization each time the function is called.

And will be the great compromise.

IRDonch: A static array would still better than a static vector in this instance, because it would be initialized at compile time. And I don't see why I should apply a half-measure.

IRDonch: OK, how about this? I just used `sizeof` to avoid pulling in the dependency.

eizamaliev: LGTM

PR #633, opened 2019-11-20
Add descriptions for upcoming reid models, remove old models

sovrasov:
@DmitriySidnev
@snosov1

PR #656, opened 2019-11-29
Accuracy checker configs for new re-identification models

DmitriySidnev:


DmitriySidnev:
@snosov1, @sovrasov

tools/accuracy_checker/configs/person-reidentification-retail-0103.yml
```diff
@@ -0,0 +1,44 @@
...
+        model:   intel/person-reidentification-retail-0103/INT8/person-reidentification-retail-0103.xml
+        weights: intel/person-reidentification-retail-0103/INT8/person-reidentification-retail-0103.bin
+        adapter: reid
+
+    datasets:
+      - name: globalme-reid
+        reader: pillow_imread
+
+        preprocessing:
+          - type: bgr_to_rgb
```
eaidova: original model has RGB input, but will be converted to IR with BGR, right?
pillow_imread reads images in RGB, so you need RGB -> BGR conversion. I suggest to rename preprocessor in order to reduce confusing.
```suggestion
          - type: rgb_to_bgr
```

tools/accuracy_checker/dataset_definitions.yml
```diff
@@ -385,6 +385,13 @@ datasets:
       data_dir: Market-1501-v15.09.15
     annotation: market1501_reid.pickle
 
+  - name: globalme-reid
+    data_source: GlobalMe-reID
```
eaidova: according dataset location, data_source should be globalme-reid/GlobalMe-reID
and data_dir for converter also.

tools/accuracy_checker/configs/person-reidentification-retail-0103.yml
DmitriySidnev: Fixed.

tools/accuracy_checker/dataset_definitions.yml
DmitriySidnev: Fixed.

eaidova:
jenkins ok to test

tools/accuracy_checker/configs/person-reidentification-retail-0103.yml
```diff
@@ -0,0 +1,44 @@
...
+        model:   intel/person-reidentification-retail-0103/INT8/person-reidentification-retail-0103.xml
+        weights: intel/person-reidentification-retail-0103/INT8/person-reidentification-retail-0103.bin
+        adapter: reid
+
+    datasets:
+      - name: globalme-reid
+        reader: pillow_imread
+
+        preprocessing:
+          - type: rgb_to_bgr
```
IRDonch: Since we established in #633 that the models have BGR inputs after all, this should no longer be needed.

DmitriySidnev: Fixed.

PR #669, opened 2019-12-04
AC: support Yolo v3 with mask parameter in output

eaidova:


eaidova:
@IRDonch , @vladimir-dudnik it is fix for `master` described via email, due to updated pylint version, validation fails (you can see that  errors detected in different from modified file).
Should I fix them or can stay as is?

IRDonch:
> Should I fix them or can stay as is?

I think it's fine as-is.

This is a backport, right? Could you link the commit it's backported from?

eaidova:
>This is a backport, right? Could you link the commit it's backported from?

Yes, it is backport: https://github.com/opencv/open_model_zoo/pull/649

eaidova:
Great news! Problem was solved without OMZ master patching :)

PR #670, opened 2019-12-04
demos: remove useless comments from command line flags and help messages

IRDonch:
They take up space and carry no information that isn't already present in the help messages themselves.

### DIFF
--- a/demos/segmentation_demo/segmentation_demo.h
+++ b/demos/segmentation_demo/segmentation_demo.h
@@ -19,8 +19,7 @@ static const char custom_cldnn_message[] = "Required for GPU custom kernels. "\
                                            "Absolute path to the .xml file with the kernels descriptions.";
 static const char custom_cpu_library_message[] = "Required for CPU custom layers. " \
                                                  "Absolute path to a shared library with the kernels implementations.";
-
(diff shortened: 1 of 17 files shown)


demos/segmentation_demo/segmentation_demo.h
```diff
@@ -9,54 +9,24 @@
...
-
-/// @brief message for user library argument
+                                           "Absolute path to the .xml file with the kernels descriptions.";
 static const char custom_cpu_library_message[] = "Required for CPU custom layers. " \
-"Absolute path to a shared library with the kernels implementations.";
-
-/// @brief message for config argument
-static constexpr char config_message[] = "Path to the configuration file. Default value: \"config\".";
+                                                 "Absolute path to a shared library with the kernels implementations.";
+static const char config_message[] = "Path to the configuration file. Default value: \"config\".";
```
Wovchena: Why not replace all `const` with `constexpr` instead? :)

IRDonch: Because it would make no difference and the diff would be longer.

IRDonch:
> Whether these comments are used somewhere, for example in https://docs.openvinotoolkit.org/?

I had this thought too, but I couldn't find anything.

PR #676, opened 2019-12-05
remove quantized models

vladimir-dudnik:
remove quantized models. they will be replaced by recipy, described hot to quantize model from IR


eaidova:
@vladimir-dudnik, does it mean that I also need remove accuracy checker configs for these models from repo?

vladimir-dudnik:
@eaidova ops, missed this. I'll do that

PR #680, opened 2019-12-06
AC: replace usage deprecated APi from sklearn

eaidova:
in upcoming sklearn version the metric.base API become a private and can not be called in the user code. Replace `_average_binary_score` with own implementation

### DIFF
--- a/tools/accuracy_checker/accuracy_checker/metrics/reid.py
+++ b/tools/accuracy_checker/accuracy_checker/metrics/reid.py
@@ -34,23 +34,17 @@ def _average_binary_score(binary_metric, y_true, y_score):
         return not (len(np.unique(y)) > 2) or (y.ndim >= 2 and len(y[0]) > 1)
 
     if binary_target(y_true):
-        return binary_metric(y_true, y_score, sample_weight=None)
+        return binary_metric(y_true, y_score)
 
     y_true = y_true.ravel()
     y_score = y_score.ravel()
-
-    if y_true.ndim == 1:
-        y_true = y_true.reshape((-1, 1))
-
-    if y_score.ndim == 1:
-        y_score = y_score.reshape((-1, 1))
 
     n_classes = y_score.shape[1]
     score = np.zeros((n_classes,))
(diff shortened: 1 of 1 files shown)


tools/accuracy_checker/accuracy_checker/metrics/reid.py
```diff
@@ -31,6 +29,32 @@
...
+    if binary_target(y_true):
+        return binary_metric(y_true, y_score, sample_weight=None)
+
+    y_true = y_true.ravel()
+    y_score = y_score.ravel()
+
+    if y_true.ndim == 1:
+        y_true = y_true.reshape((-1, 1))
+
+    if y_score.ndim == 1:
```
IRDonch: Are all these conditions necessary? I would expect the number of dimensions to be known in advance.

tools/accuracy_checker/accuracy_checker/metrics/reid.py
```diff
@@ -31,6 +29,32 @@
...
+        y_score = y_score.reshape((-1, 1))
+
+    n_classes = y_score.shape[1]
+    score = np.zeros((n_classes,))
+    for c in range(n_classes):
+        y_true_c = y_true.take([c], axis=1).ravel()
+        y_score_c = y_score.take([c], axis=1).ravel()
+        score[c] = binary_metric(y_true_c, y_score_c, None)
```
IRDonch: It would seem that `binary_metric` is always called with `sample_weight=None`, so you could just remove that parameter.

tools/accuracy_checker/accuracy_checker/metrics/reid.py
eaidova: no, I checked, we no need that, thanks

tools/accuracy_checker/accuracy_checker/metrics/reid.py
eaidova: removed and from `precision_recall_curve` too

eaidova:
jenkins please retry a build

PR #700, opened 2019-12-12
Replace tools/downloader/license.txt with per-model licensing notices

IRDonch:
This finally makes it so that all information about the model is contained within that model's directory.

Some notes about the changes:

* I embedded most of the licenses into the model's documentation pages   to make them easily accessible. The exceptions are long licenses (Apache and Creative Commons), which would dominate the documentation page if I embedded them too. I split them off into separate files instead.

* Several projects have made trivial modifications to the Apache license. This shouldn't have any legal significance, but to be on the safe side, I included every such modified license into the `licenses` directory as a separate file.

* Several models had the wrong license links in the documentation. I fixed those.

* `googlenet-v1` even had the wrong license link in `model.yml`. I fixed that too.

* I didn't update the documentation for Intel models, since the overall project license already implicitly applies to them.


IRDonch:
CC @vladimir-dudnik

eizamaliev:
LGTM

IRDonch:
Jenkins please retry a build

PR #704, opened 2019-12-13
Fix static analysis warnings in monitor code

IRDonch:


demos/common/monitors/cpu_monitor.cpp
```diff
@@ -190,6 +190,9 @@ std::deque<std::vector<double>> CpuMonitor::getLastHistory() const {
 }
 
 std::vector<double> CpuMonitor::getMeanCpuLoad() const {
+    if (samplesNumber == 0)
+        return std::vector<double>(cpuLoadSum.size());
```
Wovchena: Where can I find that `double()` is always 0?

IRDonch:
> What's wrong with getting nan when no data is collected? It is clearer than reporting 0.

Yeah, I suppose you're right. I'll remove that commit.

demos/common/monitors/cpu_monitor.cpp
IRDonch: In the C++ standard, of course. :-)

5.2.3(2):

> The expression `T()`, where T is a _simple-type-specifier_ or _typename-specifier_ for a non-array complete object type or the (possibly cv-qualified) `void` type, creates a prvalue of the specified type, whose value is that produced by value-initializing (8.5) an object of type `T`; no initialization is done for the `void()` case.

8.5(8):

> To _value-initialize_ an object of type `T` means:
>
> * if `T` is a (possibly cv-qualified) class type (Clause 9) with either no default constructor (12.1) or a default constructor that is user-provided or deleted, then the object is default-initialized;
> * if `T` is a (possibly cv-qualified) class type without a user-provided or deleted default constructor, then the object is zero-initialized and the semantic constraints for default-initialization are checked, and if `T` has a non-trivial default constructor, the object is default-initialized;
> * if `T` is an array type, then each element is value-initialized;
> * otherwise, the object is zero-initialized.

8.5(6):

> To _zero-initialize_ an object or reference of type T means:
> * if T is a scalar type (3.9), the object is initialized to the value obtained by converting the integer literal 0 (zero) to `T`;
> * [...]

All citations are from the N4296 draft.

Wovchena: Thank you :)

PR #717, opened 2019-12-17
downloader: check the size of the received file before checking the hash

IRDonch:
This has two benefits:

1. it makes sure file sizes are set correctly in the config files and therefore that we are reporting correct progress information;

2. if a download fails because the file on the remote server was replaced with an unrelated file, a size mismatch message will communicate more information than a hash mismatch message.

Also, break early if we receive more data than we expect. It's more efficient and it prevents a server from filling up the local file system by sending a very large (or even infinite) file.

### DIFF
--- a/tools/downloader/downloader.py
+++ b/tools/downloader/downloader.py
@@ -51,6 +51,12 @@ def process_download(reporter, chunk_iterable, size, file):
                     reporter.emit_event('model_file_download_progress', size=progress_size)
 
                 file.write(chunk)
+
+                # don't attempt to finish a file if it's bigger than expected
+                if progress_size > size:
+                    break
+
+        return progress_size
     finally:
         reporter.end_progress()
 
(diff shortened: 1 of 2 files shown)


tools/downloader/downloader.py
```diff
@@ -65,8 +71,19 @@ def try_download(reporter, file, num_attempts, start_download, size):
             chunk_iterable = start_download()
             file.seek(0)
             file.truncate()
-            process_download(reporter, chunk_iterable, size, file)
-            return True
+            actual_size = process_download(reporter, chunk_iterable, size, file)
+
+            if actual_size > size:
+                reporter.log_error("Remote file is longer than expected ({} B)", size)
```
eizamaliev:
```suggestion
                reporter.log_error("Remote file is longer than expected ({} B), download aborted", size)
```

IRDonch: Sure; updated.

PR #729, opened 2019-12-19
Add a script that determines which CI jobs to run for a pull request

IRDonch:
Currently, this determination is done by a similar script in an internal repository used by the CI jobs. There are, however, a few advantages to having this script in the main repo:

* It's easier to effect project layout changes, since we can update the attributes and/or the script in the same PR in which we update the layout, thus avoiding periods where the project is desynchronized from the CI logic.

* Different branches can contain different versions of the script, which lets us configure CI logic per-branch without special-casing it in the CI scripts.

* We can use Git attributes to specify most of the job triggering policy, which lets us easily create fine-grained rules.

* It's more transparent to external users.

Admittedly, since the actual definitions of CI jobs are still in an internal repository, it'll still be necessary sometimes to synchronize changes to the main and CI repos. I hope to eventually move more of the CI logic to the main repository to rectify this.

There are five possible jobs that the script can request: "ac", "demos", "documentation", "downloader", and "models". We don't actually have separate "downloader" and "models" CI jobs (both types of changes are handled by the downloader job), but I think that:

a) it's more logical to report these types of changes separately; and 
b) we might eventually create a "models" job to separate the testing of models from the testing of the downloader itself.

[skip-ci] (This doesn't affect any jobs (yet), and I don't want to take up CI machines.)

### DIFF
--- a/ci/get-jobs-for-changes.py
+++ b/ci/get-jobs-for-changes.py
@@ -1,0 +1,72 @@
+#!/usr/bin/python3
+
+# Copyright (c) 2019 Intel Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+A script that prints a JSON description of the CI jobs necessary to validate
+the changes made between a base commit and the current commit.
+
+The output format is a an object where each key is the identifier of the job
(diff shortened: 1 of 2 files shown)


ci/get-jobs-for-changes.py
```diff
@@ -0,0 +1,72 @@
...
+    for changed_file in changed_files:
+        if models_dir in changed_file.parents and changed_file.name == "model.yml":
+            if Path(changed_file).exists(): # it might've been deleted in the branch
+                jobs.setdefault("models", []).append(changed_file.parent.name)
+
+    git_check_attr_output = subprocess.run(
+        ["git", "check-attr", "--stdin", "-z", "--all"],
+        input=git_diff_output, stdout=subprocess.PIPE, check=True).stdout
+
+    for path, attribute, value in group_by_n(git_check_attr_output.decode()[:-1].split("\0"), 3):
```
eizamaliev: The variable `path` is unused here. Will it be used in the future?

IRDonch: I don't expect to use it, although who knows what'll happen. `path` is just here because the syntax requires me to put that value somewhere. :-)

PR #734, opened 2019-12-20
AC: revise async execution

eaidova:
move async from wait_for_any fo callback based api
TO DO:
- [x] check flags for other devices
- [x] check flags for multi device case
- [x] update tests

Next steps
- [x] add command line interface to switch between sync/async mode
- [x] add command line interface to select the number of infer requests
- [x] special value for automatic async requests
- [x] auto + multi
- [x] print info about num requests


eaidova:
Jenkins please retry a build

PR #736, opened 2019-12-23
AC configs: adopt to new layout

eaidova:
- changed paths to INT8 and INT1 models
- removed explicit device setting
- removed extra empy lines and some unused parameters (mo_params when IR used)

### DIFF
--- a/tools/accuracy_checker/configs/face-detection-retail-0044.yml
+++ b/tools/accuracy_checker/configs/face-detection-retail-0044.yml
@@ -3,7 +3,6 @@ models:
 
     launchers:
       - framework: caffe
-        device: CPU
         model:   public/face-detection-retail-0044/face-detection-retail-0044.prototxt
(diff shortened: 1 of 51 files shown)


tools/accuracy_checker/configs/face-detection-retail-0044.yml
```diff
@@ -3,7 +3,6 @@ models:
 
     launchers:
       - framework: caffe
-        device: CPU
         model:   public/face-detection-retail-0044/face-detection-retail-0044.prototxt
```
vladimir-dudnik: should not be FP32 subfolder there?

eaidova: No, It is running original model on Caffe

eaidova:
@vladimir-dudnik 
> please run validation job with /md option from this branch to test changes

I am not sure that I can do it, because currently MD containsdownloads models with old layout.
Besides that INT8 models will be extended with new, so I suggest just wait new OMZ package readiness.

PR #741, opened 2019-12-24
remove resnet18-dorefa

vladimir-dudnik:
remove resnet18-dorefa from R4 release


vladimir-dudnik:
Jenkins please retry a build

PR #752, opened 2019-12-30
remove old models from test cases

vladimir-dudnik:
remove old models from test cases. new models should be added later


vladimir-dudnik:
ignore not related issues

PR #759, opened 2020-01-09
AC: update config for brain-tumor-segmentation-0001

eaidova:


eizamaliev:
@eaidova thanks for fix!

PR #761, opened 2020-01-10
AC: fix initialization order

eaidova:
necessary for calibration metrics and postprocessing overriding


eaidova:
@vladimir-dudnik could you please take a look, I need this chages to update OMZ submodule in POT, it does not affect anything (needed only for POT development case)

PR #762, opened 2020-01-13
Fixed deprecation warnings

ilyachur:


demos/crossroad_camera_demo/main.cpp
```diff
@@ -161,23 +161,19 @@ struct PersonDetection : BaseDetection{
...
-    CNNNetwork read() override {
+    CNNNetwork read(const Core& ie) override {
         slog::info << "Loading network files for PersonDetection" << slog::endl;
-        CNNNetReader netReader;
         /** Read network model **/
-        netReader.ReadNetwork(FLAGS_m);
-        /** Extract model name and load it's weights **/
-        std::string binFileName = fileNameNoExt(FLAGS_m) + ".bin";
-        netReader.ReadWeights(binFileName);
+        auto network = ie.ReadNetwork(FLAGS_m);
```
IRDonch: I think you should specify the path to the weights explicitly here. The documentation says that if the second argument is omitted and a file with the automatically-derived name isn't found, the function will not load the weights at all, which is not what we want. We want it to fail with a file-not-found error.

ilyachur: @IRDonch if IR contains some weighs but you don't provide **\*.bin** file, the network won't be read and you will get error that some layer cannot set weights.

IRDonch: I have experimented with this, and the errors you get when you specify the weights path and when you don't are quite different.

When you don't, `ReadNetwork` just crashes.

When you do, it throws an exception:

```
terminate called after throwing an instance of 'InferenceEngine::details::InferenceEngineException'
  what():  filesize for: test.bin - -1<0. Please, check weights file existence.
```

The latter is certainly preferable to the former, as it hints as to what went wrong.

ilyachur: I created a PR to OpenVINO which add more informative error message if model requires weights, but bin file wasn't found.

IRDonch: All right then, sounds good to me.

demos/mask_rcnn_demo/main.cpp
```diff
@@ -103,14 +103,10 @@ int main(int argc, char *argv[]) {
         // --------------------Load network (Generated xml/bin files)-------------------------------------------
         slog::info << "Loading network files" << slog::endl;
 
-        InferenceEngine::CNNNetReader networkReader;
         /** Read network model **/
-        networkReader.ReadNetwork(FLAGS_m);
+        auto network = ie.ReadNetwork(FLAGS_m);
 
         /** Extract model name and load weights **/
```
IRDonch: Orphaned comment.

demos/multi_channel/object_detection_demo_yolov3/main.cpp
```diff
@@ -244,20 +244,20 @@ DisplayParams prepareDisplayParams(size_t count) {
     return params;
 }
 
-std::map<std::string, YoloParams> GetYoloParams(const std::vector<std::string>& outputDataBlobNames, 
-                                                InferenceEngine::CNNNetReader &netReader) {
+std::map<std::string, YoloParams> GetYoloParams(const std::vector<std::string>& outputDataBlobNames,
+                                               InferenceEngine::CNNNetwork &network) {
```
IRDonch: Restore the alignment, please.

demos/object_detection_demo_faster_rcnn/main.cpp
```diff
@@ -112,19 +112,12 @@ int main(int argc, char *argv[]) {
         // -----------------------------------------------------------------------------------------------------
 
         // --------------------------- 2. Read IR Generated by ModelOptimizer (.xml and .bin files) ------------
-        std::string binFileName = fileNameNoExt(FLAGS_m) + ".bin";
         slog::info << "Loading network files:"
             "\n\t" << FLAGS_m <<
-            "\n\t" << binFileName <<
             slog::endl;
```
IRDonch: It's misleading to say "network files" but only list one file.

demos/pedestrian_tracker_demo/main.cpp
```diff
@@ -77,7 +76,7 @@ CreatePedestrianTracker(const std::string& reid_model,
         tracker->set_descriptor_strong(descriptor_strong);
         tracker->set_distance_strong(distance_strong);
     } else {
-        std::cout << "WARNING: Either reid model or reid weights "
+        std::cout << "WARNING: Either reid model "
```
IRDonch: You can't use "either" with only one item.

IRDonch:
Jenkins: test this please

IRDonch:
@Wovchena Please review.

The build failures on CI are likely due to an old OpenVINO version. I'll update it soon.

IRDonch:
Jenkins please retry a build

demos/human_pose_estimation_demo/src/human_pose_estimator.cpp
```diff
@@ -34,10 +34,7 @@ HumanPoseEstimator::HumanPoseEstimator(const std::string& modelPath,
         ie.SetConfig({{InferenceEngine::PluginConfigParams::KEY_PERF_COUNT,
                        InferenceEngine::PluginConfigParams::YES}});
     }
-    netReader.ReadNetwork(modelPath);
-    std::string binFileName = fileNameNoExt(modelPath) + ".bin";
-    netReader.ReadWeights(binFileName);
-    network = netReader.getNetwork();
+    auto network = ie.ReadNetwork(modelPath);
```
IRDonch:
```suggestion
    network = ie.ReadNetwork(modelPath);
```

You've shadowed the member variable with the same name.

IRDonch:
Jenkins: test this please

PR #767, opened 2020-01-14
instance_segmentation: fix

Wovchena:
For Ubuntu16 the demo drew wrong masks sometimes.
I was deleting and returning back parts of the demo's code to find what causes it.
It appeared, that deleting the line `del net` fixes the issue.
Delete `del exec_net` and `del ie` just because they are not really needed.

### DIFF
--- a/demos/python_demos/instance_segmentation_demo/instance_segmentation_demo.py
+++ b/demos/python_demos/instance_segmentation_demo/instance_segmentation_demo.py
@@ -155,7 +155,6 @@ def main():
 
     log.info('Loading IR to the plugin...')
     exec_net = ie.load_network(network=net, device_name=args.device, num_requests=2)
-    del net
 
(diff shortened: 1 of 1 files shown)


IRDonch:
This is very strange. Did you file a bug?

Wovchena:
There is a bug for the demo, we can just reassign it.

eaidova:
@Ilya-Krylov could it be the reason of non-deterministic result on text-spotting model in AC?
The detector has similar architecture, we also remove net after dataset processing, but I can not just use suggested workaround because AC run in the cycle (several models e.g. with different precision can be placed in one config) and it leads to some troubles if we do not remove it.

Ilya-Krylov:
@eaidova I don't know. Moreover it is observed in case of IR_V10 but not in case of IR_V7

PR #768, opened 2020-01-15
Add ctc beam search decoder for text_detection demo

FengYen-Chang:
Add another decoder method for text recognition. Just using the flag `-b` to assign bandwidth, default is `0`. If the bandwidth larger than `0` will use CTC beam search decoder, bandwidth is `0` will use CTC greedy decoder, and bandwidth is lower than `0` will show an error message and stop the program.

### DIFF
--- a/demos/text_detection_demo/src/text_recognition.cpp
+++ b/demos/text_detection_demo/src/text_recognition.cpp
@@ -27,11 +27,13 @@ namespace  {
         *prob = 1.0f / static_cast<float>(sum);
     }
 
-    void softmax(const std::vector<float>::const_iterator& begin, const std::vector<float>::const_iterator& end, std::vector<float> &prob) {
+    std::vector<float> softmax(const std::vector<float>::const_iterator& begin, const std::vector<float>::const_iterator& end) {
+        std::vector<float> prob(end - begin, 0.f);
         std::transform(begin, end, prob.begin(), static_cast<double(*)(double)>(std::exp));
         float sum = std::accumulate(prob.begin(), prob.end(), 0.0f);
         for (int i = 0; i < static_cast<int>(prob.size()); i++)
             prob[i] /= sum;
+        return prob;
     }
 
     struct BeamElement {
@@ -87,8 +89,7 @@ std::string CTCBeamSearchDecoder(const std::vector<float> &data, const std::stri
     for (std::vector<float>::const_iterator it = data.begin(); it != data.end(); it += num_classes) {
         curr.clear();
 
-        std::vector<float> prob = std::vector<float>(num_classes, 0.f);
-        softmax(it, it + num_classes, prob);
+        std::vector<float> prob = softmax(it, it + num_classes);
 
         for(const auto& candidate: last) {
             float prob_not_blank = 0.f;
--- a/demos/text_detection_demo/src/text_recognition.cpp
+++ b/demos/text_detection_demo/src/text_recognition.cpp
@@ -41,7 +41,7 @@ namespace  {
         float prob_not_blank;        //!< The probability that the last char in CTC sequence
                                      //!< for the beam element is NOT the special blank char
 
-        float prob() {               //!< The probability of the beam element.
+        float prob() const {         //!< The probability of the beam element.
             return prob_blank + prob_not_blank;
         }
     };
@@ -90,9 +90,9 @@ std::string CTCBeamSearchDecoder(const std::vector<float> &data, const std::stri
         std::vector<float> prob = std::vector<float>(num_classes, 0.f);
         softmax(it, it + num_classes, prob);
 
-        for(auto& candidate: last) {
+        for(const auto& candidate: last) {
             float prob_not_blank = 0.f;
-            std::vector<int> candidate_sentence = candidate.sentence;
+            const std::vector<int>& candidate_sentence = candidate.sentence;
             if (!candidate_sentence.empty()) {
                 int n = candidate_sentence.back();
                 prob_not_blank = candidate.prob_not_blank * prob[n];
@@ -105,14 +105,10 @@ std::string CTCBeamSearchDecoder(const std::vector<float> &data, const std::stri
             if (check_res == std::end(curr)) {
                 curr.push_back(BeamElement{candidate.sentence, prob_blank, prob_not_blank});
             } else {
-                // auto index = std::distance(curr.begin(), check_res);
-                // curr[index].prob_not_blank  += prob_not_blank;
                 check_res->prob_not_blank  += prob_not_blank;
-                // if (curr[index].prob_blank != 0.f) {
                 if (check_res->prob_blank != 0.f) {
                     throw std::logic_error("Probability that the last char in CTC-sequence is the special blank char must be zero here");
                 }
-                // curr[index].prob_blank = prob_blank;
                 check_res->prob_blank = prob_blank;
             }
 
@@ -133,8 +129,6 @@ std::string CTCBeamSearchDecoder(const std::vector<float> &data, const std::stri
                 if (check_res == std::end(curr)) {
                     curr.push_back(BeamElement{extend, 0.f, prob_not_blank});
                 } else {
-                    // auto index = std::distance(curr.begin(), check_res);
-                    // curr[index].prob_not_blank += prob_not_blank;
                     check_res->prob_not_blank += prob_not_blank;
                 }
             }
--- a/demos/text_detection_demo/src/text_recognition.cpp
+++ b/demos/text_detection_demo/src/text_recognition.cpp
@@ -29,7 +29,7 @@ namespace  {
 
     std::vector<float> softmax(const std::vector<float>::const_iterator& begin, const std::vector<float>::const_iterator& end) {
         std::vector<float> prob(end - begin, 0.f);
-        std::transform(begin, end, prob.begin(), static_cast<double(*)(double)>(std::exp));
+        std::transform(begin, end, prob.begin(), [](float x) { return std::exp(x); });
         float sum = std::accumulate(prob.begin(), prob.end(), 0.0f);
         for (int i = 0; i < static_cast<int>(prob.size()); i++)
             prob[i] /= sum;
@@ -77,8 +77,6 @@ std::string CTCGreedyDecoder(const std::vector<float> &data, const std::string& 
 }
 
 std::string CTCBeamSearchDecoder(const std::vector<float> &data, const std::string& alphabet, char pad_symbol, double *conf, int bandwidth) {
-    std::string res = "";
-
     const int num_classes = alphabet.length();
 
     std::vector<BeamElement> curr;
@@ -98,9 +96,9 @@ std::string CTCBeamSearchDecoder(const std::vector<float> &data, const std::stri
                 int n = candidate_sentence.back();
                 prob_not_blank = candidate.prob_not_blank * prob[n];
             }
-            float prob_blank = candidate.prob() * prob[(num_classes - 1)];
+            float prob_blank = candidate.prob() * prob[num_classes - 1];
 
-            auto check_res = std::find_if(curr.begin(), curr.end(), [candidate_sentence](const BeamElement& n) {
+            auto check_res = std::find_if(curr.begin(), curr.end(), [&candidate_sentence](const BeamElement& n) {
                 return n.sentence == candidate_sentence;
             });
             if (check_res == std::end(curr)) {
@@ -123,7 +121,7 @@ std::string CTCBeamSearchDecoder(const std::vector<float> &data, const std::stri
                     prob_not_blank = prob[i] * candidate.prob();
                 }
                 
-                auto check_res = std::find_if(curr.begin(), curr.end(), [extend](const BeamElement &n) {
+                auto check_res = std::find_if(curr.begin(), curr.end(), [&extend](const BeamElement &n) {
                     return n.sentence == extend;
                 });
 
@@ -147,6 +145,7 @@ std::string CTCBeamSearchDecoder(const std::vector<float> &data, const std::stri
     }
 
     *conf = last[0].prob();
+    std::string res="";
     for (const auto& idx: last[0].sentence) {
         res += alphabet[idx];
     }
--- a/demos/text_detection_demo/src/text_recognition.cpp
+++ b/demos/text_detection_demo/src/text_recognition.cpp
@@ -10,6 +10,7 @@
 #include <vector>
 #include <limits>
 #include <stdexcept>
+#include <numeric>
 
 namespace  {
     void softmax_and_choose(const std::vector<float>::const_iterator& begin, const std::vector<float>::const_iterator& end, int *argmax, float *prob) {
@@ -26,28 +27,24 @@ namespace  {
         *prob = 1.0f / static_cast<float>(sum);
     }
 
-    void softmax(const std::vector<float>::const_iterator& begin, const std::vector<float>::const_iterator& end, std::vector<float> &prob)
-    {
+    void softmax(const std::vector<float>::const_iterator& begin, const std::vector<float>::const_iterator& end, std::vector<float> &prob) {
         std::transform(begin, end, prob.begin(), static_cast<double(*)(double)>(std::exp));
         float sum = std::accumulate(prob.begin(), prob.end(), 0.0f);
         for (int i = 0; i < static_cast<int>(prob.size()); i++)
             prob[i] /= sum;
     }
 
-    struct BeamElement
-    {
-        std::string sentence;   //!< The sequence of chars that will be result of the beam element
-        float prob_blank;       //!< The probability that the last char in CTC sequence
-                                //!< for the beam element is the special blank char
-        float prob_not_blank;   //!< The probability that the last char in CTC sequence
-                                //!<  for the beam element is NOT the special blank char
+    struct BeamElement {
+        std::vector<int> sentence;   //!< The sequence of chars that will be a result of the beam element
+        float prob_blank;            //!< The probability that the last char in CTC sequence
+                                     //!< for the beam element is the special blank char
+        float prob_not_blank;        //!< The probability that the last char in CTC sequence
+                                     //!< for the beam element is NOT the special blank char
 
-        float prob()            //!< The probability of the beam element.
-        {
+        float prob() {               //!< The probability of the beam element.
             return prob_blank + prob_not_blank;
         }
     };
-
 }  // namespace
 
 std::string CTCGreedyDecoder(const std::vector<float> &data, const std::string& alphabet, char pad_symbol, double *conf) {
@@ -85,7 +82,7 @@ std::string CTCBeamSearchDecoder(const std::vector<float> &data, const std::stri
     std::vector<BeamElement> curr;
     std::vector<BeamElement> last;
 
-    last.push_back(BeamElement{"", 1.f, 0.f});
+    last.push_back(BeamElement{std::vector<int>(), 1.f, 0.f});
 
     for (std::vector<float>::const_iterator it = data.begin(); it != data.end(); it += num_classes) {
         curr.clear();
@@ -93,13 +90,12 @@ std::string CTCBeamSearchDecoder(const std::vector<float> &data, const std::stri
         std::vector<float> prob = std::vector<float>(num_classes, 0.f);
         softmax(it, it + num_classes, prob);
 
-        for (int candidate_num = 0; candidate_num < static_cast<int>(last.size()); candidate_num++) {
+        for(auto& candidate: last) {
             float prob_not_blank = 0.f;
-            auto candidate = last[candidate_num];
-            std::string candidate_sentence = candidate.sentence;
-            if (candidate_sentence != "") {
-                int n = static_cast<int>(candidate_sentence.back());
-                prob_not_blank = candidate.prob_blank * prob[n];
+            std::vector<int> candidate_sentence = candidate.sentence;
+            if (!candidate_sentence.empty()) {
+                int n = candidate_sentence.back();
+                prob_not_blank = candidate.prob_not_blank * prob[n];
             }
             float prob_blank = candidate.prob() * prob[(num_classes - 1)];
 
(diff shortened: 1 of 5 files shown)


Wovchena:
@FengYen-Chang, I haven't yet reviewed the patch, but @LeonidBeynenson's comments sound reasonable. Can you, please, address them?

FengYen-Chang:
@Wovchena Sure and sorry for the late reply. 
@LeonidBeynenson Thanks for your recommendation, I have patched this, Thanks.

demos/text_detection_demo/src/text_recognition.cpp
```diff
@@ -53,3 +61,91 @@ std::string CTCGreedyDecoder(const std::vector<float> &data, const std::string&
...
+            float _pB = __can.pT * prob[(num_classes - 1)];
+
+            auto check_res = std::find_if(curr.begin(), curr.end(), [__can_sentance](beam const& n) {
+                return n.sentance == __can_sentance;
+            });
+            if (check_res == std::end(curr)) {
+                curr.push_back(beam(__can.sentance, _pB, _pNB, _pB + _pNB));
+            } else {
+                auto __i = std::distance(curr.begin(), check_res);
+                curr[__i].pNB += _pNB;
```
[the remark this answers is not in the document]
FengYen-Chang: I wouldn't add this into code since it's very close to zero, but not really zero.

demos/text_detection_demo/src/text_recognition.cpp
```diff
@@ -53,3 +76,83 @@ std::string CTCGreedyDecoder(const std::vector<float> &data, const std::string&
...
+
+    last.push_back(BeamElement{"", 1.f, 0.f});
+
+    for (std::vector<float>::const_iterator it = data.begin(); it != data.end(); it += num_classes) {
+        curr.clear();
+
+        std::vector<float> prob = std::vector<float>(num_classes, 0.f);
+        softmax(it, it + num_classes, prob);
+
+        for (int candidate_num = 0; candidate_num < static_cast<int>(last.size()); candidate_num++) {
```
LeonidBeynenson: Could you use here
```cpp
for(const auto& candidate: last) {
...
}
```

demos/text_detection_demo/src/text_recognition.cpp
```diff
@@ -25,6 +25,29 @@ namespace  {
...
+    {
+        std::transform(begin, end, prob.begin(), static_cast<double(*)(double)>(std::exp));
+        float sum = std::accumulate(prob.begin(), prob.end(), 0.0f);
+        for (int i = 0; i < static_cast<int>(prob.size()); i++)
+            prob[i] /= sum;
+    }
+
+    struct BeamElement
+    {
+        std::string sentence;   //!< The sequence of chars that will be result of the beam element
```
LeonidBeynenson: Could you use `vector<int>` instead of string here?
We considered with @Wovchena : It is natural to keep indexes in vectors of integers, not in strings.
(And IMHO integer type here is better than `char` type -- it would be required if we have strict memory restrictions)

demos/text_detection_demo/src/text_recognition.cpp
```diff
@@ -25,6 +25,29 @@ namespace  {
         }
         *prob = 1.0f / static_cast<float>(sum);
     }
+
+    void softmax(const std::vector<float>::const_iterator& begin, const std::vector<float>::const_iterator& end, std::vector<float> &prob)
```
LeonidBeynenson: Could you move `{` char to the same line as the function declaration?
(Everywhere in the demo this style is used.)

demos/text_detection_demo/src/text_recognition.cpp
```diff
@@ -25,6 +25,29 @@ namespace  {
...
+    {
+        std::transform(begin, end, prob.begin(), static_cast<double(*)(double)>(std::exp));
+        float sum = std::accumulate(prob.begin(), prob.end(), 0.0f);
+        for (int i = 0; i < static_cast<int>(prob.size()); i++)
+            prob[i] /= sum;
+    }
+
+    struct BeamElement
```
LeonidBeynenson: Could you move `{ ` char to the same line as the struct declaration?
(Everywhere in the demo this style is used.)

demos/text_detection_demo/src/text_recognition.cpp
```diff
@@ -25,6 +25,29 @@ namespace  {
...
+    {
+        std::string sentence;   //!< The sequence of chars that will be result of the beam element
+        float prob_blank;       //!< The probability that the last char in CTC sequence
+                                //!< for the beam element is the special blank char
+        float prob_not_blank;   //!< The probability that the last char in CTC sequence
+                                //!<  for the beam element is NOT the special blank char
+
+        float prob()            //!< The probability of the beam element.
```
LeonidBeynenson: Could you move `{ ` char to the same line as the function declaration?

demos/text_detection_demo/src/text_recognition.cpp
```diff
@@ -53,3 +76,83 @@ std::string CTCGreedyDecoder(const std::vector<float> &data, const std::string&
...
+        curr.clear();
+
+        std::vector<float> prob = std::vector<float>(num_classes, 0.f);
+        softmax(it, it + num_classes, prob);
+
+        for (int candidate_num = 0; candidate_num < static_cast<int>(last.size()); candidate_num++) {
+            float prob_not_blank = 0.f;
+            auto candidate = last[candidate_num];
+            std::string candidate_sentence = candidate.sentence;
+            if (candidate_sentence != "") {
```
LeonidBeynenson: If `sentence` field is `vector<int>`, then this line should be replaced with `if (!candidate_sentence.empty())`

demos/text_detection_demo/src/text_recognition.cpp
```diff
@@ -53,3 +76,83 @@ std::string CTCGreedyDecoder(const std::vector<float> &data, const std::string&
...
+                curr.push_back(BeamElement{candidate.sentence, prob_blank, prob_not_blank});
+            } else {
+                auto index = std::distance(curr.begin(), check_res);
+                curr[index].prob_not_blank  += prob_not_blank;
+                curr[index].prob_blank = prob_blank;
+            }
+
+            for (int i = 0; i < num_classes - 1; i++) {
+                auto extend = candidate_sentence + static_cast<char>(i);
+                if (candidate_sentence.length() > 0 && candidate.sentence.back() == static_cast<char>(i)) {
```
LeonidBeynenson: If you make `sentence` field be `vector<int>`, then `static_cast` is not required here

demos/text_detection_demo/src/text_recognition.cpp
```diff
@@ -53,3 +76,83 @@ std::string CTCGreedyDecoder(const std::vector<float> &data, const std::string&
...
+            if (check_res == std::end(curr)) {
+                curr.push_back(BeamElement{candidate.sentence, prob_blank, prob_not_blank});
+            } else {
+                auto index = std::distance(curr.begin(), check_res);
+                curr[index].prob_not_blank  += prob_not_blank;
+                curr[index].prob_blank = prob_blank;
+            }
+
+            for (int i = 0; i < num_classes - 1; i++) {
+                auto extend = candidate_sentence + static_cast<char>(i);
```
LeonidBeynenson: If you make `sentence` field be `vector<int>`, then `push_back` here should be used

demos/text_detection_demo/src/text_recognition.cpp
```diff
@@ -25,6 +25,29 @@ namespace  {
         }
         *prob = 1.0f / static_cast<float>(sum);
     }
+
+    void softmax(const std::vector<float>::const_iterator& begin, const std::vector<float>::const_iterator& end, std::vector<float> &prob)
+    {
+        std::transform(begin, end, prob.begin(), static_cast<double(*)(double)>(std::exp));
+        float sum = std::accumulate(prob.begin(), prob.end(), 0.0f);
```
LeonidBeynenson: Please, add `#include <numeric>` in the include-s section in the file, since `std::accumulate` requires it on some platforms

demos/text_detection_demo/src/text_recognition.cpp
```diff
@@ -53,3 +76,83 @@ std::string CTCGreedyDecoder(const std::vector<float> &data, const std::string&
...
+        std::vector<float> prob = std::vector<float>(num_classes, 0.f);
+        softmax(it, it + num_classes, prob);
+
+        for (int candidate_num = 0; candidate_num < static_cast<int>(last.size()); candidate_num++) {
+            float prob_not_blank = 0.f;
+            auto candidate = last[candidate_num];
+            std::string candidate_sentence = candidate.sentence;
+            if (candidate_sentence != "") {
+                int n = static_cast<int>(candidate_sentence.back());
+                prob_not_blank = candidate.prob_blank * prob[n];
```
LeonidBeynenson: 2 @FengYen-Chang : I'm sorry, but this is a mistake.
The following code should be here:
```suggestion
                prob_not_blank = candidate.prob_not_blank * prob[n];
```
The reason is as follows: if we keep the `candidate_sentence` the same
(that is, if the new `BeamElement` added to the vector `curr` has the same field `sentence` as in the current `candidate`), 
then one of the following conditions should be fulfilled:
* either the last element in CTC sequence is the blank char -- in this case `prob_blank` field is increased
* or the last element in CTC sequence is the same as the the previous one AND the previous one is NOT a blank char -- in this case `prob_not_blank` field is increased, and for the new element 
    ```cpp
    prob_not_blank = candidate.prob_not_blank[index_of_last_symbol]
    ```
The latter case is handled in this line of the code.

demos/text_detection_demo/src/text_recognition.cpp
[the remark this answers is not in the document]
LeonidBeynenson: I'm sorry, but it MUST be zero in this line of the code. This condition follows from the consistency of the algorithm.

The reason is as follows:
* when a next symbol is received, the algorithm runs through all candidate elements of the vector `last` and adds new elements to the vector `curr`
    (and vector `curr` is cleared at the begin of each iteration)
* elements are added to the vector `curr` in such a way, that different elements are *always* have different values of the field `sentence`
* the vector `last` is a part of the vector `curr` on the previous iteration. 
    Thus the different elements in the vector `last` are *always* have different values of the field `sentence`
* There are only 2 places in the algorithm where an element may be added into the vector `curr` or an element from vector `curr` is modified: 
   * lines 99-115  -- when the new element has the same field `sentence` as the current candidate https://github.com/opencv/open_model_zoo/blob/c2e691890405f9642713dfe474fe18cdc1434b50/demos/text_detection_demo/src/text_recognition.cpp#L99-L115
   * lines 118-135, when a new char is added to the end of the field `sentence` of the current candidate https://github.com/opencv/open_model_zoo/blob/c2e691890405f9642713dfe474fe18cdc1434b50/demos/text_detection_demo/src/text_recognition.cpp#L117-L136
* At that in lines 118-135 the field `prob_blank` is neither modified nor set to non-zero values
    (the new elements which are added in the part of the code always have the field `prob_blank = 0` )
* So, the only part of the code where the field `prob_blank` may be changed to non-zero value is the part in the lines 99-115
* But in this part the field `sentence` of the new element (in the vector `curr`) is the same as the field `sentence` in the candidate element  (in the vector `last`)
* Since for different elements from the vector `last` the field `sentence` is different, this part of code (lines 99-115) may be fulfilled only once for each sentence
* Thus when the algorithm enters this part of the code, the field `prob_blank` MUST be zero (if the algorithm works properly)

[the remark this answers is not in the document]
LeonidBeynenson: One more possible question: why such code is required here, if we can prove (as I did it above) that the field `prob_blank` must be zero here
```cpp
if curr[index].prob_blank != 0.f
    throw std::logic_error("Probability that the last char in CTC-sequence is the special blank char must be zero here");
```

My answer is as follows:  the code will be changed and refactored (it always happens, sooner or later), and it is important that even if the refactoring is made by another person, there are some consistency checks in the algorithm to avoid dummy bugs in refactoring.

demos/text_detection_demo/src/text_recognition.cpp
LeonidBeynenson: BTW, @Wovchena and me debugged this part of the code and found out that if we fix the mistake, the final confidence values of the results are better.

demos/text_detection_demo/src/text_recognition.cpp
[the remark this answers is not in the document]
LeonidBeynenson: Please, note that @Wovchena and me added this check in the code and found that it works properly.

demos/text_detection_demo/src/text_recognition.cpp
```diff
@@ -53,3 +76,83 @@ std::string CTCGreedyDecoder(const std::vector<float> &data, const std::string&
...
+            float prob_blank = candidate.prob() * prob[(num_classes - 1)];
+
+            auto check_res = std::find_if(curr.begin(), curr.end(), [candidate_sentence](const BeamElement& n) {
+                return n.sentence == candidate_sentence;
+            });
+            if (check_res == std::end(curr)) {
+                curr.push_back(BeamElement{candidate.sentence, prob_blank, prob_not_blank});
+            } else {
+                auto index = std::distance(curr.begin(), check_res);
+                curr[index].prob_not_blank  += prob_not_blank;
```
LeonidBeynenson: One more comment (I made it to this part of code earlier, but it is outdated)
I recommend to add here the following check:
```cpp
if curr[index].prob_blank != 0.f
    throw std::logic_error("Probability that the last char in CTC-sequence is the special blank char must be zero here");
```
This is an additional check on the consistency of the algorithm.

See detailed explanation why the check is required in my comments
* here https://github.com/opencv/open_model_zoo/pull/768#discussion_r368610795
* and here https://github.com/opencv/open_model_zoo/pull/768#discussion_r368614569

demos/text_detection_demo/src/text_recognition.cpp
FengYen-Chang: Really thanks, it is my mistake, sorry. Thanks a lot.

FengYen-Chang:
Thanks for your review and recommendations. I have updated it, and sorry for the mistake and negligence. Really thanks.

demos/text_detection_demo/src/text_recognition.cpp
```diff
@@ -53,3 +76,83 @@ std::string CTCGreedyDecoder(const std::vector<float> &data, const std::string&
...
+                
+                auto check_res = std::find_if(curr.begin(), curr.end(), [extend](const BeamElement &n) {
+                    return n.sentence == extend;
+                });
+
+                if (check_res == std::end(curr)) {
+                    curr.push_back(BeamElement{extend, 0.f, prob_not_blank});
+                } else {
+                    auto index = std::distance(curr.begin(), check_res);
+                    curr[index].prob_not_blank += prob_not_blank;
```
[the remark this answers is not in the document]
Wovchena: The commented code should just be removed

demos/text_detection_demo/src/text_recognition.cpp
```diff
@@ -53,3 +73,88 @@ std::string CTCGreedyDecoder(const std::vector<float> &data, const std::string&
...
+
+    last.push_back(BeamElement{std::vector<int>(), 1.f, 0.f});
+
+    for (std::vector<float>::const_iterator it = data.begin(); it != data.end(); it += num_classes) {
+        curr.clear();
+
+        std::vector<float> prob = std::vector<float>(num_classes, 0.f);
+        softmax(it, it + num_classes, prob);
+
+        for(auto& candidate: last) {
```
LeonidBeynenson: Could you use `const auto&` here?
It will emphasize that the algorithm does not change the candidate in `last`
```suggestion
        for(const auto& candidate: last) {
```

demos/text_detection_demo/src/text_recognition.cpp
```diff
@@ -53,3 +73,88 @@ std::string CTCGreedyDecoder(const std::vector<float> &data, const std::string&
...
+                }
+                
+                auto check_res = std::find_if(curr.begin(), curr.end(), [extend](const BeamElement &n) {
+                    return n.sentence == extend;
+                });
+
+                if (check_res == std::end(curr)) {
+                    curr.push_back(BeamElement{extend, 0.f, prob_not_blank});
+                } else {
+                    // auto index = std::distance(curr.begin(), check_res);
```
LeonidBeynenson: Could you remove the commented part of code here?

demos/text_detection_demo/src/text_recognition.cpp
```diff
@@ -53,3 +73,88 @@ std::string CTCGreedyDecoder(const std::vector<float> &data, const std::string&
...
+
+    for (std::vector<float>::const_iterator it = data.begin(); it != data.end(); it += num_classes) {
+        curr.clear();
+
+        std::vector<float> prob = std::vector<float>(num_classes, 0.f);
+        softmax(it, it + num_classes, prob);
+
+        for(auto& candidate: last) {
+            float prob_not_blank = 0.f;
+            std::vector<int> candidate_sentence = candidate.sentence;
```
LeonidBeynenson: Could you use `const std::vector<int>&` here? It will allow to avoid copying of data (not so important here, just a bit better code style). Also it will emphasize for readers that this variable won't be changed (and point it to compiler -- just to avoid dummy bugs during future refactoring) 
```suggestion
            const std::vector<int>& candidate_sentence = candidate.sentence;
```

demos/text_detection_demo/src/text_recognition.cpp
```diff
@@ -53,3 +73,88 @@ std::string CTCGreedyDecoder(const std::vector<float> &data, const std::string&
...
+            }
+            float prob_blank = candidate.prob() * prob[(num_classes - 1)];
+
+            auto check_res = std::find_if(curr.begin(), curr.end(), [candidate_sentence](const BeamElement& n) {
+                return n.sentence == candidate_sentence;
+            });
+            if (check_res == std::end(curr)) {
+                curr.push_back(BeamElement{candidate.sentence, prob_blank, prob_not_blank});
+            } else {
+                // auto index = std::distance(curr.begin(), check_res);
```
LeonidBeynenson: Could you remove commented code here?

demos/text_detection_demo/text_detection_demo.hpp
```diff
@@ -43,6 +43,7 @@ static const char input_data_type_message[] = "Required. Input data type: \"imag
                                               "\"video\" (for a saved video), "
                                               "\"webcam\" (for a webcamera device). By default, it is \"image\".";
 static const char utilization_monitors_message[] = "Optional. List of monitors to show initially.";
+static const char decoder_bandwidth_message[] = "Optional. Default bandwidth is 0 it will use CTC Greedy Decoder.";
```
LeonidBeynenson: 2 @FengYen-Chang: I'm sorry, but it requires more detailed explanation, otherwise it won't be clear for users.
```suggestion
static const char decoder_bandwidth_message[] = "Optional. Bandwidth for CTC beam search decoder. Default value is 0, in this case CTC greedy decoder will be used.";
```
I'm sorry, but IMHO it is important.

P.S. Please, change README.md correspondingly.

demos/text_detection_demo/src/text_recognition.cpp
```diff
@@ -25,6 +26,25 @@ namespace  {
         }
         *prob = 1.0f / static_cast<float>(sum);
     }
+
+    void softmax(const std::vector<float>::const_iterator& begin, const std::vector<float>::const_iterator& end, std::vector<float> &prob) {
```
LeonidBeynenson: Also one more minor change to improve style a bit:
```suggestion
    std::vector<float> softmax(const std::vector<float>::const_iterator& begin, const std::vector<float>::const_iterator& end) {
```
If the function returns probabilities instead of using a reference parameter, and returns as a value `std::vector<float>`, not as a reference `std::vector<float>&` (NB: t is impoirtant!), then the code will looks cleaner and slightly more optimal  -- it won't create a temporary object due to copy elision and return value optimization
(See details here: https://en.wikipedia.org/wiki/Copy_elision#Return_value_optimization)

In our case performance difference is not very important, so the change is more about the code style.

FengYen-Chang:
@LeonidBeynenson Thank you, and I am very sorry about I forgot to remove the comments. Really thanks.

demos/text_detection_demo/src/text_recognition.cpp
```diff
@@ -25,6 +26,27 @@ namespace  {
         }
         *prob = 1.0f / static_cast<float>(sum);
     }
+
+    std::vector<float> softmax(const std::vector<float>::const_iterator& begin, const std::vector<float>::const_iterator& end) {
+        std::vector<float> prob(end - begin, 0.f);
+        std::transform(begin, end, prob.begin(), static_cast<double(*)(double)>(std::exp));
```
IRDonch:
```suggestion
        std::transform(begin, end, prob.begin(), [](float x) { return std::exp(x); });
```
The current draft of the C++ standard [prohibits](http://eel.is/c++draft/namespace.std#6) forming a pointer to most standard library functions. It would be best to start avoiding doing that now. The lambda is shorter anyway.

I've also changed `double` to `float`, since `prob` is a `vector<float>`.

demos/text_detection_demo/src/text_recognition.cpp
```diff
@@ -53,3 +75,81 @@ std::string CTCGreedyDecoder(const std::vector<float> &data, const std::string&
...
+        for(const auto& candidate: last) {
+            float prob_not_blank = 0.f;
+            const std::vector<int>& candidate_sentence = candidate.sentence;
+            if (!candidate_sentence.empty()) {
+                int n = candidate_sentence.back();
+                prob_not_blank = candidate.prob_not_blank * prob[n];
+            }
+            float prob_blank = candidate.prob() * prob[(num_classes - 1)];
+
+            auto check_res = std::find_if(curr.begin(), curr.end(), [candidate_sentence](const BeamElement& n) {
```
IRDonch:
```suggestion
            auto check_res = std::find_if(curr.begin(), curr.end(), [&candidate_sentence](const BeamElement& n) {
```

It seems unnecessary to copy the whole sentence into the lambda.

demos/text_detection_demo/src/text_recognition.cpp
```diff
@@ -53,3 +75,81 @@ std::string CTCGreedyDecoder(const std::vector<float> &data, const std::string&
...
+        std::vector<float> prob = softmax(it, it + num_classes);
+
+        for(const auto& candidate: last) {
+            float prob_not_blank = 0.f;
+            const std::vector<int>& candidate_sentence = candidate.sentence;
+            if (!candidate_sentence.empty()) {
+                int n = candidate_sentence.back();
+                prob_not_blank = candidate.prob_not_blank * prob[n];
+            }
+            float prob_blank = candidate.prob() * prob[(num_classes - 1)];
```
IRDonch:
```suggestion
            float prob_blank = candidate.prob() * prob[num_classes - 1];
```

Redundant parentheses.

demos/text_detection_demo/src/text_recognition.cpp
```diff
@@ -53,3 +75,81 @@ std::string CTCGreedyDecoder(const std::vector<float> &data, const std::string&
...
+                auto extend = candidate_sentence;
+                extend.push_back(i);
+
+                if (candidate_sentence.size() > 0 && candidate.sentence.back() == i) {
+                    prob_not_blank = prob[i] * candidate.prob_blank;
+                } else {
+                    prob_not_blank = prob[i] * candidate.prob();
+                }
+                
+                auto check_res = std::find_if(curr.begin(), curr.end(), [extend](const BeamElement &n) {
```
IRDonch:
```suggestion
                auto check_res = std::find_if(curr.begin(), curr.end(), [&extend](const BeamElement &n) {
```

demos/text_detection_demo/src/text_recognition.cpp
```diff
@@ -53,3 +75,81 @@ std::string CTCGreedyDecoder(const std::vector<float> &data, const std::string&
     }
     return res;
 }
+
+std::string CTCBeamSearchDecoder(const std::vector<float> &data, const std::string& alphabet, char pad_symbol, double *conf, int bandwidth) {
+    std::string res = "";
```
IRDonch: `res` is not used until the very end of the function, so it would be better for readability to move this declaration to the bottom. Also, the initializer is redundant.

FengYen-Chang:
To all, Thanks for your recommendations and review.

IRDonch:
Jenkins: test this please

PR #777, opened 2020-01-17
AC: sharing bitstream from previous run

eaidova:
workaround for sg2, where we can not program several bitstream in config (but we need be able validate it)


vladimir-dudnik:
@eaidova do we still need this workaround?

eaidova:
@vladimir-dudnik ,as i know yes, nobody fix this problem and i think the problem will not be fixed in near future. This WR allows to run models where several datasets provided. Besides that, it makes sense do not program already programmed bitstream twice.

PR #781, opened 2020-01-20
AC: integrate Fashion MNIST annotation converter

eaidova:


eaidova:
> The file name should be fashion_mnist.py, rather than fashon_mnist.py.

@asutic, thank you, corrected

PR #782, opened 2020-01-20
Update for person-detection-asl-0001

druzhkov-paul:


druzhkov-paul:
@vladimir-dudnik please take a look.

PR #783, opened 2020-01-20
text-image-super-resolution-0001: fix the documented input shape

IRDonch:


IRDonch:
My bad, didn't realize that that's where it was broken.

PR #787, opened 2020-01-22
AC: fix brain-tumor-segmentation-0002

eaidova:
fix typo in reader name


eaidova:
@vladimir-dudnik  could you please approve as soon as possible, it affects quantization

PR #790, opened 2020-01-22
Reset person-reidentification-retail-0031 back to the R3 files

IRDonch:
There is a known conversion problem with current MO which makes the resulting model unable to support dynamic batching. So for now, we'll use the files from R3.


IRDonch:
After an internal discussion we decided to abandon this idea and use the new models despite the problems with dynamic batching (and to point anyone who needs dynamic batching to the R3 models). Closing.

PR #800, opened 2020-01-28
demos/monitors: beautify

Wovchena:
Make plot lines reach edges of graphs.
Make graph titles darker to contrast with light background.w

### DIFF
--- a/demos/common/monitors/presenter.cpp
+++ b/demos/common/monitors/presenter.cpp
@@ -47,10 +47,13 @@ Presenter::Presenter(const std::string& keys, int yPos, cv::Size graphSize, std:
     Presenter{strKeysToMonitorSet(keys), yPos, graphSize, historySize} {}
 
 void Presenter::addRemoveMonitor(MonitorType monitor) {
-    int sampleStep = std::max(1, static_cast<int>(graphSize.width / historySize));
-    // add an extra element to place it on the edge of a graph or out of it
-    unsigned updatedHistorySize = graphSize.width / sampleStep + 1;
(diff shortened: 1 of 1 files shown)


demos/common/monitors/presenter.cpp
```diff
@@ -48,7 +48,8 @@ Presenter::Presenter(const std::string& keys, int yPos, cv::Size graphSize, std:
 
 void Presenter::addRemoveMonitor(MonitorType monitor) {
     int sampleStep = std::max(1, static_cast<int>(graphSize.width / historySize));
```
IRDonch: Doesn't the `sampleStep` need to be changed too? E.g. if `historySize` is 2, `sampleStep` ought to be equal to `graphSize.width` (since there would be one point on the very left and one point on the very right).

Wovchena: Sorry, I don't understand your proposal.
Given `historySize == 2`, `sampleStep` becomes `graphSize.width/2` or `1` if 2 >= graphSize.width. So the only case when `sampleStep == graphSize.width` is when `graphSize.width == 1`. In that case `updatedHistorySize` will be 2, which I think is fine.

IRDonch: > Given historySize == 2, sampleStep becomes graphSize.width/2 or 1 if 2 >= graphSize.width.

Yeah, but isn't this wrong? `sampleStep` is the distance between points in the graph, right? So if there are only two points, then the left point will be on the very left, the right point will be on the very right, so the distance between them should be the entire width of the graph.

Wovchena: updated

PR #803, opened 2020-01-29
Bump all dependency versions

IRDonch:


ci/requirements-ac.txt
```diff
@@ -4,15 +4,14 @@
 #
 #    pip-compile --output-file=ci/requirements-ac.txt tools/accuracy_checker/requirements.in
 #
-joblib==0.14.0            # via scikit-learn
-nibabel==2.5.1
-numpy==1.17.2
-pillow==6.2.0
+joblib==0.14.1            # via scikit-learn
+nibabel==3.0.1
+numpy==1.17.5
```
eaidova: possibly we can change numpy version since pycocotools was adopted to work with numpy 1.18+ (now in setup.py numpy<18 used)
anyway windows users should install pycocotools manually and handle this problem

IRDonch: That's up to you. This file is regenerated from `tools/accuracy_checker/requirements.in`, so if you update that, I'll update this.

eaidova: well, I update it https://github.com/opencv/open_model_zoo/pull/806

IRDonch: Alright, I've updated this file too.

PR #805, opened 2020-01-30
Add ac configs for googlenet-v4, resnet-50

jkamelin:


tools/accuracy_checker/configs/resnet-50-official.yml
```diff
@@ -0,0 +1,49 @@
+models:
+  - name: resnet-50-official
```
eaidova: hm... does it real should be named official?
historicalyl,  original resnet is caffe model if I right know

eaidova: https://github.com/KaimingHe/deep-residual-networks

tools/accuracy_checker/configs/googlenet-v4-tf.yml
```diff
@@ -0,0 +1,45 @@
+models:
+  - name: googlenet-v4-tf
+    launchers:
+      - framework: tf
+        model: public/inception_v4.frozen.pb
```
eaidova:
```suggestion
        model: public/googlenet-v4-tf/inception_v4.frozen.pb
```

tools/accuracy_checker/configs/googlenet-v4-tf.yml
```diff
@@ -0,0 +1,45 @@
...
+        adapter: classification
+
+    dataset: 
+      - name: imagenet_1001_classes
+        preprocessing:
+          - type: bgr_to_rgb
+          - type: normalization
+            mean: [127.5,127.5,127.5]
```
eaidova: you can use just 127.5 if all values are the same

tools/accuracy_checker/configs/googlenet-v4-tf.yml
```diff
@@ -0,0 +1,45 @@
...
+            std: 127.5
+          - type: resize
+            size: 300
```
eaidova: are you sure about 300? crop only one pixel looks strange :)

eaidova: Doe it should not be a standart inception preprocessing (crop central fraction 0.875 + resize to input shape)?
https://github.com/tensorflow/models/blob/master/research/inception/inception/image_processing.py#L279

eaidova:
@jkamelin please do not forget synchronize with internal configs (I mean changes in preprocessing)

PR #807, opened 2020-01-30
segmentation: update

Wovchena:
Rewrite the demo to use OpnCV.
Delete the stuff that became unused after rewriting.


IRDonch:
The Windows build failed.

demos/segmentation_demo/README.md
```diff
@@ -1,14 +1,12 @@
...
-segmentation networks like FCN8.
+This topic demonstrates how to run the Image Segmentation demo application, which does inference using semantic segmentation networks.
 
 > **NOTE:** This topic describes usage of C++ implementation of the Image Segmentation Demo. For the Python* implementation, refer to [Image Segmentation Python* Demo](../python_demos/segmentation_demo/README.md).
 
 ## How It Works
 
-Upon the start-up the demo application reads command line parameters and loads a network and an image to the
-Inference Engine plugin. When inference is done, the application creates an output image.
+Upon the start-up the demo application reads command line parameters and loads a network. The demo runs inference and shows results for each image captured from an input and.
```
IRDonch: > from an input and.

Truncated sentence?

demos/segmentation_demo/README.md
```diff
@@ -17,36 +15,37 @@ Inference Engine plugin. When inference is done, the application creates an outp
...
+    -no_show                  Optional. Do not visualize inference results.
+    -u                        Optional. List of monitors to show initially.
 ```
 
-Running the application with the empty list of options yields the usage message given above and an error message.
+Running the application with the empty list of options yields an error message.
 
-To run the demo, you can use public or pre-trained models. To download the pre-trained models, use the OpenVINO [Model Downloader](../../tools/downloader/README.md) or go to [https://download.01.org/opencv/](https://download.01.org/opencv/).
+To run the demo, you can use public or pre-trained models. To download the pre-trained models, use the OpenVINO [Model Downloader](../../tools/downloader/README.md) or go to [https://download.01.org/opencv/](https://download.01.org/opencv/). The list of pre-trained models supported by the demo is in [models.lst](./models.lst).
```
IRDonch:
```suggestion
To run the demo, you can use public or pre-trained models. To download the pre-trained models, use the OpenVINO [Model Downloader](../../tools/downloader/README.md) or go to [https://download.01.org/opencv/](https://download.01.org/opencv/). The list of models supported by the demo is in [models.lst](./models.lst).
```

Given that you've added `deeplabv3` to the list, I don't think "pre-trained" is appropriate.

demos/tests/cases.py
```diff
@@ -225,15 +225,17 @@ def single_option_cases(key, *args):
...
+        TestCase(options={'-no_show': None, **MONITORS}),
         [
             TestCase(options={
                 '-m': ModelArg('road-segmentation-adas-0001'),
-                '-i': DataDirectoryArg('road-segmentation-adas'),
-            }),
-            TestCase(options={
-                '-m': ModelArg('semantic-segmentation-adas-0001'),
-                '-i': DataDirectoryArg('semantic-segmentation-adas'),
+                '-i': DataPatternArg('semantic-segmentation-adas'),
```
IRDonch:
```suggestion
                '-i': DataPatternArg('road-segmentation-adas'),
```

demos/segmentation_demo/main.cpp
```diff
@@ -2,29 +2,23 @@
...
+#include <samples/ocv_common.hpp>
 #include <samples/slog.hpp>
-#include <samples/args_helper.hpp>
-
-#include <vpu/vpu_tools_common.hpp>
-#include <vpu/vpu_plugin_config.hpp>
 
 #include "segmentation_demo.h"
 
-using namespace InferenceEngine;
```
IRDonch: Could you not remove this, at least for now? It makes the diff a lot harder to read. Plus, we seem to have this declaration in every other demo, so it would be inconsistent to just remove it from here.

demos/segmentation_demo/README.md
Wovchena: I will accept it to keep consistency. But I've never thought about it until now: this is strange to have formulations like `public or pre-trained models`. It should be `public or Intel models`. `deeplabv3` is still pre-trained, just not by Intel.

demos/segmentation_demo/README.md
Wovchena: Updated

demos/segmentation_demo/main.cpp
Wovchena: Okay

demos/segmentation_demo/README.md
IRDonch: I agree, we need to revamp the terminology at some point.

demos/segmentation_demo/main.cpp
```diff
@@ -89,178 +64,141 @@ int main(int argc, char *argv[]) {
...
         }
 
-        /** Printing device version **/
-
         slog::info << "Device info" << slog::endl;
         std::cout << ie.GetVersions(FLAGS_d);
 
-        // -----------------------------------------------------------------------------------------------------
-
-        // --------------------------- 2. Read IR Generated by ModelOptimizer (.xml and .bin files) ------------
```
IRDonch: Why did you remove these headings?

demos/segmentation_demo/main.cpp
```diff
@@ -89,178 +64,141 @@ int main(int argc, char *argv[]) {
...
-            }
-
-            item.second->setPrecision(Precision::FP32);
+        ICNNNetwork::InputShapes inputShapes = network.getInputShapes();
+        if (inputShapes.size() != 1)
+            throw std::runtime_error("Demo supports topologies only with 1 input");
+        const std::string& inName = inputShapes.begin()->first;
+        SizeVector& inSizeVector = inputShapes.begin()->second;
+        if (inSizeVector.size() != 4 || inSizeVector[1] != 3)
+            throw std::runtime_error("3 channel 4 dimentional model's input is expected");
```
IRDonch:
```suggestion
            throw std::runtime_error("3-channel 4-dimensional model's input is expected");
```

demos/segmentation_demo/main.cpp
```diff
@@ -89,178 +64,141 @@ int main(int argc, char *argv[]) {
...
+        inputInfo.getPreProcess().setResizeAlgorithm(ResizeAlgorithm::RESIZE_BILINEAR);
+        inputInfo.setLayout(Layout::NHWC);
+        inputInfo.setPrecision(Precision::U8);
+
+        const OutputsDataMap& outputsDataMap = network.getOutputsInfo();
+        if (outputsDataMap.size() != 1) throw std::runtime_error("Demo supports topologies only with 1 output");
+        const std::string& outName = outputsDataMap.begin()->first;
+        Data& data = *outputsDataMap.begin()->second;
+        // alternative to resetting the output precision is to have a switch
+        // statement in postprocessing handling other precision types
```
IRDonch: This comment doesn't quite make sense. If that alternative is better, why isn't it implemented instead? If _this_ alternative is better, then why have this comment at all?

demos/segmentation_demo/main.cpp
```diff
@@ -89,178 +64,141 @@ int main(int argc, char *argv[]) {
...
+                    maskImg.at<cv::Vec3b>(rowId, colId) = colors[classId];
                 }
             }
-            /** Dump resulting image **/
-            std::string fileName = "out_" + std::to_string(image) + ".bmp";
-            std::ofstream outFile(fileName, std::ofstream::binary);
-            if (!outFile.is_open()) {
-                throw std::logic_error("Can't open file : " + fileName);
+            cv::resize(maskImg, resImg, inImg.size());
+            resImg = inImg * blending + resImg * (1 - blending);
```
IRDonch: Why not `cv::addWeighted`?

demos/segmentation_demo/main.cpp
```diff
@@ -89,178 +64,141 @@ int main(int argc, char *argv[]) {
...
-                    for (size_t ch = 0; ch < num_channels; ++ch) {
-                        /**          [images stride + channels stride + pixel id ] all in bytes            **/
-                        data[image_id * image_size * num_channels + ch * image_size + pid] = imagesData.at(image_id).get()[pid*num_channels + ch];
-                    }
-                }
-            }
+        ExecutableNetwork executableNetwork = ie.LoadNetwork(network, FLAGS_d);
+        InferRequest inferRequest = executableNetwork.CreateInferRequest();
+
+        cv::VideoCapture cap;
```
IRDonch: I noticed that you removed image input support. I don't think we should do that. It's useful to be able to run the demo on individual images.

demos/segmentation_demo/main.cpp
Wovchena: Demos are not IE API examples. If one doesn't get what is happening at that level (like read net, load net, create InferRequest), one should refer to samples. Though it can be useful to have such comments explaining pre/prost processing steps and overall processing pipeline (from reading an image to showing results) solutions, such comments shouldn't focus on API.

demos/segmentation_demo/main.cpp
Wovchena: It was unexpected for me to get different types of data for deeplabv3 and any of 2 Intels' models. I think it can be useful to point it out.

As for picking a better solution, I can't say which is better. It is simpler to write/read just `data.setPrecision(Precision::FP32);`, but having switch can possibly bring faster processing.

demos/segmentation_demo/main.cpp
Wovchena: What's the difference?
Even docs say `The function can be replaced with a matrix expression: dst = src1*alpha + src2*beta + gamma;`.

demos/segmentation_demo/main.cpp
Wovchena: `cv::VideoCapture` can process single image as input. New implementation lacks only ability to read from a folder (BTW it seems `VideoCapture` can't read from folders because it is unclear how to do it: what the order of images should be, should it be recursive. And that pattern `%04d.jpg` for images was introduced instead). Supporting both ways will make the code harder to read because there will be `if` deciding how to iterate over the input. The solution could be to introduce a wrapper handling different types of inputs, maybe even implementing iterators for it. But until then I think it is better not to add such functionality, especially taking into account that most demos also don't have that feature.

demos/segmentation_demo/main.cpp
IRDonch: Fair enough. I thought it could've been more efficient, since it's a single operation.

demos/segmentation_demo/main.cpp
IRDonch: > Demos are not IE API examples. If one doesn't get what is happening at that level (like read net, load net, create InferRequest), one should refer to samples.

Fair point.

> Though it can be useful to have such comments explaining pre/prost processing steps and overall processing pipeline (from reading an image to showing results) solutions, such comments shouldn't focus on API.

Sure, but you didn't add such comments. :-)

I'm not saying that these headings should stay as-is, I just think they're useful in principle to break up the function into smaller chunks.

demos/segmentation_demo/main.cpp
IRDonch: > I think it can be useful to point it out.

But you didn't point it out. This comment doesn't say why one might expect different data types.

> but having switch can possibly bring faster processing.

Can it, really? You're doing post-processing on the CPU, and x86 doesn't have half-precision floating-point arithmetic.

demos/segmentation_demo/main.cpp
IRDonch: I don't think we need folder support (the ability to use patterns should make up for that).

However, single image support is useful, and I don't think the current implementation supports it well enough. I just tried it and when you give the demo an image it quits immediately, before the output window even has time to appear. I think we should at least start the demo in a paused state if we detect that a single image was given.

demos/segmentation_demo/main.cpp
Wovchena: > Sure, but you didn't add such comments. :-)

That's why I said can, but not must :)

> they're useful in principle to break up the function into smaller chunks.

I split blocks of code with a single line.

demos/segmentation_demo/main.cpp
Wovchena: > Can it, really? You're doing post-processing on the CPU, and x86 doesn't have half-precision floating-point arithmetic.

I think the default output type of deeplabv3 is sort of int representing class ids, so picking a solution with switch will eliminate casting from int to float by IE, and back from float to int by the demo.

> But you didn't point it out. This comment doesn't say why one might expect different data types.

I will consider extending the comment.

demos/segmentation_demo/main.cpp
Wovchena: I agree. A check, that VideoCapture has only one image will be not clear because there is no such method answering exactly that question, and we can't just always start the demo paused because of tests. This pausing can be regulated by `-no_show` or delay key could be introduced. Delay key would also enable kind of slide show with a number like 1000. And it will be clearer that pausing depends on delay but not `-no_show`. But I don't know whether such functionality with delay key is needed.

demos/segmentation_demo/main.cpp
Wovchena: Updated

demos/segmentation_demo/main.cpp
Wovchena: Updated

IRDonch: OK, I think this is a reasonable solution. Although, FWIW:

> A check, that VideoCapture has only one image will be not clear because there is no such method answering exactly that question

Wouldn't comparing `CAP_PROP_FRAME_COUNT` against 1 answer it?

demos/segmentation_demo/main.cpp
IRDonch: Alright, this is much better.

demos/segmentation_demo/main.cpp
Wovchena: A video with a single frame will also has `CAP_PROP_FRAME_COUNT ` 1. We could pretend we don't care whether it is a short video or an image, but a video of 2 frames should probably also run paused. It becomes unclear how to decide at which value to run the demo paused.

IRDonch: OK, fair enough.

IRDonch:
Jenkins please retry a build

PR #813, opened 2020-01-31
[DOC] Descriptions for public modes are auto updated

eizamaliev:
Descriptions was updated according to descriptions from documentation.
Additionally, description made more readable

### DIFF
--- a/models/public/caffenet/model.yml
+++ b/models/public/caffenet/model.yml
@@ -13,8 +13,7 @@
 # limitations under the License.
 
 description: >-
-  CaffeNet model. Used for classification. For details see paper
-  <https://arxiv.org/abs/1408.5093>.
+  CaffeNet* model is used for classification. For details see paper <https://arxiv.org/abs/1408.5093>.
 task_type: classification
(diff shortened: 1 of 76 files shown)


models/public/caffenet/model.yml
```diff
@@ -13,8 +13,7 @@
 # limitations under the License.
 
 description: >-
-  CaffeNet model. Used for classification. For details see paper
-  <https://arxiv.org/abs/1408.5093>.
+  CaffeNet* model is used for classification. For details see paper <https://arxiv.org/abs/1408.5093>.
```
IRDonch: I don't think these asterisks make sense here, since there is no footnote to explain them.

eizamaliev: These asterisks exist in documentation files and script just takes it as difference.

IRDonch: This doesn't really address my comment... but I suppose it would be difficult to automatically distinguish between footnotes and asterisks used for other reasons, so never mind.

PR #819, opened 2020-02-03
AC: retrun back to numpy 1.17.5

eaidova:


eaidova:
since in repo cocoAPI updated, on pypi not, we need to use old numpy

PR #820, opened 2020-02-03
mask_rcnn_demo: use the documented output of the models

IRDonch:
The documented output is `reshape_do_2d`, not `detection_output`. The current code has worked so far, because a layer named `detection_output` also exists in the network; however, it's an internal layer and we shouldn't be relying on it.

Note that currently, `detection_output` is also marked by MO as an output (unless you use IR v7). This might, however, be a bug. An internal report already exists for it.

### DIFF
--- a/demos/mask_rcnn_demo/README.md
+++ b/demos/mask_rcnn_demo/README.md
@@ -28,8 +28,8 @@ Options:
       -l "<absolute_path>"            Required for CPU custom layers. Absolute path to a shared library with the kernels implementations.
           Or
       -c "<absolute_path>"            Required for GPU custom kernels. Absolute path to the .xml file with the kernels descriptions.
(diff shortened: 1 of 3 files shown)


IRDonch:
> Where can I find it?

[Model documentation](https://github.com/opencv/open_model_zoo/blob/8d8d5f0b0c71386d31c43cc53af853d41bc65305/models/public/mask_rcnn_inception_v2_coco/mask_rcnn_inception_v2_coco.md).

> There is also do_reshaped_OutputOp (which has type of "Result", but it is not output layer) after reshape_do_2d in mask_rcnn_resnet50_atrous_coco.xml for me.

I don't think that's relevant. The only thing that matters is what IE considers to be an output.

IRDonch:
Jenkins please retry a build

PR #821, opened 2020-02-04
AC: support SRGAN

eaidova:
update adapter and add some postprocessing steps for enabling [SRGAN](https://github.com/tensorlayer/srgan) model

improvements:
  - [x] allow value range as mean and std + precomputed (for cifar and imagenet) like in preprocessing
  - [x]   support size specification in resize


eaidova:
> please note on pre-commit test fail

I forse pushed the fix for this failure now

PR #824, opened 2020-02-04
Fix pillow resize

eaidova:


eaidova:
@vladimir-dudnik 

> there is pre-commit fail

I forgot update tests, currently should pass

PR #829, opened 2020-02-06
demos: Fix various Cppcheck warnings

IRDonch:
See the individual commits for more details.


demos/interactive_face_detection_demo/main.cpp
```diff
@@ -289,7 +287,7 @@ int main(int argc, char *argv[]) {
                     float intensity_mean = calcMean(prev_frame(rect));
 
                     if ((face == nullptr) ||
-                        ((face != nullptr) && ((std::abs(intensity_mean - face->_intensity_mean) / face->_intensity_mean) > 0.07f))) {
+                        ((std::abs(intensity_mean - face->_intensity_mean) / face->_intensity_mean) > 0.07f)) {
```
Wovchena: Why was it removed? The behavior for `operator->` is undefined if the stored pointer is null.

demos/security_barrier_camera_demo/main.cpp
```diff
@@ -65,22 +65,19 @@ struct BboxAndDescr {
...
-        if (this != &other) {  // self-assignment check expected
-            this->actualInferRequests = other.actualInferRequests;
-            for (auto& ir : this->actualInferRequests) {
-                this->inferRequests.container.push_back(ir);
-            }
+    InferRequestsContainer(const InferRequestsContainer&) = delete;
+    InferRequestsContainer& operator=(const InferRequestsContainer&) = delete;
+
+    void assign(const std::vector<InferRequest>& inferRequests) {
+        actualInferRequests = std::move(inferRequests);
```
Wovchena:
```suggestion
        actualInferRequests = inferRequests;
```
There is no `operator=( const vector&& other )` for `vector`. And `operator=( const vector& other );` will be used any way.

demos/interactive_face_detection_demo/main.cpp
IRDonch: The second subexpression will only be evaluated if the first subexpression is false, and thus `face != nullptr`.

demos/security_barrier_camera_demo/main.cpp
IRDonch: Good catch; fixed. It must've been a leftover from an earlier approach I tried.

PR #847, opened 2020-02-12
downloader: hash files while they're downloading

IRDonch:
This avoid rereading the entire file from disk once it's written.


IRDonch:
(made a tiny bugfix)

PR #853, opened 2020-02-13
demos: fix the languages specified for code blocks in the docs

IRDonch:
Mostly, just remove `sh` from blocks with help text, since they are not in any formal language. But there are also a few cases where a block was labeled with the wrong language.

### DIFF
--- a/demos/README.md
+++ b/demos/README.md
@@ -167,14 +167,14 @@ The recommended Windows* build environment is the following:
 
 To build the demo applications for Windows, go to the directory with the `build_demos_msvc.bat`
 batch file and run it:
-```sh
+```bat
 build_demos_msvc.bat
 ```
 
 By default, the script automatically detects the highest Microsoft Visual Studio version installed on the machine and uses it to create and build
 a solution for a demo code. Optionally, you can also specify the preffered Microsoft Visual Studio version to be used by the script. Supported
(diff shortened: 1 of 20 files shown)


demos/README.md
```diff
@@ -265,7 +265,7 @@ For example, for the **Debug** configuration, go to the project's
 **Configuration Properties** to the **Debugging** category and set the `PATH`
 variable in the **Environment** field to the following:
 
-```sh
+```
```
Wovchena:
```suggestion
```bat
```

IRDonch: Why? It's not a command.

Wovchena: It is. `PATH=<INSTALL_DIR>\deployment_tools\inference_engine\bin\intel64\Debug;<INSTALL_DIR>\opencv\bin;%PATH%` exports a new value for `PATH` var. Except for `<INSTALL_DIR>` part. But there is a similar update for `<INSTALL_DIR>\bin\setupvars.bat` above.

IRDonch: Read the text above. This is supposed to be a value for the user to set in a configuration dialog, not a command.

It's not even a valid Batch command. :-) You can't do variable assignments with just an equals sign.

Wovchena: My bad

PR #855, opened 2020-02-13
Add a script to update ci/requirements-*.txt

IRDonch:
This automates the process I've been doing manually so far. This will make it easier, and will enable everyone else to do it too.

### DIFF
--- a/ci/update-requirements.py
+++ b/ci/update-requirements.py
@@ -1,0 +1,58 @@
+#!/usr/bin/env python3
+
+"""
+This script updates all of the requirements-*.txt files in this directory
+with the most recent package versions.
+
+It uses pip-compile (https://github.com/jazzband/pip-tools), so install that
+before running it.
+"""
+
+import os
+import subprocess
+import sys
+
+from pathlib import Path
+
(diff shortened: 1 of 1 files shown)


ci/update-requirements.py
```diff
@@ -0,0 +1,58 @@
...
+
+    openvino_dir = Path(os.environ['INTEL_OPENVINO_DIR'])
+
+    pip_compile('ci/requirements-ac.txt',
+        'tools/accuracy_checker/requirements.in')
+    pip_compile('ci/requirements-ac-test.txt',
+        'tools/accuracy_checker/requirements.in', 'tools/accuracy_checker/requirements-test.in')
+    pip_compile('ci/requirements-conversion.txt',
+        'tools/downloader/requirements-pytorch.in', 'tools/downloader/requirements-caffe2.in',
+        openvino_dir / 'deployment_tools/model_optimizer/requirements.txt')
```
eaidova: if I right remember there is special var for MO directory. Possibly you can use it directly instead openvino_dir

eaidova: ok, sorry, I missed that  you also use openvino_dir for another prerequisite

PR #861, opened 2020-02-18
Upgrade ASL Recognition demo

evgeny-izutov:
The upgrade for the ASL Recognition demo includes:
- Tracking of each person in a scene (instead of very first only).
- Action recognition for the arbitrary specified person.
- Switching between active persons by pressing button with appropriate numeric key (according bbox ID).
- Switching between tracking and recognition modes by pressing `space` button.
- Visualization of video streams with different frame rate (action prediction and sample playing) in a separate process.
- Library of ASL video samples.


evgeny-izutov:
@yssemaev, we add ASL video samples, so please take a look.

evgeny-izutov:
@vladimir-dudnik, video samples are moved to [PR](https://github.com/intel-iot-devkit/sample-videos/pull/2).

demos/python_demos/asl_recognition_demo/asl_recognition_demo/video_library.py
```diff
@@ -74,48 +75,95 @@ def parse_source_paths(input_dir, valid_names):
...
 
-        source_name, source_path = self.cur_source
+        if self._play_process is not None and self._play_process.is_alive():
+            return
 
-        if self.cap is None:
-            self.cap = cv2.VideoCapture(source_path)
+        self._play_process = \
+            Process(target=self._play,
+                    args=(self.visualizer, self.cur_source_id, self.source_paths,
```
IRDonch: Is the visualizer picklable? It doesn't look like it would be, given that it stores a process handle, and if so, this won't work on Windows.

demos/python_demos/asl_recognition_demo/asl_recognition_demo/video_library.py
```diff
@@ -74,48 +75,95 @@ def parse_source_paths(input_dir, valid_names):
...
+
+    @staticmethod
+    def _play(visualizer, cur_source_id, source_paths, changed_source_flag, max_image_size, trg_time_step):
+        """Produces live frame from the active video source"""
+
+        cap = None
+
+        start_time = time.perf_counter()
+        while True:
+            if changed_source_flag.value:
```
IRDonch: Is the flag even necessary? Couldn't you store the value of `cur_source_id` in a local variable and check if `cur_source_id` changed on each iteration?

demos/python_demos/asl_recognition_demo/asl_recognition_demo/visualizer.py
```diff
@@ -0,0 +1,112 @@
...
+        self._last_key = Value('i', -1, lock=True)
+        self._need_stop = Value('i', False, lock=False)
+        self._worker_process = None
+        self._tasks = dict()
+
+    def register_window(self, name):
+        """Allocates resources for the new window"""
+
+        if self._worker_process is not None and self._worker_process.is_alive():
+            raise EnvironmentError('Cannot add the window for running visualizer')
```
IRDonch: `EnvironmentError` is an alias of `OSError`, which is not the appropriate exception here.

demos/python_demos/asl_recognition_demo/asl_recognition_demo/visualizer.py
```diff
@@ -0,0 +1,112 @@
...
+
+        self._tasks[name] = Queue(1)
+
+    def get_key(self):
+        """Returns the value of pressed key"""
+
+        out_key = self._last_key.value
+        if out_key != -1:
```
IRDonch: Couldn't this condition be removed? If it's already `-1`, setting it to `-1` should be fine.

demos/python_demos/asl_recognition_demo/asl_recognition_demo/visualizer.py
```diff
@@ -0,0 +1,112 @@
...
+            with self._last_key.get_lock():
```
IRDonch: This `with` should contain the `out_key` assignment as well, since otherwise it's possible to miss keypresses that happen between the read and write of `self._last_key.value`.

demos/python_demos/asl_recognition_demo/asl_recognition_demo/visualizer.py
```diff
@@ -0,0 +1,112 @@
...
+    def show(self, frame, name):
+        """Shows frame in the specified window"""
+
+        if name not in self._tasks.keys():
+            raise ValueError('Cannot show unregistered window: {}'.format(name))
+
+        self._tasks[name].put(np.copy(frame), True)
+
+    def start(self):
+        """Starts internal threads"""
```
IRDonch: Why is it necessary to do the visualization in a background process? Couldn't the main process do it?

demos/python_demos/asl_recognition_demo/asl_recognition_demo/visualizer.py
```diff
@@ -0,0 +1,112 @@
...
+        while not need_stop.value:
+            for name, frame_queue in tasks.items():
+                if not frame_queue.empty():
+                    frame = frame_queue.get(True)
+
+                    cv2.imshow(name, frame)
+
+            key = cv2.waitKey(1)
+            if key != -1:
+                with last_key.get_lock():
```
IRDonch: Unnecessary lock.

demos/python_demos/asl_recognition_demo/asl_recognition_demo/visualizer.py
```diff
@@ -0,0 +1,112 @@
...
+                    last_key.value = key
+
+            end_time = time.perf_counter()
+            elapsed_time = end_time - start_time
+            start_time = end_time
```
IRDonch: If `start_time` is reset here, won't `elapsed_time` contain the wait time from the previous iteration? If so, the current iteration's wait time will be calculated incorrectly.

demos/python_demos/asl_recognition_demo/asl_recognition_demo/visualizer.py
```diff
@@ -0,0 +1,112 @@
...
+            rest_time = trg_time_step - elapsed_time
+            if rest_time > 0.0:
+                time.sleep(rest_time)
+
+        for frame_queue in tasks.values():
+            frame_queue.close()
```
IRDonch: `Queue.close` is supposed to be called by the producer, not the consumer; although in this case I don't see why it's necessary to call it at all.

demos/python_demos/asl_recognition_demo/asl_recognition_demo.py
```diff
@@ -128,9 +143,20 @@ def main():
...
             break
 
-        person_roi = person_tracker.get_roi(frame)
-        if person_roi is not None:
-            recognizer_result = action_recognizer(batch, person_roi)
+        detections, tracker_labels_map = person_tracker.add_frame(
+            frame, len(OBJECT_IDS), tracker_labels_map)
+        if detections is None:
+           active_object_id = -1
+           last_caption = None
```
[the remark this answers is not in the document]
evgeny-izutov: Fixed

demos/python_demos/asl_recognition_demo/asl_recognition_demo.py
```diff
@@ -143,41 +169,55 @@ def main():
...
+            if local_bbox_id in tracker_labels:
+                active_object_id = local_bbox_id
 
         if samples_library is not None:
-            sample_frame = samples_library.get_frame()
-            if sample_frame is not None:
-                cv2.imshow('Sample', sample_frame)
-
-            if key == ord('n'):
+            if key == 83:  # right
```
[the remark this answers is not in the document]
evgeny-izutov: Fixed

demos/python_demos/asl_recognition_demo/asl_recognition_demo/tracker.py
```diff
@@ -0,0 +1,225 @@
...
+            return out_detections
+
+        last_rois = np.concatenate([det.roi for det in last_detections], axis=0)
+        affinity_matrix = self._matrix_iou(last_rois, filtered_rois)
+        cost_matrix = 1.0 - affinity_matrix
+
+        row_ind, col_ind = linear_sum_assignment(cost_matrix)
+        affinity_values = 1.0 - cost_matrix[row_ind, col_ind]
+
+        valid_mathces = affinity_values > self._iou_threshold
```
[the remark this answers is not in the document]
evgeny-izutov: Fixed

demos/python_demos/asl_recognition_demo/asl_recognition_demo/tracker.py
```diff
@@ -0,0 +1,225 @@
...
+        for src_id, trg_id in zip(row_ind, col_ind):
+            det = last_detections[src_id]
+            det.waiting = 0
+            det.duration += 1
+            new_roi = filtered_rois[trg_id]
+            det.roi = self._smooth_roi(det.roi, new_roi.reshape(1, -1), self._smooth_weight)
+            det.conf = filtered_conf[trg_id]
+            out_detections.append(det)
+
+        unmatched_src_ind = set(list(range(len(last_detections)))) - set(row_ind.tolist())
```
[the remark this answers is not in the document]
evgeny-izutov: Fixed

demos/python_demos/asl_recognition_demo/asl_recognition_demo/tracker.py
```diff
@@ -0,0 +1,225 @@
...
+
+            clipped_roi = self._clip_roi(det.roi, frame_size)
+            out_det = Detection(det.id, clipped_roi, det.conf, det.waiting, det.duration)
+            out_detections.append(out_det)
+
+        if len(out_detections) > max_num_detections:
+            out_detections.sort(key=lambda x: x.conf, reverse=True)
+            out_detections = out_detections[:max_num_detections]
+
+        src_det_ids = set([det.id for det in out_detections])
```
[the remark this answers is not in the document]
evgeny-izutov: Fixed

demos/python_demos/asl_recognition_demo/asl_recognition_demo/tracker.py
```diff
@@ -0,0 +1,225 @@
...
+        trg_det_ids = set(list(labels_map.keys()))
+        matched_det_ids = src_det_ids.intersection(trg_det_ids)
```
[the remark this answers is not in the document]
evgeny-izutov: Thanks, changed.

demos/python_demos/asl_recognition_demo/asl_recognition_demo/tracker.py
```diff
@@ -0,0 +1,225 @@
...
+        unused_det_ids = list(set(list(range(max_num_detections))) - matched_det_ids)
+        unused_det_ids.sort()
```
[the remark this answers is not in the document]
evgeny-izutov: Changed

demos/python_demos/asl_recognition_demo/asl_recognition_demo/video_library.py
evgeny-izutov: I'm not sure that we really need to make it picklable because some other parameters (multiprocessing.Value, multiprocessing.Array) are not picklable too but it looks like it able to work on Windows?

demos/python_demos/asl_recognition_demo/asl_recognition_demo/video_library.py
evgeny-izutov: Fixed

demos/python_demos/asl_recognition_demo/asl_recognition_demo/visualizer.py
evgeny-izutov: Replaced on `RuntimeError`.

demos/python_demos/asl_recognition_demo/asl_recognition_demo/visualizer.py
evgeny-izutov: Done.

demos/python_demos/asl_recognition_demo/asl_recognition_demo/visualizer.py
evgeny-izutov: Moved.

demos/python_demos/asl_recognition_demo/asl_recognition_demo/visualizer.py
evgeny-izutov: We need to show video streams with different frame rate (video samples with actual frame rate and action recognition with real one).

demos/python_demos/asl_recognition_demo/asl_recognition_demo/visualizer.py
evgeny-izutov: Removed.

demos/python_demos/asl_recognition_demo/asl_recognition_demo/visualizer.py
evgeny-izutov: Moved outside the sleep section.

demos/python_demos/asl_recognition_demo/asl_recognition_demo/visualizer.py
evgeny-izutov: Removed.

demos/python_demos/asl_recognition_demo/asl_recognition_demo/tracker.py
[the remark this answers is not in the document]
IRDonch: There's another one of these below.

demos/python_demos/asl_recognition_demo/asl_recognition_demo/video_library.py
IRDonch: Have you tested this on Windows? If yes (and it works), then sure, no problem.

> some other parameters (multiprocessing.Value, multiprocessing.Array) are not picklable too

FWIW, `multiprocessing` implements special logic that allows these objects to be shared between processes.

demos/python_demos/asl_recognition_demo/asl_recognition_demo/visualizer.py
IRDonch: There's another similarly inappropriate `EnvironmentError` elsewhere, please replace that too.

demos/python_demos/asl_recognition_demo/asl_recognition_demo/visualizer.py
IRDonch: It would be simpler to move it to the front of the loop body. Then you could also remove the other assignment to `start_time` in this function.

demos/python_demos/asl_recognition_demo/asl_recognition_demo/tracker.py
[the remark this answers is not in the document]
evgeny-izutov: Thanks. Fixed too.

demos/python_demos/asl_recognition_demo/asl_recognition_demo/video_library.py
evgeny-izutov: Ok, I'll check it.

demos/python_demos/asl_recognition_demo/asl_recognition_demo/visualizer.py
evgeny-izutov: Oh, thanks. Fixed.

demos/python_demos/asl_recognition_demo/asl_recognition_demo/visualizer.py
evgeny-izutov: Agree with you, the idea to take into account the time for checking the complex (related to locking a variable) condition is unjustifiably. Changed.

demos/python_demos/asl_recognition_demo/README.md
```diff
@@ -73,9 +73,11 @@ python asl_recognition_demo.py \
 -c ./classes.json
 ```
 
+The demo starts in person tracking mode and to switch it in the action recognition mode you should press `0-9` button with appropriate detection ID (the number in top-left of each bounding box). After that you can switch back to tracking mode by pressing space button.
```
Wovchena: Space usually says for pause. I suggest Tab.

demos/python_demos/asl_recognition_demo/asl_recognition_demo/tracker.py
```diff
@@ -0,0 +1,222 @@
...
+
+        intersect_heights = np.maximum(0.0, intersect_ymax - intersect_ymin)
+        intersect_widths = np.maximum(0.0, intersect_xmax - intersect_xmin)
+
+        intersect_areas = intersect_heights * intersect_widths
+        areas_set_a = ((set_a[:, 2] - set_a[:, 0]) * (set_a[:, 3] - set_a[:, 1])).reshape([-1, 1])
+        areas_set_b = ((set_b[:, 2] - set_b[:, 0]) * (set_b[:, 3] - set_b[:, 1])).reshape([1, -1])
+
+        areas_set_a[np.less(areas_set_a, 0.0)] = 0.0
+        areas_set_b[np.less(areas_set_b, 0.0)] = 0.0
```
Wovchena: 1. It can be replaced with clip: `areas_set_a.clip(0.0, out=areas_set_a)`

2. These lines make me think that it is possible that `set_a[:, 2] - set_a[:, 0] < 0` or `set_a[:, 3] - set_a[:, 1] < 0` but that means that `set_a[:, 2]` and `set_a[:, 0]` should be just swapped (or `abs()` should be used) to get correct results instead of pretending that the rect is a line or a dot.

demos/python_demos/asl_recognition_demo/README.md
evgeny-izutov: IMHO, the `space` button is big and attractive for pressing and also it doesn't require to aim on it :-).

demos/python_demos/asl_recognition_demo/asl_recognition_demo/tracker.py
evgeny-izutov: Regarding the second item, the negative values of `set_a[:, 2] - set_a[:, 0]` or `set_a[:, 3] - set_a[:, 1]` are Ok and mean that bboxes are not intersected and therefore metric value is `0`.

Wovchena: Are bboxes inside of one set intersected with each other? I think your aim is to intersect `set_a` with `set_b`. `set_a[:, 2] - set_a[:, 0]` seems to be computing width (or height).

demos/python_demos/asl_recognition_demo/README.md
Wovchena: OK

demos/python_demos/asl_recognition_demo/asl_recognition_demo/tracker.py
evgeny-izutov: Oh, yes, I've mixed up it. But I prefer to save this logic due to possible mistakes in the bbox prediction part (I'm not sure that it can appear with current detector but some other detectors can filter boxes by setting negative limits).

Wovchena: This is different: you should clip `set_a` than, but not width. And control not only the lower bound but the upper bound as well.

evgeny-izutov: No, negative values are the intentional marker of incorrect detections (in case when we don't want to change the confidence score for any reason).

Wovchena: > I prefer to save this logic due to possible mistakes in the bbox prediction part
negative values are the intentional marker of incorrect detections

So are negative values mistakes or intentional? Or that phrases refer to different objects?

If negative values can be intentional predictions of some detectors, will you point me out where I can learn about that idea?

demos/python_demos/asl_recognition_demo/asl_recognition_demo/video_library.py
evgeny-izutov: I've fixed demo. Now, it works on Windows too.

demos/python_demos/asl_recognition_demo/asl_recognition_demo/tracker.py
evgeny-izutov: An author of a network can do everything (including negative bbox values if it can speed up the future post-processing stage) but in this situation I agree it's overhead. Fixed on `abs`.

Wovchena: The similar update should be done for `areas_set_b`.

But I have realized this `abs` doesn't solve the problem completely because another part of function assumes that `set_a[:, 0]` is upper edge of bbox, `set_a[:, 1]` is left edge of bbox and so on. This comes from `intersect_ymin = np.maximum(set_a[:, 0].reshape([-1, 1]), set_b[:, 0].reshape([1, -1]))`. If `set_a[:, 0]` was lower edge, it would be `intersect_ymin = np.MINimum(set_a[:, 0].reshape([-1, 1]), set_b[:, 0].reshape([1, -1]))`.

I think it can be easier to swap elements if they are in wrong order (or reject such objects) in the very beginning and then remove `intersect_heights = np.maximum(0.0, intersect_ymax - intersect_ymin)`, `intersect_widths = np.maximum(0.0, intersect_xmax - intersect_xmin)`, `iou_values[np.less_equal(union_areas, 0.0)] = 0.0` lines and `abs`.

Actually current implementation already sets IoU for bboxes with wrong vertex order to 0 because of that `intersect_heights = np.maximum(0.0, intersect_ymax - intersect_ymin)`. So `abs` is not needed in that approach too as well as `iou_values[np.less_equal(union_areas, 0.0)] = 0.0` because IoU will be 0 any way. But that implementations looks tricky and unclear.

evgeny-izutov: I've added bbox filtering before the call of IoU function, so there is no need in processing invalid sizes. Regarding the suggestion to remove `intersect_heights = np.maximum(0.0, intersect_ymax - intersect_ymin), intersect_widths = np.maximum(0.0, intersect_xmax - intersect_xmin)` lines (if I correctly understand the suggestion) - negative values mean that there is not intersection... , so we replace it on zeros values to make the final metric for this boxes to be zero too.

Wovchena: You are right

demos/python_demos/asl_recognition_demo/asl_recognition_demo.py
```diff
@@ -98,28 +102,40 @@ def main():
...
+    visualizer = Visualizer(VISUALIZER_TRG_FPS)
+    visualizer.register_window('Demo')
 
     samples_library = None
     if args.samples_dir is not None and exists(args.samples_dir):
-        samples_library = VideoLibrary(args.samples_dir, SAMPLES_WINDOW_SIZE, list(class_map.values()))
+        visualizer.register_window('Gesture library')
+        visualizer.start()
+
+        librarry_queue = visualizer.get_queue('Gesture library')
```
IRDonch:
```suggestion
        library_queue = visualizer.get_queue('Gesture library')
```

evgeny-izutov: Fixed.

PR #886, opened 2020-02-26
Fix YoloV3 detection method.

ngaloppo:


eizamaliev:
Already fixed in [PR 875](https://github.com/opencv/open_model_zoo/pull/875).

PR #888, opened 2020-02-26
midasnet: put the source files in a subdirectory

IRDonch:
This allows us to remove one of the postprocessing steps.

### DIFF
--- a/models/public/midasnet/model.yml
+++ b/models/public/midasnet/model.yml
@@ -23,15 +23,15 @@ description: >-
   The model output is an inverse depth map that is defined up to an unknown scale factor.
 task_type: monocular_depth_estimation
 files:
-  - name: midas_net.py
+  - name: models/midas_net.py
     size: 2662
     sha256: 0bf2e5295b69b3af9adafaf97e154186d29a808325435440aa68c7aa4d25f753
     source: https://raw.githubusercontent.com/intel-isl/MiDaS/ffb70fd13361434114383ce7eb898c2f5bec6176/models/midas_net.py
-  - name: base_model.py
+  - name: models/base_model.py
     size: 355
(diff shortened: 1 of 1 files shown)


models/public/midasnet/model.yml
```diff
@@ -44,26 +44,22 @@ files:
...
-    file: blocks.py
+    file: models/blocks.py
     pattern: 'torch.hub.load\("facebookresearch/WSL-Images", "resnext101_32x8d_wsl"\)'
     replacement: 'models.resnext101_32x8d()'
 conversion_to_onnx_args:
   - --model-name=MidasNet
   - --model-path=$dl_dir
   - --weights=$dl_dir/model.pt
-  - --import-module=midas_net
+  - --import-module=models.midas_net
```
eizamaliev: I think, it is better to set `--model-path=$dl_dir/model`. Same result but key `--model-path` more about path to source code, which is `$dl_dir/models` in that case.

models/public/midasnet/model.yml
```diff
@@ -23,15 +23,15 @@ description: >-
   The model output is an inverse depth map that is defined up to an unknown scale factor.
 task_type: monocular_depth_estimation
 files:
-  - name: midas_net.py
+  - name: models/midas_net.py
```
eizamaliev: I would prefer `model`, since there is only one model in it, not counting base blocks.

models/public/midasnet/model.yml
IRDonch: It would not be the same result, as then I wouldn't be able to remove that one post-processing step, which was the reason I did this change to begin with. :-)

eizamaliev: Understood

PR #899, opened 2020-02-27
AC: write result in csv

eaidova:
simple accuracy checker results writing. Can be useful for manual testing

### DIFF
--- a/tools/accuracy_checker/accuracy_checker/main.py
+++ b/tools/accuracy_checker/accuracy_checker/main.py
@@ -18,7 +18,6 @@ import sys
 from pathlib import Path
 from argparse import ArgumentParser
 from functools import partial
-from copy import deepcopy
 from csv import DictWriter
 
 import cv2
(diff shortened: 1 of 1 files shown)


tools/accuracy_checker/accuracy_checker/main.py
```diff
@@ -266,5 +277,32 @@ def print_processing_info(model, launcher, device, tags, dataset):
...
+            'metric_name': metric_result['name'],
+            'metric_type': metric_result['type'],
+            'metric_value': metric_result['value']
+        })
+        rows.append(result)
+    with open(csv_file, 'a+', newline='') as f:
+        writer = DictWriter(f, fieldnames=field_names)
+        if new_file:
+            writer.writeheader()
+        writer.writerows(rows)
```
IRDonch: Is it really necessary to accumulate the rows in memory first? You could just move the loop here, and write them into the file one by one.

tools/accuracy_checker/accuracy_checker/main.py
```diff
@@ -266,5 +277,32 @@ def print_processing_info(model, launcher, device, tags, dataset):
...
+        'dataset': dataset
+    }
+    rows = []
+    for metric_result in metric_results:
+        result = deepcopy(main_info)
+        result.update({
+            'metric_name': metric_result['name'],
+            'metric_type': metric_result['type'],
+            'metric_value': metric_result['value']
+        })
```
IRDonch: `deepcopy` seems excessive, why not just `result = {**main_info, metric_name: metric_result['name'], ...}`?

eaidova: fixed

tools/accuracy_checker/accuracy_checker/main.py
eaidova: fixed

PR #919, opened 2020-03-04
unify yolo naming

vladimir-dudnik:
rename yolo according to existing ones trained by intel


vladimir-dudnik:
ok, seems original naming looks more consistent. There is YOLO topology achitecture, v1, v2 or v3 is kind of generation of this architecture and tiny is lightweight variant of the particular generation (reduced num of conv). So I'll close this PR.

PR #927, opened 2020-03-05
unet-camvid-onnx-0001 accuracy checker config

vladimir-dudnik:
add unet-camvid-onnx-0001 accuracy checker config

### DIFF
--- a/tools/accuracy_checker/configs/unet-camvid-onnx-0001.yml
+++ b/tools/accuracy_checker/configs/unet-camvid-onnx-0001.yml
@@ -1,0 +1,55 @@
+models:
+  - name: unet-camvid-onnx-0001
+
+    launchers:
+      - framework: dlsdk
+        tags:
+          - FP32
+        model:   intel/unet-camvid-onnx-0001/FP32/unet-camvid-onnx-0001.xml
+        weights: intel/unet-camvid-onnx-0001/FP32/unet-camvid-onnx-0001.bin
+        adapter: segmentation
+
+      - framework: dlsdk
+        tags:
+          - FP16
+        model:   intel/unet-camvid-onnx-0001/FP16/unet-camvid-onnx-0001.xml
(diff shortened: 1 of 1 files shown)


tools/accuracy_checker/configs/unet-camvid-onnx-0001.yml
```diff
@@ -0,0 +1,55 @@
...
+    datasets:
+      - name: CamVid
+
+        preprocessing:
+          - type: resize
+            dst_height: 368
+            dst_width: 480
+            use_pillow: True
+            interpolation: BILINEAR
+          - type: bgr_to_rgb
```
eaidova: why bgr_to_rgb used? There is no way to convert model in BGR?

eaidova: if original model trained with pillow usage, than you need:
```yml
reader: pillow_imread
preprocessing:
         -  type: resize
            dst_height: 368
            dst_width: 480
            use_pillow: True
            interpolation: BILINEAR
         -  type: rgb_to_bgr

vladimir-dudnik: this model was a candidate for Open Model Zoo about year ago. Re bgr_to_rgb it is a good question, need to ask @AlexKoff88 on details.

eaidova: my old config has reading in PIL, so the configuration which I described abowe is correct
https://github.com/opencv/open_model_zoo/pull/694

Probably bgr_to_rgb was used because AC has not support of rgb_to_bgr preprocessor (they makes the same changes, just name mislead)

AlexKoff88: @vshampor, could you please comment on that?

PR #930, opened 2020-03-06
Multi-channel demo video frame loss

AnthonyQuantum:


demos/multi_channel/common/input.cpp
```diff
@@ -484,10 +484,11 @@ void VideoSourceOCV::thread_fn(VideoSourceOCV *vs) {
         if (!result) {
             vs->running = false; // stop() also affects running, so override it only when out of frames
         }
-        if (vs->queue.size() < vs->queueSize || !result) { // queue has space or source run out of frames
```
Wovchena: Why was the comment removed? It is still valid.

demos/multi_channel/face_detection_demo/main.cpp
```diff
@@ -313,7 +313,7 @@ int main(int argc, char* argv[]) {
             }
 
 
-            std::vector<Detections> detections(getTensorHeight(output->getTensorDesc()) / 200);
+            std::vector<Detections> detections(batchSize);
```
Wovchena: The function has a way to extract batch size, so there is no need to introduce an extra argument. The way the function extracts the value is weird, but this is different problem why dldt extends height instead of batch dimension, so introducing `batchSize` argument should be reverted.

PR #934, opened 2020-03-06
New face detection models

Ilya-Krylov:


Ilya-Krylov:
@IRDonch

Ilya-Krylov:
@eizamaliev

tools/accuracy_checker/configs/face-detection-0105.yml
```diff
@@ -0,0 +1,39 @@
+models:
+  - name: face-detection-0105
+
+    launchers:
+      - framework: dlsdk
+        tags:
+          - FP32
+        model:   intel/face-detection-0105/FP32/face-detection-0105.xml
```
eaidova: please add also FP16 section for all new AC configs

tools/accuracy_checker/configs/face-detection-0105.yml
```diff
@@ -0,0 +1,39 @@
...
+        weights: intel/face-detection-0105/FP32/face-detection-0105.bin
+        adapter:
+          type: fcos_person
```
eaidova: hmm... looks like we need to rename adapter to `fcos_single_class` since it used not only for person detection

models/intel/face-detection-0100/description/face-detection-0100.md
```diff
@@ -0,0 +1,53 @@
...
+    - B - batch size
+    - C - number of channels
+    - H - image height
+    - W - image width
+
+   Expected color order - BGR.
+
+## Outputs
+
+1. The "boxes" is a blob with shape: [N, 5], where N is the number of detected
```
eaidova: are you test IRs and sure that they contains the same output names? (my experience of work with `person-detection-asl` says that it can be different...)

models/intel/face-detection-0100/description/face-detection-0100.md
```diff
@@ -0,0 +1,53 @@
+# face-detection-0100
+
+## Use Case and High-Level Description
+
+Face detector based on MobileNetV2 as a backbone with a
```
eizamaliev: Are there any differences bietween *01 - *04 models? They all have same description.

eaidova: input size, backbone, scale

demos/interactive_face_detection_demo/detectors.cpp
```diff
@@ -143,63 +143,75 @@ CNNNetwork FaceDetection::read(const InferenceEngine::Core& ie)  {
...
+        if (objectSize != 7) {
+            throw std::logic_error("Face Detection network output layer should have 7 as a last dimension");
+        }
+        if (outputDims.size() != 4) {
+            throw std::logic_error("Face Detection network output dimensions not compatible shoulld be 4, but was " +
+                                   std::to_string(outputDims.size()));
+        }
+        _output->setPrecision(Precision::FP32);
+    } else {
+        for (const auto& outputLayer: outputInfo) {
```
Wovchena: You should check that there are only 2 outputs.
if no output found with `outputDims.size() == 2 && outputDims[outputDims.size() - 1] == 5`, an exception should be thrown.
Number of dimensions for `labels` blob should also be verified if the output is used.
I also suggest explicitly setting precision for outputs.

demos/interactive_face_detection_demo/detectors.cpp
```diff
@@ -143,63 +143,75 @@ CNNNetwork FaceDetection::read(const InferenceEngine::Core& ie)  {
...
+            const SizeVector outputDims = outputLayer.second->getTensorDesc().getDims();
+            if (outputDims.size() == 2 && outputDims[outputDims.size() - 1] == 5) {
```
Wovchena:
```suggestion
            if (outputDims.size() == 2 && outputDims.back() == 5) {
```

demos/interactive_face_detection_demo/detectors.cpp
```diff
@@ -143,63 +143,75 @@ CNNNetwork FaceDetection::read(const InferenceEngine::Core& ie)  {
...
+                output = outputLayer.first;
+                maxProposalCount = outputDims[0];
+                objectSize = outputDims[outputDims.size() - 1];
```
Wovchena:
```suggestion
                objectSize = outputDims.back();
```

demos/interactive_face_detection_demo/detectors.cpp
```diff
@@ -213,7 +225,50 @@ void FaceDetection::fetchResults() {
     resultsFetched = true;
     const float *detections = request->GetBlob(output)->buffer().as<float *>();
 
-    for (int i = 0; i < maxProposalCount; i++) {
+    for (int i = 0; i < maxProposalCount && objectSize == 5; i++) {
+        Result r;
+        r.label = 1;
```
Wovchena: Should the value be extracted from `labels`?

Wovchena: I have an assumption. These are single class detectors so the label is always 1. And dldt forced the model to have that `labels` output even though it always predicts 1. Than if it is possible, you should assure that `num_classes` is 1 (like the one extracted in `if (outputInfo.size() == 1)` with `num_classes = detOutput->get_attrs().num_classes;`). I also suggest clarifying that models' feature in models' descriptions. (Assuming, if it was possible to remove that output from IR, it would have been done.)

demos/interactive_face_detection_demo/detectors.cpp
Ilya-Krylov: Recently added models actually have more than 2 layers, so I would not check exact number of outputs. I'll better add info about unused (they can not be removed from IR) outputs.

The requested exception has been added if required outputs have not been determined

Explicitly set precision for `boxes`

demos/interactive_face_detection_demo/detectors.cpp
Ilya-Krylov: Changed the code to get label from model output. 

Our current understanding about how output of a alternative (to DetectionOutput) detection model should look like is  to have at least `labels` and `boxes` and despite the fact that current models are single class I would still keep label extraction here.

tools/accuracy_checker/configs/face-detection-0105.yml
Ilya-Krylov: actually this even more common than fcos only. renamed it to 'class_agnostic_detection'

models/intel/face-detection-0100/description/face-detection-0100.md
Ilya-Krylov: added

tools/accuracy_checker/configs/face-detection-0105.yml
Ilya-Krylov: added

models/intel/face-detection-0100/description/face-detection-0100.md
Ilya-Krylov: replaced models by their variants with DetectionOutput

Ilya-Krylov:
@eaidova @eizamaliev @Wovchena  could you please have a look one more time

vladimir-dudnik:
Jenkins please retry a build

tools/accuracy_checker/configs/face-detection-0105.yml
```diff
@@ -0,0 +1,48 @@
...
+  - name: face-detection-0105
+
+    launchers:
+      - framework: dlsdk
+        tags:
+          - FP32
+        model:   intel/face-detection-0105/FP32/face-detection-0105.xml
+        weights: intel/face-detection-0105/FP32/face-detection-0105.bin
+        adapter:
+          type: class_agnostic_detection
```
eaidova: you forgot replace adapter or output is not changed for this model?

eaidova: the same question about 0106

Ilya-Krylov: no, i did not forget. models 0105 and 0106 should use class_agnostic_detection adapter

p.s. I checked their outputs, they are `boxes` and `labels` (as expected)

eaidova: ok, I understand

PR #951, opened 2020-03-12
text_detection remove setBatchSize()

Wovchena:
IE doesn't recomend using setBatchSize() and for that demo using it leads to errors.


Wovchena:
Jenkins please retry a build

PR #957, opened 2020-03-13
AC: use fixed number of samples from kinetics for reproducibility

eaidova:


eaidova:
@eizamaliev  initially I wanted to create samples list, but dataset conversion very time consuming and it does not garantee public reproducibility, so I add command for first n samples extraction from dataset

eaidova:
Jenkins please retry a build

PR #964, opened 2020-03-16
ci/update-requirements.py: don't update versions by default

IRDonch:
This lets people update `ci/requirements-*.txt` when they change one of the input requirements files without introducing unrelated changes to package versions.

### DIFF
--- a/ci/update-requirements.py
+++ b/ci/update-requirements.py
@@ -8,6 +8,8 @@ It uses pip-compile (https://github.com/jazzband/pip-tools), so install that
 before running it.
 """
 
+import argparse
+import functools
 import os
 import subprocess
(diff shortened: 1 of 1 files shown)


ci/update-requirements.py
```diff
@@ -42,16 +48,19 @@ def main():
 
     openvino_dir = Path(os.environ['INTEL_OPENVINO_DIR'])
 
-    pip_compile('ci/requirements-ac.txt',
+    def pc(target, *sources):
```
eaidova: maybe partial function usage will be better?

IRDonch: Yeah, good point. Updated.

PR #965, opened 2020-03-16
Publish candidate models for the 2020.2 release

IRDonch:


eaidova:
I have two questions:
* FP32-INT8 models are not part OMZ anymore? Published only FP16-INT8. I right understand, that they not will be added in future?
* what about excluding pipelining models FP16-INT8 from package? (it is available only for encoder (decoded is not supported by POT now) and nobody validate its accuracy yet)

PR #992, opened 2020-03-24
AC: support RFCN Caffe with class agnostic bounding box regression

eaidova:


vladimir-dudnik:
@jkamelin could you please test this locally?

PR #1000, opened 2020-03-26
AC: added command line argument for vocab prefix

eaidova:


tools/accuracy_checker/accuracy_checker/config/config_reader.py
```diff
@@ -784,6 +787,12 @@ def process_launchers(launchers_configs):
 
 
 def merge_entry_paths(keys, value, args, value_id=0):
+    def select_argument_by_priority(argument_list):
+        for current_arg in argument_list:
+            if current_arg in args and args[current_arg]:
+                return current_arg
+        return argument_list[-1]
```
IRDonch: I would suggest `return next(filter(args.get, argument_list), argument_list[-1])`. It's simple enough that you might not even need a function.

tools/accuracy_checker/accuracy_checker/main.py
```diff
@@ -235,6 +235,11 @@ def build_arguments_parser():
         required=False,
         type=cast_to_bool
     )
+    parser.add_argument(
+        '-v', '--vocab', help="Prefix for vocabulary files for NLP models",
```
IRDonch: This seems really specific. Why do we need a separate prefix just for this?

eaidova: usuually it is stored with model and when the original model stored separatly with IR we need to copy this file to every directory with converted model (e.g. model package for internal validation, IR cache in OV validation and so on) if we not provide the path to model in config it can become a problem to find vocab

IRDonch: OK, but couldn't a more generic solution be implemented? E.g. an ability to add extra paths in which all model-related files would be searched for if they're not found in the main model directory.

eaidova: What do you mean under model-specific files?
It was just fast prototype in format as it was asked in validation,
I want also add opportunity to provide the full path in command line without specification arg in the config, it is the reason why I add this in command line with specific appropriate name

IRDonch: > What do you mean under model-specific files?

I never said "model-specific files".

> It was just fast prototype in format as it was asked in validation,

If you want it to be merged, then it's not a prototype, but a complete feature. :-)

> I want also add opportunity to provide the full path in command line without specification arg in the config, it is the reason why I add this in command line with specific appropriate name

I don't understand what you mean by this.

eaidova: I want to merge it in the next release only, so I have a lot of time to rewrite it : )

IRDonch: Okay, see you in the next release, then. :-)

(Also, consider opening draft pull requests for changes that are not meant to be merged ASAP.)

eaidova: It was prepared as fast workarouund for validation team, but they found another.
In the next case I take in account it and create draft, of course :)

eaidova:
I prefer to close it now to not confuse

PR #1008, opened 2020-03-31
AC: get layout from executble network if compiled model provided

eaidova:
some VPU plugins has NHWC as default layout inside and when we use compiled (using vpu compiler or compile_tool) model instead IR. It can lead to unexpected behavior if we use NCHW as default layout


eaidova:
closed as no need in this workaround anymore

PR #1012, opened 2020-04-01
AC: update check for mask_rcnn adapter's parameters

Anna-Mironova:


tools/accuracy_checker/accuracy_checker/adapters/mask_rcnn.py
```diff
@@ -74,8 +74,6 @@ def parameters(cls):
     def configure(self):
         box_outputs = ['classes_out', 'scores_out', 'boxes_out']
         detection_out = 'detection_out'
-        if contains_all(self.launcher_config, [*box_outputs, detection_out]):
-            raise ConfigError('only detection output or [{}] should be provided'.format(', '.join(box_outputs)))
```
eaidova, line 78: but we still need to check that both output formats are no provided toghther
What if `detection_out` and classes_out', 'scores_out', 'boxes_out' are not empty at the same time?

Anna-Mironova: Fixed, please review again

tools/accuracy_checker/accuracy_checker/adapters/mask_rcnn.py
```diff
@@ -71,12 +71,23 @@ def parameters(cls):
 
         return parameters
 
+    def check_params(self):
+        detection_out = self.get_value_from_config('detection_out')
+        classes_out = self.get_value_from_config('classes_out')
+        scores_out = self.get_value_from_config('scores_out')
+        boxes_out = self.get_value_from_config('boxes_out')
```
eaidova, line 78: it can be combined with class fields filling

tools/accuracy_checker/accuracy_checker/adapters/mask_rcnn.py
```diff
@@ -71,12 +71,23 @@ def parameters(cls):
 
         return parameters
 
+    def check_params(self):
```
eaidova, line 74: not clear name for me... Why `check_params` check only part of params and we have additional checks in `configure`?

Besides that, I think it is more logical to separate the realization selection logic than just check of this parameter

tools/accuracy_checker/accuracy_checker/adapters/mask_rcnn.py
Anna-Mironova: Added new function for set values

tools/accuracy_checker/accuracy_checker/adapters/mask_rcnn.py
Anna-Mironova: Added new functions for checking parameters

tools/accuracy_checker/accuracy_checker/adapters/mask_rcnn.py
```diff
@@ -71,26 +70,46 @@ def parameters(cls):
...
+
+        return False
+
+    def is_box_outputs(self):
+        if self.classes_out and self.scores_out and self.boxes_out:
+            return True
+
+        return False
+
+    def is_num_detections_out(self):
```
eaidova, line 85: it is not what I meant.
1. The checks should be used before setting parameters or in the same place where they set
2. I am not sure that we need these methods at all in your context, we always can check property directly where it is necessary instead of methods.

tools/accuracy_checker/accuracy_checker/adapters/mask_rcnn.py
```diff
@@ -72,13 +72,27 @@ def parameters(cls):
         return parameters
 
     def configure(self):
+        def is_detection_out(config):
+            if config.get('detection_out'):
+                return True
+
+            return False
```
eaidova, line 79:
```suggestion
          return bool(config.get('detection_out'))
```

tools/accuracy_checker/accuracy_checker/adapters/mask_rcnn.py
```diff
@@ -72,13 +72,27 @@ def parameters(cls):
...
+            return True
+
         box_outputs = ['classes_out', 'scores_out', 'boxes_out']
-        detection_out = 'detection_out'
-        if contains_all(self.launcher_config, [*box_outputs, detection_out]):
+        if is_detection_out(self.launcher_config) and is_box_outputs(self.launcher_config, box_outputs):
             raise ConfigError('only detection output or [{}] should be provided'.format(', '.join(box_outputs)))
-        self.detection_out = self.get_value_from_config(detection_out)
+
+        self.detection_out = self.get_value_from_config('detection_out')
```
eaidova, line 92: at this moment, you already know which output should be, please use check function results for configure outputs

tools/accuracy_checker/accuracy_checker/adapters/mask_rcnn.py
Anna-Mironova: Done

tools/accuracy_checker/accuracy_checker/adapters/mask_rcnn.py
Anna-Mironova: Done

eaidova:
jenkins please retry a build

PR #1013, opened 2020-04-02
models/public/resnet-50-tf: convert directly from the SavedModel

IRDonch:
No need to use an intermediate conversion step. This results in a model of the same size, with the same accuracy metrics.


tools/accuracy_checker/configs/resnet-50-tf.yml
```diff
@@ -2,11 +2,7 @@ models:
   - name: resnet-50-tf
     launchers:
       - framework: tf
-        model: public/resnet-50-tf/resnet_v1-50.pb
-        adapter: classification
-        inputs:
-          - name: map/TensorArrayStack/TensorArrayGatherV3
-            type: INPUT
+        # TODO
```
IRDonch, line 5: @eaidova Is there a way to make Accuracy Checker work with the original SavedModel?

eaidova: not sure about this model specific parameters, but you can try to use
`saved_model_dir` instead of `model`

eaidova: it is experimental feature in AC, so I am checking it.

eaidova: not sure that it is possible, saved_model contains image loading nodes if i right understand, AC can not remove it.
the same model without jpg suffix loaded, but fails due to uninitilized variables...

IRDonch: Would it maybe be possible to replicate the functionality of [`strip_unused_lib`](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/tools/strip_unused_lib.py) to remove the unneeded nodes?

IRDonch: Actually, if we could use the non-JPEG model, that would be even better. I checked just now and MO can convert it without issues, so clearly the model itself is fine. Maybe AC just doesn't load it correctly?

IRDonch:
I've updated the PR to use the non-JPEG original model, which is easier to convert than the JPEG one, and yields an identical IR (except for layer names). I'll try to make Accuracy Checker work with it later.

IRDonch:
OK, I give up. I tried to make AC work with the SavedModel, and I did manage to load it correctly, but it's saved with a fixed batch size of 64, and I don't know of any way to change it to 1 without advanced TensorFlow hackery. The JPEG model _does_ have a flexible batch size, but to use it, you need to insert a new placeholder in the middle of the graph (and strip away the nodes that are no longer connected to the outputs because of this), and I don't think AC can do that (and I don't think it's worth my time to implement it). So I'm abandoning this idea (for now, at least).

eaidova:
@IRDonch the inability of AC to handle original model should not be show-stopper. It is not required for model enabling, support of IR validation should be enough, so I think it is not reason to change your plans to replace model.

IRDonch:
@eaidova I don't really want to replace a model which we can validate (in original form) with one we can't. I don't think the benefits of the change outweigh the drawbacks in this case.

PR #1015, opened 2020-04-03
Added link to script for description update

eizamaliev:


eizamaliev:
@vladimir-dudnik, of course we can, and I can add it quickly. But task solved by script is very rare, so its location here  would be unnatural. Nevertheless, I agree with you, we need some additional helper scripts.
@IRDonch, what is your opinion?

IRDonch:
I think the script should be in this repository.

eizamaliev:
Fine. But I warn you, in future we should add any similar script to the repo.

CONTRIBUTING.md
```diff
@@ -72,7 +72,7 @@ The detailed descriptions of file entries provided below.
 
 **`description`**
 
-Description of the model. Must match with the description from the model [documentation](#documentation).
+Description of the model. Must match with the description from the model [documentation](#documentation). Use [this](./tools/description_updater/documentation_updater.py) script for easy update.
```
IRDonch, line 75: I think you should put it in `/ci/` instead. It's not a user-facing tool. (And it'll be reasonable to plug it into CI in the future.)

eizamaliev: I'm disagree with that. I suppose this would be a developer helper tool.  Or you talking just about relocation?

IRDonch: >  I suppose this would be a developer helper tool.

Exactly, so not for users. :-) We don't have a directory for developer tools; I suppose we could make one, but given that this could be of use in CI, I think it would be easier to just stick it into `/ci/`.

tools/documentation_updater/documentation_updater.py
```diff
@@ -0,0 +1,214 @@
...
+    return files
+
+
+def convert(lines):
+    def flatten_links(s):
+        links = re.findall(r"\[.*?\]\(.*?\)", s)
+        for link in links:
+            plain_link = re.sub('[\[\]]', '', link.replace('](', ' (').replace('(', '<').replace(')', '>'))
+            s = s.replace(link, plain_link)
+        return s
```
IRDonch, line 71: Seems like all of this could be done with a single replacement operation, something like `re.sub(r'\[(.*?\)]\((.*?)\)', r'\1 <\2>', s)`.

tools/documentation_updater/documentation_updater.py
```diff
@@ -0,0 +1,214 @@
...
+
+    result = ''
+    list_signs = ['-', '*']
+    for line in lines:
+        if len(line.strip()) == 0:
+            result += '\n'
+        elif line.lstrip()[0] in list_signs:
+            result += '\n'
+        if len(result) > 0 and result[len(result)-1] != '\n':
```
IRDonch, line 80:
```suggestion
        if result and not result.endswith('\n'):
```

tools/documentation_updater/documentation_updater.py
```diff
@@ -0,0 +1,214 @@
...
+            result += ' '
+        result += line.rstrip('\n')
+        result = flatten_links(result)
+        result = result.replace("`", "\"").replace("\*", "*")
```
IRDonch, line 84: This will reprocess all of the lines added so far on each iteration. Shouldn't you only need to do this once?

tools/documentation_updater/documentation_updater.py
```diff
@@ -0,0 +1,214 @@
...
+        for line in readme:
+            if line[0:2] == '##':
+                if not started:
+                    started = True
+                    continue
+                else:
+                    break
+            if started:
+                desc.append(line)
+        desc = convert(desc).strip('\n')
```
IRDonch, line 103: `convert` already strips all whitespace at the end so this `strip` is redundant.

tools/documentation_updater/documentation_updater.py
```diff
@@ -0,0 +1,214 @@
...
+        if desc != '':
+            descriptions[name] = desc
+        else:
+            logging.warning('No description found in {} file. '
+                            'Note, that file must be marked and description must be first chapter'.format(file))
+    return descriptions
+
+
+def get_topologies_from_configs(directory):
```
IRDonch, line 112:
```suggestion
def get_models_from_configs(directory):
```

tools/documentation_updater/documentation_updater.py
```diff
@@ -0,0 +1,214 @@
...
+        logging.error('File with list of topologies {} not found'.format(file))
+        quit(-1)
+
+
+def find_models(topologies, name):
+    models = []
+    for model in topologies:
+        if name == model['name']:
+            models.append(model)
+    return models
```
IRDonch, line 143: All of this seems to be replaceable with a simple list comprehension.

tools/documentation_updater/documentation_updater.py
```diff
@@ -0,0 +1,214 @@
...
+    
+
+def update_topologies(topologies, descriptions, compare=lambda lhs, rhs: rhs == lhs):
```
IRDonch, line 146: The last parameter is redundant, it's never explicitly specified.

ci/documentation_updater/documentation_updater.py
```diff
@@ -0,0 +1,198 @@
...
+    parser.add_argument('-d', '--model-dir', type=str, default='../../models',
+                        help='Path to root directory with models documentation and configuration files')
+    parser.add_argument('--mode', type=str, choices=MODES, default='check',
+                        help='Script work mode: "check" only finds diffs, "update" - updates values')
+    parser.add_argument('--log-level', choices=logging._levelToName.values(), default='WARNING',
+                        help='Level of logging')
+    parser.add_argument('--ignored-files', type=str,
+                        help='List of files which will be ignored')
+    parser.add_argument('--ignored-files-list', type=Path,
+                        help='Path to file with ignored files')
```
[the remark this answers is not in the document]
eizamaliev: Let consider it as a feature for the future. For example, for models with Chinese or  Japanese symbols in description (see [https://github.com/opencv/open_model_zoo/blob/develop/models/intel/handwritten-japanese-recognition-0001/description/handwritten-japanese-recognition-0001.md](https://github.com/opencv/open_model_zoo/blob/develop/models/intel/handwritten-japanese-recognition-0001/description/handwritten-japanese-recognition-0001.md)) or else.

tools/documentation_updater/documentation_updater.py
IRDonch: You fixed it for the backtick and asterisk replacement, but the same comment also applies to the link flattening.

ci/documentation_updater/documentation_updater.py
```diff
@@ -0,0 +1,200 @@
...
+
+    return models
+
+
+def update_model_descriptions(models, descriptions, mode):
+    update_models = []
+    for name, desc in descriptions.items():
+        model = models.get(name, None)
+        if model is None:
+            logging.warning('For description file {}.md no model found in topologies list'.format(name))
```
IRDonch, line 142:
```suggestion
            logging.warning('For description file {}.md no model found'.format(name))
```

They're called models rather than topologies, and no matter what you call them, there is no list of them in the first place.

ci/documentation_updater/documentation_updater.py
```diff
@@ -0,0 +1,200 @@
...
+    return descriptions
+
+
+def get_models_from_configs(directory):
+    yaml = ruamel.yaml.YAML()
+    yaml.preserve_quotes = True
+    models = {}
+    model_configs = Path(directory).glob('**/model.yml')
+    for model in model_configs:
+        with open(str(model), "r") as file:
```
IRDonch, line 131:
```suggestion
        with model.open("r") as file:
```

ci/documentation_updater/documentation_updater.py
```diff
@@ -0,0 +1,200 @@
...
+            logging.warning('No description found in {} file. '
+                            'Check  compliance with the OMZ Contribution Guide.'.format(file))
+    return descriptions
+
+
+def get_models_from_configs(directory):
+    yaml = ruamel.yaml.YAML()
+    yaml.preserve_quotes = True
+    models = {}
+    model_configs = Path(directory).glob('**/model.yml')
```
IRDonch, line 129:
```suggestion
    model_configs = directory.glob('**/model.yml')
```

Looks like `directory` is already a `Path`.

ci/documentation_updater/documentation_updater.py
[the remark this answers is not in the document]
IRDonch: First, I have to object to your example. If a model has non-ASCII characters in the doc, then those same characters should also be used in the config file. There is nothing preventing this.

But second, even if we presuppose that the description updater should ignore certain files, why should it be an _option_? Surely the set of such files will be the same for every developer, so the list should just be hardcoded in the script, and updated when necessary.

ci/documentation_updater/README.md
```diff
@@ -0,0 +1,42 @@
...
+  -h, --help            show this help message and exit
+  -d MODEL_DIR, --model-dir MODEL_DIR
+                        Path to root directory with models documentation and
+                        configuration files
+  --mode {check,update}
+                        Script work mode: "check" only finds diffs, "update" -
+                        updates values
+  --log-level {CRITICAL,ERROR,WARNING,INFO,DEBUG}
+                        Level of logging
+```
```
IRDonch, line 29: Make sure to update the help output after the recent option changes.

(IMO, there's no need to replicate the help output in the README at all. But if you want to have it here, then keep it in sync.)

ci/documentation_updater/README.md
```diff
@@ -0,0 +1,26 @@
...
+
+To check descriptions of all public models:
+```
+python documentation_updater.py -d <OMZ dir>/models --mode check
+```
+OR
+```
+python documentation_updater.py -d <OMZ dir>/models --mode check
+```
```
IRDonch, line 26: These two alternatives seem to be the same. Also, both the `-d` and `--mode` options are set to their default values and can be omitted.

IRDonch: Well, you deduplicated them, but what about my second suggestion?

eizamaliev: Yes, thanks, I mixed default values up

PR #1030, opened 2020-04-14
Added support for audio classification models

eizamaliev:
Support includes:
- reading wav files
- transforming audio: clipping, resampling
- multi-infer

### DIFF
--- a/tools/accuracy_checker/accuracy_checker/adapters/classification.py
+++ b/tools/accuracy_checker/accuracy_checker/adapters/classification.py
@@ -53,7 +53,10 @@ class ClassificationAdapter(Adapter):
         Returns:
             list of ClassificationPrediction objects
         """
+        multi_infer = frame_meta[-1].get('multi_infer', False) if frame_meta else False
         prediction = self._extract_predictions(raw, frame_meta)[self.output_blob]
+        if multi_infer:
+            prediction = np.mean(prediction, axis=0)
         if len(np.shape(prediction)) == 1:
             prediction = np.expand_dims(prediction, axis=0)
         prediction = np.reshape(prediction, (prediction.shape[0], -1))
@@ -67,3 +70,16 @@ class ClassificationAdapter(Adapter):
             result.append(single_prediction)
 
         return result
+
+    @staticmethod
+    def _extract_predictions(outputs_list, meta):
+        is_multi_infer = meta[-1].get('multi_infer', False) if meta else False
+        if not is_multi_infer:
+            return outputs_list[0] if not isinstance(outputs_list, dict) else outputs_list
+
+        output_map = {}
+        for output_key in outputs_list[0].keys():
+            output_data = np.asarray([output[output_key] for output in outputs_list])
+            output_map[output_key] = output_data
+
+        return output_map
--- a/tools/accuracy_checker/accuracy_checker/data_readers/data_reader.py
+++ b/tools/accuracy_checker/accuracy_checker/data_readers/data_reader.py
@@ -36,6 +36,11 @@ try:
 except ImportError:
     nib = None
 
+try:
+    from scipy.io import wavfile
+except ImportError:
+    wavfile = None
+
 from ..utils import get_path, read_json, zipped_transform, set_image_metadata, contains_all
 from ..dependency import ClassProvider
 from ..config import BaseField, StringField, ConfigValidator, ConfigError, DictField, ListField, BoolField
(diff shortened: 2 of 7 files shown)


tools/accuracy_checker/accuracy_checker/adapters/classification.py
```diff
@@ -67,3 +70,15 @@ def process(self, raw, identifiers=None, frame_meta=None):
             result.append(single_prediction)
 
         return result
+
+    def _extract_predictions(self, outputs_list, is_multi_infer=False):
```
eaidova, line 74: please use common method signature. In some cases, it can be necessary to get some extra info from meta for correct multi infer processing, it is the reason why meta provided to this method.

tools/accuracy_checker/accuracy_checker/adapters/classification.py
```diff
@@ -67,3 +70,15 @@ def process(self, raw, identifiers=None, frame_meta=None):
...
+        if not is_multi_infer:
+            return outputs_list[0] if not isinstance(outputs_list, dict) else outputs_list
+
+        output_keys = list(outputs_list[0].keys())
```
eaidova, line 78: if you use output_keys only as iterable object, you do not need conversion to list, dictionary keys also support iteration as well.

tools/accuracy_checker/accuracy_checker/data_readers/data_reader.py
```diff
@@ -20,6 +20,7 @@
 import re
 import cv2
 import numpy as np
+from scipy.io import wavfile
```
eaidova, line 23: scipy is hard dependency for processing on some plathorms (i mean ARM at least). If it is possible, could you please wrap its import to stay it optional if user do not want to work with your reader (see pillow_imread and tf_imread as example)

tools/accuracy_checker/accuracy_checker/preprocessor/audio_preprocessing.py
```diff
@@ -0,0 +1,102 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+
+import numpy as np
+
+from ..config import ConfigError, BaseField, NumberField, ListField, StringField
```
eaidova, line 19: why do you need so many imports, as I see you use only `NumberField`

tools/accuracy_checker/accuracy_checker/preprocessor/audio_preprocessing.py
```diff
@@ -0,0 +1,102 @@
...
+        return parameters
+
+    def configure(self):
+        self.duration = self.get_value_from_config('duration')
+        self.max_clips = self.get_value_from_config('max_clips')
+
+    def process(self, image, annotation_meta=None):
+        data = image.data
+        try:
+            sample_rate = image.metadata['sample_rate']
```
eaidova, line 87: from my view, this record is more clear:
```
sample_rate = image.metadata.get('sample_rate')
if sample_rate is None:
   ....

File: tools/accuracy_checker/accuracy_checker/data_readers/data_reader.py [hunk 6]
```diff
@@ -405,3 +406,16 @@ def _read_list(self, data_id):
     def reset(self):
         self.subset = range(len(self.data_source))
         self.counter = 0
+
+
+class WavReader(BaseReader):
```
eaidova, line 411: please also add reader to __init__.py

tools/accuracy_checker/accuracy_checker/preprocessor/audio_preprocessing.py
eaidova: please also look on pylint errors

tools/accuracy_checker/accuracy_checker/data_readers/data_reader.py
IRDonch: It might be possible to implement this without depending on SciPy by using the builtin `wave` module.

eizamaliev: Since we already have SciPy in dependencies, I won't reinvent the wheel and continue using `scipy.io.wavfile` instead of `wave`

eaidova: @eizamaliev  it leads to trouble and require invent additional workaround for model validation.
 I want remaind you that this model support requiest have a one smalll yocto arm device as target platform, where no opportunity to build scipy (due to lack of some core libs for it)...

eizamaliev: Done

eaidova: thanks

eaidova: please fix import order and I think, it is ready to merge.
P.S. please update model support in @vladimir-dudnik  fork, if it is necessary

eizamaliev: Import order is fixed, changes don't affect anything outside

PR #1034, opened 2020-04-14
Drop Python 2 support

IRDonch:
It was never really supported in the first place, but now that the OpenVINO toolkit has dropped the Python 2 bindings, it's _definitely_ not supported.

### DIFF
--- a/demos/python_demos/human_pose_estimation_3d_demo/modules/draw.py
+++ b/demos/python_demos/human_pose_estimation_3d_demo/modules/draw.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 """
  Copyright (c) 2019 Intel Corporation
  Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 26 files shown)


demos/python_demos/human_pose_estimation_3d_demo/modules/draw.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
```
Wovchena, line 1: `human_pose_estimation_3d_demo`'s modules are not supposed to be run separately. I suggest removing the line for all its modules.

IRDonch:
> There are some demos without execution permission set or without shebang line. Will you fix them or should I do it separately?

Yeah, I already have a branch for that. I'll submit the PR after this one.

> There are READMEs suggesting to run demos via python command. Can we do something about it? The problem is that for Windows python is usually python3.

Hmm, tricky. I'll see if I can do something about it. Probably not in this PR, though.

demos/python_demos/human_pose_estimation_3d_demo/modules/draw.py
IRDonch: Yeah, I already removed it in my "shebang line fix" branch. I'll submit that change afterwards.

PR #1040, opened 2020-04-16
Monitors fix

Wovchena:


demos/common/monitors/cpu_monitor.cpp
```diff
@@ -52,8 +52,17 @@ class CpuMonitor::PerformanceCounter {
...
+            switch (status) {
+                case ERROR_SUCCESS: break;
+                case PDH_INVALID_ARGUMENT:
+                case PDH_INVALID_DATA:
+                case PDH_INVALID_HANDLE:
+                    throw std::system_error(status, std::system_category(), "PdhGetFormattedCounterValue() failed");
+                default:
+                    // It looks that frequent calls of PdhCollectQueryData() for Processor Time may result in
+                    // PdhGetFormattedCounterValue() returning status which is not equal to any of {ERROR_SUCCESS,
+                    // PDH_INVALID_ARGUMENT, PDH_INVALID_DATA, PDH_INVALID_HANDLE} sometimes.
```
IRDonch, line 64: Could you give an example?

Wovchena: Just run unpatched version. For me it happens very seldom and only in the beginning of executions (probably because presenter waits for a 1 sec before next call). But there is an evidence that it may happen more frequently for others. I will summon you to a corresponding issue.

IRDonch: Okay, can confirm. I got a `PDH_CALC_NEGATIVE_DENOMINATOR` error.

That said, maybe if would be better to ignore specific errors than unknown errors? I would assume that most of the time, errors are still going to be fatal.

Wovchena: Yes, that's better. Thank you.

IRDonch:
Could you retarget this (and other fixes of yours that aren't merged yet) to the `release` branch?

Wovchena:
Retargeted

PR #1041, opened 2020-04-16
GitHub Action to lint Python code

cclauss:
`plugin` is an _undefined name_.

https://github.com/cclauss/open_model_zoo/actions

$ `flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics`
```
./demos/python_demos/face_recognition_demo/ie_module.py:40:29: F821 undefined name 'plugin'
                    .format(plugin.device, ', '.join(not_supported_layers)))
                            ^
1     F821 undefined name 'plugin'
1
```

https://flake8.pycqa.org/en/latest/user/error-codes.html

On the flake8 test selection, this PR does _not_ focus on "_style violations_" (the majority of flake8 error codes that [__psf/black__](https://github.com/psf/black) can autocorrect).  Instead these tests are focus on runtime safety and correctness:
* E9 tests are about Python syntax errors usually raised because flake8 can not build an Abstract Syntax Tree (AST).  Often these issues are a sign of unused code or code that has not been ported to Python 3.  These would be compile-time errors in a compiled language but in a dynamic language like Python they result in the script halting/crashing on the user.
* F63 tests are usually about the confusion between identity and equality in Python.  Use ==/!= to compare str, bytes, and int literals is the classic case.  These are areas where __a == b__ is True but __a is b__ is False (or vice versa).  Python >= 3.8 will raise SyntaxWarnings on these instances.
* F7 tests logic errors and syntax errors in type hints
* F82 tests are almost always _undefined names_ which are usually a sign of a typo, missing imports, or code that has not been ported to Python 3.  These also would be compile-time errors in a compiled language but in Python a __NameError__ is raised which will halt/crash the script on the user.

### DIFF
--- a/.github/workflows/lint_python.yml
+++ b/.github/workflows/lint_python.yml
@@ -1,0 +1,27 @@
+name: lint_python
+on:
+  pull_request:
+  push:
+  #  branches: [master]
+jobs:
+  lint_python:
+    runs-on: ubuntu-latest
+    # strategy:
+    #  matrix:
+    #    os: [ubuntu-latest, macos-latest, windows-latest]
+    #    python-version: [2.7, 3.5, 3.6, 3.7, 3.8]  # , pypy3]
+    steps:
+      - uses: actions/checkout@master
+      - uses: actions/setup-python@master
+      - run: pip install black codespell flake8 isort pytest
+      - run: black --check . || true
+      # - run: black --diff . || true
+      # - if: matrix.python-version >= 3.6
+      #  run: |
+      #    pip install black
+      #    black --check .
+      - run: codespell --quiet-level=2 || true  # --ignore-words-list="" --skip=""
+      - run: flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
+      - run: isort --recursive . || true
+      - run: pip install -r requirements.txt || true
+      - run: pytest . || true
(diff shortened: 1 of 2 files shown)


IRDonch:
Hi.

I think I'd rather not use GitHub Actions, because it interferes with the branching model (the configuration defined on the default branch affects all branches, which makes things confusing). That said, I am heavily considering implementing a style check job on our Jenkins instance, and I can probably add something like this to it, so thanks for the idea.

I don't think I'll merge this, but if you could submit the actual code fix you made as a separate PR, that would be much appreciated.

cclauss:
The branches can be configured on lines 3-5 of .github/workflows/lint_python.yml

IRDonch:
> The branches can be configured on lines 3-5 of .github/workflows/lint_python.yml

Yeah, but that's not the point here. The point is that the configuration file on master applies to every branch, which conflicts with the existing development process: normally, we submit changes to the `develop` branch, but if someone tries to amend the CI configuration in the `develop` branch, it won't work; it would have to be done in the `master` branch.

Ideally, either each branch should contain the CI pipelines that are used when a PR targets that branch, or there should be a dedicated branch that contains just the CI pipelines. From what I understand, neither is currently possible with GitHub Actions.

PR #1042, opened 2020-04-16
human_pose_estimation_3d: improve pose_extractor

Wovchena:


demos/python_demos/human_pose_estimation_3d_demo/pose_extractor/wrapper.cpp
```diff
@@ -79,14 +79,13 @@ PyModuleDef pose_extractor_module = {
...
-    }
     import_array();
     if (PyErr_Occurred()) {
         return nullptr;
     }
-
+    PyObject* module = PyModule_Create(&pose_extractor_module);
+    if (module == nullptr) {
+        return nullptr;
+    }
```
IRDonch, line 89: Looks unnecessary. You can just return `module` in all cases. In fact, you don't even need the `module` variable.

PR #1047, opened 2020-04-20
multi_camera_multi_person_tracking: it was possible to get latency ==…

Wovchena:
… 0.0 resulting in ZeroDivisionError: float division by zero in fps = round(1. / latency, 1)


IRDonch:
It's a good change, but it doesn't fix the problem fundamentally - if the performance clock is imprecise enough or if the program executes quickly enough, then the difference between successive clock values could still be zero. IMO, for a proper fix there should either be a special case for zero, or the denominator should be clamped from below to a non-zero value.

Wovchena:
Clamped

PR #1049, opened 2020-04-21
Redesign the pipeline of object_detection_demo_yolov3_async.py

AnthonyQuantum:


demos/python_demos/object_detection_demo_yolov3_async/README.md
```diff
@@ -30,11 +30,9 @@ The command yields the following usage message:
...
 usage: object_detection_demo_yolov3_async.py [-h] -m MODEL -i INPUT
                                              [-l CPU_EXTENSION] [-d DEVICE]
-                                             [--labels LABELS]
-                                             [-t PROB_THRESHOLD]
-                                             [-iout IOU_THRESHOLD]
-                                             [-ni NUMBER_ITER] [-pc] [-r]
-                                             [--no_show]
+                                             [--labels LABELS] [-t PROB_THRESHOLD]
+                                             [-iout IOU_THRESHOLD] [-r]
+                                             [-nireq]
```
Wovchena, line 35: It lacks some keys

demos/python_demos/object_detection_demo_yolov3_async/object_detection_demo_yolov3_async.py
```diff
@@ -219,154 +321,172 @@ def main():
...
         (round(cap.get(cv2.CAP_PROP_FRAME_WIDTH) / 4), round(cap.get(cv2.CAP_PROP_FRAME_HEIGHT) / 8)))
-    while cap.isOpened():
-        # Here is the first asynchronous point: in the Async mode, we capture frame to populate the NEXT infer request
-        # in the regular mode, we capture frame to the CURRENT infer request
-        if is_async_mode:
-            ret, next_frame = cap.read()
-        else:
+
+    while cap.isOpened() or completed_request_results or active_requests_data:
+
```
Wovchena, line 355:
```suggestion
```

demos/python_demos/object_detection_demo_yolov3_async/object_detection_demo_yolov3_async.py
```diff
@@ -219,154 +321,172 @@ def main():
...
+        elif cap.isOpened():
+            start_time = perf_counter()
             ret, frame = cap.read()
+            if not ret:
+                if args.loop_input:
+                    if number_input_frames != 1:
+                        cap.set(cv2.CAP_PROP_POS_FRAMES, 0)
+                    else:
+                        cap.open(input_stream)
```
Wovchena, line 436: Since `CAP_PROP_POS_FRAMES` doesn't work for an image, I think it is simpler just to reopen `cap` independently on `number_input_frames`

demos/python_demos/object_detection_demo_yolov3_async/object_detection_demo_yolov3_async.py
```diff
@@ -219,154 +321,172 @@ def main():
...
+                if status != StatusCode.OK:
+                    raise Exception("Wait for idle request failed")
+                infer_request_id = exec_nets[mode.current].get_idle_request_id()
+                if infer_request_id < 0:
+                    raise Exception("Invalid request id")
+                for request, frame_id, *_ in active_requests_data.values():
+                    if request == exec_nets[mode.current].requests[infer_request_id]:
+                        add_completion_result(active_requests_data.pop(frame_id))
+                        break
+            request = exec_nets[mode.current].requests[infer_request_id]
```
Wovchena, line 453: This part at should go before frame reading because it increases latency for the frame that was just read.

demos/python_demos/object_detection_demo_yolov3_async/README.md
```diff
@@ -60,16 +62,30 @@ Options:
...
+                        Optional. Iterate over input infinitely
+  -no_show, --no_show   Optional. Don't show output
   -u UTILIZATION_MONITORS, --utilization_monitors UTILIZATION_MONITORS
                         Optional. List of monitors to show initially.
 ```
 
+The number of InferRequests is specified by -nireq flag. An increase of this number usually leads to an increase of performance, since in this case several InferRequests can be processed simultaneously if the device supports parallelization. However, a large number of InferRequests increases the latency because each frame still has to wait before being sent for inference.
+
+For higher FPS, it is recommended that you set -nireq to slightly exceed the -nstreams value, summed across all devices used.
+
```
Wovchena, line 88: You should leave a note about performance degradation.

demos/python_demos/object_detection_demo_yolov3_async/object_detection_demo_yolov3_async.py
```diff
@@ -224,156 +338,162 @@ def main():
...
-        # Here is the first asynchronous point: in the Async mode, we capture frame to populate the NEXT infer request
-        # in the regular mode, we capture frame to the CURRENT infer request
-        if is_async_mode:
-            ret, next_frame = cap.read()
-        else:
-            ret, frame = cap.read()
 
-        if not ret:
-            break
+    while cap.isOpened() \
```
Wovchena, line 372: `callback_exceptions` is ignored by this loop.

demos/python_demos/object_detection_demo_yolov3_async/object_detection_demo_yolov3_async.py
```diff
@@ -224,156 +338,162 @@ def main():
...
-                is_async_mode = not is_async_mode
-                log.info("Switched to {} mode".format("async" if is_async_mode else "sync"))
-            else:
-                presenter.handleKey(key)
-    print(presenter.reportMeans())
-
-    cv2.destroyAllWindows()
+            event.wait()
+    
+    if not callback_exceptions:
```
Wovchena, line 479: Just raise the exception

demos/python_demos/object_detection_demo_yolov3_async/object_detection_demo_yolov3_async.py
```diff
@@ -89,9 +101,29 @@ def __init__(self, param, side):
...
+
+class Mode():
+    def __init__(self, value):
+        self.current = value
+
+    def next(self):
+        if self.current.value + 1 < len(Modes):
+            self.current = Modes(self.current.value + 1)
+        else:
+            self.current = Modes(0)
```
IRDonch, line 118: Looks like this entire class could be replaced with `itertools.cycle(Modes)`.

AnthonyQuantum: @IRDonch,
The iterator returned by itertools.cycle(Modes) does not allow me to get the "current" value, only the "next" one. I need the "current" value most of the times.

Of course, I can create a variable for storing the current value, but in that case I will also need to wrap it into dict or something in order to be able to pass it into callback function by reference. Or I can make the variable global, which is worse.
So, I'm not sure that replacing this class with itertools.cycle will result in better code.

IRDonch: OK, fair enough.

PR #1059, opened 2020-04-22
YOLO python demo: proportional resize

eizamaliev:
*UPDATE:*
The demo will be updated to proportional resize.

### DIFF
--- a/demos/python_demos/object_detection_demo_yolov3_async/object_detection_demo_yolov3_async.py
+++ b/demos/python_demos/object_detection_demo_yolov3_async/object_detection_demo_yolov3_async.py
@@ -133,15 +133,22 @@ def entry_index(side, coord, classes, location, entry):
     return int(side_power_2 * (n * (coord + classes + 1) + entry) + loc)
 
 
-def scale_bbox(x, y, input_height, input_width, class_id, confidence, h_scale, w_scale):
-    xmin = int((x - input_width / 2) * w_scale)
-    ymin = int((y - input_height / 2) * h_scale)
-    xmax = int(xmin + input_width * w_scale)
-    ymax = int(ymin + input_height * h_scale)
+def scale_bbox(x, y, h, w, class_id, confidence, h_scale, w_scale):
+    xmin = int((x - w / 2) * w_scale)
+    ymin = int((y - h / 2) * h_scale)
+    xmax = int(xmin + w * w_scale)
+    ymax = int(ymin + h * h_scale)
     return dict(xmin=xmin, xmax=xmax, ymin=ymin, ymax=ymax, class_id=class_id, confidence=confidence)
 
 
 def parse_yolo_region(blob, resized_image_shape, original_im_shape, params, threshold):
+    def argmax(array):
+        index, max_val = -1, -1
+        for i, val in enumerate(array):
+            if val > max_val:
+                index, max_val = i, val
+        return index
+
     # ------------------------------------------ Validating output parameters ------------------------------------------
     _, _, out_blob_h, out_blob_w = blob.shape
     assert out_blob_w == out_blob_h, "Invalid size of output blob. It sould be in NCHW layout and height should " \
@@ -176,16 +183,20 @@ def parse_yolo_region(blob, resized_image_shape, original_im_shape, params, thre
             except OverflowError:
                 continue
             # Depends on topology we need to normalize sizes by feature maps (up to YOLOv3) or by input shape (YOLOv3)
-            input_width = w_exp * params.anchors[2 * n] / (resized_image_w if params.isYoloV3 else params.side)
-            input_height = h_exp * params.anchors[2 * n + 1] / (resized_image_h if params.isYoloV3 else params.side)
+            w = w_exp * params.anchors[2 * n] / (resized_image_w if params.isYoloV3 else params.side)
+            h = h_exp * params.anchors[2 * n + 1] / (resized_image_h if params.isYoloV3 else params.side)
+            classes = []
             for j in range(params.classes):
                 class_index = entry_index(params.side, params.coords, params.classes, n * side_square + i,
                                           params.coords + 1 + j)
-                confidence = scale * predictions[class_index]
-                if confidence < threshold:
-                    continue
-                objects.append(scale_bbox(x=x, y=y, input_height=input_height, input_width=input_width, class_id=j,
-                                          confidence=confidence, h_scale=orig_im_h, w_scale=orig_im_w))
+                classes.append(predictions[class_index])
+
+            label = argmax(classes)
+            confidence = classes[label]*scale
+            if confidence < threshold:
+                continue
+            objects.append(scale_bbox(x=x, y=y, h=h, w=w, class_id=label, confidence=confidence,
+                                      h_scale=orig_im_h, w_scale=orig_im_w))
     return objects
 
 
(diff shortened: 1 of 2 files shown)


demos/python_demos/object_detection_demo_yolov3_async/object_detection_demo_yolov3_async.py
```diff
@@ -146,14 +153,18 @@ def parse_yolo_region(blob, resized_image_shape, original_im_shape, params, thre
...
                 class_index = entry_index(params.side, params.coords, params.classes, n * side_square + i,
                                           params.coords + 1 + j)
-                confidence = scale * predictions[class_index]
-                if confidence < threshold:
-                    continue
-                objects.append(scale_bbox(x=x, y=y, h=h, w=w, class_id=j, confidence=confidence,
-                                          h_scale=orig_im_h, w_scale=orig_im_w))
+                classes.append(predictions[class_index])
+
+            label = argmax(classes)
```
IRDonch, line 162: Why are you selecting the _greatest_ class index? Is there some significance to it?

demos/python_demos/object_detection_demo_yolov3_async/object_detection_demo_yolov3_async.py
```diff
@@ -310,8 +321,10 @@ def main():
         presenter.drawGraphs(frame)
         for obj in objects:
             # Validation bbox of detected object
-            if obj['xmax'] > origin_im_size[1] or obj['ymax'] > origin_im_size[0] or obj['xmin'] < 0 or obj['ymin'] < 0:
-                continue
+            obj['xmax'] = min(obj['xmax'], origin_im_size[1])
+            obj['ymax'] = min(obj['ymax'], origin_im_size[0])
+            obj['xmin'] = max(obj['xmin'], 0)
+            obj['ymin'] = max(obj['ymin'], 0)
```
IRDonch, line 327: Shouldn't there still be some check as to whether the object overlaps the image at all?

eizamaliev: Box always intersects with image, because center of it **always** lies inside image.

demos/python_demos/object_detection_demo_yolov3_async/object_detection_demo_yolov3_async.py
eizamaliev: I'm not sure what do yo mean. But as you know, for each bbox model predicts objectness score and class probabilities. So, the real score is derived as multiplication of objectness score and highest class probability. Another way, you need only class with highest probability.

IRDonch: Right, I misread the code. Maybe you could rename `classes` to `class_probabilities` to make it clearer?

IRDonch:
I converted the PR into a draft since you added "WIP" to the title. Feel free to convert it back when it's ready.

Wovchena:
You have merge conflict

demos/python_demos/object_detection_demo_yolov3_async/object_detection_demo_yolov3_async.py
```diff
@@ -172,6 +184,24 @@ def intersection_over_union(box_1, box_2):
...
+        return cv2.resize(image, size, interpolation=interpolation)
+
+    iw, ih = image.shape[0:2][::-1]
+    w, h = size
+    scale = min(w/iw, h/ih)
+    nw = int(iw*scale)
+    nh = int(ih*scale)
+    image = cv2.resize(image, (nw, nh), interpolation=interpolation)
+    new_image = np.zeros((size[1], size[0], 3), np.uint8)
+    new_image.fill(128)
```
IRDonch, line 198: Why not `np.full`?

demos/python_demos/object_detection_demo_yolov3_async/object_detection_demo_yolov3_async.py
```diff
@@ -146,14 +154,18 @@ def parse_yolo_region(blob, resized_image_shape, original_im_shape, params, thre
...
                 class_index = entry_index(params.side, params.coords, params.classes, n * side_square + i,
                                           params.coords + 1 + j)
-                confidence = scale * predictions[class_index]
-                if confidence < threshold:
-                    continue
-                objects.append(scale_bbox(x=x, y=y, h=h, w=w, class_id=j, confidence=confidence,
-                                          h_scale=orig_im_h, w_scale=orig_im_w))
+                class_probabilities.append(predictions[class_index])
+
+            label = np.argmax(class_probabilities)
```
IRDonch, line 163: Maybe this should be called `class_id`? It's used to initialize a field with that name, after all.

demos/python_demos/object_detection_demo_yolov3_async/object_detection_demo_yolov3_async.py
```diff
@@ -310,10 +339,12 @@ def main():
...
             # Validation bbox of detected object
-            if obj['xmax'] > origin_im_size[1] or obj['ymax'] > origin_im_size[0] or obj['xmin'] < 0 or obj['ymin'] < 0:
-                continue
+            obj['xmax'] = min(obj['xmax'], origin_im_size[1])
+            obj['ymax'] = min(obj['ymax'], origin_im_size[0])
+            obj['xmin'] = max(obj['xmin'], 0)
+            obj['ymin'] = max(obj['ymin'], 0)
             color = (int(min(obj['class_id'] * 12.5, 255)),
-                     min(obj['class_id'] * 7, 255), min(obj['class_id'] * 5, 255))
+                     int(min(obj['class_id'] * 7, 255)), int(min(obj['class_id'] * 5, 255)))
```
IRDonch, line 347: I can see why these `int`s are here, but it's not obvious at all. IMO, you should instead convert the class ID to `int` _before_ you put it into `obj` (and explain in a comment why you're doing it).

demos/python_demos/object_detection_demo_yolov3_async/object_detection_demo_yolov3_async.py
```diff
@@ -345,6 +376,7 @@ def main():
         start_time = time()
         if not args.no_show:
             cv2.imshow("DetectionResults", frame)
+        cv2.imwrite('out.jpg', frame)
```
IRDonch, line 379: I think you forgot to remove some debugging code.

demos/python_demos/object_detection_demo_yolov3_async/object_detection_demo_yolov3_async.py
```diff
@@ -176,16 +185,20 @@ def parse_yolo_region(blob, resized_image_shape, original_im_shape, params, thre
             except OverflowError:
                 continue
             # Depends on topology we need to normalize sizes by feature maps (up to YOLOv3) or by input shape (YOLOv3)
-            input_width = w_exp * params.anchors[2 * n] / (resized_image_w if params.isYoloV3 else params.side)
-            input_height = h_exp * params.anchors[2 * n + 1] / (resized_image_h if params.isYoloV3 else params.side)
+            w = w_exp * params.anchors[2 * n] / size_normalizer[0]
+            h = h_exp * params.anchors[2 * n + 1] / size_normalizer[1]
```
Wovchena, line 189: Can you rename single char vars. It's hard to search for their usages with Ctrl+f. Here and in `def scale_bbox(x, y, h, w, class_id, confidence, im_h, im_w, is_proportional):`

demos/python_demos/object_detection_demo_yolov3_async/object_detection_demo_yolov3_async.py
IRDonch: I see that you removed the `int` casts. Were they not necessary after all? I thought they were needed to ensure that we pass `int` objects to OpenCV, rather than `np.int32`/`np.float64`?

eizamaliev: I moved `int` cast to parsing function since it should return `class_id` as int.

IRDonch: Ah, I see now, you've used the `item` method to convert the value to a native Python type. Yeah, that's much better, thanks.

Issue #1061, opened 2020-04-23
YoloV3 Custom Feature Scale Detection is Incorrect


jnazaren:
The detection of a feature map's width and height seems to be backwards here: https://github.com/opencv/open_model_zoo/blob/2020.2/tools/accuracy_checker/accuracy_checker/adapters/yolo.py#L277

I believe this should be `cells_y, cells_x = prediction.shape[1:]`, as the output format of the model (at least by default) is NC**HW**. Although it doesn't make a difference with square inputs, this presents a fatal flaw for custom, non-square input sizes.

jnazaren:
Solved by merge request #1062

PR #1062, opened 2020-04-23
Update Yolo Adapter to Fix Feature Scale Detection

jnazaren:
As mentioned in issue #1061


eaidova:
@jnazaren thank you for your fix! I hope,  #1061 also can be cloesd?

jnazaren:
Yes, I've closed it

PR #1066, opened 2020-04-23
Add whiteboard inpainting demo

DmitriySidnev:
This is a whiteboard inpainting demo.
@snosov1


demos/python_demos/whiteboard_inpainting_demo/README.md
```diff
@@ -0,0 +1,76 @@
+# Whiteboard inpainting demo
+
+This demo demonstrates how to disable a person from a whiteboard and show
```
druzhkov-paul, line 3: What do you mean by disabling a person? :astonished:

druzhkov-paul, line 3: IMHO, "demo shows" sounds better.

demos/python_demos/whiteboard_inpainting_demo/README.md
```diff
@@ -0,0 +1,76 @@
...
+text on the whiteboard using OpenVINO<sup>TM</sup>.
+
+## How it works
+
+The demo expects one of the next models in the Intermediate Representation (IR) format:
```
druzhkov-paul, line 8: "one of the following models"

demos/python_demos/whiteboard_inpainting_demo/README.md
```diff
@@ -0,0 +1,76 @@
...
+
+    * Instance segmentation model
+    * Semantic segmentation model
+
+It can be your own models or pre-trained model from OpenVINO Open Model Zoo.
+In the `models.lst` are the list of appropriate models for this demo that can
```
druzhkov-paul, line 14: @snosov1 do we still have technical writers' reviews?

demos/python_demos/whiteboard_inpainting_demo/README.md
```diff
@@ -0,0 +1,76 @@
...
+be obtained via `Model downloader`. Please see more information about
+`Model downloader` [here](../../../tools/downloader/README.md).
+
+As input, the demo application takes:
+
+    * paths to a video file
```
druzhkov-paul, line 20: How many paths is enough? :smile:

demos/python_demos/whiteboard_inpainting_demo/README.md
```diff
@@ -0,0 +1,76 @@
...
+                                     [-ms M_SEMANTIC_SEGMENTATION] [-t THRESHOLD]
+                                     [--output_video OUTPUT_VIDEO] [--no_show] [-d DEVICE]
+                                     [-l CPU_EXTENSION]
+
+Whiteboard inpainting demo
+
+optional arguments:
+  -h, --help            show this help message and exit
+  -i I                  Input sources (index of camera or path to video file)
+  -mi M_INSTANCE_SEGMENTATION, --m_instance_segmentation M_INSTANCE_SEGMENTATION
```
druzhkov-paul, line 48: I believe that other samples use underscores for alike arguments: `-m_i` and `-m_s`.

demos/python_demos/whiteboard_inpainting_demo/README.md
```diff
@@ -0,0 +1,76 @@
...
+
+Example of command:
+
+```
+python whiteboard_inpainting_demo.py \
+    -i path/to/video.avi \
+    -mi path/to/instance-segmentation-security-0050.xml \
+```
+
+## Using
```
druzhkov-paul, line 71: A common template has "How It Works", "Running", "Demo output" and "See Also" sections. Why don't you follow this convention here?

demos/python_demos/whiteboard_inpainting_demo/requirements.txt
```diff
@@ -0,0 +1,2 @@
+opencv-python>=3.4.5.20
```
druzhkov-paul, line 1: OpenVINO includes OpenCV, why do you need it here?

demos/python_demos/whiteboard_inpainting_demo/models.lst
```diff
@@ -0,0 +1,3 @@
+# This file can be used with the --list option of the model downloader.
```
druzhkov-paul, line 1: Why do you need the `__init__.py` file in this directory? Can't leave a comment for this change, 'cause it's empty, so leaving it here.

demos/python_demos/whiteboard_inpainting_demo/utils/capture.py
```diff
@@ -0,0 +1,45 @@
...
+            mode = 'cam'
+        except ValueError:
+            mode = 'video'
+        if mode == 'cam':
+            log.info('Connection to cam#{}'.format(cam_id))
+            self.capture = cv2.VideoCapture(cam_id)
+            self.capture.set(cv2.CAP_PROP_FRAME_WIDTH, 1280)
+            self.capture.set(cv2.CAP_PROP_FRAME_HEIGHT, 720)
+            self.capture.set(cv2.CAP_PROP_FPS, 30)
+            self.capture.set(cv2.CAP_PROP_FOURCC, cv2.VideoWriter_fourcc(*'MJPG'))
```
druzhkov-paul, line 31: What does this fourcc code influence here?

demos/python_demos/whiteboard_inpainting_demo/utils/ie_tools.py
```diff
@@ -0,0 +1,99 @@
...
+        return outputs
+
+    def get_input_shape(self):
+        """Returns an input shape of the wrapped IE model"""
+        return self.inputs_info[self.input_key]
+
+
+def load_ie_model(ie, model_xml, device, plugin_dir, cpu_extension='', num_reqs=1):
+    """Loads a model in the Inference Engine format"""
+    print(model_xml)
```
druzhkov-paul, line 64: Maybe use log here for consistency.

demos/python_demos/whiteboard_inpainting_demo/utils/ie_tools.py
```diff
@@ -0,0 +1,99 @@
...
+    # Read IR
+    log.info("Loading network files:\n\t%s\n\t%s", model_xml, model_bin)
+    net = IENetwork(model=model_xml, weights=model_bin)
+
+    if "CPU" in device:
+        supported_layers = ie.query_network(net, "CPU")
+        not_supported_layers = [l for l in net.layers.keys() if l not in supported_layers]
+        if not_supported_layers:
+            log.error("Following layers are not supported by the plugin for specified device %s:\n %s",
+                      device, ', '.join(not_supported_layers))
```
druzhkov-paul, line 80: List layers' types here would be useful.

demos/python_demos/whiteboard_inpainting_demo/utils/network_wrappers.py
```diff
@@ -0,0 +1,123 @@
...
+ limitations under the License.
+"""
+
+import json
+import logging as log
+
+import cv2
+import numpy as np
+
+from utils.ie_tools import load_ie_model
```
druzhkov-paul, line 20: Use relative import here please.

demos/python_demos/whiteboard_inpainting_demo/utils/misc.py
```diff
@@ -0,0 +1,58 @@
...
+        self.crop_available = False
+
+    def get_points(self, event, x, y, flags, param):
+        if event == cv2.EVENT_LBUTTONDOWN:
+            self.points[0] = (x, y)
+        elif event == cv2.EVENT_LBUTTONUP:
+            self.points[1] = (x, y)
+        elif event == cv2.EVENT_RBUTTONDOWN:
+            self.points = {}
+        self.crop_available = True if len(self.points) == 2 else False
```
druzhkov-paul, line 32: Just out of curiosity, wouldn't `LBUTTONDOWN` -> `RBUTTONDOWN` -> `LBUTTONUP` sequence break this logic?

demos/python_demos/whiteboard_inpainting_demo/utils/misc.py
```diff
@@ -0,0 +1,58 @@
...
+        return osp.isdir(full_path)
+    else:
+        return osp.isfile(full_path)
+
+
+def check_pressed_keys(key):
+    if key == 32:  # Pause
+        while True:
+            key = cv2.waitKey(0)
+            if key == 27 or key == 32 or key == 13:  # enter: resume, space: next frame, esc: exit
```
druzhkov-paul, line 53: Please use named constants instead of this magic values.

demos/python_demos/whiteboard_inpainting_demo/utils/network_wrappers.py
```diff
@@ -0,0 +1,123 @@
...
+class MaskRCNN:
+    """Wrapper class for a network returning masks of objects"""
+
+    def __init__(self, ie, model_path, conf=.6, device='CPU', ext_path='',
+                 max_reqs=100):
+        self.max_reqs = max_reqs
+        self.confidence = conf
+        self.net = load_ie_model(ie, model_path, device, None, ext_path, num_reqs=self.max_reqs)
+
+        required_input_keys = [{'im_info', 'im_data'}, {'im_data', 'im_info'}]
```
druzhkov-paul, line 33: Sets are unordered.

demos/python_demos/whiteboard_inpainting_demo/utils/network_wrappers.py
```diff
@@ -0,0 +1,123 @@
...
+
+    def forward(self, im_data, im_info):
+        if (self.h - im_data.shape[1] < 0) or (self.w - im_data.shape[2] < 0):
+            raise ValueError('Input image should resolution of {}x{} or less, '
+                             'got {}x{}.'.format(self.w, self.h, im_data.shape[2], im_data.shape[1]))
+        im_data = np.pad(im_data, ((0, 0),
+                                   (0, self.h - im_data.shape[1]),
+                                   (0, self.w - im_data.shape[2])),
+                         mode='constant', constant_values=0).reshape(1, self.c, self.h, self.w)
+        im_info = im_info.reshape(1, *im_info.shape)
```
druzhkov-paul, line 63: `im_info = im_info[None, ]`

demos/python_demos/whiteboard_inpainting_demo/utils/network_wrappers.py
```diff
@@ -0,0 +1,123 @@
...
+
+        classes = output['classes']
+        valid_detections_mask = classes > 0
+        classes = classes[valid_detections_mask]
+        boxes = output['boxes'][valid_detections_mask]
+        scores = output['scores'][valid_detections_mask]
+        masks = output['raw_masks'][valid_detections_mask]
+        return boxes, classes, scores, np.full(len(classes), 0, dtype=np.int32), masks
+
+    def get_detections(self, frames, return_cropped_masks=False, only_class_person=True):
```
druzhkov-paul, line 74: `only_class_person` sounds strange.

demos/python_demos/whiteboard_inpainting_demo/utils/network_wrappers.py
```diff
@@ -0,0 +1,123 @@
...
+            scores, classes, boxes, masks = postprocess(scores, classes, boxes, masks,
+                                                        im_h=meta['original_size'][0],
+                                                        im_w=meta['original_size'][1],
+                                                        im_scale_y=meta['processed_size'][0] / meta['original_size'][0],
+                                                        im_scale_x=meta['processed_size'][1] / meta['original_size'][1],
+                                                        full_image_masks=True, encode_masks=False,
+                                                        confidence_threshold=self.confidence)
+            frame_output = []
+            for i in range(len(scores)):
+                if only_class_person and classes[i] != 1:
```
druzhkov-paul, line 92: Please use a named constant instead of just `1`.

demos/python_demos/whiteboard_inpainting_demo/utils/network_wrappers.py
```diff
@@ -0,0 +1,123 @@
...
+    def get_detections(self, frames, only_class_person=True):
+        outputs = []
+        for frame in frames:
+            out_h, out_w = frame.shape[:-1]
+            res = self.net.forward(frame)
+            output = []
+            for batch, data in enumerate(res):
+                data = data.transpose((1, 2, 0)).astype('uint8')
+                data = cv2.resize(data, (out_w, out_h))
+                data = np.where(data > 9, 1, 0)
```
druzhkov-paul, line 119: What is `9`?

demos/python_demos/whiteboard_inpainting_demo/whiteboard_inpainting_demo.py
```diff
@@ -0,0 +1,184 @@
...
+                        help='Path to the instance segmentation model')
+    parser.add_argument('-ms', '--m_semantic_segmentation', type=str, required=False,
+                        help='Path to the semantic segmentation model')
+    parser.add_argument('-t', '--threshold', type=float, default=0.6,
+                        help='Threshold for person instance segmentation model')
+    parser.add_argument('--output_video', type=str, default='', required=False,
+                        help='Optional. Path to output video')
+    parser.add_argument("--no_show", help="Optional. Don't show output", action='store_true')
+
+    parser.add_argument('-d', '--device', type=str, default='CPU')
```
druzhkov-paul, line 74: Help message is absent.

demos/python_demos/whiteboard_inpainting_demo/whiteboard_inpainting_demo.py
```diff
@@ -0,0 +1,184 @@
...
+    args = parser.parse_args()
+
+    if not path_exist(args.i, type='file') and not str.isdigit(args.i):
+        raise ValueError('File not found or wrong camera id: {}'.format(args.i))
+    else:
+        capture = VideoCapture(args.i)
+
+    if (args.m_instance_segmentation and args.m_semantic_segmentation) or \
+       (not args.m_instance_segmentation and not args.m_semantic_segmentation):
+        raise ValueError('Set up exectly one of segmentation models: '\
```
druzhkov-paul, line 87: `exactly`

demos/python_demos/whiteboard_inpainting_demo/whiteboard_inpainting_demo.py
```diff
@@ -0,0 +1,184 @@
...
+                              path to a shared library with the kernels impl.')
+    args = parser.parse_args()
+
+    if not path_exist(args.i, type='file') and not str.isdigit(args.i):
+        raise ValueError('File not found or wrong camera id: {}'.format(args.i))
+    else:
+        capture = VideoCapture(args.i)
+
+    if (args.m_instance_segmentation and args.m_semantic_segmentation) or \
+       (not args.m_instance_segmentation and not args.m_semantic_segmentation):
```
druzhkov-paul, line 86: `bool(args.m_instance_segmentation) != bool(args.m_semantic_segmentation)` should also do the job. As an alternative you can define these arguments as mutually exclusive by means of argparse.

demos/python_demos/whiteboard_inpainting_demo/whiteboard_inpainting_demo.py
```diff
@@ -0,0 +1,184 @@
...
+    cv2.namedWindow(WINNAME)
+    cv2.setMouseCallback(WINNAME, mouse.get_points)
+
+    if args.output_video:
+        fourcc = cv2.VideoWriter_fourcc(*'XVID')
+        output_video = cv2.VideoWriter(args.output_video, fourcc, fps, out_frame_size)
+    else:
+        output_video = None
+
+    log.info("Creating Inference Engine")
```
druzhkov-paul, line 103: "Initializing" or something, not "creating".

demos/python_demos/whiteboard_inpainting_demo/whiteboard_inpainting_demo.py
```diff
@@ -0,0 +1,184 @@
...
+    frame_number = 0
+    key = -1
+
+    while has_frame:
+        start = time.time()
+        if not args.no_show:
+            key = check_pressed_keys(key)
+            if key == 27:
+                break
+            elif key == 105:
```
druzhkov-paul, line 124: What is `105`?

demos/python_demos/whiteboard_inpainting_demo/utils/ie_tools.py
```diff
@@ -0,0 +1,99 @@
...
+import cv2
+import logging as log
+import numpy as np
+import os
+import sys
+
+from openvino.inference_engine import IENetwork, IECore # pylint: disable=import-error,E0611
+
+
+class IEModel:
```
druzhkov-paul, line 23: The wrappers don't look thoroughly designed, making it (sometimes) not straightforward to understand what is going on.

demos/python_demos/whiteboard_inpainting_demo/README.md
DmitriySidnev: The description updated.

DmitriySidnev: Fixed.

demos/python_demos/whiteboard_inpainting_demo/README.md
DmitriySidnev: Fixed.

demos/python_demos/whiteboard_inpainting_demo/README.md
DmitriySidnev: Fixed.

demos/python_demos/whiteboard_inpainting_demo/whiteboard_inpainting_demo.py
DmitriySidnev: Help message added.

demos/python_demos/whiteboard_inpainting_demo/whiteboard_inpainting_demo.py
DmitriySidnev: Fixed.

demos/python_demos/whiteboard_inpainting_demo/whiteboard_inpainting_demo.py
DmitriySidnev: Applied your suggestion.

demos/python_demos/whiteboard_inpainting_demo/whiteboard_inpainting_demo.py
DmitriySidnev: Fixed.

demos/python_demos/whiteboard_inpainting_demo/whiteboard_inpainting_demo.py
DmitriySidnev: It is an 'i' key. Fixed and added a comment.

demos/python_demos/whiteboard_inpainting_demo/utils/ie_tools.py
DmitriySidnev: what does exactly confuse you? `IEModel` is a wrapper for inference engine model with two main methods: `preprocess` and `forward`.

druzhkov-paul: There are too many unnecessary levels of abstraction. I mean that this wrapper is further wrapped with a wrapper (`MaskRCNN`, `SemanticSegmentation`). And `MaskRCNN` class does not use methods of `IEModel`, but directly calls underlying IE API.

demos/python_demos/whiteboard_inpainting_demo/README.md
snosov1: Alina should take a look

demos/python_demos/whiteboard_inpainting_demo/utils/ie_tools.py
DmitriySidnev: Reimplemented with `IEModel` as a base class for both of segmentation models.

demos/python_demos/whiteboard_inpainting_demo/README.md
```diff
@@ -0,0 +1,80 @@
...
+                        MKLDNN (CPU)-targeted custom layers.Absolute path to a
+                        shared library with the kernels impl.
+```
+
+Example of command:
+
+```
+python whiteboard_inpainting_demo.py \
+    -i path/to/video.avi \
+    -mi path/to/instance-segmentation-security-0050.xml \
```
aalborov, line 72: This last slash is extra, isn't it?

demos/python_demos/whiteboard_inpainting_demo/utils/ie_tools.py
```diff
@@ -0,0 +1,84 @@
...
+        _, _, h, w = self.get_input_shape().shape
+        img = np.expand_dims(cv2.resize(img, (w, h)).transpose(2, 0, 1), axis=0)
+        return img
+
+    def forward(self, img):
+        """Performs forward pass of the wrapped IE model"""
+        res = self.net.infer(inputs={self.input_key: self._preprocess(img)})
+        return np.copy(res[self.output_key])
+
+    def get_dedections(self, input):
```
druzhkov-paul, line 39: `get_detections`

demos/python_demos/whiteboard_inpainting_demo/utils/ie_tools.py
```diff
@@ -0,0 +1,84 @@
...
+            supported_layers = ie.query_network(net, "CPU")
+            not_supported_layers = [l for l in net.layers.keys() if l not in supported_layers]
+            if not_supported_layers:
+                log.error("Following layers are not supported by the plugin for specified device %s:\n %s",
+                          device, ', '.join(not_supported_layers))
+                log.error("Please try to specify cpu extensions library path in sample's command line parameters using -l "
+                          "or --cpu_extension command line argument")
+                sys.exit(1)
+
+        assert len(net.inputs.keys()) == 1 or len(net.inputs.keys()) == 2, \
```
druzhkov-paul, line 68: This is strange to have this logic here. Preprocessing and forward pass logic is restricted to support only a single input and single output.

demos/python_demos/whiteboard_inpainting_demo/utils/ie_tools.py
```diff
@@ -0,0 +1,84 @@
...
+        img = np.expand_dims(cv2.resize(img, (w, h)).transpose(2, 0, 1), axis=0)
+        return img
+
+    def forward(self, img):
+        """Performs forward pass of the wrapped IE model"""
+        res = self.net.infer(inputs={self.input_key: self._preprocess(img)})
+        return np.copy(res[self.output_key])
+
+    def get_dedections(self, input):
+        pass
```
druzhkov-paul, line 40: If it has to be implemented in child class, then `NotImplementedError` should be raised.

demos/python_demos/whiteboard_inpainting_demo/utils/network_wrappers.py
```diff
@@ -0,0 +1,117 @@
...
+                else:
+                    mask = masks[i]
+                frame_output.append([bbox, scores[i], mask])
+            outputs.append(frame_output)
+        return outputs
+
+
+class SemanticSegmentation(IEModel):
+    def __init__(self, ie, model_path, conf=.6, device='CPU', ext_path='', max_reqs=1):
+        super().__init__(ie, model_path, conf, device, ext_path, max_reqs)
```
druzhkov-paul, line 102: Why do you need this constructor explicitly here?

demos/python_demos/whiteboard_inpainting_demo/utils/ie_tools.py
druzhkov-paul: Looks better now.

demos/python_demos/whiteboard_inpainting_demo/README.md
DmitriySidnev: Removed.

demos/python_demos/whiteboard_inpainting_demo/utils/ie_tools.py
DmitriySidnev: `MaskRCNN` has 2 inputs and 4 outputs (in one of previous releases the model had 5 outputs), `SemanticSegmentation` has 1 input and 1 output. So, here I just check that input model has supported format.

demos/python_demos/whiteboard_inpainting_demo/utils/ie_tools.py
DmitriySidnev: Fixed.

demos/python_demos/whiteboard_inpainting_demo/utils/network_wrappers.py
DmitriySidnev: Because it does not work without explicit calling of the base class constructor.

demos/python_demos/whiteboard_inpainting_demo/utils/ie_tools.py
druzhkov-paul: This is clear. But it's reasonable to place this checks in a corresponding wrapper, not in a base class.

DmitriySidnev: Fixed.

demos/python_demos/whiteboard_inpainting_demo/utils/capture.py
```diff
@@ -0,0 +1,45 @@
...
+        else:
+            log.info('Open video file {}'.format(source))
+            self.capture = cv2.VideoCapture(source)
```
Wovchena, line 34: You can eliminate the var `mode` by putting everything in `if mode == 'cam'` into the `try` block and putting everything in `else` into `except ValueError` block.

demos/python_demos/whiteboard_inpainting_demo/utils/capture.py
```diff
@@ -0,0 +1,45 @@
...
+
+    def get_frame(self):
+        return self.capture.read()
+
+    def get_source_parameters(self):
+        frame_size = (int(self.capture.get(cv2.CAP_PROP_FRAME_WIDTH)),
+                      int(self.capture.get(cv2.CAP_PROP_FRAME_HEIGHT)))
+        fps = self.capture.get(cv2.CAP_PROP_FPS)
+        return frame_size, fps
+
```
Wovchena, line 45:
```suggestion
```

demos/python_demos/whiteboard_inpainting_demo/utils/ie_tools.py
```diff
@@ -0,0 +1,92 @@
...
+import numpy as np
+import os
+import sys
+
+from openvino.inference_engine import IENetwork, IECore # pylint: disable=import-error,E0611
+
+
+class IEModel:
+    """Class for inference of models in the Inference Engine format"""
+    def __init__(self, ie, model_path, conf=.6, device='CPU', ext_path='', num_reqs=1):
```
Wovchena, line 25: `IEModel` can't utilize more than one  `num_reqs`. This argument should be removed. The same is for `load_ie_model()` and all subclasses.

demos/python_demos/whiteboard_inpainting_demo/utils/ie_tools.py
```diff
@@ -0,0 +1,92 @@
...
+
+    def _preprocess(self, img):
+        _, _, h, w = self.get_input_shape().shape
+        img = np.expand_dims(cv2.resize(img, (w, h)).transpose(2, 0, 1), axis=0)
+        return img
+
+    def forward(self, img):
+        """Performs forward pass of the wrapped IE model"""
+        res = self.net.infer(inputs={self.input_key: self._preprocess(img)})
+        return np.copy(res[self.output_key])
```
Wovchena, line 37:
```suggestion
        return res[self.output_key]
```

demos/python_demos/whiteboard_inpainting_demo/utils/ie_tools.py
```diff
@@ -0,0 +1,92 @@
...
+        """Returns an input shape of the wrapped IE model"""
+        return self.inputs_info[self.input_key]
+
+    def get_allowed_inputs_len(self):
+        return (1, )
+
+    def get_allowed_outputs_len(self):
+        return (1, )
+
+    def load_ie_model(self, ie, model_xml, device, plugin_dir, cpu_extension='', num_reqs=1):
```
Wovchena, line 52: `plugin_dir` is never used.

demos/python_demos/whiteboard_inpainting_demo/utils/ie_tools.py
```diff
@@ -0,0 +1,92 @@
...
+        """Loads a model in the Inference Engine format"""
+        model_bin = os.path.splitext(model_xml)[0] + ".bin"
+        # Plugin initialization for specified device and load extensions library if specified
+        log.info("Initializing Inference Engine plugin for %s ", device)
+
+        if cpu_extension and 'CPU' in device:
+            ie.add_extension(cpu_extension, 'CPU')
+        # Read IR
+        log.info("Loading network files:\n\t%s\n\t%s", model_xml, model_bin)
+        net = IENetwork(model=model_xml, weights=model_bin)
```
Wovchena, line 62: Reading network using constructor is deprecated. Please, use `IECore.read_network()` method instead

demos/python_demos/whiteboard_inpainting_demo/utils/ie_tools.py
```diff
@@ -0,0 +1,92 @@
...
+        out_blob = next(iter(net.outputs))
+        net.batch_size = 1
+
+        # Loading model to the plugin
+        log.info("Loading model to the plugin")
+        self.net = ie.load_network(network=net, device_name=device, num_requests=num_reqs)
+        self.inputs_info = net.inputs
+        self.input_key = input_blob
+        self.output_key = out_blob
+
```
Wovchena, line 92:
```suggestion
```

demos/python_demos/whiteboard_inpainting_demo/utils/misc.py
```diff
@@ -0,0 +1,58 @@
...
+                    datefmt='%Y-%m-%d %H:%M:%S', level=log.DEBUG)
+
+
+def path_exist(name, type='dir', root=None):
+    assert type in ['dir', 'file']
+    full_path = osp.join(root, name) if root is not None else name
+    if type == 'dir':
+        return osp.isdir(full_path)
+    else:
+        return osp.isfile(full_path)
```
Wovchena, line 46: This is just
```suggestion
def path_exist(name):
    return osp.isfile(name)
```
I even doubt you need to introduce a separate function for that.

demos/python_demos/whiteboard_inpainting_demo/utils/misc.py
```diff
@@ -0,0 +1,58 @@
...
+def check_pressed_keys(key):
+    if key == 32:  # Pause
+        while True:
+            key = cv2.waitKey(0)
+            if key == 27 or key == 32 or key == 13:  # enter: resume, space: next frame, esc: exit
+                break
+    else:
+        key = cv2.waitKey(1)
+    return key
+
```
Wovchena, line 58:
```suggestion
```

demos/python_demos/whiteboard_inpainting_demo/utils/network_wrappers.py
```diff
@@ -0,0 +1,123 @@
...
+"""
+
+import json
+import logging as log
+
+import cv2
+import numpy as np
+
+from utils.ie_tools import IEModel
+from .segm_postprocess import postprocess
```
Wovchena, line 21: Why not `from utils.segm_postprocess import postprocess`

demos/python_demos/whiteboard_inpainting_demo/utils/network_wrappers.py
```diff
@@ -0,0 +1,123 @@
...
+        super().__init__(ie, model_path, conf, device, ext_path, max_reqs)
+
+        required_input_keys = [{'im_info', 'im_data'}, {'im_data', 'im_info'}]
+        current_input_keys = set(self.inputs_info.keys())
+        assert current_input_keys in required_input_keys
+        required_output_keys = {'boxes', 'scores', 'classes', 'raw_masks'}
+        assert required_output_keys.issubset(self.net.outputs)
+
+        self.n, self.c, self.h, self.w = self.inputs_info['im_data'].shape
+        assert self.n == 1, 'Only batch 1 is supported.'
```
Wovchena, line 35:
```suggestion
```
You've set batch to one in `__init__` of super.

demos/python_demos/whiteboard_inpainting_demo/utils/network_wrappers.py
```diff
@@ -0,0 +1,123 @@
...
+        scale = min(self.h / image_height, self.w / image_width)
+        processed_image = cv2.resize(frame, None, fx=scale, fy=scale)
+        processed_image = processed_image.astype('float32').transpose(2, 0, 1)
+
+        sample = dict(original_image=frame,
+                      meta=dict(original_size=frame.shape[:2],
+                                processed_size=processed_image.shape[1:3]),
+                      im_data=processed_image,
+                      im_info=np.array([processed_image.shape[1], processed_image.shape[2], 1.0], dtype='float32'))
+        return sample
```
Wovchena, line 54: I suggest returning a tuple instead of dict and unpack it later. Or if you insist on giving names to each of element, you should use namedtuple.
You don't need to give a name to an object you return. Just return whatever you construct.

demos/python_demos/whiteboard_inpainting_demo/utils/network_wrappers.py
```diff
@@ -0,0 +1,123 @@
...
+
+    def forward(self, im_data, im_info):
+        if (self.h - im_data.shape[1] < 0) or (self.w - im_data.shape[2] < 0):
+            raise ValueError('Input image should resolution of {}x{} or less, '
```
Wovchena, line 58:
```suggestion
            raise ValueError('Input image should have the resolution of {}x{} or less, '
```

demos/python_demos/whiteboard_inpainting_demo/utils/network_wrappers.py
```diff
@@ -0,0 +1,123 @@
...
+                             'got {}x{}.'.format(self.w, self.h, im_data.shape[2], im_data.shape[1]))
+        im_data = np.pad(im_data, ((0, 0),
+                                   (0, self.h - im_data.shape[1]),
+                                   (0, self.w - im_data.shape[2])),
```
Wovchena, line 62: Why not place the image in the center?

demos/python_demos/whiteboard_inpainting_demo/utils/network_wrappers.py
```diff
@@ -0,0 +1,123 @@
...
+
+class SemanticSegmentation(IEModel):
+    def __init__(self, ie, model_path, conf=.6, device='CPU', ext_path='', max_reqs=1):
+        super().__init__(ie, model_path, conf, device, ext_path, max_reqs)
+
+    def get_detections(self, frames, only_class_person=True):
+        outputs = []
+        for frame in frames:
+            out_h, out_w = frame.shape[:-1]
+            res = self.forward(frame)
```
Wovchena, line 114: I think there should be resize like in `_preprocess()`

demos/python_demos/whiteboard_inpainting_demo/utils/network_wrappers.py
```diff
@@ -0,0 +1,123 @@
...
+            output = []
+            for batch, data in enumerate(res):
```
Wovchena, line 116:
```suggestion
            for data in res:
```

demos/python_demos/whiteboard_inpainting_demo/whiteboard_inpainting_demo.py
```diff
@@ -0,0 +1,185 @@
...
+            board = output_frame[y0: y1, x0: x1, :]
+            if board.shape[0] > 0 and board.shape[1] > 0:
+                cv2.namedWindow('Board', cv2.WINDOW_KEEPRATIO)
+                cv2.imshow('Board', board)
+            
+        end = time.time()
+        print('\rProcessing frame: {}, fps = {:.3}' \
+            .format(frame_number, 1. / (end - start)), end="")
+        frame_number += 1
+
```
Wovchena, line 178: Please, integrate `monitors.Presenter`: https://github.com/opencv/open_model_zoo/blob/e157f0ac4e98aa1e6ea21c56bc0bdc2d666d4cee/demos/python_demos/colorization_demo/colorization_demo.py#L150

demos/python_demos/whiteboard_inpainting_demo/whiteboard_inpainting_demo.py
```diff
@@ -0,0 +1,185 @@
...
+            .format(frame_number, 1. / (end - start)), end="")
+        frame_number += 1
+    
+    if output_video is not None:
+        output_video.release()
+        
+
+if __name__ == '__main__':
+    main()
+
```
Wovchena, line 185:
```suggestion
```

demos/python_demos/whiteboard_inpainting_demo/whiteboard_inpainting_demo.py
```diff
@@ -0,0 +1,185 @@
+"""
```
Wovchena, line 1: Please, set execution permissions for that file and add shebang line.

demos/python_demos/whiteboard_inpainting_demo/README.md
DmitriySidnev: Fixed.

demos/python_demos/whiteboard_inpainting_demo/README.md
DmitriySidnev: Fixed.

demos/python_demos/whiteboard_inpainting_demo/utils/capture.py
DmitriySidnev: As I know for some webcams using this codec instead default one improves frame rate. May be it is wrong. I have never faced with this problem. But if it is true, why not?

demos/python_demos/whiteboard_inpainting_demo/utils/ie_tools.py
DmitriySidnev: It was just a debug print. Removed.

demos/python_demos/whiteboard_inpainting_demo/utils/misc.py
DmitriySidnev: No 😃 I have checked.

demos/python_demos/whiteboard_inpainting_demo/utils/misc.py
DmitriySidnev: Fixed.

demos/python_demos/whiteboard_inpainting_demo/utils/network_wrappers.py
DmitriySidnev: So, there are two options with both sequences.

demos/python_demos/whiteboard_inpainting_demo/utils/network_wrappers.py
DmitriySidnev: The argument was removed. Logic was reimplemented with list of named classes which will be hided.

demos/python_demos/whiteboard_inpainting_demo/utils/network_wrappers.py
DmitriySidnev: Logic was reimplemented with list of named classes which will be hided using text files with labels according to a model.

demos/python_demos/whiteboard_inpainting_demo/utils/network_wrappers.py
DmitriySidnev: Logic was reimplemented with list of named classes which will be hided using text files with labels according to a model.

demos/python_demos/whiteboard_inpainting_demo/whiteboard_inpainting_demo.py
```diff
@@ -0,0 +1,185 @@
...
+                        help='MKLDNN (CPU)-targeted custom layers.Absolute \
+                              path to a shared library with the kernels impl.')
+    args = parser.parse_args()
+
+    if not path_exist(args.i, type='file') and not str.isdigit(args.i):
+        raise ValueError('File not found or wrong camera id: {}'.format(args.i))
+    else:
+        capture = VideoCapture(args.i)
+
+    print(bool(args.m_instance_segmentation), bool(args.m_semantic_segmentation))
```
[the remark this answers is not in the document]
DmitriySidnev: Already fixed.

demos/python_demos/whiteboard_inpainting_demo/utils/network_wrappers.py
DmitriySidnev: Fixed.

demos/python_demos/whiteboard_inpainting_demo/utils/network_wrappers.py
DmitriySidnev: Resize is used in the `_preprocess` method implemented in the base class.

demos/python_demos/whiteboard_inpainting_demo/README.md
```diff
@@ -0,0 +1,78 @@
...
+```
+python whiteboard_inpainting_demo.py \
+    -i path/to/video.avi \
+    -mi path/to/instance-segmentation-security-0050.xml
+```
+
+## Usage
+
+    * Invert colors on the resulting frame by pressing the `i` key.
+    * Select a part of the frame be shown in a separate window by using your left mouse button.
```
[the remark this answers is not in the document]
DmitriySidnev: Fixed.

demos/python_demos/whiteboard_inpainting_demo/whiteboard_inpainting_demo.py
```diff
@@ -0,0 +1,185 @@
...
+
+from utils.capture import VideoCapture
+from utils.network_wrappers import MaskRCNN, SemanticSegmentation
+from utils.misc import MouseClick, set_log_config, path_exist, check_pressed_keys
+
+set_log_config()
+WINNAME = 'Whiteboard_inpainting_demo'
+
+
+def upsample_mask(detection):
```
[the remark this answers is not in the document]
DmitriySidnev: Renamed.

demos/python_demos/whiteboard_inpainting_demo/whiteboard_inpainting_demo.py
```diff
@@ -0,0 +1,185 @@
...
+    while has_frame:
+        start = time.time()
+        if not args.no_show:
+            key = check_pressed_keys(key)
+            if key == 27:  # 'Esc'
+                break
+            elif key == ord('i'):  # catch pressing of key 'i'
+                black_board = not black_board
+                if output_frame is not None:
+                    output_frame = 255 - output_frame
```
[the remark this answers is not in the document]
DmitriySidnev: Moved.

demos/python_demos/whiteboard_inpainting_demo/whiteboard_inpainting_demo.py
```diff
@@ -0,0 +1,185 @@
...
+
+        has_frame, frame = capture.get_frame()
+        mask = None
+        if frame is not None:
```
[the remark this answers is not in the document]
DmitriySidnev: Yes, you are right. Fixed.

demos/python_demos/whiteboard_inpainting_demo/whiteboard_inpainting_demo.py
```diff
@@ -0,0 +1,185 @@
...
+            mask = np.stack([mask, mask, mask], axis=-1)
+        else:
+            mask = np.zeros(frame.shape, dtype='uint8')
+
+        clear_frame = remove_background(frame, invert_colors=not black_board)
+
+        if output_frame is None:
+            fill_value = 0 if black_board else 255
+            white_black = np.full(frame.shape, fill_value, dtype='uint8')
+            output_frame = np.where(mask, white_black, clear_frame)
```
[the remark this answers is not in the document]
DmitriySidnev: Before the loop there is not an initial value for the `output_frame`.

demos/python_demos/whiteboard_inpainting_demo/utils/network_wrappers.py
DmitriySidnev: This padding is like in instance_segmentation demo.

demos/python_demos/whiteboard_inpainting_demo/whiteboard_inpainting_demo.py
DmitriySidnev: Added.

demos/python_demos/whiteboard_inpainting_demo/utils/network_wrappers.py
DmitriySidnev: Relative paths are used for both of imported modules.

demos/python_demos/whiteboard_inpainting_demo/utils/ie_tools.py
druzhkov-paul: Only the names of unsupported operations are listed here, providing no explicit information of what kind of operations are not supported.

Though, I took a look at other demos and it seems to be the common way to handle unsupported layers. Ok to leave it as is then.

demos/python_demos/whiteboard_inpainting_demo/utils/network_wrappers.py
druzhkov-paul: "Unordered" means that order does not matter. Following this definition, these two options are the same.

This being said, the following code should do the job:
```python
required_input_keys = {'im_info', 'im_data'}
current_input_keys = set(self.net.inputs_info.keys())
assert current_input_keys == required_input_keys
```

demos/python_demos/whiteboard_inpainting_demo/utils/network_wrappers.py
druzhkov-paul: It sounds too strange. Have you actually tried removing these two lines? What kind of error is raised?

demos/python_demos/whiteboard_inpainting_demo/utils/capture.py
DmitriySidnev: Fixed.

demos/python_demos/whiteboard_inpainting_demo/utils/capture.py
DmitriySidnev: Fixed.

demos/python_demos/whiteboard_inpainting_demo/utils/ie_tools.py
DmitriySidnev: Fixed.

demos/python_demos/whiteboard_inpainting_demo/utils/ie_tools.py
DmitriySidnev: Fixed.

demos/python_demos/whiteboard_inpainting_demo/utils/ie_tools.py
DmitriySidnev: Removed.

demos/python_demos/whiteboard_inpainting_demo/utils/ie_tools.py
DmitriySidnev: Fixed.

demos/python_demos/whiteboard_inpainting_demo/utils/ie_tools.py
DmitriySidnev: Fixed.

demos/python_demos/whiteboard_inpainting_demo/utils/misc.py
DmitriySidnev: Fixed.

demos/python_demos/whiteboard_inpainting_demo/utils/misc.py
DmitriySidnev: Fixed.

demos/python_demos/whiteboard_inpainting_demo/utils/network_wrappers.py
DmitriySidnev: Fixed.

demos/python_demos/whiteboard_inpainting_demo/utils/network_wrappers.py
DmitriySidnev: Fixed with tuple.

demos/python_demos/whiteboard_inpainting_demo/utils/network_wrappers.py
DmitriySidnev: Fixed.

demos/python_demos/whiteboard_inpainting_demo/utils/network_wrappers.py
DmitriySidnev: Fixed.

demos/python_demos/whiteboard_inpainting_demo/whiteboard_inpainting_demo.py
DmitriySidnev: Done, but I did not realize how it works.

demos/python_demos/whiteboard_inpainting_demo/whiteboard_inpainting_demo.py
DmitriySidnev: Fixed.

demos/python_demos/whiteboard_inpainting_demo/utils/network_wrappers.py
DmitriySidnev: Oh, yes, my bad, just looked into another class. Fixed.

demos/python_demos/whiteboard_inpainting_demo/whiteboard_inpainting_demo.py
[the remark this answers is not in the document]
Wovchena: I meant the whole block under `if not args.no_show`

demos/python_demos/whiteboard_inpainting_demo/whiteboard_inpainting_demo.py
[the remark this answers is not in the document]
Wovchena: Why do you `continue` instead of `break` if `frame` is `None`? There is no reason trying to read another frame if the current one fails.

demos/python_demos/whiteboard_inpainting_demo/whiteboard_inpainting_demo.py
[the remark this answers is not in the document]
Wovchena: `output_frame` can be initialized with `np.full((int(frame_size[1]), int(frame_size[0]), 3), 255, dtype='uint8')`

demos/python_demos/whiteboard_inpainting_demo/utils/misc.py
Wovchena: Why do you keep the function?

demos/python_demos/whiteboard_inpainting_demo/whiteboard_inpainting_demo.py
Wovchena: Try `presenter = monitors.Presenter(args.utilization_monitors, 20, (out_frame_size[0] // 4, out_frame_size[1] // 16))` instead of current construction.

demos/python_demos/whiteboard_inpainting_demo/utils/network_wrappers.py
DmitriySidnev: Fixed.

demos/python_demos/whiteboard_inpainting_demo/whiteboard_inpainting_demo.py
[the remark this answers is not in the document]
DmitriySidnev: OK, moved.

demos/python_demos/whiteboard_inpainting_demo/whiteboard_inpainting_demo.py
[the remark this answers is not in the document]
DmitriySidnev: Replaced by `break`.

demos/python_demos/whiteboard_inpainting_demo/whiteboard_inpainting_demo.py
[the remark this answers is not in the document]
DmitriySidnev: Updated with your suggestion.

demos/python_demos/whiteboard_inpainting_demo/utils/misc.py
DmitriySidnev: The function removed.

demos/python_demos/whiteboard_inpainting_demo/whiteboard_inpainting_demo.py
DmitriySidnev: Fixed.

demos/python_demos/whiteboard_inpainting_demo/whiteboard_inpainting_demo.py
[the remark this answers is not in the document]
Wovchena: Please replace `while has_frame` with `while True`. `has_frame = True` before the loop is not needed.

demos/python_demos/whiteboard_inpainting_demo/whiteboard_inpainting_demo.py
```diff
@@ -0,0 +1,197 @@
...
+        presenter.drawGraphs(merged_frame)
+        if not args.no_show:
+            cv2.imshow(WINNAME, merged_frame)
+            key = check_pressed_keys(key)
+            presenter.handleKey(key)
+            if key == 27:  # 'Esc'
+                break
+            elif key == ord('i'):  # catch pressing of key 'i'
+                black_board = not black_board
+                output_frame = 255 - output_frame
```
Wovchena, line 172:
```suggestion
            if key == 27:  # 'Esc'
                break
            if key == ord('i'):  # catch pressing of key 'i'
                black_board = not black_board
                output_frame = 255 - output_frame
            else:
                presenter.handleKey(key)
```

DmitriySidnev: Fixed.

demos/python_demos/whiteboard_inpainting_demo/whiteboard_inpainting_demo.py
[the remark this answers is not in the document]
DmitriySidnev: Fixed.

vladimir-dudnik:
@DmitriySidnev do we have a suitable publicly available video which we can refer to in demo description? I think we need this as demo work with quite specific scenario (person in front of whiteboard).

DmitriySidnev:
@vladimir-dudnik, I can only provide link (youtube) to a suitable public video.

demos/python_demos/whiteboard_inpainting_demo/README.md
```diff
@@ -0,0 +1,85 @@
...
+This demo focuses on a whiteboard text overlapped by a person. The demo shows
+how to use the OpenVINO<sup>TM</sup> toolkit to detect and hide a person on a
+video so that all text on a whiteboard is visible.
+
+## How It Works
+
+The demo expects one of the following models in the Intermediate Representation (IR) format:
+
+    * Instance segmentation model
+    * Semantic segmentation model
```
eizamaliev, line 12: Redundant indents. Here and below.

demos/python_demos/whiteboard_inpainting_demo/whiteboard_inpainting_demo.py
```diff
@@ -0,0 +1,198 @@
...
+                        help='Optional. Specify a target device to infer on. CPU, GPU, FPGA, HDDL or MYRIAD is '
+                             'acceptable. The demo will look for a suitable plugin for the device specified')
+    parser.add_argument('-l', '--cpu_extension', type=str, default=None,
+                        help='MKLDNN (CPU)-targeted custom layers.Absolute \
+                              path to a shared library with the kernels impl.')
+    parser.add_argument('-u', '--utilization_monitors', default='', type=str,
+                        help='Optional. List of monitors to show initially')
+    args = parser.parse_args()
+
+    if not osp.isfile(args.i) and not str.isdigit(args.i):
```
IRDonch, line 90: You shouldn't check that the file exists, because OpenCV supports RTSP URIs as inputs, which aren't files. Also, AFAIK, the camera ID needs to actually be an integer for OpenCV to treat it as the camera ID.

The usual approach here it to cast the input to an `int` and if that raises a `ValueError`, keep it as a string.

demos/python_demos/whiteboard_inpainting_demo/whiteboard_inpainting_demo.py
```diff
@@ -0,0 +1,198 @@
...
+    key = -1
+
+    while True:
+        start = time.time()
+        _, frame = capture.get_frame()
+
+        mask = None
+        if frame is not None:
+            detections = segmentation.get_detections([frame])
+            print(frame_size[0])
```
IRDonch, line 140: This looks like debugging output.

demos/python_demos/whiteboard_inpainting_demo/whiteboard_inpainting_demo.py
```diff
@@ -0,0 +1,198 @@
...
+                     max(mouse.points[0][1], mouse.points[1][1])
+            x1, y1 = min(x1, output_frame.shape[1] - 1), min(y1, output_frame.shape[0] - 1)
+            board = output_frame[y0: y1, x0: x1, :]
+            if board.shape[0] > 0 and board.shape[1] > 0:
+                cv2.namedWindow('Board', cv2.WINDOW_KEEPRATIO)
+                cv2.imshow('Board', board)
+            
+        end = time.time()
+        print('\rProcessing frame: {}, fps = {:.3}' \
+            .format(frame_number, 1. / (end - start)), end="")
```
IRDonch, line 188: Since there is no line feed printed at the end, at the end of the loop the caret will still be at the end of the text, so any text printed afterwards will be appended to the line. You should fix it by printing a blank line after the loop.

demos/python_demos/whiteboard_inpainting_demo/README.md
```diff
@@ -0,0 +1,85 @@
+# Whiteboard Inpainting Demo
+
+This demo focuses on a whiteboard text overlapped by a person. The demo shows
+how to use the OpenVINO<sup>TM</sup> toolkit to detect and hide a person on a
+video so that all text on a whiteboard is visible.
```
IRDonch, line 5: It seems to me that the demo is not really whiteboard-specific. Wouldn't it work with any background? If so, then it would be better to rename it to something more generic, e.g. `background_inpainting_demo` and rephrase the description to only use whiteboards as an example.

demos/python_demos/whiteboard_inpainting_demo/README.md
DmitriySidnev: Fixed.

demos/python_demos/whiteboard_inpainting_demo/utils/network_wrappers.py
```diff
@@ -0,0 +1,116 @@
...
+from .ie_tools import IEModel
+from .segm_postprocess import postprocess
+
+
+class MaskRCNN(IEModel):
+    def __init__(self, ie, model_path, labels_file,  conf=.6, device='CPU', ext_path=''):
+        super().__init__(ie, model_path, labels_file, conf, device, ext_path)
+
+        required_input_keys = {'im_info', 'im_data'}
+        current_input_keys = set(self.inputs_info.keys())
```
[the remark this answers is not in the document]
DmitriySidnev: Fixed.

demos/python_demos/whiteboard_inpainting_demo/utils/segm_postprocess.py
```diff
@@ -0,0 +1,102 @@
...
+    if scale is None:
+        assert (im_scale_x is not None) and (im_scale_y is not None)
+        scale = [im_scale_x, im_scale_y, im_scale_x, im_scale_y]
+
+    confidence_filter = scores > confidence_threshold
+    scores = scores[confidence_filter]
+    classes = classes[confidence_filter]
+    boxes = boxes[confidence_filter]
+    if raw_cls_masks is not None:
+        raw_cls_masks = list(segm for segm, is_valid in zip(raw_cls_masks, confidence_filter) if is_valid)
```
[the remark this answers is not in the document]
DmitriySidnev: Fixed.

demos/python_demos/whiteboard_inpainting_demo/whiteboard_inpainting_demo.py
DmitriySidnev: The check removed. Proposed approach is implemented in the `VideoCapture` class.

demos/python_demos/whiteboard_inpainting_demo/whiteboard_inpainting_demo.py
DmitriySidnev: Yes, thank you. Removed.

demos/python_demos/whiteboard_inpainting_demo/whiteboard_inpainting_demo.py
DmitriySidnev: Print of empty line added.

demos/python_demos/whiteboard_inpainting_demo/README.md
DmitriySidnev: Actually we did not find more appropriate name. `Background inpainting` is not suitable for the demo because our target is text on a white or black board behind a person.

IRDonch: > our target is text on a white or black board behind a person.

Well... doesn't that make it the background? If it was something other than a blackboard, wouldn't the demo still work?

DmitriySidnev: It works independently of background's type. We just focused on a text. @snosov1, can we use the proposed name: `background_inpainting_demo`?

IRDonch: I discussed it with @snosov1 and we decided to keep the name as-is.

demos/python_demos/whiteboard_inpainting_demo/README.md
```diff
@@ -0,0 +1,85 @@
...
+  -u UTILIZATION_MONITORS, --utilization_monitors UTILIZATION_MONITORS
+                        Optional. List of monitors to show initially
+```
+
+Example of a command:
+
+```
+python whiteboard_inpainting_demo.py \
+    -i path/to/video.avi \
+    -mi path/to/instance-segmentation-security-0050.xml
```
eizamaliev, line 76:
```suggestion
    -m_i path/to/instance-segmentation-security-0050.xml
```

DmitriySidnev:
@eizamaliev, could you please approve this PR if you do not have any more comments?

demos/python_demos/whiteboard_inpainting_demo/README.md
DmitriySidnev: Fixed.

IRDonch:
Jenkins: test this please

PR #1068, opened 2020-04-23
Add a model quantizer script

IRDonch:
This script is akin to the model converter, but instead of converting, it quantizes full-precision models to INT8 using POT with a simple configuration. For now, only certain models (for which it's known that the resulting model has good accuracy) are marked as quantizable.

(This is a draft PR for now, because I'll be adding documentation later.)

### DIFF
--- a/models/public/alexnet/model.yml
+++ b/models/public/alexnet/model.yml
@@ -46,4 +46,5 @@ model_optimizer_args:
   - --input_model=$dl_dir/alexnet.caffemodel
   - --input_proto=$dl_dir/alexnet.prototxt
 framework: caffe
+quantizable: yes
 license: https://raw.githubusercontent.com/BVLC/caffe/master/models/bvlc_alexnet/readme.md
--- a/tools/downloader/README.md
+++ b/tools/downloader/README.md
@@ -67,16 +67,17 @@ The basic usage is to run the script like this:
 ./downloader.py --all
 ```
 
-This will download all models into a directory tree rooted in the current
-directory. To download into a different directory, use the `-o`/`--output_dir`
-option:
+This will download all models. The `--all` option can be replaced with
+other filter options to download only a subset of models. See the "Shared options"
+section.
+
+By default, the script will download models into a directory tree rooted
+in the current directory. To download into a different directory, use
+the `-o`/`--output_dir` option:
 
 ```sh
 ./downloader.py --all --output_dir my/download/directory
 ```
-
-The `--all` option can be replaced with other filter options to download only
-a subset of models. See the "Shared options" section.
 
 You may use `--precisions` flag to specify comma separated precisions of weights
 to be downloaded.
@@ -221,6 +222,9 @@ This will convert all models into the Inference Engine IR format. Models that
 were originally in that format are ignored. Models in PyTorch and Caffe2 formats will be
 converted in ONNX format first.
 
+The `--all` option can be replaced with other filter options to convert only
+a subset of models. See the "Shared options" section.
+
 The current directory must be the root of a download tree created by the model
 downloader. To specify a different download tree path, use the `-d`/`--download_dir`
 option:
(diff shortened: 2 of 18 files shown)


IRDonch:
CC @vladimir-dudnik

vladimir-dudnik:
Jenkins please retry a build

models/public/alexnet/model.yml
```diff
@@ -46,4 +46,5 @@ model_optimizer_args:
   - --input_model=$dl_dir/alexnet.caffemodel
   - --input_proto=$dl_dir/alexnet.prototxt
 framework: caffe
+quantizable: yes
```
eizamaliev, line 49: @IRDonch, how many models should be `quantizable`? Could we set `yes` as default?

tools/downloader/README.md
```diff
@@ -11,6 +11,9 @@ based on configuration files in the models' directories.
 * `converter.py` (model converter) converts the models that are not in the
   Inference Engine IR format into that format using Model Optimizer.
 
+* `quantizer.py` (model quantizer) quantizes full-precision models in the IR
+  format into low-precision versions using Post-Training Optimization Toolkit.
```
eizamaliev, line 15: What do you think about adding link to toolkit?

tools/downloader/README.md
```diff
@@ -286,15 +290,109 @@ executed commands, or "auto", in which case the number of CPUs in the system is
...
+section for information on the expected contents of that directory.
+
+The basic usage is to run the script like this:
+
+```sh
+./quantizer.py --all --dataset_dir <DATASET_DIR>
+```
+
+This will quantize all models for which quantization is supported. Other models
+are ignored.
```
eizamaliev, line 317: Question: where (how) can I find list of supported model?

demos/python_demos/3d_segmentation_demo/3d_segmentation_demo.py
```diff
@@ -28,7 +28,7 @@
 from scipy.ndimage import interpolation
 from sys import stdout
 
-from openvino.inference_engine import IENetwork, IECore
+from openvino.inference_engine import IECore
```
eizamaliev, line 31: How do these changes in demo relate to quantizer tool?

IRDonch: I think I opened the PR against `develop` by accident. It's actually supposed to go into `release`. Fixed now.

tools/downloader/README.md
IRDonch: How would I link to it?

eizamaliev: For example, link to documentation. I don't insist on it, it is just a suggestion.

tools/downloader/README.md
IRDonch: You could use `--all --dry_run` to see which models will get quantized. I'm considering adding a field to the info dumper's output to indicate which precisions each model can be quantized to, but I haven't done so yet, because it seems of secondary importance and there's not much time before the release.

models/public/alexnet/model.yml
IRDonch: I got the current list from @vladimir-dudnik. We may extend it later as we determine that they still have acceptable quality after quantizing.

>  Could we set yes as default?

No, because we can't trust that an arbitrary model will still have good quality after quantizing (or that it will quantize at all).

tools/downloader/README.md
IRDonch: Hmm... it wouldn't be a bad idea, but the problem is that URL for the the POT documentation is currently "https://docs.openvinotoolkit.org/latest/_README.html", which seems rather unstable to me. I don't want to link to something like that might end up pointing to a completely different page later.

models/public/alexnet/model.yml
eizamaliev: Ok, thanks!

tools/downloader/README.md
eizamaliev: Ok, I got it.

tools/downloader/README.md
eizamaliev: Should we add something to documentation?
E.g.:
| Metric            | Value         |
|-------------------|---------------|
| Type              | Detection     |
| GFLOPs            | 5.424         |
| MParams           | 11.229        |
| **Quantazible** | **Yes** |
| Source framework  | TensorFlow.js\*  |

@vladimir-dudnik what do you think?

vladimir-dudnik: I'd prefer to avoid manual maintenance of this info in different places.

vladimir-dudnik: I'd prefer to avoid manual maintenance of this info in different places

PR #1074, opened 2020-04-27
Replace deprecated buffer() and cbuffer() from ie_blob with rwmap() and rmap()

AnthonyQuantum:


demos/common/samples/ocv_common.hpp
```diff
@@ -27,7 +27,7 @@ void matU8ToBlob(const cv::Mat& orig_image, InferenceEngine::Blob::Ptr& blob, in
     if (static_cast<size_t>(orig_image.channels()) != channels) {
         THROW_IE_EXCEPTION << "The number of channels for net input and image must match";
     }
-    T* blob_data = blob->buffer().as<T*>();
+    T* blob_data = InferenceEngine::as<InferenceEngine::MemoryBlob>(blob)->rwmap().as<T*>();
```
IRDonch, line 30: Pretty sure you can't do that. The pointer returned from `as` will become invalid as soon as the object returned by `rwmap` is destroyed.

AnthonyQuantum: @IRDonch,

Can I fix it like this? Would you approve?
Ofc I'll fix other demos too.
```
InferenceEngine::LockedMemory<void> mappedBlob = InferenceEngine::as<InferenceEngine::MemoryBlob>(blob)->rwmap();
T* blob_data = mappedBlob.as<T*>();
```

IRDonch: Sure, just watch the object lifetimes. You need to make sure that all blobs are unmapped before any inference happens.

I would also suggest using `auto` for the `LockedMemory` object, otherwise the code gets really cumbersome.

Wovchena: I disagree about `auto`. `LockedMemory` makes it clearer what is returned by `rmap()`. I think `auto` should be used when it is obvious what the actual type is. In this case `auto` makes a reader find the declaration of `rmap()`. While demos don't focus on on IE API, I believe it still will be easier to read with `LockedMemory`.

IRDonch: OK. Doesn't matter that much to me.

demos/common/samples/ocv_common.hpp
```diff
@@ -27,7 +27,8 @@ void matU8ToBlob(const cv::Mat& orig_image, InferenceEngine::Blob::Ptr& blob, in
     if (static_cast<size_t>(orig_image.channels()) != channels) {
         THROW_IE_EXCEPTION << "The number of channels for net input and image must match";
     }
-    T* blob_data = blob->buffer().as<T*>();
+    InferenceEngine::LockedMemory<void> blobMapped = InferenceEngine::as<InferenceEngine::MemoryBlob>(blob)->rwmap();
```
IRDonch, line 30: Doesn't this just need `wmap`?

demos/gaze_estimation_demo/src/ie_wrapper.cpp
```diff
@@ -84,8 +84,8 @@ void IEWrapper::setInputBlob(const std::string& blobName,
     if (dimsProduct != data.size()) {
         throw std::runtime_error("Input data does not match size of the blob");
     }
-    auto inputBlob = request.GetBlob(blobName);
-    auto buffer = inputBlob->buffer().as<InferenceEngine::PrecisionTrait<InferenceEngine::Precision::FP32>::value_type *>();
+    LockedMemory<void> blobMapped = as<MemoryBlob>(request.GetBlob(blobName))->rwmap();
+    auto buffer = blobMapped.as<PrecisionTrait<Precision::FP32>::value_type *>();
```
IRDonch, line 88: Might as well change the type to `float`...

demos/human_pose_estimation_demo/src/human_pose_estimator.cpp
```diff
@@ -120,15 +120,19 @@ void HumanPoseEstimator::reshape(const cv::Mat& image){
 
 void HumanPoseEstimator::frameToBlobCurr(const cv::Mat& image) {
     CV_Assert(image.type() == CV_8UC3);
-    InferenceEngine::Blob::Ptr input = requestCurr->GetBlob(network.getInputsInfo().begin()->first);
-    auto buffer = input->buffer().as<InferenceEngine::PrecisionTrait<InferenceEngine::Precision::U8>::value_type *>();
+    InferenceEngine::LockedMemory<void> requestCurrBlobMapped = InferenceEngine::as<
+        InferenceEngine::MemoryBlob>(requestCurr->GetBlob(network.getInputsInfo().begin()->first))->rwmap();
```
IRDonch, line 124: This is quite messy. I think you should keep the `input` variable.

demos/interactive_face_detection_demo/detectors.cpp
```diff
@@ -188,8 +188,10 @@ void FaceDetection::fetchResults() {
     results.clear();
     if (resultsFetched) return;
     resultsFetched = true;
-    const float *detections = request->GetBlob(output)->buffer().as<float *>();
-    const int32_t *labels = !labels_output.empty() ? request->GetBlob(labels_output)->buffer().as<int32_t *>() : nullptr;
+    LockedMemory<void> outputMapped = as<MemoryBlob>(request->GetBlob(output))->rwmap();
+    const float *detections = outputMapped.as<float *>();
+    const int32_t *labels = !labels_output.empty()
+                            ? as<MemoryBlob>(request->GetBlob(labels_output))->rwmap().as<int32_t *>() : nullptr;
```
IRDonch, line 194: This still has the dangling pointer problem.

demos/smart_classroom_demo/src/action_detector.cpp
```diff
@@ -131,23 +131,32 @@ DetectedActions ActionDetection::fetchResults() {
...
     const auto det_conf_blob_name = new_network_ ? config_.new_det_conf_blob_name : config_.old_det_conf_blob_name;
 
-    const cv::Mat priorbox_out =
-        new_network_
-          ? cv::Mat()
-          : cv::Mat(ieSizeToVector(request->GetBlob(config_.old_priorbox_blob_name)->getTensorDesc().getDims()),
-                    CV_32F, request->GetBlob(config_.old_priorbox_blob_name)->buffer());
+    cv::Mat priorboxMat;
+    if (new_network_) {
+        priorboxMat = cv::Mat();
```
IRDonch, line 136: This is a useless assignment. `priorboxMat` is already default-initialized.

demos/smart_classroom_demo/src/action_detector.cpp
```diff
@@ -131,23 +131,32 @@ DetectedActions ActionDetection::fetchResults() {
...
+    } else {
+        LockedMemory<void> priorboxOutBlobMapped = as<MemoryBlob>(request->
+                                                                  GetBlob(config_.old_priorbox_blob_name))->rwmap();
+        priorboxMat = cv::Mat(ieSizeToVector(request->
+                              GetBlob(config_.old_priorbox_blob_name)->getTensorDesc().getDims()), CV_32F,
+                              priorboxOutBlobMapped);
```
IRDonch, line 142: I don't think this is correct. `priorboxMat` is initialized with a pointer to the blob data, which becomes invalid once the `else` block ends.

demos/smart_classroom_demo/src/action_detector.cpp
```diff
@@ -131,23 +131,32 @@ DetectedActions ActionDetection::fetchResults() {
...
+    }
+    const cv::Mat priorbox_out = priorboxMat;
```
IRDonch, line 144: Is there a reason to duplicate the `Mat` object?

demos/text_detection_demo/src/cnn.cpp
```diff
@@ -68,8 +68,9 @@ void Cnn::Init(const std::string &model_path, Core & ie, const std::string & dev
     // --------------------------- Preparing input -------------------------------------------------------
 
     /* Resize manually and copy data from the image to the input blob */
-    Blob::Ptr input = infer_request_.GetBlob(input_name);
-    input_data_ = input->buffer().as<PrecisionTrait<Precision::FP32>::value_type *>();
+    InferenceEngine::LockedMemory<void> inputMapped = InferenceEngine::as<
+        InferenceEngine::MemoryBlob>(infer_request_.GetBlob(input_name))->rwmap();
+    input_data_ = inputMapped.as<InferenceEngine::PrecisionTrait<InferenceEngine::Precision::FP32>::value_type *>();
```
IRDonch, line 73: This produces a dangling pointer after `Init` returns.

demos/gaze_estimation_demo/src/ie_wrapper.cpp
AnthonyQuantum: @IRDonch 

I agree that this change will make the code less complicated, but shouldn't we use PrecisionTrait<Precision::FP32>::value_type everywhere instead of direct type names?

What if we change the type behind "FP32" sometime in the future? This is unlikely, but still possible.

Wovchena: It's unlikely FP32 is going to change. IE just needs that type for functions like `InputInfo::setPrecision()`.

demos/smart_classroom_demo/src/action_detector.cpp
```diff
@@ -131,26 +131,34 @@ DetectedActions ActionDetection::fetchResults() {
...
-                                CV_32F, request->GetBlob(det_conf_blob_name)->buffer());
+                                CV_32F, detConfBlobMapped.as<float*>());
 
     std::vector<cv::Mat> add_conf_out;
     for (int glob_anchor_id = 0; glob_anchor_id < num_glob_anchors_; ++glob_anchor_id) {
         const auto& blob_name = glob_anchor_names_[glob_anchor_id];
+        LockedMemory<const void> blobMapped = as<MemoryBlob>(request->GetBlob(blob_name))->rmap();
         add_conf_out.emplace_back(ieSizeToVector(request->GetBlob(blob_name)->getTensorDesc().getDims()),
-                                  CV_32F, request->GetBlob(blob_name)->buffer());
+                                  CV_32F, blobMapped.as<float*>());
```
IRDonch, line 147: This also creates a dangling pointer once `blobMapped` goes out of scope.

demos/text_detection_demo/include/cnn.hpp
```diff
@@ -14,7 +14,7 @@ using namespace InferenceEngine;
 
 class Cnn {
   public:
-    Cnn():is_initialized_(false), channels_(0), input_data_(nullptr), time_elapsed_(0), ncalls_(0) {}
+    Cnn():is_initialized_(false), channels_(0), input_name_(""), input_data_(nullptr), time_elapsed_(0), ncalls_(0) {}
```
IRDonch, line 17: This initializer is unnecessary; `input_name_` would get initialized to an empty string anyway.

demos/text_detection_demo/src/cnn.cpp
```diff
@@ -65,18 +65,17 @@ void Cnn::Init(const std::string &model_path, Core & ie, const std::string & dev
...
     is_initialized_ = true;
 }
 
 InferenceEngine::BlobMap Cnn::Infer(const cv::Mat &frame) {
     std::chrono::steady_clock::time_point begin = std::chrono::steady_clock::now();
 
+    /* Resize manually and copy data from the image to the input blob */
+    InferenceEngine::LockedMemory<void> inputMapped =
+        InferenceEngine::as<InferenceEngine::MemoryBlob>(infer_request_.GetBlob(input_name_))->wmap();
+    input_data_ = inputMapped.as<float *>();
```
IRDonch, line 77: Seems like `input_data_` could now be turned into a local variable.

demos/security_barrier_camera_demo/net_wrappers.hpp
```diff
@@ -271,7 +277,9 @@ class Lpr {
             InferenceEngine::Blob::Ptr seqBlob = inferRequest.GetBlob(LprInputSeqName);
             // second input is sequence, which is some relic from the training
             // it should have the leading 0.0f and rest 1.0f
-            float* blob_data = seqBlob->buffer().as<float*>();
+            InferenceEngine::LockedMemory<void> seqBlobMapped =
+                InferenceEngine::as<InferenceEngine::MemoryBlob>(seqBlob)->rwmap();
```
IRDonch, line 281: `wmap`, maybe?

demos/smart_classroom_demo/src/action_detector.cpp
```diff
@@ -131,26 +131,34 @@ DetectedActions ActionDetection::fetchResults() {
     const auto loc_blob_name = new_network_ ? config_.new_loc_blob_name : config_.old_loc_blob_name;
     const auto det_conf_blob_name = new_network_ ? config_.new_det_conf_blob_name : config_.old_det_conf_blob_name;
 
-    const cv::Mat priorbox_out =
-        new_network_
-          ? cv::Mat()
-          : cv::Mat(ieSizeToVector(request->GetBlob(config_.old_priorbox_blob_name)->getTensorDesc().getDims()),
-                    CV_32F, request->GetBlob(config_.old_priorbox_blob_name)->buffer());
-
+    LockedMemory<void> locBlobMapped = as<MemoryBlob>(request->GetBlob(loc_blob_name))->rwmap();
```
IRDonch, line 134: `rmap`?

demos/smart_classroom_demo/src/action_detector.cpp
```diff
@@ -131,26 +131,34 @@ DetectedActions ActionDetection::fetchResults() {
...
 
     /** Parse detections **/
+    if (new_network_) {
+        const cv::Mat priorbox_out;
+        return GetDetections(loc_out, main_conf_out, priorbox_out, add_conf_out,
+                             cv::Size(static_cast<int>(width_), static_cast<int>(height_)));
+    }
+
+    LockedMemory<void> priorboxOutBlobMapped =
+        as<MemoryBlob>(request->GetBlob(config_.old_priorbox_blob_name))->rwmap();
```
IRDonch, line 158: `rmap`?

demos/smart_classroom_demo/src/action_detector.cpp
AnthonyQuantum: I've fixed the problem, but I'm still a bit unsure whether I should pass LockedMemory objects to GetDetections function or even make them global variables. 

loc_out, main_conf_out, priorbox_out, add_conf_out are used in GetDetections, but this is outside the scope in which LockedMemory objects were created.

demos/text_detection_demo/src/cnn.cpp
IRDonch: The trailing underscore at the end of the name should be removed, since it's not a member variable anymore.

demos/smart_classroom_demo/src/action_detector.cpp
IRDonch: > loc_out, main_conf_out, priorbox_out, add_conf_out are used in GetDetections, but this is outside the scope in which LockedMemory objects were created.

Technically true, but irrelevant. The problem is not scope per se, but object lifetime. To take `locBlobMapped` as an example, it's destroyed at the end of the function, and since `GetDetections`  completes _before_ the end of the function, it's okay for it to access `loc_out`. 

When I said "once `blobMapped` goes out of scope", it was a shorthand for "once  `blobMapped` is destroyed at the end of the block".

demos/smart_classroom_demo/src/action_detector.cpp
```diff
@@ -131,26 +131,35 @@ DetectedActions ActionDetection::fetchResults() {
...
-                          CV_32F, request->GetBlob(loc_blob_name)->buffer());
+                          CV_32F, locBlobMapped.as<float*>());
 
+    LockedMemory<const void> detConfBlobMapped = as<MemoryBlob>(request->GetBlob(det_conf_blob_name))->rmap();
     const cv::Mat main_conf_out(ieSizeToVector(request->GetBlob(det_conf_blob_name)->getTensorDesc().getDims()),
-                                CV_32F, request->GetBlob(det_conf_blob_name)->buffer());
+                                CV_32F, detConfBlobMapped.as<float*>());
 
     std::vector<cv::Mat> add_conf_out;
+    std::vector<LockedMemory<const void>> blobsMapped;
```
IRDonch, line 143: I would suggest to swap these lines. It makes no practical difference, but it's better style-wise, because:

1. you add elements to `blobsMapped` before `add_conf_out`, and
1. since objects are destroyed in reverse order, there would no longer be a moment at the end of the function when the `Mat`s are alive but the `LockedMemory`s they are referring to are not.

demos/human_pose_estimation_demo/src/human_pose_estimator.cpp
```diff
@@ -121,14 +121,18 @@ void HumanPoseEstimator::reshape(const cv::Mat& image){
 void HumanPoseEstimator::frameToBlobCurr(const cv::Mat& image) {
     CV_Assert(image.type() == CV_8UC3);
     InferenceEngine::Blob::Ptr input = requestCurr->GetBlob(network.getInputsInfo().begin()->first);
-    auto buffer = input->buffer().as<InferenceEngine::PrecisionTrait<InferenceEngine::Precision::U8>::value_type *>();
+    InferenceEngine::LockedMemory<const void> requestCurrBlobMapped =
+        InferenceEngine::as<InferenceEngine::MemoryBlob>(input)->rmap();
```
Wovchena, line 125:
```suggestion
    InferenceEngine::LockedMemory<void> requestCurrBlobMapped =
        InferenceEngine::as<InferenceEngine::MemoryBlob>(input)->wmap();
```
The same for `void HumanPoseEstimator::frameToBlobNext(const cv::Mat& image)`.

PR #1088, opened 2020-04-30
Action recognition demo update

druzhkov-paul:
Summary of the changes:

- [x] Decoder model is now optional. If it's not specified, then encoder's outputs within a time window are averaged.
- [x] Label smoothing has been made optional and configurable. Default behavior is still the same as before, though.
- [x] Length of sequence that decoder takes as input has also been made configurable.

### DIFF
--- a/demos/python_demos/action_recognition/action_recognition.py
+++ b/demos/python_demos/action_recognition/action_recognition.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python3
 """
- Copyright (c) 2019 Intel Corporation
+ Copyright (c) 2020 Intel Corporation
 
  Licensed under the Apache License, Version 2.0 (the "License");
  you may not use this file except in compliance with the License.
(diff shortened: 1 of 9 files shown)


druzhkov-paul:
@snosov1

druzhkov-paul:
@Wovchena Yes, this is the reason.

demos/python_demos/action_recognition/action_recognition_demo/models.py
```diff
@@ -115,6 +111,28 @@ def wait_request(self, req_id):
...
+        self.requests[req_id] = self._average(model_input)
+    
+    def infer(self, model_input):
+        return self._average(model_input)
+
+    def wait_request(self, req_id):
+        assert req_id in self.requests
+        output = self.requests[req_id]
+        del self.requests[req_id]
+        return output
```
IRDonch, line 133:
```suggestion
        return self.requests.pop(req_id)
```

demos/python_demos/action_recognition/action_recognition_demo/models.py
```diff
@@ -115,6 +111,28 @@ def wait_request(self, req_id):
         return self.exec_net.requests[req_id].outputs[self.output_name]
 
 
+class DummyDecoder:
+    def __init__(self, *args, num_requests=1, **kwargs):
```
IRDonch, line 115: `args` and `kwargs` are unused.

demos/python_demos/action_recognition/action_recognition_demo/steps.py
```diff
@@ -27,11 +27,11 @@
 from .queue import Signal
 
 
-def run_pipeline(video, encoder, decoder, render_fn, fps=30):
+def run_pipeline(video, encoder, decoder, render_fn, decoder_seq_size=16, fps=30):
     pipeline = AsyncPipeline()
     pipeline.add_step("Data", DataStep(video), parallel=False)
     pipeline.add_step("Encoder", EncoderStep(encoder), parallel=False)
-    pipeline.add_step("Decoder", DecoderStep(decoder), parallel=False)
+    pipeline.add_step("Decoder", DecoderStep(decoder, sequence_size=decoder_seq_size), parallel=False)
```
IRDonch, line 34: I can understand the sequence size being configurable when you're using the dummy decoder, but does it make sense to override it when using a real decoder? IMO, the code should only allow the `--seq` option to be set when `-m_de` is unspecified, and otherwise derive the sequence size from the decoder's input shape.

demos/python_demos/action_recognition/action_recognition_demo/models.py
druzhkov-paul: :+1: Done.

demos/python_demos/action_recognition/action_recognition_demo/models.py
druzhkov-paul: Removed those.

demos/python_demos/action_recognition/action_recognition_demo/steps.py
druzhkov-paul: Implemented.

demos/python_demos/action_recognition/action_recognition.py
```diff
@@ -88,27 +91,37 @@ def main():
...
     encoder = IEModel(encoder_xml, encoder_bin, ie, encoder_target_device,
                       num_requests=(3 if args.device == 'MYRIAD' else 1))
-    decoder = IEModel(decoder_xml, decoder_bin, ie, decoder_target_device, num_requests=2)
-    video_demo(encoder, decoder, videos, args.no_show, args.utilization_monitors, args.fps, labels)
+
+    if args.m_decoder is not None:
+        decoder_xml = args.m_decoder
+        decoder_bin = args.m_decoder.replace('.xml', '.bin')
+        decoder = IEModel(decoder_xml, decoder_bin, ie, decoder_target_device, num_requests=2)
+        decoder_seq_size = decoder.input_size[1]
```
IRDonch, line 115: You should make sure that the `--seq` option was not specified, so that the user isn't led to believe that it does something when it does not.

druzhkov-paul: These arguments are in a mutually exclusive group. Isn't it enough?

IRDonch: Oh, my bad. Didn't realize you did that.

PR #1091, opened 2020-05-06
AC: use input dtype for onnxruntime launcher

eaidova:
resolves problem with BERT processing  (with int64 input precision) via ONNX Runtime launcher


eaidova:
Jenkins please retry a build

PR #1094, opened 2020-05-07
AC: Yolo adapter refactoring

eizamaliev:


tools/accuracy_checker/accuracy_checker/adapters/yolo.py
```diff
@@ -1,13 +1,40 @@
...
+from ..config import BoolField, NumberField, StringField, ConfigValidator, ListField, ConfigError
 from ..representation import DetectionPrediction
 from ..topology_types import YoloV1Tiny, YoloV2, YoloV2Tiny, YoloV3, YoloV3Tiny
 from ..utils import get_or_parse_value
 
+DetectionBox = namedtuple('DetectionBox', ["x", "y", "w", "h", "confidence", "probabilities"])
+
+
+class YoloOutputProcessor:
+    def __init__(self, coord_correct=lambda x: x, size_correct=lambda x: np.exp(x), conf_correct=lambda x: x,
```
eaidova, line 15: I agree with pylint that lambda transition x->x  by default may be excess, probably we can stay None and perform correction only if provided

tools/accuracy_checker/accuracy_checker/adapters/README.md
```diff
@@ -44,6 +46,8 @@ AccuracyChecker supports following set of adapters:
   * `threshold` - minimal objectness score value for valid detections (default 0.001).
   * `input_width` and `input_height` - network input width and height correspondingly (default 416).
   * `outputs` - the list of output layers names (optional), if specified there should be exactly 3 output layers provided.
```
eaidova, line 48: cells missed for yolo_v3 adapter

tools/accuracy_checker/accuracy_checker/adapters/yolo.py
```diff
@@ -235,7 +286,18 @@ def parameters(cls):
                 description="The list of output layers names (optional),"
                             " if specified there should be exactly 3 output layers provided."
             ),
-            'anchor_masks': ListField(optional=True, description='per layer used anchors mask')
+            'anchor_masks': ListField(optional=True, description='per layer used anchors mask'),
+            'cells': ListField(
+                optional=True, default=[13, 26, 52],
```
eaidova, line 291: please note that usage this values as default will require update at least config for mobilefacedet-v1-mxnet since it has different cell size. Also not sure about model yolo-v3-spp (with 608x608 input resolution)

eizamaliev: I solved this problem, may be little bit messy, but should work for both models, if they keep 4d shape. Also original config for yolo-v3-tf should work with flat output.

eaidova: what about check `mobilefacedet-v1-mxnet`? This model also use this adapter and have different from standart cells

tools/accuracy_checker/accuracy_checker/adapters/README.md
```diff
@@ -33,6 +33,8 @@ AccuracyChecker supports following set of adapters:
   * `coords` - number of bbox coordinates (default 4).
   * `num` - num parameter from DarkNet configuration file (default 5).
   * `cells` - number of cells across width and height (default 13).
+  * `raw_output` - enabling additional preprocessing for raw YOLO output format (default `False`).
+  * `output_format` - setting output layer format - boxes first (`BHW`)(default, also default for generated IRs), boxes last (`HWB`). Applicable only if network output not 3D (4D with batch) tensor.
```
eaidova, line 37:
```suggestion
  * `output_format` - setting output layer format:
      - `BHW` - boxes first (default, also default for generated IRs).
      - `HWB` - boxes last. 
      Applicable only if network output not 3D (4D with batch) tensor.
```

tools/accuracy_checker/accuracy_checker/adapters/yolo.py
eizamaliev: Well, the idea is to use `cells` option only if output not 3d or `do_reshape=True`. The `mobilefacedet-v1-mxnet` has 3d output tensor, so default cells will be ignored. This is not good idea, more silver bullet, but it is preferable than new adapter for now.

eaidova: ok, now I am patient that we do not break something :)

tools/accuracy_checker/accuracy_checker/adapters/yolo.py
```diff
@@ -1,13 +1,40 @@
...
 from ..utils import get_or_parse_value
 
+DetectionBox = namedtuple('DetectionBox', ["x", "y", "w", "h", "confidence", "probabilities"])
+
+
+class YoloOutputProcessor:
+    def __init__(self, coord_correct=None, size_correct=None, conf_correct=None,
+                 prob_correct=None, coord_normalizer=(1, 1), size_normalizer=(1, 1)):
+        self.coord_correct = coord_correct if coord_correct else lambda x: x
+        self.size_correct = size_correct if size_correct else lambda x: np.exp(x)
```
eaidova, line 18: no need in lambda, you can directly call numpy function
```suggestion
        self.size_correct = size_correct if size_correct else  np.exp
```

PR #1099, opened 2020-05-12
Use face-detection models in onnx format

AlexanderDokuchaev:


AlexanderDokuchaev:
@snosov1

vladimir-dudnik:
I think you need to move models in source framework format to public models. OMZ agreement is that intel models redistributed in IR form (and so do not require conversion to IR step).

PR #1116, opened 2020-05-18
Unify the reporting mechanisms used by the model tools

IRDonch:
The main reason I'm doing this is to prepare for implementing parallel downloading in the downloader, but as a side benefit, this also enforces consistent formatting in the converter and quantizer.

The info dumper only has machine-readable output, so it has no use for these mechanisms and is thus unaffected.

See the commit messages for more details.

### DIFF
--- a/tools/downloader/converter.py
+++ b/tools/downloader/converter.py
@@ -36,9 +36,9 @@ class QueuedOutputContext(common.JobContext):
     def print(self, value, *, end='\n', flush=False):
         self._output_queue.put(value + end)
 
-    def subprocess(self, args):
+    def subprocess(self, args, **kwargs):
         with subprocess.Popen(args, stdout=subprocess.PIPE, stderr=subprocess.STDOUT,
-                universal_newlines=True) as p:
+                universal_newlines=True, **kwargs) as p:
             for line in p.stdout:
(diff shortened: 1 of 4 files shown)


tools/downloader/converter.py
```diff
@@ -75,20 +60,20 @@ def cancel(self):
...
 
     conversion_to_onnx_args = [string.Template(arg).substitute(conv_dir=output_dir / model.subdirectory,
                                                                dl_dir=args.download_dir / model.subdirectory)
                                for arg in model.conversion_to_onnx_args]
     cmd = [str(args.python), str(Path(__file__).absolute().parent / model.converter_to_onnx), *conversion_to_onnx_args]
 
-    context.printf('Conversion to ONNX command: {}', common.command_string(cmd))
-    context.printf('')
+    reporter.print('Conversion to ONNX command: {}', common.command_string(cmd))
+    reporter.print(flush=True)
```
eizamaliev, line 73: Will it be equal to `reporter.print('Conversion to ONNX command: {}', common.command_string(cmd), flush=True)`?

IRDonch: No, the second `print` is to print a blank line before the output of the command.

PR #1117, opened 2020-05-18
demos/monitors: fix cpu monitors for Linux

Wovchena:


Wovchena:
@IRDonch, don't you know where it finds `::move()`? :)

IRDonch:
> @IRDonch, don't you know where it finds `::move()`? :)

Argument-dependent lookup. I would still suggest using `std::move` for clarity, though.

Wovchena:
Thank you!

PR #1119, opened 2020-05-19
Pipeline redesign for object_detection_demo_ssd_async (C++)

AnthonyQuantum:


demos/object_detection_demo_ssd_async/main.cpp
```diff
@@ -229,100 +281,93 @@ int main(int argc, char *argv[]) {
...
 
-        // --------------------------- 6. Do inference ---------------------------------------------------------
-        slog::info << "Start inference " << slog::endl;
-
-        bool isLastFrame = false;
-        bool isAsyncMode = false;  // execution is always started using SYNC mode
-        bool isModeChanged = false;  // set to TRUE when execution mode is changed (SYNC<->ASYNC)
+        // --------------------------- 6. Init variables -------------------------------------------------------
+        typedef std::chrono::duration<double, std::chrono::milliseconds::period> ms;
+        typedef std::chrono::duration<double, std::chrono::seconds::period> sec;
```
Wovchena, line 297: They represent types, so the names should be Ms and Sec.

demos/object_detection_demo_ssd_async/main.cpp
```diff
@@ -229,100 +281,93 @@ int main(int argc, char *argv[]) {
...
+        struct RequestResult {
+            cv::Mat frame;
+            const float* output;
+            std::chrono::high_resolution_clock::time_point startTime;
+            bool isSameMode;
+        };
+
+        struct ModeInfo {
+           int framesCount = 0;
+           double latencySum = 0;
```
Wovchena, line 308: You should use there the original type `std::chrono::steady_clock::duration`. Extract printable value only when you need to report it. With that approach you won't accumulate errors produced by double representation.

demos/object_detection_demo_ssd_async/main.cpp
```diff
@@ -229,100 +281,93 @@ int main(int argc, char *argv[]) {
...
+                     "press ESC or 'q' key" << std::endl;
+        std::cout << "To switch between min_latency/user_specified modes, press TAB key in the output window" 
+                  << std::endl;
+
+        while ((cap.isOpened()
+                || !completedRequestResults.empty()
+                || ((isUserSpecifiedMode && emptyRequests.size() < FLAGS_nireq)
+                    || (!isUserSpecifiedMode && emptyRequests.size() < 1)))
+               && callbackException == nullptr) {
+            if (callbackException) std::rethrow_exception(callbackException);
```
Wovchena, line 350: You should protect some of these vars because they are modified by another thread. I suggest moving the condition under the scope of the following `lock_guard`.

demos/object_detection_demo_ssd_async/main.cpp
```diff
@@ -346,59 +391,188 @@ int main(int argc, char *argv[]) {
...
+
+                std::ostringstream out;
+                out << "FPS: " << std::fixed << std::setprecision(1) << modeInfo[isUserSpecifiedMode].framesCount / 
+                    std::chrono::duration_cast<sec>(std::chrono::high_resolution_clock::now() -
+                                                    modeInfo[isUserSpecifiedMode].lastStartTime).count();
+                cv::putText(requestResult.frame, out.str(), cv::Point2f(10, 25), cv::FONT_HERSHEY_TRIPLEX, 0.6, 
+                            cv::Scalar(0, 0, 255), 1);
+                out.str("");
+                modeInfo[isUserSpecifiedMode].latencySum += std::chrono::duration_cast<sec>(
+                    std::chrono::high_resolution_clock::now() - requestResult.startTime).count();
```
Wovchena, line 414: You should have a variable storing the result of `now()`, so you don't need to call it twice and you won't get different values.

Actually the same idea should have been applied to `object_detection_demo_yolov3_async.py`, but I didn't notice that back than.

demos/object_detection_demo_ssd_async/main.cpp
```diff
@@ -346,59 +391,188 @@ int main(int argc, char *argv[]) {
...
+                out << "Latency: " << std::fixed << std::setprecision(1) << (modeInfo[isUserSpecifiedMode].latencySum /
+                    modeInfo[isUserSpecifiedMode].framesCount) * 1e3 << " ms";
+                cv::putText(requestResult.frame, out.str(), cv::Point2f(10, 50), cv::FONT_HERSHEY_TRIPLEX, 0.6,
```
Wovchena, line 417: I think you should add `put_highlighted_text()` as for `object_detection_demo_yolov3_async.py`.

demos/object_detection_demo_ssd_async/main.cpp
```diff
@@ -346,59 +391,188 @@ int main(int argc, char *argv[]) {
...
+                    cv::imshow("Detection Results", requestResult.frame);
+                    
+                    const int key = cv::waitKey(1);
+
+                    if (27 == key || 'q' == key || 'Q' == key) {  // Esc
+                        break;
+                    }
+                    else if (9 == key) {  // Tab
+                        bool prevMode = isUserSpecifiedMode;
+                        isUserSpecifiedMode ^= true;
```
Wovchena, line 430:
```suggestion
                        isUserSpecifiedMode = !isUserSpecifiedMode;
```

demos/object_detection_demo_ssd_async/main.cpp
```diff
@@ -346,59 +391,188 @@ int main(int argc, char *argv[]) {
...
+                        } else emptyRequests.push(minLatencyInferRequest);
+
+                        modeInfo[prevMode].lastEndTime = std::chrono::high_resolution_clock::now();
+                        modeInfo[isUserSpecifiedMode] = ModeInfo();
+                    } else {
+                        presenter.handleKey(key);
+                    }
+                }
             }
+            else if (!emptyRequests.empty() && cap.isOpened()) {
```
Wovchena, line 453:
```suggestion
            } else if (!emptyRequests.empty() && cap.isOpened()) {
```

demos/object_detection_demo_ssd_async/main.cpp
```diff
@@ -346,59 +391,188 @@ int main(int argc, char *argv[]) {
...
-            t1 = std::chrono::high_resolution_clock::now();
-            ocv_render_time = std::chrono::duration_cast<ms>(t1 - t0).count();
+                InferRequest::Ptr request;
+                {
+                    std::lock_guard<std::mutex> lock(mutex);
 
-            if (isLastFrame) {
-                break;
-            }
+                    request = emptyRequests.front();
```
Wovchena, line 472:
```suggestion
                    request = std::move(emptyRequests.front());
```

demos/object_detection_demo_ssd_async/main.cpp
```diff
@@ -346,59 +391,188 @@ int main(int argc, char *argv[]) {
...
+                }
+                frameToBlob(frame, request, imageInputName);
+                
+                bool frameMode = isUserSpecifiedMode;
+                request->SetCompletionCallback([&mutex,
+                                                &completedRequestResults,
+                                                nextFrameId,
+                                                frame,
+                                                request,
+                                                outputName,
```
Wovchena, line 483:
```suggestion
                                                &outputName,
```

demos/object_detection_demo_ssd_async/main.cpp
```diff
@@ -346,59 +391,188 @@ int main(int argc, char *argv[]) {
...
+                                                &emptyRequests,
+                                                &callbackException,
+                                                &condVar] {
+                    {
+                        std::lock_guard<std::mutex> callbackLock(mutex);
+                    
+                        try {
+                            completedRequestResults.insert(
+                                std::pair<int, RequestResult>(nextFrameId, RequestResult{
+                                    frame, request->GetBlob(outputName)->buffer().as<float*>(),
```
Wovchena, line 496: it should be rmap()

demos/object_detection_demo_ssd_async/main.cpp
```diff
@@ -346,59 +391,188 @@ int main(int argc, char *argv[]) {
...
+                    }
+                    condVar.notify_one();
+                });
 
-            if (isModeChanged) {
-                isModeChanged = false;
+                request->StartAsync();
+                nextFrameId++;
             }
+            else {
```
Wovchena, line 515:
```suggestion
            } else {
```

demos/object_detection_demo_ssd_async/README.md
```diff
@@ -3,16 +3,13 @@
 This demo showcases Object Detection with SSD and new Async API.
 Async API usage can improve overall frame-rate of the application, because rather than wait for inference to complete,
 the app can continue doing things on the host, while accelerator is busy.
-Specifically, this demo keeps two parallel infer requests and while the current is processed, the input frame for the next
-is being captured. This essentially hides the latency of capturing, so that the overall framerate is rather
-determined by the `MAXIMUM(detection time, input capturing time)` and not the `SUM(detection time, input capturing time)`.
+Specifically, this demo keeps the number of Infer Requests that you have set using input flag. While some of the Infer Requests are processed, the other ones can be filled with new frame data and asynchronously started or the next output can be taken from the Infer Request and displayed.
```
Wovchena, line 6:
```suggestion
Specifically, this demo keeps the number of Infer Requests that you have set using `nireq` flag. While some of the Infer Requests are processed by IE, the other ones can be filled with new frame data and asynchronously started or the next output can be taken from the Infer Request and displayed.
```

demos/object_detection_demo_ssd_async/main.cpp
Wovchena: There the problem with lifetime of `LockedMemory`. `outputMapped.as<float*>()` is stored in `RequestResult` while `outputMapped` is destroyed.

demos/object_detection_demo_ssd_async/main.cpp
```diff
@@ -354,58 +403,194 @@ int main(int argc, char *argv[]) {
...
+                                                &condVar] {
+                    {
+                        std::lock_guard<std::mutex> callbackLock(mutex);
+                    
+                        try {
+                            completedRequestResults.insert(
+                                std::pair<int, RequestResult>(nextFrameId, RequestResult{
+                                    frame,
+                                    std::unique_ptr<LockedMemory<const void>>(new
+                                        LockedMemory<const void>(as<MemoryBlob>(request->GetBlob(outputName))->rmap())),
```
IRDonch, line 509: Why do you map the blob here? IMO, it would be a lot clearer to do it where the contents of the blob are actually used.

demos/object_detection_demo_ssd_async/main.cpp
```diff
@@ -354,58 +403,194 @@ int main(int argc, char *argv[]) {
...
-                isModeChanged = true;
+                request->StartAsync();
+                nextFrameId++;
             } else {
-                presenter.handleKey(key);
+                std::unique_lock<std::mutex> lock(mutex);
+
+                while (callbackException == nullptr && emptyRequests.empty() && completedRequestResults.empty()) {
+                    condVar.wait(lock);
+                }
```
IRDonch, line 534: This seems odd. You wait for one of these objects to be updated from a background thread, but then you immediately release the lock and access those objects while the lock is _not_ held - isn't that thread-unsafe?

demos/object_detection_demo_ssd_async/main.cpp
```diff
@@ -235,102 +298,88 @@ int main(int argc, char *argv[]) {
...
+            emptyRequests.assign(userSpecifiedInferRequests.begin(), userSpecifiedInferRequests.end());
+        } else emptyRequests = {minLatencyInferRequest};
+
+        std::unordered_map<int, RequestResult> completedRequestResults;
+        int nextFrameId = 0;
+        int nextFrameIdToShow = 0;
+        std::exception_ptr callbackException = nullptr;
+        std::mutex mutex;
+        std::condition_variable condVar;
+        std::map<bool, ModeInfo> modeInfo = {{true, ModeInfo()}, {false, ModeInfo()}};
```
IRDonch, line 342: The usage of `bool` for the execution mode makes the code hard to understand, because it's not obvious which is `true` and which is `false`. I would suggest introducing an `enum class ExecutionMode` instead.

demos/object_detection_demo_ssd_async/main.cpp
```diff
@@ -354,58 +403,194 @@ int main(int argc, char *argv[]) {
...
+                            completedRequestResults.insert(
+                                std::pair<int, RequestResult>(nextFrameId, RequestResult{
+                                    frame,
+                                    std::unique_ptr<LockedMemory<const void>>(new
+                                        LockedMemory<const void>(as<MemoryBlob>(request->GetBlob(outputName))->rmap())),
+                                    startTime,
+                                    frameMode == isUserSpecifiedMode
+                                }));
+                            
+                            if (isUserSpecifiedMode == frameMode) {
```
IRDonch, line 514: I don't quite understand why this condition is here. Why wouldn't we want to return `request` to `emptyRequests` if the execution mode changed?

demos/object_detection_demo_ssd_async/main.cpp
```diff
@@ -113,14 +122,64 @@ int main(int argc, char *argv[]) {
...
+                }
+
+                // for CPU execution, more throughput-oriented execution via streams
+                userSpecifiedConfig.insert({ CONFIG_KEY(CPU_THROUGHPUT_STREAMS),
+                                (deviceNstreams.count(device) > 0 ? std::to_string(deviceNstreams.at(device))
+                                                                  : CONFIG_VALUE(CPU_THROUGHPUT_AUTO)) });
+
+                minLatencyConfig.insert({ CONFIG_KEY(CPU_THROUGHPUT_STREAMS), "1" });
+
+                deviceNstreams[device] = std::stoi(
```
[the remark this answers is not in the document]
AnthonyQuantum: Removed

demos/object_detection_demo_ssd_async/main.cpp
AnthonyQuantum: Done

demos/object_detection_demo_ssd_async/main.cpp
AnthonyQuantum: I've replaced `while` with `if` because `while` is not actually useful here, as experiments have shown.

security_barrier_camera_demo and classification_demo might need to be fixed too, because it looks like they have similar problem:
```
std::unique_lock<std::mutex> lk(tasksMutex);
while (running && tasks.empty()) {
        tasksCondVar.wait(lk);
}
```

```
{
        std::unique_lock<std::mutex> lock(mutex);

        while (!irCallbackException
                  && emptyInferRequests.empty() && completedInferRequests.empty()) {
                condVar.wait(lock);
        }
}
```

demos/object_detection_demo_ssd_async/main.cpp
AnthonyQuantum: It's because the `emptyRequests` container is only for requests from current mode. I've changed the name to make it clearer.

demos/object_detection_demo_ssd_async/main.cpp
IRDonch: > I've replaced while with if because while is not actually useful here, as experiments have shown.

No, `while` is necessary because of spurious wakeups.

demos/object_detection_demo_ssd_async/main.cpp
IRDonch: > It's because the emptyRequests container is only for requests from current mode.

Okay, but why, though? When you switch modes, you wait for all pending requests to complete and then reinitialize `emptyRequests`, so does it even matter which requests it contains at that point?

demos/object_detection_demo_ssd_async/main.cpp
```diff
@@ -354,58 +401,197 @@ int main(int argc, char *argv[]) {
...
+                                                &callbackException,
+                                                &condVar] {
+                    {
+                        std::lock_guard<std::mutex> callbackLock(mutex);
+                    
+                        try {
+                            completedRequestResults.insert(
+                                std::pair<int, RequestResult>(nextFrameId, RequestResult{
+                                    frame,
+                                    request->GetBlob(outputName),
```
IRDonch, line 509: Given that `request` is later moved to `currentModeEmptyRequests`, isn't it possible that it'll get reused before the result is processed?

demos/object_detection_demo_ssd_async/main.cpp
```diff
@@ -235,102 +292,92 @@ int main(int argc, char *argv[]) {
...
+           int framesCount = 0;
+           std::chrono::steady_clock::duration latencySum;
+           std::chrono::steady_clock::time_point lastStartTime;
+           std::chrono::steady_clock::time_point lastEndTime;
+        };
+
+        auto totalT0 = std::chrono::steady_clock::now();
+
+        // execution always starts in USER_SPECIFIED mode
+        enum class ExecutionMode {userSpecified, minLatency};
```
IRDonch, line 324:
```suggestion
        enum class ExecutionMode {USER_SPECIFIED, MIN_LATENCY};
```

Not only is this the convention for enum constants, but you even wrote `USER_SPECIFIED` yourself above. :)

demos/object_detection_demo_ssd_async/main.cpp
IRDonch: You haven't really fixed it yet. I can see that both `callbackException` and `emptyRequests` are accessed in the main loop while the mutex is not held.

demos/object_detection_demo_ssd_async/main.cpp
```diff
@@ -354,58 +401,197 @@ int main(int argc, char *argv[]) {
...
+                    {
+                        std::lock_guard<std::mutex> callbackLock(mutex);
+                    
+                        try {
+                            completedRequestResults.insert(
+                                std::pair<int, RequestResult>(nextFrameId, RequestResult{
+                                    frame,
+                                    std::make_shared<TBlob<float>>(*as<TBlob<float>>(request->GetBlob(outputName))),
+                                    startTime,
+                                    frameMode == currentMode
```
IRDonch, line 510: It would be simpler to just save `frameMode` and compare it to `currentMode` in the main loop (when you're processing the completed request). This way, you won't have to capture `currentMode` in the callback, nor access it from multiple threads without synchronization.

demos/object_detection_demo_ssd_async/main.cpp
```diff
@@ -354,58 +401,197 @@ int main(int argc, char *argv[]) {
...
-                break;
-            if (9 == key) {  // Tab
-                isAsyncMode ^= true;
-                isModeChanged = true;
+                request->StartAsync();
+                nextFrameId++;
             } else {
-                presenter.handleKey(key);
+                while (true) {
+                    std::unique_lock<std::mutex> lock(mutex);
```
IRDonch, line 528: Why did you move the lock inside the loop? There's no reason to repeatedly relock the mutex, the original loop structure was fine.

demos/object_detection_demo_ssd_async/main.cpp
AnthonyQuantum: Now it will be reused before the result is processed, but `output` was already copied into the new `RequestResult` object here:
```
completedRequestResults.insert(
                            std::pair<int, RequestResult>(nextFrameId, RequestResult{
                                frame,
                                std::make_shared<TBlob<float>>(*as<TBlob<float>>(request->GetBlob(outputName))),
                                startTime,
                                frameMode
                            }));
```
So, I suppose that everything is safe now.

PR #1122, opened 2020-05-20
AC: Fix coco_orig_metric

Anna-Mironova:
Calculation of coco_orig_metric has been fixed.   AP has been reduced to 66.6% from 69%. After fix AP for mentioned model is 68.37%. 

In this PR usage of original and generated annotation file has been combined by condition. If there is annotation file, it is used when calculating the metric. If annotation file is missing or there is subsample_size option, use generated annotation file.

### DIFF
--- a/tools/accuracy_checker/accuracy_checker/metrics/coco_orig_metrics.py
+++ b/tools/accuracy_checker/accuracy_checker/metrics/coco_orig_metrics.py
@@ -18,6 +18,7 @@ import os
 import tempfile
 import json
 import warnings
+import numpy as np
 from pathlib import Path
 try:
     from pycocotools.coco import COCO
@@ -92,7 +93,7 @@ class MSCOCOorigBaseMetric(FullDatasetEvaluationMetric):
 
         return data_to_store
 
-    def _prepare_coco_structures(self, annotations):
+    def _prepare_generated_coco_structures(self, annotations):
         label_map = self.dataset.metadata.get('label_map')
         if self.dataset.metadata.get('background_label') is not None:
             label_map.pop(self.dataset.metadata.get('background_label'))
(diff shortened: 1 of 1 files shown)


tools/accuracy_checker/accuracy_checker/metrics/coco_orig_metrics.py
```diff
@@ -252,9 +333,32 @@ def _run_coco_evaluation(coco, coco_res, iou_type='bbox', threshold=None):
...
+                    return True
+                else:
+                    return False
+            else:
+                return False
+
     def compute_precision_recall(self, annotations, predictions):
-        coco = self._prepare_coco_structures(annotations)
-        coco_data_to_store = self._convert_data_to_coco_format(predictions)
+        if(self._use_original_coco()):
```
eaidova, line 352:
```suggestion
        if self._use_original_coco():
```

tools/accuracy_checker/accuracy_checker/metrics/coco_orig_metrics.py
```diff
@@ -252,9 +333,32 @@ def _run_coco_evaluation(coco, coco_res, iou_type='bbox', threshold=None):
...
+            coco, map_coco_img_file_name_to_img_id, map_pred_label_id_to_coco_cat_id = \
+                self._prepare_original_coco_structures()
```
eaidova, line 354:
```suggestion
            structures = self._prepare_original_coco_structures()
            coco, map_coco_img_file_name_to_img_id, map_pred_label_id_to_coco_cat_id = structures
                
```

tools/accuracy_checker/accuracy_checker/metrics/coco_orig_metrics.py
```diff
@@ -107,7 +108,56 @@ def _prepare_coco_structures(self, annotations):
...
+
+    def _prepare_original_coco_structures(self):
+        annotation_conversion_parameters = self.dataset.config.get('annotation_conversion')
+        annotation_file = annotation_conversion_parameters.get('annotation_file')
+        has_background = annotation_conversion_parameters.get('has_background', False)
+        use_full_label_map = annotation_conversion_parameters.get('use_full_label_map', False)
+
+        meta = self.dataset.metadata
+
+        if COCO is None:
```
eaidova, line 128: If I right remember we move this check to configure, is not it?
If yes, possibly we do not need to check it twice

Anna-Mironova: there are checks for self.dataset.metadata and self.dataset.metadata.get('label_map') in configure. I can move check for COCO in configure too

PR #1124, opened 2020-05-20
YOLO pydemo: confusing output blob flattening is deleted

eizamaliev:


eizamaliev:
cc @vladimir-dudnik

demos/python_demos/object_detection_demo_yolov3_async/object_detection_demo_yolov3_async.py
```diff
@@ -161,44 +161,37 @@ def parse_yolo_region(blob, resized_image_shape, original_im_shape, params, thre
...
-            if confidence < threshold:
-                continue
-            objects.append(scale_bbox(x=x, y=y, height=height, width=width, class_id=class_id, confidence=confidence,
-                                      im_h=orig_im_h, im_w=orig_im_w, is_proportional=is_proportional))
+    for row, col, n in np.ndindex(params.side, params.side, params.num):
+        # Getting raw values for each detection bounding box
+        bbox = predictions[0, n*bbox_size:(n+1)*bbox_size, row, col]
+        x, y = bbox[0:2]
+        width, height = bbox[2:4]
+        object_probability = bbox[4]
```
Wovchena, line 172:
```suggestion
        x, y, width, height, object_probability = bbox[:5]
```

demos/python_demos/object_detection_demo_yolov3_async/object_detection_demo_yolov3_async.py
```diff
@@ -161,44 +161,35 @@ def parse_yolo_region(blob, resized_image_shape, original_im_shape, params, thre
...
-    predictions = blob.flatten()
-    side_square = params.side * params.side
     size_normalizer = (resized_image_w, resized_image_h) if params.isYoloV3 else (params.side, params.side)
+    bbox_size = params.coords + 1 + params.classes
     # ------------------------------------------- Parsing YOLO Region output -------------------------------------------
-    for i in range(side_square):
-        row = i // params.side
-        col = i % params.side
-        for n in range(params.num):
-            obj_index = entry_index(params.side, params.coords, params.classes, n * side_square + i, params.coords)
```
IRDonch, line 172: Looks like the `entry_index` function could now be deleted.

demos/python_demos/object_detection_demo_yolov3_async/object_detection_demo_yolov3_async.py
```diff
@@ -161,44 +161,35 @@ def parse_yolo_region(blob, resized_image_shape, original_im_shape, params, thre
...
-            confidence = class_probabilities[class_id]*object_probability
-            if confidence < threshold:
-                continue
-            objects.append(scale_bbox(x=x, y=y, height=height, width=width, class_id=class_id, confidence=confidence,
-                                      im_h=orig_im_h, im_w=orig_im_w, is_proportional=is_proportional))
+    for row, col, n in np.ndindex(params.side, params.side, params.num):
+        # Getting raw values for each detection bounding box
+        bbox = predictions[0, n*bbox_size:(n+1)*bbox_size, row, col]
+        x, y, width, height, object_probability = bbox[:5]
+        class_probabilities = bbox[5:]
```
IRDonch, line 171:
```suggestion
        x, y, width, height, object_probability, *class_probabilities = bbox
```

eizamaliev: Well, now I think, that initial code (with indexes) was much more clearer because it shows us how exactly values extracted from the blob. If anyone supports me, I'll revert this change, otherwise agree with you

demos/python_demos/object_detection_demo_yolov3_async/object_detection_demo_yolov3_async.py
eizamaliev: Thanks

demos/python_demos/object_detection_demo_yolov3_async/object_detection_demo_yolov3_async.py
Wovchena: That suggestion changes `class_probabilities`'s type. It will be a list. So `bbox[5:]` will be cast to list and than back to ndarray by np.argmax (but I'm not actually sure if argmax will cast list to ndarray before processing).

The current state and the suggestion also shows how values are extracted from the array: the fist 5 elements are special with given names and the rest are class_probabilities.

IRDonch: > That suggestion changes class_probabilities's type. It will be a list. So bbox[5:] will be cast to list and than back to ndarray by np.argmax (but I'm not actually sure if argmax will cast list to ndarray before processing).

Yeah, that's a good point. It would be better if it remained as an `ndarray`, if for no other reason than to be memory-efficient. I'm cancelling my suggestion. :-)

> Well, now I think, that initial code (with indexes) was much more clearer because it shows us how exactly values extracted from the blob.

I don't see how it was clearer. It obscured all accesses behind complicated index calculations. To me, it seems much clearer now.

pra-dan: Exactly. The `entry_index` function left me and some of my peers at the Intel OpenVINO-related project spooked. It is very difficult to understand the way the indices are calculated from the flattened blob. 

Either the process should be documented or the function should be removed.

PR #1127, opened 2020-05-20
AC: revert changes in coco orig metric behaviour

eaidova:


eaidova:
we make sure that everything works now correctly, so reverting changes unnecessary

PR #1132, opened 2020-05-22
AC: update adapter configuration

eaidova:
remove users confusing behaviour

### DIFF
--- a/tools/accuracy_checker/accuracy_checker/adapters/yolo.py
+++ b/tools/accuracy_checker/accuracy_checker/adapters/yolo.py
@@ -284,8 +284,7 @@ class YoloV3Adapter(Adapter):
                                      description="Minimal objectiveness score value for valid detections."),
             'outputs': ListField(
                 optional=False,
-                description="The list of output layers names (optional),"
-                            " if specified there should be exactly 3 output layers provided."
+                description="The list of output layers names."
             ),
             'anchor_masks': ListField(optional=True, description='per layer used anchors mask'),
             'do_reshape': BoolField(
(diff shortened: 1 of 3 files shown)


tools/accuracy_checker/accuracy_checker/adapters/yolo.py
```diff
@@ -283,7 +283,7 @@ def parameters(cls):
             'threshold': NumberField(value_type=float, optional=True, min_value=0, default=0.001,
                                      description="Minimal objectiveness score value for valid detections."),
             'outputs': ListField(
-                optional=True, default=[],
+                optional=False,
```
eizamaliev, line 286: By default `optional=False`, I think you should remove it.

tools/accuracy_checker/accuracy_checker/adapters/yolo.py
```diff
@@ -283,7 +283,7 @@ def parameters(cls):
...
                 description="The list of output layers names (optional),"
                             " if specified there should be exactly 3 output layers provided."
```
eizamaliev, line 288:
```suggestion
                description="The list of output layers names."
```

eizamaliev:
What if user describes only two layer for yolo-v3? Should we test this case?

tools/accuracy_checker/accuracy_checker/adapters/yolo.py
eaidova: fixed

eaidova:
> What if user describes only two layer for yolo-v3? Should we test this case?

It already tested by one user and it works :)

PR #1138, opened 2020-05-25
Add an option of restricted output characters for hjr0001

cocokale:
We want to add an option of restricted output characters for this demo.

As the modification in README.md says:

> 
When the designated_character argument is provided, the script will check Top k steps in looking up the decoded character, until a designated one is found. By doing so, the output character will be restricted to a designated region.


demos/python_demos/handwritten_japanese_recognition_demo/utils/codec.py
```diff
@@ -3,8 +3,9 @@
 
 class CTCCodec(object):
     """ Convert between text-label and text-index """
-    def __init__(self, characters):
+    def __init__(self, characters, args):
```
eizamaliev, line 6: Since you use only `designated_characters` and `top_k` values, you should cut `args` to only these values. Or may be even move to `decode` method.

demos/python_demos/handwritten_japanese_recognition_demo/handwritten_japanese_recognition_demo.py
```diff
@@ -42,6 +42,8 @@ def build_argparser():
     args.add_argument("-ni", "--number_iter", type=int, default=1,
                       help="Optional. Number of inference iterations")
     args.add_argument("-cl", "--charlist", type=str, default=os.path.join(os.path.dirname(__file__), "data/kondate_nakayosi_char_list.txt"), help="Path to the decoding char list file")
+    args.add_argument("-dc", "--designated_characters", type=str, default=None, help="Path to the designated character file")
+    args.add_argument("-tk", "--top_k", type=int, default=20, help="Top k steps in looking up the decoded character, until a designated one is found")
```
eizamaliev, line 46: Add `Optional.` to help description, see `-ni` key as example.

demos/python_demos/handwritten_japanese_recognition_demo/README.md
```diff
@@ -50,13 +52,27 @@ Options:
...
 ```
 
 
 For example:
 ```
 python handwritten_japanese_recognition_demo.py -i data/test.png -m path/to/ir_xml/model.xml
 
+```
+
+When the designated_character argument is provided, the script will check Top k steps in looking up the decoded character, until a designated one is found. By doing so, the output character will be restricted to a designated region. K is set to 20 by default.
```
eizamaliev, line 69:
```suggestion
When the `designated_character` argument is provided, the script will check Top k steps in looking up the decoded character, until a designated one is found. By doing so, the output character will be restricted to a designated region. K is set to 20 by default.
```

demos/python_demos/handwritten_japanese_recognition_demo/utils/codec.py
```diff
@@ -20,26 +22,43 @@ def decode(self, preds):
         texts = []
         index = 0
```
Wovchena, line 23:
```suggestion
```

demos/python_demos/handwritten_japanese_recognition_demo/README.md
```diff
@@ -50,13 +52,27 @@ Options:
...
 
 For example:
 ```
 python handwritten_japanese_recognition_demo.py -i data/test.png -m path/to/ir_xml/model.xml
 
+```
+
+When the `designated_characters` argument is provided, the script will check Top k steps in looking up the decoded character, until a designated one is found. By doing so, the output character will be restricted to a designated region. K is set to 20 by default.
+
+For example, if we want to restrict the output character to a digit and hyphen circumstance, we need to provide the path to the designated character file, then the command line will be like:
```
Wovchena, line 71:
```suggestion
For example, if we want to restrict the output character only to digits and hyphen circumstance, we need to provide the path to the designated character file and set K to the number of rows in charlist file, then the command line will be like:
```
The initial version misled me. I expected that only the chars from `digit_hyphen.txt` will appear, but other characters are still allowed if none of `digit_hyphen.txt` is in first K chosen elements.

demos/python_demos/handwritten_japanese_recognition_demo/README.md
```diff
@@ -50,13 +52,27 @@ Options:
...
+
+```
+python handwritten_japanese_recognition_demo.py -i data/test.png -m path/to/ir_xml/model.xml -dc data/digit_hyphen.txt
```
Wovchena, line 74:
```suggestion
python handwritten_japanese_recognition_demo.py -i data/test.png -m path/to/ir_xml/model.xml -dc data/digit_hyphen.txt -tk 4442
```

demos/python_demos/handwritten_japanese_recognition_demo/utils/codec.py
```diff
@@ -20,26 +22,43 @@ def decode(self, preds):
...
+        preds_index = preds_index.transpose(1, 0) # WB -> BW
+        preds_index_reshape = preds_index.reshape(-1) # B*W
+
+        char_list = []
+        if self.designated_characters != None:
+            designated_character_list = []
+            with open(self.designated_characters, encoding='utf-8') as f:
+                lines = f.readlines()
+                for line in lines:
+                    designated_character_list.append(line.strip())
```
Wovchena, line 35: You should move that reading to `__init__`. Otherwise the file is reread if `-ni` key > 1.

demos/python_demos/handwritten_japanese_recognition_demo/README.md
cocokale: Actually, your initial expectation is right. This pull request is based on the charlist file.  But if an additional designated_characters file is provided, this script will perform a **post**-filtering processing, which means it is possible that other characters are still allowed if none of digit_hyphen.txt is in first K chosen elements. Thus, the -tk argument has little to do with the charlist file, instead it is an extra argument for post-processing.

demos/python_demos/handwritten_japanese_recognition_demo/README.md
cocokale: Such an large `-tk 4442` does not make sense here, as it is a parameter for post-processing.

demos/python_demos/handwritten_japanese_recognition_demo/README.md
Wovchena: I think I was unclear with my comment. I meant that

I expected that only the chars from `digit_hyphen.txt` will appear, but *the reality turned out to be that* other characters are still allowed if none of `digit_hyphen.txt` is in first K chosen elements.

> post-filtering processing, which means it is possible that other characters are still allowed if none of digit_hyphen.txt is in first K chosen elements.

So that's what I was trying to point out, but your version is clearer :) Could you add it to readme?

cocokale: Sure. Thanks your careful review

demos/python_demos/handwritten_japanese_recognition_demo/utils/codec.py
Wovchena: Now `self.designated_characters` is not needed to be stored. In `__init()__` you can use `designated_characters` directly and in `decode()` you can check if `self.designated_character_list` is not None instead of `self.designated_characters`.

demos/python_demos/handwritten_japanese_recognition_demo/utils/codec.py
```diff
@@ -3,8 +3,17 @@
...
-    def __init__(self, characters):
+    def __init__(self, characters, designated_characters, top_k):
         # characters (str): set of the possible characters.
+        self.designated_character_list = None
+        if designated_characters != None:
+            self.designated_character_list = []
+            with open(designated_characters, encoding='utf-8') as f:
+                lines = f.readlines()
+                for line in lines:
+                    self.designated_character_list.append(line.strip())
```
IRDonch, line 14:
```suggestion
            with open(designated_characters, encoding='utf-8') as f:
                self.designated_character_list = [line.strip() for line in f]
```

demos/python_demos/handwritten_japanese_recognition_demo/README.md
```diff
@@ -50,13 +52,28 @@ Options:
...
 
 
 For example:
 ```
 python handwritten_japanese_recognition_demo.py -i data/test.png -m path/to/ir_xml/model.xml
 
+```
+When the `designated_characters` argument is provided, if the output character is not included in the designated characters, the script will check Top k steps in looking up the decoded character, until a designated one is found. By doing so, the output character will be restricted to a designated region. K is set to 20 by default.
+
+For example, if we want to restrict the output character to an only digits and hyphens circumstance, we need to provide the path to the designated character file, e.g. `digit_hyphen.txt`. Then the script will perform a post-filtering processing on the ouptut characters, but please note that it is possible that other characters are still allowed if none of `digit_hyphen.txt` is in first K chosen elements.
```
IRDonch, line 70:
```suggestion
For example, if we want to restrict the output characters to only digits and hyphens, we need to provide the path to the designated character file, e.g. `digit_hyphen.txt`. Then the script will perform a post-filtering processing on the ouptut characters, but please note that it is possible that other characters are still allowed if none of `digit_hyphen.txt` is in first K chosen elements.
```

IRDonch:
Jenkins: test this please

PR #1139, opened 2020-05-25
Model downloader: implement parallel downloading

IRDonch:
This adds a `-j` option to the model downloader, which will tell it to download multiple models at the same time, similarly to the model converter. Also similarly to the converter, the output from the concurrent downloading jobs will be serialized, so that it'll be indistinguishable from downloading all models sequentially. This, I will admit, is not ideal, but it's better than just having the jobs chaotically writer over each other's output. Ideally, we should have multiple progress bars updating at the same time, but that seems quite tricky to implement.

See the individual commits for more details (and easier-to-review diffs :-) ).

### DIFF
--- a/tools/downloader/README.md
+++ b/tools/downloader/README.md
@@ -126,6 +126,13 @@ human-readable format.
 You can also set this option to `text` to explicitly request the default text
 format.
 
+The script can download files for multiple models concurrently. To enable this,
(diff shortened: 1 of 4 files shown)


IRDonch:
I made a little edit to the docs to clarify that machine-readable events can get interleaved in parallel mode.

PR #1141, opened 2020-05-26
Initial ver of the BERT question-answering demo

myshevts:


myshevts:
Should work fine with the OMZ's Bert models by @vshampor

demos/python_demos/bert_question_answering/question_answering_demo.py
```diff
@@ -0,0 +1,338 @@
...
+    log.info("Loading network files:\n\t{}\n\t{}".format(model_xml, model_bin))
+    ie_encoder = ie.read_network(model=model_xml, weights=model_bin)
+
+    if args.reshape:
+        # reshape the sequence length to the context + maximum question length (in tokens)
+        first_input_layer = next(iter(ie_encoder.inputs))
+        c = ie_encoder.inputs[first_input_layer].shape[1]
+        # find the closest multiple of 64
+        seq = min(c, round((len(c_tokens_id) + args.max_question_token_num) / 64) * 64)
+        if seq < c:
```
Wovchena, line 173: Why is only narrowing of input sequence allowed? Can't the net be increased in input size in the same way?

demos/python_demos/bert_question_answering/question_answering_demo.py
```diff
@@ -0,0 +1,338 @@
...
+            new_shapes = dict([])
+            for i in input_info:
+                n, c = ie_encoder.inputs[i].shape
+                new_shapes[i] = [n, seq]
+                log.info("Reshaped input {} from {} to the {}".format(i, ie_encoder.inputs[i].shape, new_shapes[i]))
+            log.info("Attempting to reshape the network to the modified inputs...")
+            try:
+                ie_encoder.reshape(new_shapes)
+                log.info("Successful!")
+            except:
```
Wovchena, line 184:
```suggestion
            except RuntimeError:
```

demos/python_demos/bert_question_answering/question_answering_demo.py
```diff
@@ -0,0 +1,338 @@
...
+                log.info("Failed...reloading the network")
+                ie_encoder = ie.read_network(model=model_xml, weights=model_bin)
```
Wovchena, line 186: That approach with reloading the net would make scene if there was no demo key managing that. I think it is simpler and clearer just to fail if the model can't be reshaped. So next time a user just shouldn't request reshaping.

demos/python_demos/bert_question_answering/question_answering_demo.py
```diff
@@ -0,0 +1,338 @@
...
+        # calculate number of tokens for context in each inference request.
+        # reserve 3 positions for special tokens
+        # [CLS] q_tokens [SEP] c_tokens [SEP]
+        c_wnd_len = max_length - (len(q_tokens_id) + 3)
+
+        # token num between two neighbour context windows
+        # 1/2 means that context windows are overlapped by half
+        c_stride = c_wnd_len // 2
+
+        t0 = time.time()
```
Wovchena, line 228: It's better to use `time.perf_counter`.

demos/python_demos/bert_question_answering/question_answering_demo.py
```diff
@@ -0,0 +1,338 @@
...
+                input_names[0]: np.array([input_ids], dtype=np.int32),
+                input_names[1]: np.array([attention_mask], dtype=np.int32),
+                input_names[2]: np.array([token_type_ids], dtype=np.int32),
+            }
+            t_start = time.time()
+            # infer by IE
+            res = ie_encoder_exec.infer(inputs=inputs)
+            t_end = time.time()
+            t_count += 1
+            log.info("Sequence of length {} is processed with {:0.2f} sentence/sec ({:0.2} sec per request)".format(
```
Wovchena, line 263:
```suggestion
            log.info("Sequence of length {} is processed with {:0.2f} request/sec ({:0.2} sec per request)".format(
```

demos/python_demos/bert_question_answering/question_answering_demo.py
```diff
@@ -0,0 +1,338 @@
...
+
+            # convert to context text start-end index
+            max_s = c_tokens_se[c_s + max_s][0]
+            max_e = c_tokens_se[c_s + max_e][1]
+
+            # check that answers list does not have duplicates (because of context windows overlapping)
+            same = [i for i, a in enumerate(answers) if a[1] == max_s and a[2] == max_e]
+            if same:
+                assert len(same) == 1
+                # update exist answer record
```
Wovchena, line 306:
```suggestion
                # update existing answer record
```

demos/python_demos/bert_question_answering/question_answering_demo.py
```diff
@@ -0,0 +1,338 @@
...
+                a = answers[same[0]]
+                answers[same[0]] = (max(max_score, a[0]), max_s, max_e)
+            else:
+                # add new record
+                answers.append((max_score, max_s, max_e))
+
+            # check that context window reach the end
```
Wovchena, line 313:
```suggestion
            # check that context window reached the end
```

demos/python_demos/bert_question_answering/question_answering_demo.py
```diff
@@ -0,0 +1,338 @@
...
+    if response.status_code == 404:
+        raise Exception('URL is invalid')
+    html = bs4.BeautifulSoup(response.text, 'html.parser')
+    heads = html.select("#firstHeading")
+    title = heads[0].text if heads else "no title"
+    paragraphs = html.select("p")
+    log.info("Article title: '{}'".format(title))
+    context = '\n'.join([par.text for par in paragraphs])
+    log.info("Size: {} chars".format(len(context)))
+    log.info("Context: \033[91m {} \033[0m".format(context))
```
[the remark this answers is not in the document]
myshevts: we really need these to markup the text for a nice demo. On plain Windows cmd they do not work, but do not harm either

demos/python_demos/bert_question_answering/question_answering_demo.py
myshevts: It potentially can (up to the maximum sequence length that the network was trained for, usually 512).
But it brings no performance upside when the sequence are small (which is exactly the case we atrying to cover with reshaping here, to save on calculations).

demos/python_demos/bert_question_answering/question_answering_demo.py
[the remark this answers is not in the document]
Wovchena: The demo stops being nice when it prints random codes. In addition some Linux terminals don't support colors. I cant name such, but by default .bashrc usually has tests for color support, so it is definitely possible.

I don't see why you need highlighting for the context at all. And for the answer you could cast that part to uppercase or put brackets around.

demos/python_demos/bert_question_answering/question_answering_demo.py
Wovchena: I agree that it's better to shrink the model when we don't need its full length.

But what prevents from exceeding the source length of 512? Is that correct that it is possible to decrease the shape but not increase? Usually it works both ways.

myshevts: It is possible to increase (to make less inference calls on the large contexts that do not fit the original sequence length), but this is not covered by this simple demo.
the limitation of the 512 comes from the fact that certain layers in the model are trained to see only 512 tokens at most.

demos/python_demos/bert_question_answering_demo/README.md
```diff
@@ -0,0 +1,107 @@
...
+## Classifying Documents with Long Texts
+Notice that when the original "context" (text from the url) together with the question do not fit the model input
+(usually 384 tokens for the Bert-Large, or 128 for the Bert-Base), the demo splits the context into overlapping segments.
+Thus, for the long texts, the network is called multiple times. The results are then sorted by the probabilities.
+
+## Demo Performance
+Even though, the demo reports inference performance (by measuring wall-clock time for individual inference calls),
+it is rather baseline performance, as certain tricks like batching,
+[throughput mode] (https://docs.openvinotoolkit.org/latest/_docs_IE_DG_Intro_to_Performance.html) can be applied.
+Please use to the full-blown [Benchmark C++ Sample](https://docs.openvinotoolkit.org/latest/_inference_engine_samples_benchmark_app_README.html)
```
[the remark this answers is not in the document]
myshevts: thanks for spotting, fixed

demos/python_demos/bert_question_answering/question_answering_demo.py
Wovchena: I think just having multiplied `c` in `seq = min(c, round((len(c_tokens_id) + args.max_question_token_num) / 64) * 64)` will cover that. With may be an inplace comment added to clarify that in general it is limited only by memory size on a device.

I also just noticed, should it be `(len(c_tokens_id) + args.max_question_token_num + 63) / 64) * 64` to round up instead of `round((len(c_tokens_id) + args.max_question_token_num) / 64) * 64`?

demos/python_demos/bert_question_answering/question_answering_demo.py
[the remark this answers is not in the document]
myshevts: Ok, good point, so let's the colors optional

demos/python_demos/bert_question_answering/question_answering_demo.py
Wovchena: So what about support of reshaping up? The rest looks fine.

myshevts: I have tested the reshaping up and it brings no performance value, but does introduce the accuracy issues, probably because there are some layers that accommodated to "see" only the sequence that the model was trained with. 
I discussed this with data scientists and they said are not aware on "upsampling" the language models.

Wovchena: Ok, thank you!

demos/python_demos/bert_question_answering_demo/README.md
```diff
@@ -0,0 +1,130 @@
...
+sequence length that the network expects. This is performance (speed) and memory footprint saving option.
+Since some networks are not-reshapable (due to limitations of the internal layers) the reshaping might fail,
+so you will need to run the demo without it.
+Please see general ['reshape intro and limitations'](https://docs.openvinotoolkit.org/latest/_docs_IE_DG_ShapeInference.html)
+
+## Demo Outputs
+The application outputs found answers to the same console.
+
+## Supported Models
+[Open Model Zoo](https://github.com/opencv/open_model_zoo) comes with example BERT-large trained on the Squad*.
```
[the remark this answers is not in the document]
myshevts: Ok, will change to the models link https://github.com/opencv/open_model_zoo/tree/master/models/intel

demos/python_demos/bert_question_answering_demo/bert_question_answering_demo.py
```diff
@@ -0,0 +1,364 @@
...
+
+# return context as one big string by given input arguments
+def get_context(url):
+    log.info("Get context from {}".format(url))
+    response = requests.get(url)
+    if response.status_code == 404:
+        raise Exception('URL is invalid')
+    html = bs4.BeautifulSoup(response.text, 'html.parser')
+    heads = html.select("#firstHeading")
+    title = heads[0].text if heads else "no title"
```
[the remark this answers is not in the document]
myshevts: it is being logged

demos/python_demos/bert_question_answering_demo/README.md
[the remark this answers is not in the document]
IRDonch: OK, but make it a relative link, and point it at the index file.

demos/python_demos/bert_question_answering_demo/bert_question_answering_demo.py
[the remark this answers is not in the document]
IRDonch: I can see that, but... why? It has no bearing on what the demo does.

demos/python_demos/bert_question_answering_demo/bert_question_answering_demo.py
```diff
@@ -0,0 +1,370 @@
...
+                sys.exit(-1)
+        else:
+            log.info("Skipping network reshaping,"
+                     " as (context length + max question length) exceeds the current (input) network sequence length")
+
+    # check input and output names
+    input_names_model = list(ie_encoder.inputs.keys())
+    output_names_model = list(ie_encoder.outputs.keys())
+    input_names = eval(args.input_names)
+    output_names = eval(args.output_names)
```
IRDonch, line 224: It seems unnecessary to enable arbitrary code execution for these arguments, when all you need is a list of names. It's also inconvenient for the user to enter Python list literals as arguments. I would suggest to just split the arguments on commas.

demos/python_demos/bert_question_answering_demo/bert_question_answering_demo.py
```diff
@@ -0,0 +1,370 @@
...
+        log.error("    Actual network input->output names: {}->{}".format(input_names_model, output_names_model))
+        raise Exception("Unexpected network input or output names")
+
+    # load model to the device
+    log.info("Loading model to the {}".format(args.device))
+    ie_encoder_exec = ie.load_network(network=ie_encoder, device_name=args.device)
+
+    # loop on user's questions
+    while True:
+        question = input('Type question (enter to exit):')
```
IRDonch, line 240:
```suggestion
        question = input('Type question (empty string to exit):')
```

demos/python_demos/bert_question_answering_demo/bert_question_answering_demo.py
```diff
@@ -0,0 +1,370 @@
...
+                 "please use the benchmark_app tool (part of the OpenVINO samples) for any actual measurements.")
+        log.info("{} requests of {} length were processed in {:0.2f}sec ({:0.2}sec per request)".format(
+            t_count,
+            max_length,
+            t1 - t0,
+            (t1 - t0) / t_count
+        ))
+
+        # print top 3 results
+        answers = list(sorted(answers, key=lambda x: -x[0]))
```
IRDonch, line 362:
```suggestion
        answers = sorted(answers, key=lambda x: -x[0])
```

`sorted` already returns a list.

demos/python_demos/bert_question_answering_demo/bert_question_answering_demo.py
```diff
@@ -0,0 +1,370 @@
...
+                log.info("Successful!")
+            except RuntimeError:
+                log.error("Failed to reshape the network, please retry the demo without '-r' option")
+                sys.exit(-1)
+        else:
+            log.info("Skipping network reshaping,"
+                     " as (context length + max question length) exceeds the current (input) network sequence length")
+
+    # check input and output names
+    input_names_model = list(ie_encoder.inputs.keys())
```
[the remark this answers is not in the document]
myshevts: we need a set one more time ( when reporting error)

demos/python_demos/bert_question_answering_demo/bert_question_answering_demo.py
myshevts: fixed

demos/python_demos/bert_question_answering_demo/models.lst
```diff
@@ -0,0 +1,3 @@
+# This file can be used with the --list option of the model downloader.
+bert-small-uncased-whole-word-masking-squad-????
+bert-large-uncased-whole-word-masking-squad-????
```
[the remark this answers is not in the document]
myshevts: https://github.com/opencv/open_model_zoo/tree/develop/models/intel/bert-small-uncased-whole-word-masking-squad-0001
https://github.com/opencv/open_model_zoo/tree/develop/models/intel/bert-large-uncased-whole-word-masking-squad-fp32-onnx-0001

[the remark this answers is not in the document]
IRDonch: Yes. So you need to update this file to account for the final name.

vladimir-dudnik:
Jenkins please retry a build

IRDonch:
Jenkins: test this please

PR #1142, opened 2020-05-26
AC: processing different format of threshold for coco_metric

Anna-Mironova:


eaidova:
@Anna-Mironova  do you check that your solution also work for https://github.com/opencv/open_model_zoo/blob/master/tools/accuracy_checker/accuracy_checker/metrics/coco_orig_metrics.py

PR #1144, opened 2020-05-26
AC: align common semantic segmentation behaviour

eaidova:


ngaloppo:
@vladimir-dudnik @eaidova 

Are you sure this works? With this change, it is impossible for me to use accuracy checker. (worked before this change) 

Since the `identifier` and `mask_id` do not encode the relative paths anymore (e.g. `images/` and `labels/`) the accuracy_check `evaluator` module can't find them. Even if I force the image source location (with `-s <source_folder>`), then it will find the image, but not the mask.

eaidova:
@ngaloppo  you need to specify `segmentation_masks_source` parameter in config, which will be works as prefix for mask_id. `data_source ` used for images only now. these changes is not our decision, they were required for enabling sementic segmentaion task in DL Workbench

ngaloppo:
@eaidova Thank you for your quick response. That makes sense now. I didn't see the change to `dataset_definitions.yml` in the same PR so I assumed nothing else had changed. I'm assuming this is captured in the documentation?

ngaloppo:
@eaidova I am curious, how come you didn't have to specify `segmentation_masks_source` for all datasets that use `common_semantic_segmentation` in the `dataset_definitions.yml`? I only see it defined on VOC, even though there are many other ones that have a separate mask folder (e.g. adas).

eaidova:
@ngaloppo , it is my mistake, we have precollected annotation from previous releases in validation pipeline, I update definition when we will be ready to update validation infra. (anyway nobody has access to our data for these datasets). Regarding docs, yes we have note about masks_source, but I think we should highlight it, thanks

PR #1146, opened 2020-05-26
pytorch_to_onnx.py: explicitly specify the ONNX opset

IRDonch:
PyTorch might change the default opset in a future release to one that is unsupported by Model Optimizer. To prevent surprises, it's better to set the opset explicitly.


IRDonch:
> I think, it might be useful to manipulate opset version via parameter

I don't think it'll be necessary. As soon as both PyTorch and MO support a newer opset, we can just bump the version in the script and use it for all models. We'll only need to parameterize it if some models require different opsets (I hope not).

PR #1149, opened 2020-05-27
pytorch_to_onnx.py: work around an incompatibility with PyTorch 1.5

IRDonch:
PyTorch 1.5 started assigning names to ONNX nodes, which causes Model Optimizer to use those names as IR output names, rather than the ONNX output names. For us, this is backwards-incompatible (our existing `--output` options stop working) and just plain inconvenient (the PyTorch-generated names are much less descriptive than our hand-picked output names).

Work around it by deleting the ONNX node names after conversion (effectively, restoring the PyTorch 1.4 behavior).

### DIFF
--- a/tools/downloader/pytorch_to_onnx.py
+++ b/tools/downloader/pytorch_to_onnx.py
@@ -101,14 +101,22 @@ def convert_to_onnx(model, input_shape, output_file, input_names, output_names):
     torch.onnx.export(model, dummy_input, str(output_file), verbose=False, opset_version=9,
                       input_names=input_names.split(','), output_names=output_names.split(','))
 
-    # Model check after conversion
     model = onnx.load(str(output_file))
+
+    # Model Optimizer takes output names from ONNX node names if they exist.
+    # However, the names PyTorch assigns to the ONNX nodes are generic and
+    # non-descriptive (e.g. "Gemm_151"). By deleting these names, we make
+    # MO fall back to the ONNX output names, which we can set to whatever we want.
+    for node in model.graph.node:
+        node.ClearField('name')
+
     try:
         onnx.checker.check_model(model)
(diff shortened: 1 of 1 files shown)


tools/downloader/pytorch_to_onnx.py
```diff
@@ -101,14 +101,22 @@ def convert_to_onnx(model, input_shape, output_file, input_names, output_names):
     torch.onnx.export(model, dummy_input, str(output_file), verbose=False, opset_version=9,
                       input_names=input_names.split(','), output_names=output_names.split(','))
 
-    # Model check after conversion
     model = onnx.load(str(output_file))
+
+    # Model Optimizer takes output names from ONNX node names if they exist.
+    # However, the names PyTorch assigns to the ONNX nodes are generic and
+    # non-descriptive (e.g. "Gemm_151"). By deleting these names, we make
+    # MO fall back to the ONNX output names, which we can set to whatever we want.
```
eizamaliev, line 109:
```suggestion
    # However, the names PyTorch assigns to the ONNX nodes since version 1.5 are generic 
    # and non-descriptive (e.g. "Gemm_151"). By deleting these names, we make
    # MO fall back to the ONNX output names, which we can set to whatever we want.
```
Sorry if I broke formatting.

IRDonch: Your wording makes it sound like before 1.5, PyTorch assigned perfectly descriptive names, but this is not the case - it didn't assign any names at all.

If you'd like, I can adjust the wording to explain it better, but IMO, it's not that relevant - the key point is that whenever PyTorch assigns node names, we don't want them. The historical details are in the commit message.

eizamaliev: Suit yourself, it is more suggestion than requirement.

PR #1154, opened 2020-05-28
Revert "Update typo error faster_rcnn_inception_v2_coco.md"

vladimir-dudnik:
Reverts opencv/open_model_zoo#1153 as we do not accept PR for master branch in general (this is branch corresponding content of the latest public release)


IRDonch:
Looks good, but now we should port this patch to `develop` so that it can be included in the next release. Will you do it?

PR #1160, opened 2020-05-29
AC: Added ovelapping while clipping audio

eizamaliev:


tools/accuracy_checker/accuracy_checker/preprocessor/README.md
```diff
@@ -119,5 +119,6 @@ Accuracy Checker supports following set of preprocessors:
 * `resample_audio` - converts audio to new sample rate
   * `sample_rate` - sets new sample rate
 * `clip_audio` - slices audio into several parts with equal duration
-  * `duration` - sets duration in seconds
-  * `max_clips` - sets the maximum number of clips (by default `1`)
+  * `duration`, `duration_in_samples` - sets duration in seconds or samples respectively (only one acceptable)
```
eaidova, line 122: it is not clear in which cases which option should be used.
Also i think it can be combined in simmilar way like it done in `subsample size` (option accepts integer values and string with % at the end for relative partition, here for seconds can be used with s postfix for example)

tools/accuracy_checker/accuracy_checker/preprocessor/audio_preprocessing.py
```diff
@@ -65,19 +65,40 @@ def parameters(cls):
...
+                value_type=float, min_value=0, optional=True, default=None,
                 description="Length of audio clip in seconds."
             ),
+            'duration_in_samples': NumberField(
+                value_type=int, min_value=0, optional=True, default=None,
+                description="Length of audio clip in samples."
+            ),
             'max_clips': NumberField(
-                value_type=int, min_value=1, default=1,
+                value_type=int, min_value=1, default=999, optional=True,
```
eaidova, line 76: it is hard to understand, 999 was set by user or auto value
what if somebody have too long input?
I suggest to use -1 as default or not set at all for understanding that user set nothing

eizamaliev: done

tools/accuracy_checker/accuracy_checker/preprocessor/README.md
eizamaliev: Thanks for comment. I delete duplication, see changes in Readme for more information.

eaidova:
Jenkins please retry a build

tools/accuracy_checker/accuracy_checker/preprocessor/audio_preprocessing.py
```diff
@@ -64,20 +64,76 @@ class ClipAudio(Preprocessor):
...
+                                      .format(self.__provider__, self.duration_in_samples))
+            else:
+                raise ConfigError("Preprocessor {}: invalid value for duration - {}.".
+                                  format(self.__provider__, duration))
+        else:
+            try:
+                self.duration = float(duration)
+            except ValueError:
+                raise ConfigError("Preprocessor {}: invalid value for duration - {}."
+                                  .format(self.__provider__, duration))
```
eaidova, line 101: probably duration configuration should be moved to independent function, too many logic it has to be a part of common configuration

tools/accuracy_checker/accuracy_checker/preprocessor/audio_preprocessing.py
```diff
@@ -64,20 +64,76 @@ class ClipAudio(Preprocessor):
...
+            if self.duration <= 0:
+                raise ConfigError("Preprocessor {}: duration should be positive value - {}."
+                                  .format(self.__provider__, self.duration))
+
+        self.max_clips = self.get_value_from_config('max_clips') or np.inf
+
+        overlap = self.get_value_from_config('overlap')
+        if isinstance(overlap, str):
```
eaidova, line 109: the same comment about function

PR #1168, opened 2020-05-31
Updated person-detection-asl-0001.md

shashwat623:
Corrected the output description for the bounding boxes coordinates.


eaidova:
@shashwat623  thank you for you patch, but your fix is incorrect.
The standard image coordinate system defined with start point in left top corner of image (0, 0). 
The abscissa axis is directed to the right, and the ordinate axis is directed down correspondingly.
So the higher y coordinate will be closer to image bottom, not top.

PR #1170, opened 2020-06-01
classification_demo: Add HBONet support, change default resolution

AnthonyQuantum:
Works with all three new HBONet-s.

I've also changed the default resolution because 1920x1080 was higher than what almost all other demos had.

### DIFF
--- a/demos/classification_demo/main.cpp
+++ b/demos/classification_demo/main.cpp
@@ -233,12 +233,13 @@ int main(int argc, char *argv[]) {
 
         // --------------------------------------Configure model output---------------------------------------
         auto outputInfo = network.getOutputsInfo();
-        if (outputInfo.size() != 1) {
-            throw std::logic_error("The network should have only one output.");
+        if (outputInfo.size() < 1 || outputInfo.size() > 3) {
+            throw std::logic_error("The network has an incorrect number of outputs.");
         }
-
-        auto outputName = outputInfo.begin()->first;
-        auto outputLayerData = outputInfo.begin()->second;
+        auto outputIterator = (outputInfo.size() == 1) ? outputInfo.begin()
+                                                       : --outputInfo.end(); // for HBONet support
+        auto outputName = outputIterator->first;
+        auto outputLayerData = outputIterator->second;
         auto layerDataDims = outputLayerData->getTensorDesc().getDims();
         if (layerDataDims.size() != 2 && layerDataDims.size() != 4) {
             throw std::logic_error("Incorrect number of dimensions in model output layer. Must be 2 or 4.");
--- a/demos/classification_demo/main.cpp
+++ b/demos/classification_demo/main.cpp
@@ -233,13 +233,12 @@ int main(int argc, char *argv[]) {
 
         // --------------------------------------Configure model output---------------------------------------
         auto outputInfo = network.getOutputsInfo();
-        if (outputInfo.size() < 1 || outputInfo.size() > 3) {
-            throw std::logic_error("The network has an incorrect number of outputs.");
+        if (outputInfo.size() != 1) {
+            throw std::logic_error("The network should have only one output.");
         }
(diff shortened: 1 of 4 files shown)


demos/classification_demo/main.cpp
```diff
@@ -233,12 +233,13 @@ int main(int argc, char *argv[]) {
 
         // --------------------------------------Configure model output---------------------------------------
         auto outputInfo = network.getOutputsInfo();
-        if (outputInfo.size() != 1) {
-            throw std::logic_error("The network should have only one output.");
+        if (outputInfo.size() < 1 || outputInfo.size() > 3) {
```
Wovchena, line 236:
```suggestion
        if (outputInfo.size() > 3) {
```
That's strange to expect that a net doesn't have any output.

demos/classification_demo/main.cpp
```diff
@@ -233,12 +233,13 @@ int main(int argc, char *argv[]) {
...
+            throw std::logic_error("The network has an incorrect number of outputs.");
         }
-
-        auto outputName = outputInfo.begin()->first;
-        auto outputLayerData = outputInfo.begin()->second;
+        auto outputIterator = (outputInfo.size() == 1) ? outputInfo.begin()
+                                                       : --outputInfo.end(); // for HBONet support
```
Wovchena, line 240: You shouldn't rely on `outputInfo` order. Better to check that output you are going to use has the shape you expect.

Docs claim there is one output: https://github.com/opencv/open_model_zoo/blob/develop/models/public/hbonet-0.25/hbonet-0.25.md Either docs are incorrect or the model is wrongly converted. @dkurt to comment.

dkurt: Here is `.xml` file. It has 3 outputs but only one is actually classification output, `1x1000`: 
[hbonet-0.25.txt](https://github.com/opencv/open_model_zoo/files/4711569/hbonet-0.25.txt)

dkurt: Looks like this is MO issue because Split is not ignored:
```xml
<layer id="53" name="510/Split" type="VariadicSplit" version="opset1">
			<input>
				<port id="0">
					<dim>1</dim>
					<dim>10</dim>
					<dim>112</dim>
					<dim>112</dim>
				</port>
				<port id="1"/>
				<port id="2">
					<dim>2</dim>
				</port>
			</input>
			<output>
				<port id="3" precision="FP32">
					<dim>1</dim>
					<dim>1</dim>
					<dim>112</dim>
					<dim>112</dim>
				</port>
				<port id="4" precision="FP32">
					<dim>1</dim>
					<dim>9</dim>
					<dim>112</dim>
					<dim>112</dim>
				</port>
			</output>
		</layer>
```

Wovchena: > Here is .xml file. It has 3 outputs

How's that 3? `if (outputInfo.size() < 1 || outputInfo.size() > 3)` suggests there are 2 :)

> Looks like this is MO issue

@dkurt, will you file a bug for MO?

AnthonyQuantum: @Wovchena 
hbonet-0.25 has three outputs, hbonet-0.5 has two and hbonet-1.0 has only one.
Since 60+ other classification models all have 1 output, I agree that it is a MO issue.

dkurt: So maybe for current PR we can just use the latest topological output? It will still work for single output networks as well.

dkurt: I just tested https://github.com/openvinotoolkit/openvino/commit/41ed6f0891fa5d5c07257bf6b920e57199e4f82a and it produces correct IR with 1 output node.

AnthonyQuantum:
Works with all three hbonet-s, MO from openvinotoolkit/openvino@8377c71.

Tests are failing because of the recent person-reidentification-retail models update (there are no model.yml files right now).

IRDonch:
Jenkins please retry a build

PR #1167, opened 2020-05-30
Person ReId models update

DmitriySidnev:
This is update for new person ReID models.
@snosov1, @sovrasov

PR #1177, opened 2020-06-02
Temporarily restore the old ReID models

IRDonch:
They were deleted in #1167, but they're still used in demo tests. We can't use the new models yet, because their files haven't been published yet.

Restore them until the new models' files are published.


eaidova:
@IRDonch there one model  was stayed (0248), is not enough for demo testing?

IRDonch:
I could temporarily reduce the test cases to just that model, but this way it's easier to remember to add the new models later.

PR #1182, opened 2020-06-03
tqdm progress reporter fix

druzhkov-paul:


eaidova:
@druzhkov-paul , thank you

PR #1183, opened 2020-06-03
AC support for person-detection-0106

druzhkov-paul:


tools/accuracy_checker/accuracy_checker/adapters/mask_rcnn.py
```diff
@@ -174,18 +194,7 @@ def _process_pytorch_outputs(self, raw_outputs, identifiers, frame_meta):
...
-                per_obj_raw_masks = []
-                for cls, raw_mask in zip(classes, raw_masks):
-                    per_obj_raw_masks.append(raw_mask[cls, ...])
-            else:
-                per_obj_raw_masks = np.squeeze(raw_masks, axis=1)
-
-            for box, raw_cls_mask in zip(boxes, per_obj_raw_masks):
-                mask = self.segm_postprocess(box, raw_cls_mask, *original_image_size, True, True)
-                masks.append(mask)
+            masks = self._process_masks_pytorch(boxes, raw_masks, identifiers, original_image_size)
```
eaidova, line 197: better to call method only if raw_mask is not None and remove check inside

tools/accuracy_checker/configs/person-detection-0106.yml
```diff
@@ -0,0 +1,58 @@
...
+
+    launchers:
+      - framework: dlsdk
+        tags:
+          - FP32
+        model:   intel/person-detection-0106/FP32/person-detection-0106.xml
+        weights: intel/person-detection-0106/FP32/person-detection-0106.bin
+        adapter:
+          type: mask_rcnn
+          boxes_out: boxes
```
eaidova, line 12: why you need mask_rcnn adapter if you have only boxes output?

druzhkov-paul: Two reasons here. First, is that some adapter is required. :smile: Second, for consistency. In OTE we currently export detection and instance segmentation models is a common way, so having a common adapter in AC looks reasonable to me.

eaidova: than `mask_rcnn` name looks confused :)

tools/accuracy_checker/accuracy_checker/adapters/mask_rcnn.py
```diff
@@ -152,12 +153,31 @@ def _process_tf_obj_detection_api_outputs(self, raw_outputs, identifiers, frame_
...
 
     def _process_pytorch_outputs(self, raw_outputs, identifiers, frame_meta):
-        classes = raw_outputs[self.classes_out]
+        classes = raw_outputs.get(self.classes_out, None)
+        boxes = raw_outputs.get(self.boxes_out, None)
+        scores = raw_outputs.get(self.scores_out, None)
+        raw_masks = raw_outputs.get(self.raw_masks_out, None)
+
+        if boxes is None:
+            for v in raw_outputs.values():
```
eaidova, line 162: It is not safe from quantization perspective (outputs dictionary extended with layers for statistics collection), I can not guarantee that in the model always only 2 layer with such shape and dims...

druzhkov-paul: I see. Changed the logic a bit, please take a look.

eizamaliev:
This PR should be merged with [PR 1184](https://github.com/opencv/open_model_zoo/pull/1184). Now you have no need to change it, but in future please publish all that model need, under single PR.

PR #1186, opened 2020-06-03
AC: Added logging of found models and weights for piplined models

Anna-Mironova:


eaidova:
@Anna-Mironova , well, could you please also update colorization evaluator?

tools/accuracy_checker/accuracy_checker/evaluators/custom_evaluators/mtcnn_evaluator.py
```diff
@@ -393,6 +395,7 @@ def auto_model_search(self, network_info):
                         raise ConfigError('Several suitable weights found, please specify required explicitly')
                     weights = weights_list[0]
             weights = Path(weights)
+            print_info('{} - Found weights: {}'.format(self.default_model_name, get_path(weights)))
```
eaidova, line 398:
```suggestion
            weights = get_path(weights)
            print_info('{} - Found weights: {}'.format(self.default_model_name, weights))
```

tools/accuracy_checker/accuracy_checker/evaluators/custom_evaluators/colorization_evaluator.py
```diff
@@ -54,6 +55,14 @@ def from_configs(cls, config, delayed_model_loading=False):
             launcher_settings['device'] = 'CPU'
         launcher = create_launcher(launcher_settings, delayed_model_loading=True)
         network_info = config.get('network_info', {})
+        colorization_network = network_info.get('colorization_network', {})
+        verification_network = network_info.get('verification_network', {})
+        colorization_network['net_type'] = 'colorization_network'
+        verification_network['net_type'] = 'verification_network'
```
eaidova, line 61: why do you need this? they have this info as keys in network info, why is it not enough for you?

tools/accuracy_checker/accuracy_checker/evaluators/custom_evaluators/sequential_action_recognition_evaluator.py
```diff
@@ -413,10 +414,11 @@ def automatic_model_search(self, network_info):
             if len(model_list) > 1:
                 raise ConfigError('Several suitable models for {} found'.format(self.default_model_suffix))
             model = model_list[0]
+            print_info('{} - Found model: {}'.format(self.default_model_suffix, model))
         if model.suffix == '.blob':
             return model, None
         weights = network_info.get('weights', model.parent / model.name.replace('xml', 'bin'))
-
+        print_info('{} - Found weights: {}'.format(self.default_model_suffix, get_path(weights)))
```
eaidova, line 421: the same comment about moving get_path.
I mean that It should be applied when we get path in first time, not when only print it.
It logically incorrect and give mess in error logs

tools/accuracy_checker/accuracy_checker/evaluators/custom_evaluators/colorization_evaluator.py
```diff
@@ -374,8 +377,8 @@ def set_input_and_output(self):
 
 
 class ColorizationCheckModel(BaseModel):
-    def __init__(self, network_info, launcher, delayed_model_loading=False):
-        super().__init__(network_info, launcher, delayed_model_loading)
+    def __init__(self, network_info, launcher, net_type, delayed_model_loading=False):
```
eaidova, line 380: inside the model you know its type, so it can be constant field like in action recognition and text spotting, no need have it as constructor argument

PR #1190, opened 2020-06-04
temporaly restore person-reidentification-retail docs

eaidova:
lack of this model in index makes false positives in docs build,
like here https://github.com/opencv/open_model_zoo/pull/1184


IRDonch:
I'm afraid you'll have to restore the doc pages themselves too, else there'll be broken links.

eaidova:
> I'm afraid you'll have to restore the doc pages themselves too, else there'll be broken links.

yes, I also return back them

PR #1191, opened 2020-06-04
[DEMO] Add labels file to CPP Object Detection demos

eizamaliev:
Adds option `-labels` for file with labels mapping. Also fixed previous errors with reading labels.

### DIFF
--- a/demos/object_detection_demo_ssd_async/main.cpp
+++ b/demos/object_detection_demo_ssd_async/main.cpp
@@ -124,19 +124,14 @@ int main(int argc, char *argv[]) {
         cnnNetwork.setBatchSize(1);
         /** Read labels (if any)**/
         std::vector<std::string> labels;
-        std::string labelFileName;
         if (!FLAGS_labels.empty()) {
-            labelFileName = FLAGS_labels;
-            std::ifstream inputFile(labelFileName);
+            std::ifstream inputFile(FLAGS_labels);
             std::string label;
             while (std::getline(inputFile, label)) {
                 labels.push_back(label);
             }
-        }
-        if (!labels.empty()) {
-            slog::info << "Loaded " << labels.size() << " labels" << slog::endl;
-        } else {
-            slog::info << "File " << labelFileName << " empty or not found. Labels are omitted." << slog::endl;
+            if (labels.empty())
+                throw std::logic_error("File empty or not found: " + FLAGS_labels);
         }
         // -----------------------------------------------------------------------------------------------------
 
@@ -201,14 +196,11 @@ int main(int argc, char *argv[]) {
             throw std::logic_error("Class labels are not supported with IR version older than 10");
         }
 
-        if (static_cast<int>(labels.size()) != num_classes) {
+        if (!labels.empty() && static_cast<int>(labels.size()) != num_classes) {
             if (static_cast<int>(labels.size()) == (num_classes - 1))  // if network assumes default "background" class, having no label
                 labels.insert(labels.begin(), "fake");
             else {
-                slog::info << "The number of labels (" << labels.size() << ") "
-                    << "is different from numbers of model classes (" << num_classes << "). "
-                    << "Labels are omitted." << slog::endl;
-                labels.clear();
+                throw std::runtime_error("The number of labels is different from numbers of model classes");
             }                
         }
         const SizeVector outputDims = output->getTensorDesc().getDims();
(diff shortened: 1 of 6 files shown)


demos/object_detection_demo_ssd_async/main.cpp
```diff
@@ -123,12 +123,21 @@ int main(int argc, char *argv[]) {
         slog::info << "Batch size is forced to  1." << slog::endl;
         cnnNetwork.setBatchSize(1);
         /** Read labels (if any)**/
-        std::string labelFileName = fileNameNoExt(FLAGS_m) + ".labels";
         std::vector<std::string> labels;
-        std::ifstream inputFile(labelFileName);
-        std::copy(std::istream_iterator<std::string>(inputFile),
-                  std::istream_iterator<std::string>(),
-                  std::back_inserter(labels));
+        std::string labelFileName;
```
Wovchena, line 127: You can use `FLAGS_labels`. No need to copy it here.

demos/object_detection_demo_ssd_async/main.cpp
```diff
@@ -123,12 +123,21 @@ int main(int argc, char *argv[]) {
...
+            std::string label;
+            while (std::getline(inputFile, label)) {
+                labels.push_back(label);
+            }
+        }
+        if (!labels.empty()) {
+            slog::info << "Loaded " << labels.size() << " labels" << slog::endl;
+        } else {
+            slog::info << "File " << labelFileName << " empty or not found. Labels are omitted." << slog::endl;
+        }
```
Wovchena, line 140: These prints seem to be useless to me. We notice that labels are not used when the demo draws numbers instead of words. And if `-labels` is omitted intentionally it becomes even annoying because the demo reports that.

As for `labels.size()`, the size itself is not interesting and you report about labels and model output mismatch later.

demos/object_detection_demo_ssd_async/main.cpp
```diff
@@ -195,8 +204,12 @@ int main(int argc, char *argv[]) {
         if (static_cast<int>(labels.size()) != num_classes) {
             if (static_cast<int>(labels.size()) == (num_classes - 1))  // if network assumes default "background" class, having no label
                 labels.insert(labels.begin(), "fake");
-            else
+            else {
+                slog::info << "The number of labels (" << labels.size() << ") "
+                    << "is different from numbers of model classes (" << num_classes << "). "
+                    << "Labels are omitted." << slog::endl;
```
Wovchena, line 210: I have an idea it is better to throw an exception and fail instead of trying to pseudo fix labels. A user can do the same by removing `-labels` and this is more explicit.

eizamaliev: Fair enough, I just tried to save old behavior.

demos/object_detection_demo_ssd_async/main.cpp
```diff
@@ -201,14 +196,11 @@ int main(int argc, char *argv[]) {
             throw std::logic_error("Class labels are not supported with IR version older than 10");
         }
 
-        if (static_cast<int>(labels.size()) != num_classes) {
+        if (!labels.empty() && static_cast<int>(labels.size()) != num_classes) {
```
Wovchena, line 199: Why is `!labels.empty()` needed? It should be fine without it

demos/object_detection_demo_ssd_async/main.cpp
Wovchena: I'm not sure how to choose an exception type, but given that the demo uses `logic_error` mostly, you should throw `logic_error`.

demos/object_detection_demo_ssd_async/main.cpp
eizamaliev: We have three different cases here:
1. Labels aren't set at all (`-labels` omitted)
2. Labels have no `background` class
3. Labels number differs from number of classes.

First case we simply ignore. In second case we add `fake` label. Third case is an error. 
If we leave this section:
```
        if (static_cast<int>(labels.size()) != num_classes) {
            if (static_cast<int>(labels.size()) == (num_classes - 1))  
                labels.insert(labels.begin(), "fake");
            else {
                throw std::runtime_error("The number of labels is different from numbers of model classes");
            }                
        }
```
it will break everytime without `-labels` option specified.

Wovchena: You are right, ok.

demos/object_detection_demo_ssd_async/main.cpp
eizamaliev: Done. There are both appropriate, but in ssd_demo all errors are logic, so be it. I dont change it in yolo_demo, since there are presented both type of errors.

PR #1198, opened 2020-06-05
 AC: Added logging of input/output info for piplined models 

Anna-Mironova:


tools/accuracy_checker/accuracy_checker/evaluators/custom_evaluators/text_spotting_evaluator.py
```diff
@@ -514,6 +543,7 @@ def load_model(self, network_info, launcher):
         )
         self.im_info_name = [x for x in input_info if len(input_info[x].shape) == 2][0]
         self.im_data_name = [x for x in input_info if len(input_info[x].shape) == 4][0]
+        self.print_input_output_info()
```
eaidova, line 546: in common pipeline this function called only when logging enabled.
There are some reason why it was done:
1. reshape require model reloading, so after reshape this message will be printed on each image
2. quantization required also several times reload network. 
So please condition on logging here

PR #1200, opened 2020-06-05
Update readmes

Wovchena:


demos/python_demos/text_spotting_demo/README.md
```diff
@@ -116,13 +115,13 @@ To run the demo, you can use public or pre-trained models. To download the pre-t
 
 > **NOTE**: Before running the demo with a trained model, make sure the model is converted to the Inference Engine format (`*.xml` + `*.bin`) using the [Model Optimizer tool](https://docs.openvinotoolkit.org/latest/_docs_MO_DG_Deep_Learning_Model_Optimizer_DevGuide.html).
 
-To run the demo, please provide paths to the model in the IR format and to an input video, image, or folder with images:
+To run the demo, please provide paths to the model in the IR format and to an input with images:
```
vladimir-dudnik, line 118: is that demo accept only video (from file or from camera) but not image?

demos/python_demos/text_spotting_demo/text_spotting_demo.py
```diff
@@ -79,7 +79,7 @@ def build_argparser():
                       required=True, type=str, metavar='"<path>"')
     args.add_argument('-i',
                       dest='input_source',
-                      help='Required. Path to an image, video file or a numeric camera ID.',
+                      help='Required. Input to process.',
```
vladimir-dudnik, line 82:
```suggestion
                      help='Required. Path to video file or numeric camera ID.',
```
is that correct statement?

demos/python_demos/text_spotting_demo/README.md
Wovchena: Depends on having ffmpeg plugin for opencv on Windows. So my change is to avoid specifying what sources are supported because it is up to opencv

demos/python_demos/text_spotting_demo/text_spotting_demo.py
Wovchena: My change is suggested by https://github.com/opencv/open_model_zoo/blob/develop/CONTRIBUTING.md#demo

demos/README.md
```diff
@@ -248,6 +248,7 @@ list above.
 ### Get Ready for Running the Demo Applications on Windows*
 
 Before running compiled binary files, make sure your application can find the Inference Engine and OpenCV libraries.
+Optionally download OpenCV community FFmpeg plugin. There is a downloader script in the OpenVINO package: `opencv\ffmpeg-download.ps1`.
```
eizamaliev, line 251: > Optionally download OpenCV community FFmpeg plugin
Not clear, why user should do this.
> There is a downloader script in the OpenVINO package: `opencv\ffmpeg-download.ps1`.
I think, it will be better `<INSTALL_DIR>\opencv\ffmpeg-download.ps1`

Wovchena: I can list only two reasons: one wants opencv to use ffmpeg backend (obvious, no need to clarify). And one wants to read an image with VideoCapture which is actually sub point for the first reason. But this approach is discouraged by Alexander Alekhin and a common module for reading inputs should use imread() instead. Following Alexander Alekhin we don't what to encourage others to use ffmpeg with videocapture to read a single image.

I don't know how to clarify why a user may want to do this.

> I think, it will be better <INSTALL_DIR>\opencv\ffmpeg-download.ps1

It's clear because of `in the OpenVINO package`.

IRDonch: > I can list only two reasons: one wants opencv to use ffmpeg backend (obvious, no need to clarify).

That is indeed obvious, but that just shifts the question: why would the user want to use the FFmpeg backend? For someone who is not well-acquainted with the various OpenCV backends, it won't be clear whether they want it or not. 

> It's clear because of in the OpenVINO package.

That may be so, but the document already uses `<INSTALL_DIR>` in various places. It would be consistent to use it here, too.

demos/python_demos/text_spotting_demo/README.md
```diff
@@ -22,11 +22,11 @@ Second model is Text Recognition Encoder that takes `text_features` as input and
 
 Third model is Text Recognition Decoder that takes `encoded text` from Text Recognition Encoder ,`previous symbol` and `hidden state`. On the first step special `Start Of Sequence (SOS)` symbol and zero `hidden state` are fed to Text Recognition Decoder. The decoder produces `symbols distribution`, `current hidden state` each step until `End Of Sequence (EOS)` symbol is generated.
 
-As input, the demo application takes a path to a single image file, a video file or a numeric ID of a web camera specified with a command-line argument `-i`.
+Examples of valid inputs to specify with a command-line argument `-i` are a video file or a numeric ID of a web camera.
```
IRDonch, line 25:
```suggestion
Examples of valid inputs to specify with a command-line argument `-i` are a path to video file or a numeric ID of a web camera.
```

demos/python_demos/text_spotting_demo/README.md
```diff
@@ -22,11 +22,11 @@ Second model is Text Recognition Encoder that takes `text_features` as input and
...
 
 The demo workflow is the following:
 
-1. The demo application reads image/video frames one by one, resizes them to fit into the input image blob of the network (`im_data`).
+1. The demo application reads frames from the povided input, resizes them to fit into the input image blob of the network (`im_data`).
```
IRDonch, line 29:
```suggestion
1. The demo application reads frames from the provided input, resizes them to fit into the input image blob of the network (`im_data`).
```

demos/README.md
Wovchena: As i said, I don't know how to clarify why a user may want to do this. Do you have an idea how to reformulate the note and clarify the usage of ffmpeg backend?

> That may be so, but the document already uses <INSTALL_DIR> in various places. It would be consistent to use it here, too.

Ok.

demos/python_demos/text_spotting_demo/README.md
Wovchena: Updated

PR #1205, opened 2020-06-08
AC: fix single class semantic segmentation adapter

eaidova:
resize does not work with boolean values


eaidova:
jenkins please retry a build

PR #1211, opened 2020-06-09
Update ci/requirements-*.txt using a prerelease build of OpenVINO 2020.4

IRDonch:
Don't bump the versions, just update the lists.


IRDonch:
@Wovchena I just added a workaround to make the YOLO Python demo work with the new version of OpenVINO, could you take a look?

PR #1213, opened 2020-06-09
AC: fix typos

druzhkov-paul:


eaidova:
@druzhkov-paul  not fully understand your idea to support single model like pipeline. is it impossible just add adapter if it required and use asl recognition or i3d model as example for configuration?

tools/accuracy_checker/dataset_definitions.yml
```diff
@@ -778,3 +779,18 @@ datasets:
       max_query_length: 64
       doc_stride: 128
       lower_case: true
+
+  - name: weld-porosity
+    data_source: weld-porosity/images
+    annotation_conversion:
+      converter: clip_weld_defect_recognition
```
eaidova, line 786: if it is something private, here no place for it...

eaidova:
Jenkins please retry a build

PR #1215, opened 2020-06-10
object_detection_demo_yolov3_async: Work around a bug in the GPU plugin

IRDonch:
Apparently, net.layers turns into an empty dict if you don't access it prior to loading the network to the device.


IRDonch:
Never mind, I just realized that this won't pass the tests unless I combine it with updating the requirements files.

PR #1218, opened 2020-06-10
AclNet - Model description

sathishkuttan:
Add model description for sound classification with AclNet

Co-authored by: Paulo Lopez Meyer <paulo.lopez.meyer@intel.com>
Signed-off-by: Sathish Kuttan <sathish.k.kuttan@intel.com>

### DIFF
--- a/models/public/aclnet/aclnet.md
+++ b/models/public/aclnet/aclnet.md
@@ -17,7 +17,7 @@ The model output for `AclNet` is the sound classifier output for the 53 differen
 | Metric            | Value         |
 |-------------------|---------------|
 | Type              | Classification|
-| GFLOPs            |            |
+| GFLOPs            | 1.4           |
 | MParams           | 2.7           |
 | Source framework  | PyTorch\*     |
 
--- a/models/public/aclnet/aclnet.md
+++ b/models/public/aclnet/aclnet.md
@@ -1,0 +1,47 @@
+# AclNet
+
+## Use Case and High-Level Description
+
+The `AclNet` model is designed to perform sound classification.
+The `AclNet` model is trained on a dataset of environmental sounds (DES-53).
+For details about the model, see this [paper](https://arxiv.org/pdf/1811.06669.pdf).
+
+The model input is a segment of PCM audio samples in [N, C, H, W] format with the shape [1, 1, 1, 16000].
+
+The model output for `AclNet` is the sound classifier output for the 53 different environmental sound classes from the DES-53 database.
+
+## Example
+
+## Specification
+
+| Metric            | Value         |
+|-------------------|---------------|
+| Type              | Classification|
+| MParams           | 2.7           |
+| Source framework  | PyTorch\*     |
+
+## Accuracy
+
+See this [publication](http://dcase.community/documents/workshop2019/proceedings/DCASE2019Workshop_Huang_52.pdf) and this [paper](https://arxiv.org/pdf/1811.06669.pdf).
+
+## Performance
+
+## Input
+
+Audio, name - `data`, shape - `1,1,1,16000`, format is `N,C,H,W` where:
+
+- `N` - number of samples
+- `C` - channel
+- `H` - height
+- `W` - width
+
+## Output
+
+Sound classifier according to DES-53 classes, name - `softmax`, shape - `1,53`, output data format is `N,C` where:
+
+- `N` - number of samples classified
+- `C` - Predicted softmax scores for each class in [0, 1] range
+
+## Legal Information
+
+The original model is distributed under [Apache License, Version 2.0](https://www.apache.org/licenses/LICENSE-2.0.html)
--- a/models/public/aclnet/aclnet.md
+++ b/models/public/aclnet/aclnet.md
@@ -1,4 +1,4 @@
-# AclNet
+# aclnet
 
 ## Use Case and High-Level Description
 
@@ -17,6 +17,7 @@ The model output for `AclNet` is the sound classifier output for the 53 differen
 | Metric            | Value         |
 |-------------------|---------------|
 | Type              | Classification|
+| GFLOPs            |            |
 | MParams           | 2.7           |
 | Source framework  | PyTorch\*     |
 
@@ -28,7 +29,18 @@ See this [publication](http://dcase.community/documents/workshop2019/proceedings
 
 ## Input
 
-Audio, name - `data`, shape - `1,1,1,16000`, format is `N,C,H,W` where:
+### Original Model
+
+Audio, name - `0`, shape - `1,1,1,16000`, format is `N,C,H,W` where:
+
+- `N` - number of samples
+- `C` - channel
+- `H` - height
+- `W` - width
+
+### Converted Model
+
+Audio, name - `0`, shape - `1,1,1,16000`, format is `N,C,H,W` where:
 
 - `N` - number of samples
 - `C` - channel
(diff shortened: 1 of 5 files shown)


eaidova:
Jenkins ok to test

models/public/aclnet/aclnet.md
```diff
@@ -0,0 +1,68 @@
...
+| Metric            | Value         |
+|-------------------|---------------|
+| Type              | Classification|
+| GFLOPs            |            |
+| MParams           | 2.7           |
+| Source framework  | PyTorch\*     |
+
+## Accuracy
+
+See this [publication](http://dcase.community/documents/workshop2019/proceedings/DCASE2019Workshop_Huang_52.pdf) and this [paper](https://arxiv.org/pdf/1811.06669.pdf).
```
eizamaliev, line 26: @sathishkuttan could you add link to dataset here?

eizamaliev:
cc @vladimir-dudnik

vladimir-dudnik:
Jenkins please retry a build

models/public/aclnet/model.yml
```diff
@@ -0,0 +1,37 @@
...
+  shape [1, 1, 1, 16000].
+
+  The model output for "AclNet" is the sound classifier output for the 53 different
+  environmental sound classes from the DES-53 database.
+task_type: audio_classification
+files:
+  - name: aclnet_des_53.onnx
+    size: 10972740
+    sha256: 33339e960c788907640e67a040ecc5355304a50eb14d22a02187e10c030bf402
+    source: https://download.01.org/opencv/public_models/062020/aclnet/15062020/aclnet_des_53_fp32.onnx
```
IRDonch, line 30: I'm confused here. If this is in `public_models`, then that suggests it's a model from a public source. But where is this source?

tools/downloader/common.py
```diff
@@ -51,6 +51,7 @@
 }
 KNOWN_TASK_TYPES = {
     'action_recognition',
+    'audio_classification',
```
IRDonch, line 54: Please update `README.txt` accordingly.

models/public/aclnet/model.yml
vladimir-dudnik: the model is similar to OTE models, trained at Intel but redistributed in source form. Potentially, training pipeline could be added later

IRDonch: OK, but then it should be in `models_contrib`, not `public_models`.

vladimir-dudnik: @eizamaliev could you please move model to models_contrib folder?

tools/downloader/common.py
eizamaliev: Done

models/public/aclnet/model.yml
eizamaliev: Done

tools/downloader/common.py
vladimir-dudnik: @eizamaliev it seems task type should be changed to `sound_classification` what is generally used name of such kind of tasks according to[ comment](https://github.com/opencv/open_model_zoo/pull/1258#pullrequestreview-448981525)

nnshah1: @sathishkuttan Can you recommend final naming of task? From our (DLStreamer) perspective we will describe it as audio_event_detection - but don't see a concern with it being listed here as sound / audio classification.
@vladimir-dudnik From a brief literature survey there doesn't seem to be as much standardized naming as for CV - but it was only a brief look

nnshah1: https://arxiv.org/pdf/1811.06669v1.pdf 

Uses Audio Classification

sathishkuttan: Let's go with "sound classification" which is what we have used in the internal efforts (JIRA, internal repo, model directory in download.01, model description etc.)

nnshah1: @sathishkuttan  - will you update the files accordingly? Currently looks like audio classification

sathishkuttan: @nnshal1 I authored only the documentation file (`aclnet.md`) where I used sound classification consistent with my comment. Other files were added by @eizamaliev I think.
@eizamaliev can you please change the naming accordingly?

nnshah1: @sathishkuttan @vladimir-dudnik @IRDonch 

Changes made - please resolve thread

vladimir-dudnik:
Jenkins please retry a build

nnshah1:
Jenkins please retry a build

nnshah1:
@vladimir-dudnik I looked at the broken links (first of all its pretty cool that you check for broken links) - and they seem unrelated to the audio model - so I assume this is a temporary / intermittent CI failure - if not and there is something I can help with - please let me know.

vladimir-dudnik:
@nnshah1 do not worry, we do separate real issues from just temporal network instability. This will not stop us from merge PR.

models/public/aclnet/model.yml
vladimir-dudnik: @IRDonch seems your concern was addressed

nnshah1:
@vladimir-dudnik Thanks for pushing on this.

models/public/aclnet/aclnet.md
```diff
@@ -0,0 +1,68 @@
+# aclnet
+
+## Use Case and High-Level Description
+
+The `AclNet` model is designed to perform sound classification.
+The `AclNet` model is trained on a dataset of environmental sounds (DES-53).
+For details about the model, see this [paper](https://arxiv.org/pdf/1811.06669.pdf).
+
+The model input is a segment of PCM audio samples in [N, C, H, W] format with the shape [1, 1, 1, 16000].
```
IRDonch, line 9: > H, W

That doesn't make sense - audio doesn't have height or width.

models/public/aclnet/aclnet.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Performance
+
+## Input
+
+### Original Model
+
+Audio, name - `0`, shape - `1,1,1,16000`, format is `N,C,H,W` where:
+
+- `N` - number of samples
```
IRDonch, line 36: Shouldn't this be the batch size? It seems like it's the last dimension that is the number of samples.

models/public/aclnet/aclnet.md
```diff
@@ -0,0 +1,68 @@
...
+### Converted Model
+
+Sound classifier according to DES-53 classes, name - `203`, shape - `1,53`, output data format is `N,C` where:
+
+- `N` - number of samples classified
+- `C` - Predicted softmax scores for each class in [0, 1] range
+
+## Legal Information
+
+The original model is distributed under [Apache License, Version 2.0](https://www.apache.org/licenses/LICENSE-2.0.html)
```
IRDonch, line 68:
```suggestion
The original model is distributed under [Apache License, Version 2.0](https://www.apache.org/licenses/LICENSE-2.0.html).
```

tools/downloader/common.py
```diff
@@ -51,6 +51,7 @@
 }
 KNOWN_TASK_TYPES = {
     'action_recognition',
+    'sound_classification',
```
IRDonch, line 54: Please move this so that the list remains sorted (ditto for the `README`).

sathishkuttan: Done.

models/public/aclnet/aclnet.md
sathishkuttan: Done

models/public/aclnet/aclnet.md
sathishkuttan: Yes batch size is more accurate. changed it to avoid confusion.

models/public/aclnet/aclnet.md
paulolm: This nomenclature follows the adopted convention format of NCHW typically used in deep learning frameworks, where the input data is represented in a 2-dimensional matrix of size Height x Width (H x W),  i.e. in the case of raw audio H = 1 and W = 16,000. Of course, strictly speaking, audio signals do not have height and width, but the input data representation following the mentioned convention facilitates the handling of the raw audio as the input data represented by a vector of size H x W = 1 x 16,000 (1 second of audio).

IRDonch: I understand what the convention is, but it only makes sense for images. We can pretend that the audio segment is an image of height 1, but that's an extra abstraction that doesn't seem to gain us anything - it just obscures the meanings of the dimensions. IMO, we should call things what they are and describe the format as [N, C, 1, L], where L is the length of the segment (or the number of samples in the segment).

paulolm: I agree that the use of [N,C,1,L] format will avoid further confusion. Additionally, L could be of variable length, not necessarily  a fixed size of 16,000 audio sample points.

sathishkuttan: @IRDonch: Pushed an update to `aclnet.md` according to the above comment from @paulolm.

PR #1222, opened 2020-06-11
PerformanceMetrics class for Python demos

AnthonyQuantum:


demos/python_demos/common/helpers.py
```diff
@@ -0,0 +1,21 @@
...
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+"""
+
+import cv2
+
+def put_highlighted_text(frame, message, position, font_face, font_scale, color, thickness):
+    cv2.putText(frame, message, position, font_face, font_scale, (255, 255, 255), thickness + 1) # white border
+    cv2.putText(frame, message, position, font_face, font_scale, color, thickness)
```
Wovchena, line 21:
```suggestion
    cv2.putText(frame, message, position, font_face, font_scale, color, thickness)

```

demos/python_demos/common/performance_metrics.py
```diff
@@ -0,0 +1,86 @@
...
+ See the License for the specific language governing permissions and
+ limitations under the License.
+"""
+
+from time import perf_counter
+import cv2
+from helpers import put_highlighted_text
+
+
+class Sum:
```
Wovchena, line 22: This is not a sum it is more like a statistic. Or you can move its attributes into PerformanceMetrics to avoid giving a name.

demos/python_demos/common/performance_metrics.py
```diff
@@ -0,0 +1,86 @@
...
+    def __init__(self):
+        self.latency = 0.0
+        self.time = 0.0
```
Wovchena, line 25: You use the word time for both duration and time points. It's better to distinguish them.

demos/python_demos/common/performance_metrics.py
```diff
@@ -0,0 +1,86 @@
...
+        self.frame_count = 0
+
+
+class PerformanceMetrics:
+    def __init__(self, time_window=1.0):    # defines the length of the timespan used for 'current fps' calculation,
+        self.time_window_size = time_window # set to '1 second' by default
+        self.last_moving_sum = Sum()
+        self.current_moving_sum = Sum()
+        self.total_sum = Sum()
+        self.last_update_time = perf_counter()
```
Wovchena, line 35: You should set`last_update_time` with actual time point in `update()`. Otherwise it becomes important when `PerformanceMetrics` is created.

demos/python_demos/common/performance_metrics.py
```diff
@@ -0,0 +1,86 @@
...
+    
+    def update(self, last_request_start_time):
+        current_time = perf_counter()
+
+        last_request_latency = current_time - last_request_start_time
+        self.current_moving_sum.latency += last_request_latency
+        self.current_moving_sum.frame_count += 1
+        if self.current_moving_sum.frame_count == 1 and self.total_sum.frame_count == 0:
+            self.last_moving_sum.latency = last_request_latency
+            self.last_moving_sum.frame_count = 1
```
Wovchena, line 45: It makes sense to set `last_moving_sum` and `total_sum` and skip `current_moving_sum` for the first frame, so `last_moving_sum` and `total_sum` will always be updated simultaneously.

demos/python_demos/common/performance_metrics.py
```diff
@@ -0,0 +1,86 @@
...
+
+        if current_time - self.last_update_time > self.time_window_size:
+            self.current_moving_sum.time = current_time - self.last_update_time
```
Wovchena, line 48: Write the value to `self.last_moving_sum` directly.

That will make `self.current_moving_sum.time` unused. `get_total_fps()` doesn't count because `self.current_moving_sum.time` is always zero there. You need a different approach to report total_fps or update `self.current_moving_sum.time` on each `update()`.

demos/python_demos/common/performance_metrics.py
```diff
@@ -0,0 +1,86 @@
...
+            self.last_update_time = current_time
+            self.last_moving_sum = self.current_moving_sum
+            self.current_moving_sum = Sum()
+            self.total_sum.latency += self.last_moving_sum.latency
+            self.total_sum.time += self.last_moving_sum.time
+            self.total_sum.frame_count += self.last_moving_sum.frame_count
+
+    def get_current_fps(self):
```
Wovchena, line 56: There is no reason to separate fps and latency: if you needed fps and latency to be kept separate you wouldn't count them in a single `update()` method.

It even makes sense to merge `update()` and `show_current_*()` methods (which actually should have been called `put_moving_*()`). I can't come up with a scenario when you don't call `show_current_*()` after `update()` except for that use case with Tab in your implementation. Which has a problem: it does't draw latency for a while after switching to min_latency mode with nireq set big enough. It will be better if you find a way to track previous and current mode more accurate in the demo.

demos/python_demos/common/performance_metrics.py
```diff
@@ -0,0 +1,86 @@
...
+        return self.last_moving_sum.frame_count / self.last_moving_sum.time
+
+    def get_current_latency(self):
+        return (1000 * self.last_moving_sum.latency) / self.last_moving_sum.frame_count
```
Wovchena, line 60: You don't need parentheses.

demos/python_demos/common/performance_metrics.py
```diff
@@ -0,0 +1,86 @@
...
+
+    def get_total_fps(self):
+        return (self.total_sum.frame_count + self.current_moving_sum.frame_count) / \
+               (self.total_sum.time + self.current_moving_sum.time)
+        
+    def get_total_latency(self):
+        return (1000 * (self.total_sum.latency + self.current_moving_sum.latency)) / \
```
Wovchena, line 67: [The preferred way of wrapping long lines is by using Python's implied line continuation inside parentheses.](https://www.python.org/dev/peps/pep-0008/#maximum-line-length)

demos/python_demos/common/performance_metrics.py
```diff
@@ -0,0 +1,86 @@
...
+               (self.total_sum.frame_count + self.current_moving_sum.frame_count)
+
+    def show_current_fps(self, frame, position=(15, 50), font_scale=0.75, color=(200, 10, 10), thickness=2):
+        if self.last_moving_sum.frame_count > 1:
```
Wovchena, line 71: After setting `last_update_time` with actual time point in `update()` it will become more correct to check that the denominator isn't 0 in `get_current_fps()` instead of numerator. The similar problem is reported by tests.

That's `get_*()` methods responsibilities to check if they can compute values. They can return None if they can't compute.

demos/python_demos/common/performance_metrics.py
```diff
@@ -0,0 +1,86 @@
...
+            put_highlighted_text(frame, "Latency: {:.1f} ms".format(self.get_current_latency()),
+                                 position, cv2.FONT_HERSHEY_COMPLEX, font_scale, color, thickness)
+    
+    def print_total_fps(self, log):
+        if self.total_sum.frame_count + self.current_moving_sum.frame_count > 1:
+            log.info("FPS: {:.1f}".format(self.get_total_fps()))
+
+    def print_total_latency(self, log):
+        if self.total_sum.frame_count + self.current_moving_sum.frame_count != 0:
+            log.info("Latency: {:.1f} ms".format(self.get_total_latency()))
```
Wovchena, line 86: [`print()` fits better.](https://docs.python.org/3/howto/logging.html#when-to-use-logging) You even name the methods `print*()`.

demos/python_demos/object_detection_demo_yolov3_async/object_detection_demo_yolov3_async.py
```diff
@@ -497,22 +479,13 @@ def main():
...
-        log.info("Latency: {:.1f} ms".format((mode_info[mode_value].latency_sum / \
-                                             mode_info[mode_value].frames_count) * 1e3))
+        log.info("Mode: {}".format(mode_name.name))
+        mode_metrics[mode_name].print_total_fps(log)
+        mode_metrics[mode_name].print_total_latency(log)
     print(presenter.reportMeans())
 
-    for exec_net in exec_nets.values():
-        await_requests_completion(exec_net.requests)
-
```
Wovchena, line 515: I think this part is usefull.

demos/python_demos/common/performance_metrics.py
Wovchena: Actually demos that use console can't make use of `PerformanceMetrics`. Usually we don't want to keep moving average for them at all.

AnthonyQuantum: > Actually demos that use console can't make use of `PerformanceMetrics`. Usually we don't want to keep moving average for them at all.

@Wovchena, do you mean that since console demos can't use PerformanceMetrics, we can safely merge update() and  put_moving_*()? Or do you mean that I should add another type of metric specially for console demos?

Wovchena: > do you mean that since console demos can't use PerformanceMetrics, we can safely merge update() and put_moving_*()?

Yes.

But if you see a nice way to modify/split `PerformanceMetrics` to use it in console only demos, we can consider that approach. For example a base class for console demos and its derivative to report moving stats.

On the other hand metric computing for console demos shall be very simple and it can be easier to use bare `perf_counter()` without introducing any helper.

The idea behind merging methods is to prevent forgetting to call one of the methods as they are always expected to be called one after another in common use case. And to prevent calling that methods far form each other like
```
performanceMetrics.update()
postprocess_detections()
performanceMetrics.put_moving_*()
```

AnthonyQuantum:
@Wovchena 

I noticed that the demo with -nireq  > 1 does not close properly very often since I started using the new Python API.

It does not depend on whether the demo has this part or not:
```
for exec_net in exec_nets.values():
        await_requests_completion(exec_net.requests)
```
sys.exit(), exit(), quit() do not stop the demo, only sys._exit(0) works, which is unsafe because it does not call cleanup handlers.

Might be a Python API bug.

Wovchena:
I can reproduce the problem. Will you submit a bug?

AnthonyQuantum:
> I can reproduce the problem. Will you submit a bug?

Sure

demos/python_demos/common/performance_metrics.py
```diff
@@ -0,0 +1,96 @@
...
+            self.current_moving_statistic.duration = current_time - self.last_update_time
+            self.current_moving_statistic.frame_count += 1
+        else:
+            # for 1st frame
+            self.last_moving_statistic.latency = last_request_latency
+            self.last_moving_statistic.frame_count = 1
+            self.total_statistic.latency = last_request_latency
+            self.total_statistic.frame_count = 1
+
+        if current_time - self.last_update_time > self.time_window_size:
```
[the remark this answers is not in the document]
Wovchena: @AnthonyQuantum, this one is left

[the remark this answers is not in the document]
AnthonyQuantum: > I think you should also check if new data was collected because if the pipeline works slower than `time_window_size` you will store 0 in `last_moving_statistic`.

Even when the pipeline works slower than `time_window_size`, there is at least one value in `last_moving_statistic`.
It's because every time `update()` is called, one value is collected: `last_request_latency = current_time - last_request_start_time`.

I checked it on smallest time_window sizes like 0.0 and 0.1 and it still has at least one value.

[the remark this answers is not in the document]
Wovchena: Ok

demos/python_demos/common/performance_metrics.py
```diff
@@ -0,0 +1,97 @@
...
+            self.last_moving_statistic.latency = self.current_moving_statistic.latency
+            self.last_moving_statistic.period = current_time - self.last_update_time
+            self.last_moving_statistic.frame_count = self.current_moving_statistic.frame_count
+            self.total_statistic.latency += self.last_moving_statistic.latency
+            self.total_statistic.period += self.last_moving_statistic.period
+            self.total_statistic.frame_count += self.last_moving_statistic.frame_count
+            self.current_moving_statistic.latency = 0.0
+            self.current_moving_statistic.frame_count = 0
```
IRDonch, line 63: Shouldn't `current_moving_statistic.period` be reset too? And if yes, then it would be simpler to just do `self.current_moving_statistic = Statistic()`.

demos/python_demos/common/performance_metrics.py
```diff
@@ -0,0 +1,97 @@
...
+class Statistic:
+    def __init__(self):
+        self.latency = 0.0
+        self.period = 0.0
+        self.frame_count = 0
+
+
+class PerformanceMetrics:
+    def __init__(self, time_window=1.0):    # defines the length of the timespan used for 'current fps' calculation,
+        self.time_window_size = time_window # set to '1 second' by default
```
IRDonch, line 31: It's not clear what this comment applies to (in fact, I initially thought there were two separate comments). I would suggest to move it to its own line and add a subject to the sentence. And also, remove the second part, because it's redundant.

demos/python_demos/common/performance_metrics.py
```diff
@@ -0,0 +1,97 @@
...
+        self.last_moving_statistic = Statistic()
+        self.current_moving_statistic = Statistic()
+        self.total_statistic = Statistic()
+        self.last_update_time = None
+    
+    def update(self, last_request_start_time, frame, position=(15, 30),
+               font_scale=0.75, color=(200, 10, 10), thickness=2):
+        current_time = perf_counter()
+        if self.last_update_time is None:
```
IRDonch, line 40: Perhaps it would be better to initialize `last_update_time` in the constructor?

AnthonyQuantum: I think it's better when the length of the period does not depend on when PerformanceMetrics object was initialized
https://github.com/opencv/open_model_zoo/pull/1222#discussion_r440455123

demos/python_demos/common/performance_metrics.py
```diff
@@ -0,0 +1,97 @@
...
+        if self.total_statistic.frame_count != 0:
+            self.current_moving_statistic.latency += last_request_latency
+            self.current_moving_statistic.period = current_time - self.last_update_time
+            self.current_moving_statistic.frame_count += 1
+        else:
+            # for 1st frame
+            self.last_moving_statistic.latency = last_request_latency
+            self.last_moving_statistic.frame_count = 1
+            self.total_statistic.latency = last_request_latency
+            self.total_statistic.frame_count = 1
```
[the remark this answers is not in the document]
AnthonyQuantum: It's for cases when the input consists of only one image.

>And it seems that this code sets frame_count to 1 while period is still 0 - doesn't that result in infinite FPS?

Yes, it results in infinite FPS, but it's handled in `get_last` and `get_total`. They return `None` in this case. So, the incorrect value is not shown.


based on Wovchena's comment (https://github.com/opencv/open_model_zoo/pull/1222#pullrequestreview-430229272)
>The special case is latency and fps for the first frame. Handling it is important because the input may consist of one image and some demos allow setting 0 as an argument for waitKey(). My opinion is that moving latency should be initialized for the first time after the first frame is encountered but not after time_window_size passes. There are two options for fps: report 1/latency or report nothing. I like the second option because in that case we don't mix approaches used to compute fps (1/latency for the first and frame frequency for the rest).

demos/python_demos/common/performance_metrics.py
IRDonch: OK, but I think the code could be rearranged to make the logic clearer. It seems to me that `self.last_update_time is None` is effectively equivalent to `self.total_statistic.frame_count == 0`, so the `self.last_update_time` initialization can be moved to the `else` branch of the `if self.total_statistic.frame_count != 0` statement.

demos/python_demos/common/performance_metrics.py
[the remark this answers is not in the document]
IRDonch: I guess I understand the motivation, but won't this skew the results in the non-single-image case? E.g., after the second frame is processed, the period will only cover the processing of the second frame, leading to the FPS being reported as being higher than it actually is. This error will decrease as more frames are processed, but this logic still feels wrong.

Maybe it would be better to record the start time in the constructor after all? (Or to have a dedicated `start()` method.) That way, no special first-frame logic would be required.

Also, regarding the demos with `waitKey(0)`... Assuming we do a similar thing to #1229 for the Python demos, won't demos no longer need to do that? I was under the impression that for images, we'll just tell the users to use `-loop_input` and have the demos process the same image over and over. That way, we'll likely get more accurate performance metrics, as well as remove the need for any special logic. @Wovchena Was that your plan or am I misunderstanding?

demos/python_demos/common/performance_metrics.py
```diff
@@ -0,0 +1,99 @@
...
+
+        if current_time - self.last_update_time > self.time_window_size:
+            self.last_moving_statistic.latency = self.current_moving_statistic.latency
+            self.last_moving_statistic.period = current_time - self.last_update_time
+            self.last_moving_statistic.frame_count = self.current_moving_statistic.frame_count
```
IRDonch, line 64:
```suggestion
            self.last_moving_statistic = self.current_moving_statistic
```

demos/python_demos/common/performance_metrics.py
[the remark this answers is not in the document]
Wovchena: > I guess I understand the motivation, but won't this skew the results in the non-single-image case? E.g., after the second frame is processed, the period will only cover the processing of the second frame, leading to the FPS being reported as being higher than it actually is. This error will decrease as more frames are processed, but this logic still feels wrong.

I think the question is where to begin counting. I assume you expect timer to be started at pipeline start for FPS. But FPS is frame frequency, and I believe we should start counting when the first frame is shown. It's similar to reporting FPS for the first frame - it's undefined because there is no previous frame.

> Also, regarding the demos with waitKey(0)... Assuming we do a similar thing to #1229 for the Python demos, won't demos no longer need to do that? I was under the impression that for images, we'll just tell the users to use -loop_input and have the demos process the same image over and over. That way, we'll likely get more accurate performance metrics, as well as remove the need for any special logic. @Wovchena Was that your plan or am I misunderstanding?

I'm not sure if we will completely remove the option to change `waitKey()` argument for now, but that makes seance. On the other hand we still need the implemented approach because the demo won't report metrics for the first `time_window` seconds otherwise.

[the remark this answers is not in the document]
IRDonch: > I assume you expect timer to be started at pipeline start for FPS. But FPS is frame frequency, and I believe we should start counting when the first frame is shown.

Yeah, that's fair enough.

I think the crux of the issue here is that the two stats we measure are properties of different things: the latency is the property of a frame, while the period is the interval _between_ frames. And so once N frames have been processed, we will naturally have N latencies, but only N-1 periods. Therefore, it is incorrect to divide by `frame_count` when calculating total averages, because no matter how you count frames, one of the averages will be wrong.

I see two reasonable ways to resolve this:

1. Just don't collect the latency for the first frame. Then we'll always have the same number of latencies and periods. This does mean that we won't be able to report the latency for the first `time_window` seconds. I don't think that's a huge problem, because the latency of the first frame is likely to be non-representative anyway (since the pipeline won't have stabilized yet). This approach has the advantage of code being simpler, and it makes sure that the stats the user sees are always calculated the same way (i.e., by averaging).

2. Instead of the block this comment thread is attached to, save the first frame's latency to a separate variable (e.g., self.first_frame_latency). In `get_total`, add `self.first_frame_latency` to the numerator of the latency, and 1 to the denominator. In `get_last`, check if `self.total_statistic.frame_count == 0`, and if so, return `(self.first_frame_latency, None)`. This is pretty ugly, but it avoids the error in the FPS calculation, and faithfully represents the fact that we calculate the latency differently for the first `time_window` seconds.

demos/python_demos/common/performance_metrics.py
```diff
@@ -0,0 +1,96 @@
...
+            self.current_moving_statistic.frame_count += 1
+        else:
+            # for the 1st frame
+            self.last_update_time = current_time
+            self.last_moving_statistic.latency = last_request_latency
+            self.last_moving_statistic.frame_count = 1
+            self.total_statistic.latency = last_request_latency
+            self.total_statistic.frame_count = 1
+
+        if current_time - self.last_update_time > self.time_window_size:
```
IRDonch, line 60: I think this `if` should be moved to the true branch of the preceding `if`, since this cannot happen on the first frame.

demos/python_demos/common/performance_metrics.py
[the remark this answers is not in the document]
Wovchena: I didn't notice that problem.

I prefer the first way of resolving.

PR #1225, opened 2020-06-12
Replace deprecated 'inputs' property in new demos and 'outputs' property of IRs

AnthonyQuantum:


demos/python_demos/multi_camera_multi_target_tracking/utils/network_wrappers.py
```diff
@@ -135,12 +135,12 @@ def __init__(self, ie, model_path, trg_classes, conf=.6,
         self.net = load_ie_model(ie, model_path, device, None, ext_path, num_reqs=self.max_reqs)
 
         required_input_keys = [{'im_info', 'im_data'}, {'im_data', 'im_info'}]
-        current_input_keys = set(self.net.inputs_info.keys())
+        current_input_keys = set(self.net.input_info)
```
IRDonch, line 138:
```suggestion
        current_input_keys = self.net.input_info.keys()
```

AnthonyQuantum:
> can you please explain why this out_blob.buffer is now needed? thanks

It's a requirement of the new version of IE Python API.
They've probably added this in order to align C++ and Python APIs openvinotoolkit/openvino#637.

PR #1232, opened 2020-06-16
Enable quantization for faster_rcnn_resnet*

IRDonch:
To do that, implement custom quantization config support in Model Quantizer.

### DIFF
--- a/tools/downloader/quantizer.py
+++ b/tools/downloader/quantizer.py
@@ -25,25 +25,38 @@ import tempfile
 
 from pathlib import Path
 
+import yaml
+
 import common
 
 OMZ_ROOT = Path(__file__).resolve().parents[2]
 
+DEFAULT_POT_CONFIG_BASE = {
+    'compression': {
+        'algorithms': [
+            {
+                'name': 'DefaultQuantization',
+                'params': {
+                    'preset': 'performance',
+                    'stat_subset_size': 300,
+                },
+            },
+        ],
+    },
+}
+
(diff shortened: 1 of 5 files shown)


eizamaliev:
@IRDonc, as far as I remember, configs used in POT have json format (correct me if I am wrong). May be we also should use json for consistency, what do you think?

IRDonch:
@eizamaliev Since these configs aren't directly passed to POT, we're not required to use JSON, and YAML is more readable, so I think it makes sense to use it.

eizamaliev:
@IRDonch, I agree with you about readability and others. But my point, that it would be easier to add config if it was created for POT. It is more about model contributing experience.
Just think about it, I do not insist on it.

tools/downloader/quantizer.py
```diff
@@ -27,23 +27,36 @@
 
 import common
 
+import yaml
```
eizamaliev, line 30: Please, move it to rest core imports.

IRDonch: I did mess up the import order (which I have now fixed), but note that `yaml` is a third-party module, so I'm not grouping it with the standard library modules.

IRDonch:
> But my point, that it would be easier to add config if it was created for POT. It is more about model contributing experience.

It is a fair point, but I'm just following the maxim that code is read more times than it's written, and therefore it's more important to optimize for readability than write-ability.

Also, it's very easy to convert POT's JSON files to YAML with a script (which is what I did to get these 🙂). Maybe I'll clean this script up and add it to the repo.

PR #1235, opened 2020-06-17
object_detection_demo_ssd_async: list pelee-coco

Wovchena:


demos/python_demos/object_detection_demo_ssd_async/models.lst
```diff
@@ -5,7 +5,8 @@ face-detection-retail-????
 pedestrian-and-vehicle-detector-adas-????
 pedestrian-detection-adas-????
 pedestrian-detection-adas-binary-????
-person-detection-retail-0013
+pelee-coco
+person-detection-retail-????
```
IRDonch, line 9: Are you sure -0002 is compatible? Looking at the code, it seems that the demo only works with 3-element `im_info` inputs, not 6-element ones (and `person-detection-retail-0002` has a 6-element one).

Wovchena: You are right. Reverted.

PR #1203, opened 2020-06-05
Update python object_detection_ssd_async demo

druzhkov-paul:

PR #1236, opened 2020-06-17
Redesign the pipeline of object_detection_demo_ssd_async.py

druzhkov-paul:
Change log:

- Demo pipeline is redesigned to align with the one for `object_detection_demo_yolov3_async.py`.
- Additional abstraction levels for `Model` and `Detector` are introduced.
- Additional output parser is added to enable models from OTE including `face-detection-????` and `person-detection-????` ones.

Notes:

- I was not caring too much about `ModeInfo` related code because of #1222. But tried to make collected runtime statistics valid and precise.
- Output parsers might be merged into (subclasses of) `Detector`, though left it as is for now.
- I don't know if you have some particular plans on refactoring of object detection demos, but IMHO we have too much of them now. I'd think about merging them together or redesigning those to use the same core. The hope is `Model` and `Detector` classes can be useful and help to reduce the amount of a boilerplate code needed.
- This PR can be considered as a replacement of #1203.


druzhkov-paul:
@Wovchena @AnthonyQuantum please take a look.

IRDonch:
> This PR can be considered as a replacement of #1203.

If #1203 is obsolete, then please close it.

Wovchena:
@druzhkov-paul, can you provide performance numbers for your setup. Specifically I expect that the old version will complete processing a video faster than the new version in latency mode. If this is fine for you, we can keep that pipeline.
The general IE recommendation for callbacks is to keep callbacks as small as possible. That is why I ask you to move postprocessing out of the callback. That should slightly boost performance.

druzhkov-paul:
> @druzhkov-paul, can you provide performance numbers for your setup. Specifically I expect that the old version will complete processing a video faster than the new version in latency mode. If this is fine for you, we can keep that pipeline.
> The general IE recommendation for callbacks is to keep callbacks as small as possible. That is why I ask you to move postprocessing out of the callback. That should slightly boost performance.

Thanks, this is a valuable note. I run a few CPU demo runs with `person-detection-retail-0013` model using

1. the version of `object_detection_demo_ssd_async` from `develop` branch is `sync` mode.
1. this PR's version with postprocessing happening inside a callback.
1. this PR's version (see the last commit) with postprocessing happening in the main loop.

Here are the results showing how much time (in seconds) the demo spent in the `while` loop.

| Run | Baseline | Update (callback) | Update (main loop) |
| ---- | --------- | ------------------- | --------------------- |
| 1 | 20.388 | 26.510 | 25.739 |
| 2 | 20.364 | 24.336 | 24.996 |
| 3 | 20.356 | 24.926 | 25.302 |
| 4 | 20.287 | 24.396 | 26.047 |
| 5 | 20.334 | 25.604 | 25.047 |

So I don't see any difference with postprocessing being performed inside the callback vs the main loop (given postprocessing required for this model), but both of these implementations lag behind the reference implementation for some reason.

Wovchena:
Thank you!
The hypothesis is that the difference between the baseline and the updated version is caused by GIL slowing down switches between main and callback threads. We don't observe that problem for C++.
Thanks to your comparison of the last to columns I just realized there is no difference to expect with `-nireq 1` which latency mode sets. There is just no much resources for the pipeline to orchestrate. But with `-nireq 2` we might have observed a difference (and maybe with heavier postprocessing) because it looks that while a callback isn't completed IE stream is not allowed to process next inferRequest. Just if you were curious.

It will take me a while to finish the review.

demos/python_demos/object_detection_demo_ssd_async/object_detection_demo_ssd_async.py
```diff
@@ -15,241 +15,581 @@
...
+    def postprocess(self, outputs, meta):
+        return outputs
+
+    def request_callback(self, status, callback_args):
+        request, frame_id, frame_meta = callback_args
+        try:
+            if status != 0:
+                raise RuntimeError('Infer Request has returned status code {}'.format(status))
+            raw_outputs = {key: blob.buffer for key, blob in request.output_blobs.items()}
+            outputs = None
```
Wovchena, line 182: That var has become pointless with moving of callback.

demos/python_demos/object_detection_demo_ssd_async/object_detection_demo_ssd_async.py
```diff
@@ -15,241 +15,581 @@
...
+        else:
+            try:
+                parser = MultipleOutputParser(net.outputs, bboxes, scores, labels)
+                log.info('Use MultipleOutputParser')
+            except ValueError:
+                h, w = net.input_info[image_blob_name].input_data.shape[2:]
+                parser = OTEParser([w, h], net.outputs)
+                log.info('Use OTEParser')
+            return parser
+        raise RuntimeError('Unsupported model outputs')
```
Wovchena, line 256:
```suggestion
```
This is never reached.

demos/python_demos/object_detection_demo_ssd_async/object_detection_demo_ssd_async.py
```diff
@@ -15,241 +15,581 @@
...
+            self.labels_layer = self.find_layer_by_name(labels_layer, all_outputs)
+            log.info('Use output "{}" as the one containing labels of detected objects.'
+                     .format(self.labels_layer))
+        except ValueError:
+            log.warning('Suitable layer for "{}" output is not found. '
+                        'Treating detector as class agnostic with output label {}'
+                        .format(labels_layer, default_label))
+            self.labels_layer = None
+            self.default_label = default_label
+            log.info('Treating detector as a class-agnostic one with output label {}.'.format(self.default_label))
```
Wovchena, line 342:
```suggestion
```
It has been reported in warning.

demos/python_demos/object_detection_demo_ssd_async/object_detection_demo_ssd_async.py
```diff
@@ -15,241 +15,581 @@
...
-        return SingleOutputPostprocessor(output_blob)
-    elif len(net.outputs) >= 3:
-        def find_layer(name, all_outputs):
-            suitable_layers = [layer_name for layer_name in all_outputs if name in layer_name]
-            if not suitable_layers:
-                raise ValueError('Suitable layer for "{}" output is not found'.format(name))
 
-            if len(suitable_layers) > 1:
-                raise ValueError('More than 1 layer matched to "{}" output'.format(name))
+def get_plugin_configs(args):
```
Wovchena, line 398: It's cleaner to pass ` args.num_streams`, `args.device` and `args.number_threads` explicitly as arguments.

demos/python_demos/object_detection_demo_ssd_async/object_detection_demo_ssd_async.py
```diff
@@ -15,241 +15,581 @@
...
+        return inputs_dict
+
+    def preprocess(self, inputs):
+        meta = {}
+        return inputs, meta
+
+    def postprocess(self, outputs, meta):
+        return outputs
+
+    def request_callback(self, status, callback_args):
```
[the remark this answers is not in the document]
druzhkov-paul: Renamed.

demos/python_demos/object_detection_demo_ssd_async/object_detection_demo_ssd_async.py
druzhkov-paul: Removed.

demos/python_demos/object_detection_demo_ssd_async/object_detection_demo_ssd_async.py
druzhkov-paul: Changed the logic a bit here.

Wovchena: `else` up there seems to be redundant.

druzhkov-paul: Thanks. Moved some checks to `SingleOutputParser` itself, made parser selection more consistent.

Wovchena:
Jenkins please retry a build

druzhkov-paul:
Jenkins please retry a build

demos/python_demos/object_detection_demo_ssd_async/object_detection_demo_ssd_async.py
```diff
@@ -15,239 +15,593 @@
...
+        ]
+        if not filter_outputs:
+            raise ValueError('Suitable output with bounding boxes is not found')
+        if len(filter_outputs) > 1:
+            raise ValueError('More than 1 candidate for output with bounding boxes.')
+        return filter_outputs[0]
+
+    @staticmethod
+    def find_layer_by_name(name, all_outputs):
+        return MultipleOutputParser.find_layer(name, all_outputs)
```
IRDonch, line 375: This suggests to me that `find_layer_by_name` should just be a free function.

demos/python_demos/object_detection_demo_ssd_async/object_detection_demo_ssd_async.py
```diff
@@ -15,239 +15,593 @@
...
+        return suitable_layers[0]
+
+    def __call__(self, outputs):
+        bboxes = outputs[self.bboxes_layer][0]
+        scores = outputs[self.scores_layer][0]
+        labels = outputs[self.labels_layer][0]
+        return [Detector.Detection(*bbox, score, label) for label, score, bbox in zip(labels, scores, bboxes)]
+
+
+class OTEParser:
```
IRDonch, line 343: Why is this called `OTEParser`? Presumably OTE stands for OpenVINO training extensions, but surely not all OTE models have this form of output. I would suggest giving it a more descriptive name.

demos/python_demos/object_detection_demo_ssd_async/object_detection_demo_ssd_async.py
```diff
@@ -15,239 +15,593 @@
...
+    config_user_specified = {}
+    config_min_latency = {}
 
-            if len(suitable_layers) > 1:
-                raise ValueError('More than 1 layer matched to "{}" output'.format(name))
+    devices_nstreams = {}
+    if num_streams:
+        devices_nstreams = {device: num_streams for device in ['CPU', 'GPU'] if device in device} \
+                           if num_streams.isdigit() \
+                           else dict([device.split(':') for device in num_streams.split(',')])
```
IRDonch, line 418:
```suggestion
                           else dict(device.split(':', 1) for device in num_streams.split(','))
```

demos/python_demos/object_detection_demo_ssd_async/object_detection_demo_ssd_async.py
```diff
@@ -15,239 +15,593 @@
...
+          and not exceptions:
+        if next_frame_id_to_show in completed_request_results:
+            frame_meta, raw_outputs = completed_request_results.pop(next_frame_id_to_show)
+            objects = exec_nets[mode].postprocess(raw_outputs, frame_meta)
+
+            frame = frame_meta['frame']
+            start_time = frame_meta['start_time']
+
+            if len(objects) and args.raw_output_message:
+                log.info(' Class ID | Confidence | XMIN | YMIN | XMAX | YMAX | COLOR ')
```
IRDonch, line 506:
```suggestion
                log.info(' Class ID | Confidence | XMIN | YMIN | XMAX | YMAX | COLOR ')
```

Why does the color need to be here? That's not an intrinsic property of the object.

demos/python_demos/object_detection_demo_ssd_async/object_detection_demo_ssd_async.py
```diff
@@ -15,239 +15,593 @@
...
+        request, frame_id, frame_meta = callback_args
+        try:
+            if status != 0:
+                raise RuntimeError('Infer Request has returned status code {}'.format(status))
+            raw_outputs = {key: blob.buffer for key, blob in request.output_blobs.items()}
+            self.completed_request_results[frame_id] = (frame_meta, raw_outputs)
+            self.empty_requests.append(request)
+        except Exception as e:
+            self.callback_exceptions.append(e)
+        self.event.set()
```
IRDonch, line 186: Something seems to be missing. Here, the event is set, however it's not _reset_ anywhere. Therefore, as soon as the first submitted `InferRequest` completes, the event will be set and all subsequent `await_any` calls will return immediately. That can't be right.

demos/python_demos/object_detection_demo_ssd_async/object_detection_demo_ssd_async.py
```diff
@@ -15,239 +15,593 @@
...
-    next_request_id = 1
+    log.info('Loading network...')
+    completed_request_results = {}
+    modes = cycle(Modes)
+    prev_mode = mode = next(modes)
+    log.info('Using {} mode'.format(mode.name))
+    mode_info = {mode: ModeInfo()}
+    exceptions = []
+
+    exec_nets = {
```
IRDonch, line 462: It seems confusing to name this `exec_net`, when the objects it contains are not `ExecutableNetwork`s.

demos/python_demos/object_detection_demo_ssd_async/object_detection_demo_ssd_async.py
```diff
@@ -15,239 +15,593 @@
...
+        Modes.USER_SPECIFIED:
+            Detector(ie, args.model, device=args.device, plugin_config=config_user_specified,
+                     results=completed_request_results, max_num_requests=args.num_infer_requests,
+                     labels_map=labels_map, keep_aspect_ratio_resize=args.keep_aspect_ratio,
+                     caught_exceptions=exceptions),
+        Modes.MIN_LATENCY:
+            Detector(ie, args.model, device=args.device.split(':')[-1].split(',')[0], plugin_config=config_min_latency,
```
IRDonch, line 469: Why is the device different in this case?

demos/python_demos/object_detection_demo_ssd_async/models.lst
```diff
@@ -1,11 +1,13 @@
 # This file can be used with the --list option of the model downloader.
+face-detection-????
 face-detection-adas-????
 face-detection-adas-binary-????
 face-detection-retail-????
 pedestrian-and-vehicle-detector-adas-????
 pedestrian-detection-adas-????
 pedestrian-detection-adas-binary-????
 pelee-coco
+person-detection-asl-????
```
IRDonch, line 10: Your PR description says that this also adds support for `person-detection-????`, so shouldn't it also be added?

demos/python_demos/object_detection_demo_ssd_async/object_detection_demo_ssd_async.py
```diff
@@ -15,239 +15,593 @@
...
+            log.warning('Following layers are not supported '
+                        'by the CPU plugin:\n\t{}'.format(unsupported_info))
+            log.warning('Please try to specify cpu extensions library path.')
+            raise ValueError('Some of the layers are not supported.')
+
+    def unify_inputs(self, inputs):
+        if not isinstance(inputs, dict):
+            if not isinstance(inputs, (list, tuple)):
+                inputs = [inputs]
+            if len(inputs) == 1:
```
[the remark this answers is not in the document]
druzhkov-paul: I'd say that this's a common pattern for different DL libs to get a list of tensors as an input. This is just a convenience wrapper that allows to simplify the stuff for a single input net case. On the other hand, it's not clear how to map a list of an arbitrary length to actual network inputs (`input_info` is unordered).

demos/python_demos/object_detection_demo_ssd_async/object_detection_demo_ssd_async.py
druzhkov-paul: We're on a way to move all (maybe some legacy models will be kept as is, but the majority of models are affected) detection models from OTE (OpenVINO Training Extensions) to a common engine (PyTorch -> Object Detection). Face and person detection models are already there and more to come.

demos/python_demos/object_detection_demo_ssd_async/object_detection_demo_ssd_async.py
druzhkov-paul: Just preserved the format as used in yolov3 demo. I could only guess that it could be useful to map printed detections and the ones drawn, when the caption is not readable. Don't mind to remove it here, though.

demos/python_demos/object_detection_demo_ssd_async/object_detection_demo_ssd_async.py
druzhkov-paul: yolov3 demo seems to be affected as well.

demos/python_demos/object_detection_demo_ssd_async/object_detection_demo_ssd_async.py
druzhkov-paul: Once again this came from a yolov3 demo. I believe that the idea is to use only primary device for a hetero regime to reduce the latency.

demos/python_demos/object_detection_demo_ssd_async/models.lst
druzhkov-paul: Yes, thanks. I removed those for a while. As long as they are available now, readded them to the list.

demos/python_demos/object_detection_demo_ssd_async/object_detection_demo_ssd_async.py
IRDonch: Sure, but I have no reason to believe that you will stay with that engine forever (or that the engine will always produce outputs in this format). At the very least, if this format comes from that engine, then the parser should be named after the engine, not OTE.

demos/python_demos/object_detection_demo_ssd_async/object_detection_demo_ssd_async.py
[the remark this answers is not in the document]
IRDonch: > On the other hand, it's not clear how to map a list of an arbitrary length to actual network inputs (input_info is unordered).

That's my point exactly. If you can't handle a list of length other than 1, why accept a list in the first place?

demos/python_demos/object_detection_demo_ssd_async/object_detection_demo_ssd_async.py
```diff
@@ -15,239 +15,589 @@
...
+        self.event.set()
+
+    def __call__(self, inputs, id, meta):
+        request = self.empty_requests.popleft()
+        inputs = self.unify_inputs(inputs)
+        inputs, preprocessing_meta = self.preprocess(inputs)
+        meta.update(preprocessing_meta)
+        request.set_completion_callback(py_callback=self.inference_completion_callback,
+                                        py_data=(request, id, meta))
+        self.event.clear()
```
IRDonch, line 194: I don't think this is correct. Consider this sequence of events:

1. Main thread starts one InputRequest.
2. It completes in the background thread, setting the event.
3. Main thread starts the next InputRequest (this clears the event).
4. Main thread calls `await_any`. This ignores the first request's completion (because the event was cleared) and waits for the second request.

demos/python_demos/object_detection_demo_ssd_async/object_detection_demo_ssd_async.py
IRDonch: @AnthonyQuantum It looks like you originally implemented this logic in the YOLO demo, could you work on a fix for that? (Or alternatively, could you explain how I'm wrong and the code is correct? 🙂)

demos/python_demos/object_detection_demo_ssd_async/object_detection_demo_ssd_async.py
druzhkov-paul: See no issues here. We are calling `await_any` right after the check that all Infer Requests are busy. In your situation we won't call `await_any` at all, which is fine.

druzhkov-paul:
@IRDonch please review one more time

demos/python_demos/object_detection_demo_ssd_async/object_detection_demo_ssd_async.py
IRDonch: OK, I see what you mean. It does seem safe (if a bit unobvious).

druzhkov-paul:
As long as this PR has been approved, don't you mind merging it?

demos/python_demos/object_detection_demo_ssd_async/object_detection_demo_ssd_async.py
AnthonyQuantum: Thank you for pointing that out. It's already fixed in ssd demo, as I checked, and here's the fix for yolov3 demo #1313

PR #1238, opened 2020-06-17
Bump versions in ci/requirements-*.txt

IRDonch:


IRDonch:
Looks like torchvision 0.6.0 is broken with Python 3.5. :-/

PR #1247, opened 2020-06-19
tools/downloader: implement download resumption

IRDonch:
Now when we retry a download, we try to resume from where we left off, and if that fails, restart from the beginning.

Currently, this is only implemented for plain HTTP sources (i.e., not Google Drive).


IRDonch:
@eizamaliev Ping?

PR #1249, opened 2020-06-21
Corrected filtering on basis of iou

pra-dan:
The IOU filtering process aims to remove duplicate predictions of the same **objects** and ends up choosing the box with highest probablity. To validate this Pull Request, kindly run prediction on this 
sample image at the confidence threshold of 0.1. 

Using the earlier code, you will notice that though the person was detected, but on comparing its IOU with cycle, led to ignoring it in the final results. This is because it makes no sense to compare a cycle with a man or any different object.


vladimir-dudnik:
@pra-dan please target your PR to `develop` branch, according to requirements listed in [CONTRIBUTING.md](https://github.com/opencv/open_model_zoo/blob/master/CONTRIBUTING.md#pull-request-requirements)

pra-dan:
Done as suggested :)

PR #1250, opened 2020-06-21
Corrected filtering on basis of IoU

pra-dan:
The IOU filtering process aims to remove duplicate predictions of the same **objects** and ends up choosing the box with highest probablity. To validate this Pull Request, kindly run prediction on this 
[sample image](https://unsplash.com/photos/Tzz4XrrdPUE) at the confidence threshold of 0.1. 

Using the earlier code, you will notice that though the person was detected, but on comparing its IOU with cycle, led to ignoring it in the final results. This is because it makes no sense to compare a cycle with a man or any different object.

### DIFF
--- a/demos/python_demos/object_detection_demo_yolov3_async/object_detection_demo_yolov3_async.py
+++ b/demos/python_demos/object_detection_demo_yolov3_async/object_detection_demo_yolov3_async.py
@@ -245,6 +245,9 @@ def filter_objects(objects, iou_threshold, prob_threshold):
         if objects[i]['confidence'] == 0:
             continue
         for j in range(i + 1, len(objects)):
+            # We perform IOU only on objects of same class 
+            if(objects[i]['class_id'] != objects[j]['class_id']): continue
+
             if intersection_over_union(objects[i], objects[j]) > iou_threshold:
                 objects[j]['confidence'] = 0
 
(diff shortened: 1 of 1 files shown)


demos/python_demos/object_detection_demo_yolov3_async/object_detection_demo_yolov3_async.py
```diff
@@ -245,6 +245,9 @@ def filter_objects(objects, iou_threshold, prob_threshold):
         if objects[i]['confidence'] == 0:
             continue
         for j in range(i + 1, len(objects)):
+            # We perform IOU only on objects of same class 
+            if(objects[i]['class_id'] != objects[j]['class_id']): continue
```
eizamaliev, line 249:
```suggestion
            if objects[i]['class_id'] != objects[j]['class_id']: 
                continue
```

eizamaliev:
Hi @pra-dan, thank you for contribution.
You are right about it, but it can cause another problem - duplication of detection with another similar class.

pra-dan:
This is something that can be controlled by choosing an appropriate value of `iou_threshold`. 

If, two objects of same class (say dog1 and dog2) are far away and compared, their IoU will certainly be low and causes no issue. But if both the dogs are sitting close, it is possible that their IoU will be high; this can be mitigated by an appropriate value for the threshold.

eizamaliev:
That's not what I meant. 
Sometimes (e.g. in road scenario) one object has several detection boxes with different, but close, classes (e.g. `car` and `truck`). In that case this algorithm would leave both box. I've already had such situations, so I want warn you about it. Sometimes it could be fixed by increasing `threshold` value.

IRDonch:
Jenkins: test this please

PR #1258, opened 2020-06-23
Audio classification demo

eizamaliev:


demos/python_demos/audio_classification_demo/audio_classification_demo.py
```diff
@@ -0,0 +1,214 @@
...
+    log.info("Starting inference")
+    outputs = []
+    clips = 0
+    infer_time = []
+    for idx, chunk in enumerate(audio.chunks(length, hop, num_chunks=batch_size)):
+        if len(chunk.shape) != len(input_shape):
+            chunk = np.reshape(chunk, newshape=input_shape)
+        infer_start_time = datetime.now()
+        output = exec_net.infer(inputs={input_blob: chunk})
+        infer_time.append(datetime.now() - infer_start_time)
```
Wovchena, line 191: Compute sum instead of storing each value

demos/python_demos/audio_classification_demo/audio_classification_demo.py
```diff
@@ -0,0 +1,214 @@
...
+            end_time = ((idx*batch_size + batch)*hop + length) / audio.framerate
+            outputs.append(data)
+            label = np.argmax(data)
+            log.info("[{:.2f}:{:.2f}] - {:s}: {:.2f}%".format(start_time, end_time,
+                                                              labels[label] if labels else "Class {}".format(label),
+                                                              data[label] * 100))
+
+    if clips == 0:
+        log.error("Audio too short for inference by that model")
+        sys.exit(1)
```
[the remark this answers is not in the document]
eizamaliev: Thanks, I missed it.

demos/python_demos/audio_classification_demo/audio_classification_demo.py
```diff
@@ -0,0 +1,214 @@
...
+    args.add_argument('-m', "--model", type=str, required=True,
+                      help="Required. Path to an .xml file with a trained model.")
+    args.add_argument("-l", "--cpu_extension", type=str, default=None,
+                      help="Optional. Required for CPU custom layers. Absolute path to a shared library with "
+                           "the kernels implementations.")
+    args.add_argument("-d", "--device", type=str, default="CPU",
+                      help="Optional. Specify the target device to infer on; CPU, GPU, FPGA, HDDL or MYRIAD is"
+                           " acceptable. The sample will look for a suitable plugin for device specified. "
+                           "Default value is CPU")
+    args.add_argument('--labels', type=str, default=None,
```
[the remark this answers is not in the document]
eizamaliev: It depends on which which model you use (mostly which dataset was used)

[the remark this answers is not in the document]
Wovchena: aclnet from models.lst

demos/python_demos/audio_classification_demo/audio_classification_demo.py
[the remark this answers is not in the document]
Wovchena: Still main function exits and other functions raise exceptions.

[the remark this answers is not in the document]
eizamaliev: I cant' agree with you: `sys.exit()` in `main` makes code more clearer. On the other hand, replacing `raise` in other function would't be appropriate. Only option is to replace `exit` in `main`. Nevertheless, most demos use this approach - `exit` for `main`, `raise` for others, and I think it is correct.

demos/python_demos/audio_classification_demo/audio_classification_demo.py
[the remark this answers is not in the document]
Wovchena: OK, as discussed.

demos/python_demos/audio_classification_demo/audio_classification_demo.py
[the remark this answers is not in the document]
Wovchena: OK, as discussed.

demos/python_demos/audio_classification_demo/audio_classification_demo.py
```diff
@@ -0,0 +1,216 @@
...
+        params = wav.getparams()
+        data = wav.readframes(params.nframes)
+        if sampwidth_types.get(params.sampwidth):
+            data = np.frombuffer(data, dtype=sampwidth_types[params.sampwidth])
+        else:
+            raise RuntimeError("Couldn't process file {}: unsupported sample width {}"
+                               .format(str(file), params.sampwidth))
+        data = np.reshape(data, (params.nframes, params.nchannels))
+        if as_float:
+            data = data / sampwidth_max[params.sampwidth]
```
IRDonch, line 124: Would this work correctly for both 8-bit and 16-bit wave files? It seems that for 8-bit ones it would produce values in the [0;1] range, while for 16-bit files it would produce values in the [-1;1] range. But the network could only accept one of those.

demos/python_demos/audio_classification_demo/audio_classification_demo.py
```diff
@@ -0,0 +1,216 @@
...
+
+    labels = []
+    if args.labels:
+        with open(args.labels, "r") as file:
+            labels = [l.rstrip() for l in file.readlines()]
+
+    batch_size, channels, _, length = input_shape
+
+    audio = AudioSource(args.input, channels=channels, samplerate=args.samplerate)
+    audio.load()
```
IRDonch, line 177: Is there a reason to separate `load` from `__init__` if you just invoke them one after the other?

demos/python_demos/audio_classification_demo/audio_classification_demo.py
```diff
@@ -0,0 +1,216 @@
...
+        log.error("Wrong value for '-ol/--overlap' argument - overlapping more than clip length")
+        sys.exit(1)
+
+    log.info("Starting inference")
+    outputs = []
+    clips = 0
+    infer_time = 0
+    for idx, chunk in enumerate(audio.chunks(length, hop, num_chunks=batch_size)):
+        if len(chunk.shape) != len(input_shape):
+            chunk = np.reshape(chunk, newshape=input_shape)
```
IRDonch, line 190: I think you could just replace this with `chunk.shape = input_shape` and let NumPy figure it out.

demos/python_demos/audio_classification_demo/audio_classification_demo.py
```diff
@@ -0,0 +1,216 @@
...
+        output = exec_net.infer(inputs={input_blob: chunk})
+        infer_time += time.perf_counter() - infer_start_time
+        clips += batch_size
+        output = output[output_blob]
+        for batch, data in enumerate(output):
+            start_time = (idx*batch_size + batch)*hop / audio.samplerate
+            end_time = ((idx*batch_size + batch)*hop + length) / audio.samplerate
+            outputs.append(data)
+            label = np.argmax(data)
+            log.info("[{:.2f}:{:.2f}] - {:s}: {:.2f}%".format(start_time, end_time,
```
IRDonch, line 201:
```suggestion
            log.info("[{:.2f}-{:.2f}] - {:6.2%} - {:s}".format(start_time, end_time,
```

For times, you would usually show a range with a hyphen. A colon suggests minutes:seconds to me.

Also, putting the probability first and padding it makes output lines align better.

demos/python_demos/audio_classification_demo/audio_classification_demo.py
```diff
@@ -0,0 +1,216 @@
...
+                                                              labels[label] if labels else "Class {}".format(label),
+                                                              data[label] * 100))
+
+    if clips == 0:
+        log.error("Audio too short for inference by that model")
```
IRDonch, line 206: I'm not sure if that's the right approach. Maybe when you're breaking up the file into chunks, it would be better to take the tail (the unprocessed part at the end) and zero-extend it to the chunk size? Then we could always process the entire file (even if it's smaller than a chunk).

demos/python_demos/audio_classification_demo/audio_classification_demo.py
```diff
@@ -0,0 +1,216 @@
...
+        sys.exit(1)
+    total = np.mean(outputs, axis=0)
```
IRDonch, line 208: Does this produce something useful? I'd expect that averaging the classification results over different chunks would just result in nonsense.

demos/python_demos/audio_classification_demo/audio_classification_demo.py
```diff
@@ -0,0 +1,216 @@
...
+    def __init__(self, source, channels=2, samplerate=None):
+        self.source = source
+        self.samplerate = samplerate
+        self.channels = channels
+
+    def load(self):
+        samplerate, audio = read_wav(self.source, as_float=True)
+        audio = audio.T
+        if audio.shape[0] != self.channels:
+            raise RuntimeError("Audio has unsupported number of channels")
```
IRDonch, line 73: It would be useful to say what the supported number is.

demos/python_demos/audio_classification_demo/audio_classification_demo.py
```diff
@@ -0,0 +1,216 @@
...
+    exec_net = ie.load_network(network=net, device_name=args.device)
+
+    log.info("Preparing input")
+
+    labels = []
+    if args.labels:
+        with open(args.labels, "r") as file:
+            labels = [l.rstrip() for l in file.readlines()]
+
+    batch_size, channels, _, length = input_shape
```
IRDonch, line 174: I assume the third component is expected to be 1, in which case there should be a check for it.

demos/python_demos/audio_classification_demo/README.md
```diff
@@ -0,0 +1,60 @@
+# Audio Classification Python\* Demo
+
+Demo application for audio classification algorithm.
+
+## How It Works
+
+Upon the start-up the demo application reads command line parameters and loads a network to Inference engine. It uses only audiofiles in `wav` format. Audio could be converted to different samplerate using `-sr/--samplerate` option, if necessary for the model. After reading the audio, it is sliced into clips to fit model input (clips are allowed to overlap with `-ol/--overlap` option) and each clip is processed separately with its own prediction.
```
sathishkuttan, line 7: Suggest that the sample rate expected by the model, ie., 16KHz is stated here. And, state that if the wav file has any other sample rate, the `-sr` option could be used to convert to 16KHz.

demos/python_demos/audio_classification_demo/README.md
```diff
@@ -0,0 +1,60 @@
...
+                        Required. Input to process
+  -m MODEL, --model MODEL
+                        Required. Path to an .xml file with a trained model.
+  -l CPU_EXTENSION, --cpu_extension CPU_EXTENSION
+                        Optional. Required for CPU custom layers. Absolute
+                        path to a shared library with the kernels
+                        implementations.
+  -d DEVICE, --device DEVICE
+                        Optional. Specify the target device to infer on; CPU,
+                        GPU, FPGA, HDDL or MYRIAD is acceptable. The sample
```
sathishkuttan, line 33: Has this model been tested on MYRIAD?

demos/python_demos/audio_classification_demo/README.md
```diff
@@ -0,0 +1,60 @@
...
+                        will look for a suitable plugin for device specified.
+                        Default value is CPU
+  --labels LABELS       Optional. Labels mapping file
```
sathishkuttan, line 36: What is the format of this mapping file?
We have a JSON file that has the label mapping in the internal repository. Let me know if you would like to use it.

vladimir-dudnik: @sathishkuttan yes, I think mapping file could be useful to distribute with OMZ

demos/python_demos/audio_classification_demo/README.md
```diff
@@ -0,0 +1,60 @@
...
+  -sr SAMPLERATE, --samplerate SAMPLERATE
+                        Optional. Set samplerate for audio input
+  -ol OVERLAP, --overlap OVERLAP
+                        Optional. Set the overlapping between audio clip in
+                        samples or percent
+```
+Running the application with the empty list of options yields the usage message given above and an error message.
+You can use the following command to do inference on GPU with a pre-trained audio classification model and conversion input audio to samplerate of 16000:
```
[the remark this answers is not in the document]
vladimir-dudnik:
```suggestion
You can use the following command to do inference on GPU with a pre-trained sound classification model and conversion of input audio to samplerate of 16000:
```

demos/python_demos/audio_classification_demo/README.md
```diff
@@ -0,0 +1,60 @@
...
+                        Default value is CPU
+  --labels LABELS       Optional. Labels mapping file
+  -sr SAMPLERATE, --samplerate SAMPLERATE
+                        Optional. Set samplerate for audio input
+  -ol OVERLAP, --overlap OVERLAP
+                        Optional. Set the overlapping between audio clip in
+                        samples or percent
+```
+Running the application with the empty list of options yields the usage message given above and an error message.
+You can use the following command to do inference on GPU with a pre-trained audio classification model and conversion of input audio to samplerate of 16000:
```
vladimir-dudnik, line 44:
```suggestion
You can use the following command to do inference on GPU with a pre-trained sound classification model and conversion of input audio to samplerate of 16000:
```

demos/python_demos/audio_classification_demo/README.md
eizamaliev: @sathishkuttan it shoud be simple txt file each line of it represent class name. I could add support for JSON files, but I tried to make demo more universal w/o model binding.
@vladimir-dudnik IMHO such file should be distributed with model

demos/python_demos/audio_classification_demo/audio_classification_demo.py
eizamaliev: That's a good point! But it can be useful for detecting main sound in audio file.
@sathishkuttan have you any comments?

demos/python_demos/audio_classification_demo/audio_classification_demo.py
eizamaliev: I tried this, but it drastically reduces accuracy of classification. I can add it, if necessary.
@sathishkuttan please comment it.

demos/python_demos/audio_classification_demo/README.md
vladimir-dudnik: @eizamaliev ok, let's put labels file with model

demos/python_demos/audio_classification_demo/audio_classification_demo.py
IRDonch: How can it reduce accuracy? It doesn't affect any full chunks, and as for the tail chunk, you're currently not classifying it at all, so there's nothing to reduce.

demos/python_demos/sound_classification_demo/sound_classification_demo.py
```diff
@@ -0,0 +1,221 @@
...
+
+class AudioSource:
+    def __init__(self, source, channels=2, samplerate=None):
+        self.source = source
+        self.samplerate = samplerate
+        self.channels = channels
+        self.load()
+
+    def load(self):
+        samplerate, audio = read_wav(self.source, as_float=True)
```
IRDonch, line 71:
```suggestion
        self.samplerate = samplerate

        samplerate, audio = read_wav(source, as_float=True)
```

(`self.channels` can be replaced with `channels` below)

demos/python_demos/sound_classification_demo/sound_classification_demo.py
```diff
@@ -0,0 +1,221 @@
...
+    data = np.array([np.interp(x_new, x_old, channel) for channel in audio])
+
+    return data
+
+
+def read_wav(file, as_float=False):
+    sampwidth_types = {
+        1: np.uint8,
+        2: np.int16,
+        3: np.int32,
```
IRDonch, line 107: Does this work? I'd assume that a 24-bit type would be needed.

demos/python_demos/audio_classification_demo/audio_classification_demo.py
IRDonch: I see you have implemented the padding, so shouldn't this message be removed? The only way `clips == 0` could hold is if the audio file had no samples at all, which isn't an error - there's just nothing to classify.

demos/python_demos/sound_classification_demo/sound_classification_demo.py
```diff
@@ -0,0 +1,226 @@
...
+            self.samplerate = samplerate
+
+        self.audio = audio
+
+    def duration(self):
+        return self.audio.shape[1] / self.samplerate
+
+    def chunks(self, size, hop=None, num_chunks=1):
+        def get_clip(pos, size):
+            if pos > self.audio.shape[1]:
```
IRDonch, line 83: Isn't this an impossible condition? `get_clip` is only called from within the while loop whose condition is `pos < self.audio.shape[1]`.

eizamaliev: It could appear for batch size > 1. In that case will be generated blank audio chunk and it won't be printed as a result.

IRDonch: Right. I missed that.

IRDonch (approved):

PR #1264, opened 2020-06-24
Develop

Demon9497:
Early implementation of the retinaface


vladimir-dudnik:
@Demon9497 what is that? What are those 770 commits stands for? Could you please follow [CONTRIBUTING.md](https://github.com/opencv/open_model_zoo/blob/master/CONTRIBUTING.md) when create PR to Open Model Zoo?
And by the way, have you noticed [retinaface-resnet50](https://github.com/opencv/open_model_zoo/blob/develop/models/public/retinaface-resnet50/retinaface-resnet50.md) model, which should be available through Open Model Zoo in the nearest OpenVINO release?

IRDonch:
I don't know what this is supposed to be, but right now it's just a merge from `develop` to `master`, which is not something we need to do right now and doesn't have anything to do with RetinaFace. So I'm closing it.

PR #1266, opened 2020-06-25
Fix VAS face detection stage 2 issue 

hojaedev:
Fix issue for when no candidates are found for face detection stage 1.
Previous implementation assumed that inference generates at least on candidate .


eaidova:
Jenkins ok to test

PR #1268, opened 2020-06-25
Readd the bert-large-uncased-whole-word-masking-squad-fp32-onnx-0001

IRDonch:
Since we added resumption support to the Model Downloader, it should now be possible to download this.

This reverts commit 2cd6d66568f1d302c1543a652d383d58dc908409.

FYI: We don't really need to merge this, since we're likely to publish a new version of the models anyway. This PR is just to make sure that the resumption support works and that the model runs.


IRDonch:
Jenkins please retry a build

PR #1279, opened 2020-06-30
Modify NMS class to support overlap ratio calculation with different base area

hojaedev:


tools/accuracy_checker/accuracy_checker/postprocessor/nms.py
```diff
@@ -46,9 +46,8 @@ def set_scores(prediction, scores):
 def set_box_scores(prediction, scores):
     prediction.bbox_scores = scores
 
-
 class NMS(Postprocessor):
-    __provider__ = 'nms'
+    __provider__ = 'face_detection_nms'
```
eaidova, line 50: why does this class renamed?

tools/accuracy_checker/accuracy_checker/postprocessor/README.md
```diff
@@ -28,6 +28,7 @@ Accuracy Checker supports following set of postprocessors:
    - the preprocessing steps contains only two operations changing input image size, and the operations are `resize` and then `padding`.
 * `nms` - non-maximum suppression. Supported representations: `DetectionAnotation`, `DetectionPrediction`, `ActionDetectionAnnotation`, `ActionDetectionPrediction`.
   * `overlap` - overlap threshold for merging detections.
+  * `use_union_area` - boolean value to determine whether to use union area of two bounding boxes to calculate overlap. If set to false, overlap ratio is calculated with the minimum size of two bounding boxes as base area.
```
eaidova, line 31: on my view, better to have flag to enable your special behavior then turn off default, it is more clear for users, because initially in majority of cases assumes that nms is intersection over union and now it is difficult to understand to which behavior it leads if you have only config (so `use_min_area: True` will be more meaningful for your case)

eaidova:
Jenkins ok to test

tools/accuracy_checker/accuracy_checker/postprocessor/nms.py
eaidova: tests failed on the CI due to this renaming

eaidova:
Jenknins ok to test

eaidova:
Jenkins ok to test

hojaedev:
@eaidova , @vladimir-dudnik Could you also merge this patch? Thank you:)

PR #1280, opened 2020-07-01
Fix error in VAS Face detection refinement stage adapter

hojaedev:
Before, each single batch prediction was matched with all the candidates detected in the previous stage. Instead, each item in raw 
 (single batch prediction) should be matched with a single corresponding candidate.


eaidova:
Jenkins ok to test

PR #1282, opened 2020-07-02
AC: added loadmat function

Anna-Mironova:


tools/accuracy_checker/accuracy_checker/annotation_converters/aflw2000_3d.py
```diff
@@ -31,7 +19,7 @@ def convert(self, check_content=False, progress_callback=None, progress_interval
                     content_errors.append('{}: does not exists'.format(str(annotation_file)))
                 continue
 
-            image_info = scipy_io.loadmat(str(annotation_file))
+            image_info = loadmat(str(annotation_file))
```
eaidova, line 22: BTW, if it is internal function, I think it can work with Path instead of str

tools/accuracy_checker/accuracy_checker/utils.py
```diff
@@ -20,16 +20,32 @@
...
 
 from pathlib import Path
 from typing import Union
 from warnings import warn
 from collections.abc import MutableSet
 
+try:
+    from collections.abc import Sequence
+except ImportError:
+    from collections import Sequence
```
eaidova, line 36: if I right understand, try-except here need for compatibility with python 2.7, but our target is python3.5 and higher, probably we can use only first statement.

tools/accuracy_checker/accuracy_checker/utils.py
```diff
@@ -20,16 +20,32 @@
...
 import pickle
+import struct
+import sys
+import zlib
 from enum import Enum
 
 from pathlib import Path
 from typing import Union
 from warnings import warn
 from collections.abc import MutableSet
```
eaidova, line 31:
```suggestion
from collections.abc import MutableSet, Sequence
```

tools/accuracy_checker/accuracy_checker/utils.py
eaidova: and look above, we already has import from `collections.abc` :)

tools/accuracy_checker/accuracy_checker/utils.py
```diff
@@ -20,16 +20,32 @@
...
+
 import numpy as np
 import yaml
 
+try:
+    from itertools import izip
+    ispy2 = True
+except ImportError:
+    izip = zip
+    ispy2 = False
```
eaidova, line 47: here also compatibility with python2

tools/accuracy_checker/accuracy_checker/utils.py
```diff
@@ -518,3 +534,306 @@ def color_format(s, color=Color.PASSED):
...
+        end = len(b) == 0
+        if not end:
+            curpos = fd.tell()
+            fd.seek(curpos - 1)
+        return end
+
+    if isinstance(filename, str):
+        fd = open(filename, 'rb')
+    else:
+        fd = filename
```
eaidova, line 808: I think we are always put filename, not file descriptor, so else statement is not necessary

tools/accuracy_checker/accuracy_checker/utils.py
```diff
@@ -518,3 +522,307 @@ def color_format(s, color=Color.PASSED):
...
+        if mc == 'mxCELL_CLASS':
+            return self._read_cell_array(fd, endian, header)
+        if mc == 'mxSTRUCT_CLASS':
+            return self._read_struct_array(fd, endian, header)
+        if mc == 'mxOBJECT_CLASS':
+            raise ParseError('Object classes not supported')
+        if mc == 'mxFUNCTION_CLASS':
+            raise ParseError('Function classes not supported')
+        if mc == 'mxOPAQUE_CLASS':
+            raise ParseError('Anonymous function classes not supported')
```
eaidova, line 645: I think that we need format if-statements (before all unsupported, then all supported)

tools/accuracy_checker/accuracy_checker/utils.py
```diff
@@ -518,3 +522,307 @@ def color_format(s, color=Color.PASSED):
...
+    fd.seek(124)
+    tst_str = fd.read(4)
+    little_endian = (tst_str[2:4] == b'IM')
+    endian = ''
+    if (sys.byteorder == 'little' and little_endian) or (sys.byteorder == 'big' and not little_endian):
+        pass
+    elif sys.byteorder == 'little':
+        endian = '>'
+    else:
+        endian = '<'
```
eaidova, line 806: I think we can simplify this
```suggestion
    if sys.byteorder == 'little' and not little_endian:
        endian = '>'
    if sys.byteorder == 'big' and little_endian:
        endian = '<'
```

tools/accuracy_checker/accuracy_checker/utils.py
```diff
@@ -518,3 +522,307 @@ def color_format(s, color=Color.PASSED):
...
+    maj_ind = int(little_endian)
+    maj_val = tst_str[maj_ind]
+    if maj_val != 1:
+        raise ParseError('Can only read from Matlab level 5 MAT-files')
+    mdict = {}
+    reader = MatlabDataReader()
+    if meta:
```
eaidova, line 813: do we use meta flag?

PR #1285, opened 2020-07-02
Horizontal-text-detection

GalyaZalesskaya:


GalyaZalesskaya:
@Ilya-Krylov Hi, could you check my version of text_detection demo please?

demos/text_detection_demo/src/text_detection.cpp
```diff
@@ -100,6 +100,34 @@ std::vector<cv::RotatedRect> maskToBoxes(const cv::Mat &mask, float min_area, fl
     return bboxes;
   }
 
+std::vector<cv::RotatedRect> coordToBoxes(const std::vector<float> &coords, float min_area, float min_height,
+                                         cv::Size image_size) {
+    std::vector<cv::RotatedRect> bboxes;
+    int max_bbox_idx = coords.size() / 5;
+    float x_scale = image_size.width / 704.0;
```
eizamaliev, line 107: What does `704.0` magic number mean?

demos/text_detection_demo/text_detection_demo.hpp
```diff
@@ -51,8 +51,8 @@ DEFINE_string(m_td, "", text_detection_model_message);
 DEFINE_string(m_tr, "", text_recognition_model_message);
 DEFINE_string(m_tr_ss, "0123456789abcdefghijklmnopqrstuvwxyz", text_recognition_model_symbols_set_message);
 DEFINE_bool(cc, false, text_central_image_crop_message);
-DEFINE_int32(w_td, 1280, image_width_for_text_detection_model_message);
-DEFINE_int32(h_td, 768, image_height_for_text_detection_model_message);
+DEFINE_int32(w_td, 0, image_width_for_text_detection_model_message);
+DEFINE_int32(h_td, 0, image_height_for_text_detection_model_message);
```
eizamaliev, line 55: This changes could affect existing behavior. Could you explain it necessity of these changes?

demos/text_detection_demo/main.cpp
```diff
@@ -294,9 +294,9 @@ int main(int argc, char *argv[]) {
           std::cout << "text detection model inference (ms) (fps): "
                     << text_detection.time_elapsed() / text_detection.ncalls() << " "
                     << text_detection.ncalls() * 1000 / text_detection.time_elapsed() << std::endl;
-        if (std::fabs(text_detection_postproc_time) < std::numeric_limits<double>::epsilon()) {
-            throw std::logic_error("text_detection_postproc_time can't be equal to zero");
-        }
+        if (std::fabs(text_detection_postproc_time) < std::numeric_limits<double>::epsilon())
```
Ilya-Krylov, line 297: @GalyaZalesskaya why did you remove curly braces?

demos/text_detection_demo/src/text_detection.cpp
```diff
@@ -191,49 +219,73 @@ std::vector<cv::RotatedRect> postProcess(const InferenceEngine::BlobMap &blobs,
...
+        cv::Mat mask = decodeImageByJoin(cls_data, new_cls_data_shape,
+                       link_data, new_link_data_shape, cls_conf_threshold, link_conf_threshold);
+        rects = maskToBoxes(mask, static_cast<float>(kMinArea),
+                                                         static_cast<float>(kMinHeight), image_size);
+
+    }
+    else {
+        // PostProcessing for Horizontal Text Detection model
+        for (const auto& blob : blobs) {
+            if (blob.second->getTensorDesc().getDims()[0] == 100)
```
Ilya-Krylov, line 270: I think this value may vary from model to model.

Ilya-Krylov:
@GalyaZalesskaya format your code please

demos/text_detection_demo/src/text_detection.cpp
```diff
@@ -100,6 +100,34 @@ std::vector<cv::RotatedRect> maskToBoxes(const cv::Mat &mask, float min_area, fl
     return bboxes;
   }
 
+std::vector<cv::RotatedRect> coordToBoxes(const std::vector<float> &coords, float min_area, float min_height,
+                                         cv::Size image_size) {
+    std::vector<cv::RotatedRect> bboxes;
+    int max_bbox_idx = coords.size() / 5;
+    float x_scale = image_size.width / 704.0;
+    float y_scale = image_size.height / 704.0;
+    //TODO: Find a way not to hardcode the certain input_shape (704, 704)
```
Ilya-Krylov, line 109: Maybe add one more param to `postProcess` ?

demos/text_detection_demo/src/text_detection.cpp
```diff
@@ -100,6 +100,34 @@ std::vector<cv::RotatedRect> maskToBoxes(const cv::Mat &mask, float min_area, fl
     return bboxes;
   }
 
+std::vector<cv::RotatedRect> coordToBoxes(const std::vector<float> &coords, float min_area, float min_height,
+                                         cv::Size image_size) {
```
Ilya-Krylov, line 104: const ref

demos/text_detection_demo/src/text_detection.cpp
```diff
@@ -100,6 +100,34 @@ std::vector<cv::RotatedRect> maskToBoxes(const cv::Mat &mask, float min_area, fl
...
+    std::vector<cv::RotatedRect> bboxes;
+    int max_bbox_idx = coords.size() / 5;
+    float x_scale = image_size.width / 704.0;
+    float y_scale = image_size.height / 704.0;
+    //TODO: Find a way not to hardcode the certain input_shape (704, 704)
+    for (int i = 1; i < max_bbox_idx; i++) {
+
```
Ilya-Krylov, line 111: remove line

demos/text_detection_demo/src/text_detection.cpp
```diff
@@ -100,6 +100,34 @@ std::vector<cv::RotatedRect> maskToBoxes(const cv::Mat &mask, float min_area, fl
...
+        std::vector<float> prediction(coords.begin() + (i-1) * 5, coords.begin() + i*5);
```
Ilya-Krylov, line 112: spacing

demos/text_detection_demo/src/text_detection.cpp
```diff
@@ -100,6 +100,34 @@ std::vector<cv::RotatedRect> maskToBoxes(const cv::Mat &mask, float min_area, fl
...
+        float confidence = prediction[4];
+
+        if (confidence < std::numeric_limits<float>::epsilon())
+            break; //predictions are sorted the way that all insignificant boxes are grouped together
+
+        cv::Point2f center = cv::Point2f((prediction[0] + prediction[2])/2 * x_scale,
+                                        (prediction[1] + prediction[3])/2 * y_scale);
```
Ilya-Krylov, line 119: spacing and indent

demos/text_detection_demo/src/text_detection.cpp
```diff
@@ -100,6 +100,34 @@ std::vector<cv::RotatedRect> maskToBoxes(const cv::Mat &mask, float min_area, fl
...
   }
 
+std::vector<cv::RotatedRect> coordToBoxes(const std::vector<float> &coords, float min_area, float min_height,
+                                         cv::Size image_size) {
+    std::vector<cv::RotatedRect> bboxes;
+    int max_bbox_idx = coords.size() / 5;
+    float x_scale = image_size.width / 704.0;
+    float y_scale = image_size.height / 704.0;
+    //TODO: Find a way not to hardcode the certain input_shape (704, 704)
+    for (int i = 1; i < max_bbox_idx; i++) {
```
Ilya-Krylov, line 110: will you consider all boxes in such range?

demos/text_detection_demo/src/text_detection.cpp
```diff
@@ -100,6 +100,34 @@ std::vector<cv::RotatedRect> maskToBoxes(const cv::Mat &mask, float min_area, fl
...
+
+        std::vector<float> prediction(coords.begin() + (i-1) * 5, coords.begin() + i*5);
+        float confidence = prediction[4];
+
+        if (confidence < std::numeric_limits<float>::epsilon())
+            break; //predictions are sorted the way that all insignificant boxes are grouped together
```
Ilya-Krylov, line 116: spacing

@eizamaliev is there any cpplint that you are maybe using?

demos/text_detection_demo/src/text_detection.cpp
GalyaZalesskaya: It is input_shape for the model, so every picture is resized to (704, 704). I use x_scale and y_scale to resize predicted bounding boxes to the initial picture resolution.

demos/text_detection_demo/text_detection_demo.hpp
GalyaZalesskaya: I doubt, that this modification will affect existing behavior because of this [line](https://github.com/GalyaZalesskaya/open_model_zoo/blob/da991e8cb0b82c714adad571291725c779dead0a/demos/text_detection_demo/src/cnn.cpp#L29)  in the text detection model initialization.
It means that size of text detection model is calculated directly from model information ([line](https://github.com/GalyaZalesskaya/open_model_zoo/blob/da991e8cb0b82c714adad571291725c779dead0a/demos/text_detection_demo/src/cnn.cpp#L27)) if some special resolution is not set with parameters (new_input_resolution!=cv::Size()). 

It is needed not to pass internal architecture information (that user doesn't need to know) through parameters and make demo more generalized for models with different input resolution.
I checked that behavior for text-detection-0004 did not change, because (768, 1280) resolution is already define with model, so you don't need to keep default shape in this file.
Please say if my check is not enough.

demos/text_detection_demo/main.cpp
GalyaZalesskaya: For the sameness with lines [above](https://github.com/GalyaZalesskaya/open_model_zoo/blob/da991e8cb0b82c714adad571291725c779dead0a/demos/text_detection_demo/main.cpp#L294) and [below](https://github.com/GalyaZalesskaya/open_model_zoo/blob/da991e8cb0b82c714adad571291725c779dead0a/demos/text_detection_demo/main.cpp#L300) where one std::cout operator is not bracketed

demos/text_detection_demo/src/text_detection.cpp
GalyaZalesskaya: You are right - it depends on the batch size. But to be honest, I don't even need to calculate kClsOutputName in the next [line](https://github.com/GalyaZalesskaya/open_model_zoo/blob/da991e8cb0b82c714adad571291725c779dead0a/demos/text_detection_demo/src/text_detection.cpp#L271), but I did it for sameness with PixelLink post-processing and to enable this [check](https://github.com/GalyaZalesskaya/open_model_zoo/blob/da991e8cb0b82c714adad571291725c779dead0a/demos/text_detection_demo/src/text_detection.cpp#L275) if passed text detection model is not supported. 
Probably I should remove or rewrite it

demos/text_detection_demo/src/text_detection.cpp
GalyaZalesskaya: This max_bbox_idx value is equal to the max number of prediction per image (in the horizontal-text-detection model it is 100). Some of them are insignificant, so i filter it out [here](https://github.com/GalyaZalesskaya/open_model_zoo/blob/da991e8cb0b82c714adad571291725c779dead0a/demos/text_detection_demo/src/text_detection.cpp#L115).
If I misunderstood the question, please say

GalyaZalesskaya:
@eaidova Hi, could you please review my accuracy checker changes to enable "focused text metrics" (hmean, precision, recall) calculation with "class_agnostic_detection" adapter

GalyaZalesskaya:
@Ilya-Krylov Hi, I've finished accuracy checker modifications!

tools/accuracy_checker/accuracy_checker/metrics/text_detection.py
```diff
@@ -20,6 +20,7 @@
 from .metric import PerImageEvaluationMetric
 from ..config import BoolField, NumberField
 from ..representation import TextDetectionPrediction, TextDetectionAnnotation
+from ..representation import DetectionPrediction
```
eaidova, line 23:
```suggestion
from ..representation import TextDetectionPrediction, TextDetectionAnnotation, DetectionPrediction
```

tools/accuracy_checker/accuracy_checker/metrics/text_detection.py
```diff
@@ -150,7 +151,13 @@ def configure(self):
 
     def update(self, annotation, prediction):
         gt_rects = list(map(rect_from_points, annotation.boxes))
-        prediction_rects = list(map(rect_from_points, prediction.boxes))
+        if 'TextDetectionPrediction' in str(type(prediction)):
```
eaidova, line 154: please use `isinstance` for type checking

Also I have alternative suggestion, you can just add boxes property in DetectionPrediction type for align behaviour between 2 types and hide boxes calculation implementation. For me, it can be more useful, because if we will need to extend support with another prediction type or introduce new metric, we will not care how to find boxes

tools/accuracy_checker/configs/horizontal-text-detection-0001.yml
```diff
@@ -0,0 +1,38 @@
...
+
+    launchers:
+      - framework: dlsdk
+        tags:
+          - FP32
+        model:   intel/horizontal-text-detection-0001/FP32/horizontal-text-detection-0001.xml
+        weights: intel/horizontal-text-detection-0001/FP32/horizontal-text-detection-0001.bin
+        adapter:
+            type: class_agnostic_detection
+            scale: 0.00141
```
eaidova, line 12: please add FP16 model config.

tools/accuracy_checker/configs/horizontal-text-detection-0001.yml
```diff
@@ -0,0 +1,38 @@
+models:
```
eaidova, line 1: @GalyaZalesskaya , we have important change connected with Accuracy Checker configs location, starting with openvino 2021.1 (it is target release for your model).
Please move this file to your model directory (where description stored) and rename it to `accuracy_check.yml`
see https://github.com/opencv/open_model_zoo/tree/develop/models/intel/text-detection-0004 as example.
in this place, you can create symlink (sometime we will need it for backward compatibility on infra side) or I can do it after model merge if we still will require this.

tools/accuracy_checker/accuracy_checker/metrics/text_detection.py
GalyaZalesskaya: Thank you for your quick and detailed response. I've fixed all your comments

tools/accuracy_checker/configs/horizontal-text-detection-0001.yml
```diff
@@ -0,0 +1 @@
+/home/kotik/open_model_zoo/models/intel/horizontal-text-detection-0001/accuracy_check.yml
```
eaidova, line 1: please use relative path for link creation :)

models/intel/horizontal-text-detection-0001/description/horizontal-text-detection-0001.md
```diff
@@ -0,0 +1,50 @@
...
+
+1. name: "input" , shape: [1x3x704x704] - An input image in the format [BxCxHxW],
+   where:
+
+    - B - batch size
+    - C - number of channels
+    - H - image height
+    - W - image width
+
+   Expected color order - RGB.
```
eaidova, line 34: all models in OMZ have BGR color order (even if original model has RGB input, it will be distributed as IR converted with `--reverse_input_channels`).
Do you have some specific reasons to stay input in RGB (e.g. MO can not convert it with this option or any other issues)?

models/intel/horizontal-text-detection-0001/description/horizontal-text-detection-0001.md
```diff
@@ -0,0 +1,50 @@
...
+![](./horizontal-text-detection-0001.png)
+
+## Specification
+
+| Metric                                                        | Value                   |
+|---------------------------------------------------------------|-------------------------|
+| F-measure (harmonic mean of precision and recall on ICDAR2013)| 88.45%                 |
+| GFlops                                                        | 7.78                  |
+| MParams                                                       | 2.26                   |
+| Source framework                                              | PyTorch              |
```
eaidova, line 19:
```suggestion
| Source framework                                              | PyTorch\*              |
```

eaidova:
Jenkins ok to test

models/intel/horizontal-text-detection-0001/description/horizontal-text-detection-0001.md
GalyaZalesskaya: No, there are no specific reasons to keep it. But if I change order to BGR, should I fix any configs? And if the answer is yes, which files must be modified?

models/intel/horizontal-text-detection-0001/accuracy-check.yml
```diff
@@ -0,0 +1,47 @@
...
+
+    launchers:
+      - framework: dlsdk
+        tags:
+          - FP32
+        model:   intel/horizontal-text-detection-0001/FP32/horizontal-text-detection-0001.xml
+        weights: intel/horizontal-text-detection-0001/FP32/horizontal-text-detection-0001.bin
+        adapter:
+            type: class_agnostic_detection
+            scale: 0.0014204
```
eaidova, line 12: please fix linting error:
`
16:58:24    11:13     error    wrong indentation: expected 10 but found 12  (indentation)`
```suggestion
          type: class_agnostic_detection
          scale: 0.0014204
```

models/intel/horizontal-text-detection-0001/accuracy-check.yml
```diff
@@ -0,0 +1,47 @@
...
+
+      - framework: dlsdk
+        tags:
+          - FP16
+        model:   intel/horizontal-text-detection-0001/FP16/horizontal-text-detection-0001.xml
+        weights: intel/horizontal-text-detection-0001/FP16/horizontal-text-detection-0001.bin
+        adapter:
+            type: class_agnostic_detection
+            scale: 0.0014204
```
eaidova, line 21:
```suggestion
          type: class_agnostic_detection
          scale: 0.0014204
```

GalyaZalesskaya:
@eizamaliev  @IRDonch Hi, could you please approve my PR

models/intel/pedestrian-and-vehicle-detector-adas-0001/description/pedestrian-and-vehicle-detector-adas-0001.md
```diff
@@ -21,7 +21,7 @@ Pedestrian and vehicle detection network based on MobileNet v1.0 + SSD.
 | Source framework                | Caffe*                |
 
 Average Precision (AP) metric is described in: Mark Everingham et al.
-[The PASCAL Visual Object Classes (VOC) Challenge](http://host.robots.ox.ac.uk/pascal/VOC/pubs/everingham10.pdf).
+[The PASCAL Visual Object Classes (VOC) Challenge](https://link.springer.com/content/pdf/10.1007/s11263-009-0275-4.pdf).
```
IRDonch, line 24: This link doesn't work for anyone without a Springer subscription. Moreover, this change has nothing to do with the topic of the PR. Please revert it.

demos/text_detection_demo/main.cpp
```diff
@@ -33,15 +33,17 @@
...
-cv::Mat cropImage(const cv::Mat &image, const std::vector<cv::Point2f> &points, const cv::Size& target_size, int top_left_point_idx);
-void setLabel(cv::Mat& im, const std::string& label, const cv::Point & p);
+std::vector<cv::Point2f> floatPointsFromRotatedRect(
+    const cv::RotatedRect &rect);
+std::vector<cv::Point> boundedIntPointsFromRotatedRect(
+    const cv::RotatedRect &rect, const cv::Size &image_size);
+cv::Point topLeftPoint(const std::vector<cv::Point2f> &points, int *idx);
+cv::Mat cropImage(const cv::Mat &image, const std::vector<cv::Point2f> &points,
+                  const cv::Size &target_size, int top_left_point_idx);
+void setLabel(cv::Mat &im, const std::string &label, const cv::Point &p);
```
IRDonch, line 43: Please don't reformat unrelated code. I can barely tell what you actually changed.

models/intel/pedestrian-and-vehicle-detector-adas-0001/description/pedestrian-and-vehicle-detector-adas-0001.md
GalyaZalesskaya: I've changed it because paper was moved, so documentation tests failed because of broken link. But okey, I'll revert it

demos/text_detection_demo/src/text_detection.cpp
```diff
@@ -183,57 +211,83 @@ cv::Mat decodeImageByJoin(const std::vector<float> &cls_data, const std::vector<
...
-    std::vector<cv::RotatedRect> rects = maskToBoxes(mask, static_cast<float>(kMinArea),
-                                                     static_cast<float>(kMinHeight), image_size);
-
+    if (!kLocOutputName.empty() && !kClsOutputName.empty()) {
+        // PostProcessing for PixelLink Text Detection model
+    	auto link_shape = blobs.at(kLocOutputName)->getTensorDesc().getDims();
+    	size_t link_data_size = link_shape[0] * link_shape[1] * link_shape[2] * link_shape[3];
+    	InferenceEngine::LockedMemory<const void> locOutputMapped = InferenceEngine::as<
+   	    InferenceEngine::MemoryBlob>(blobs.at(kLocOutputName))->rmap();
+        float *link_data_pointer = locOutputMapped.as<float *>();
```
eizamaliev, line 238: It seems that formatting is broken. Please fix it

models/intel/horizontal-text-detection-0001/description/horizontal-text-detection-0001.md
```diff
@@ -0,0 +1,50 @@
...
+| GFlops                                                        | 7.78                  |
+| MParams                                                       | 2.26                   |
+| Source framework                                              | PyTorch\*              |
+
+
+## Performance
+
+## Inputs
+
+1. name: "input" , shape: [1x3x704x704] - An input image in the format [BxCxHxW],
```
eizamaliev, line 26:
```suggestion
1. Name: `input` , shape: [1x3x704x704] - An input image in the format [BxCxHxW],
```

models/intel/horizontal-text-detection-0001/description/horizontal-text-detection-0001.md
```diff
@@ -0,0 +1,50 @@
...
+    - B - batch size
+    - C - number of channels
+    - H - image height
+    - W - image width
+
+   Expected color order - BGR.
+
+## Outputs
+
+1. The "boxes" is a blob with shape: [N, 5], where N is the number of detected
```
eizamaliev, line 38:
```suggestion
1. The `boxes` is a blob with shape: [N, 5], where N is the number of detected
```

models/intel/horizontal-text-detection-0001/description/horizontal-text-detection-0001.md
```diff
@@ -0,0 +1,50 @@
...
+   bounding boxes. For each detection, the description has the format:
+   [`x_min`, `y_min`, `x_max`, `y_max`, `conf`],
+   where:
+    - (`x_min`, `y_min`) - coordinates of the top left bounding box corner
+    - (`x_max`, `y_max`) - coordinates of the bottom right bounding box corner.
+    - `conf` - confidence for the predicted class
+2. The "labels" is a blob with shape: [N], where N is the number of detected
```
eizamaliev, line 45:
```suggestion
2. The `labels` is a blob with shape: [N], where N is the number of detected
```

models/intel/horizontal-text-detection-0001/description/horizontal-text-detection-0001.md
```diff
@@ -0,0 +1,50 @@
...
+   bounding boxes. It contains `label` per each detected box.
```
eizamaliev, line 46: What does `label` mean?

models/intel/pedestrian-and-vehicle-detector-adas-0001/description/pedestrian-and-vehicle-detector-adas-0001.md
IRDonch: FYI, it wasn't moved, `host.robots.ox.ac.uk` is just a very unstable website which keeps going down. Just ignore it.

models/intel/horizontal-text-detection-0001/description/horizontal-text-detection-0001.md
GalyaZalesskaya: It means the class of the detected box, which is useful for general object detection models. But in this case predicted "label" is always "text"

models/intel/pedestrian-and-vehicle-detector-adas-0001/description/pedestrian-and-vehicle-detector-adas-0001.md
GalyaZalesskaya: Oh, I didn't know it. Will keep it in mind, thank you

demos/text_detection_demo/src/text_detection.cpp
```diff
@@ -183,57 +211,82 @@ cv::Mat decodeImageByJoin(const std::vector<float> &cls_data, const std::vector<
...
+std::vector<cv::RotatedRect> postProcess(const InferenceEngine::BlobMap &blobs,
+                                         const cv::Size &image_size, const cv::Size &input_shape,
                                          float cls_conf_threshold, float link_conf_threshold) {
     const int kMinArea = 300;
     const int kMinHeight = 10;
 
     std::string kLocOutputName;
     std::string kClsOutputName;
 
+    std::vector<cv::RotatedRect> rects;
```
IRDonch, line 223: I don't think this variable is needed. Just `return` the appropriate value at the end of each branch.

demos/text_detection_demo/src/text_detection.cpp
```diff
@@ -98,7 +98,36 @@ std::vector<cv::RotatedRect> maskToBoxes(const cv::Mat &mask, float min_area, fl
...
+                                          float min_area, float min_height,
+                                          const cv::Size &input_shape,
+                                          const cv::Size &image_size) {
+    std::vector<cv::RotatedRect> bboxes;
+    int max_bbox_idx = coords.size() / 5;
+    float x_scale = image_size.width / float(input_shape.width);
+    float y_scale = image_size.height / float(input_shape.height);
+
+    for (int i = 1; i <= max_bbox_idx; i++) {
+        std::vector<float> prediction(coords.begin() + (i - 1) * 5,
```
IRDonch, line 113: Seems like `const float *prediction` would achieve the same effect with less code and without unnecessarily copying the data.

GalyaZalesskaya: Thank you for your detailed and very useful feedback and even additional explanation. But here I don't really understand how float pointer could replace the vector with 5  float values, which is created on 113 line. Did you meant 115 line maybe?
` float confidence = prediction[4];`

IRDonch: All you do with `prediction` is index it, so as far as I can tell, you could replace it with `const float *prediction = &coords[(i - 1) * 5];` and the rest of the code would work the same way.

demos/text_detection_demo/src/text_detection.cpp
```diff
@@ -98,7 +98,36 @@ std::vector<cv::RotatedRect> maskToBoxes(const cv::Mat &mask, float min_area, fl
...
+                                          float min_area, float min_height,
+                                          const cv::Size &input_shape,
+                                          const cv::Size &image_size) {
+    std::vector<cv::RotatedRect> bboxes;
+    int max_bbox_idx = size_data / 5;
+    float x_scale = image_size.width / float(input_shape.width);
+    float y_scale = image_size.height / float(input_shape.height);
+
+    for (int i = 1; i <= max_bbox_idx; i++) {
+        const float *prediction = &coords[(i - 1) * 5];
```
IRDonch, line 114:
```suggestion
    for (int i = 0; i < max_bbox_idx; i++) {
        const float *prediction = &coords[i * 5];
```

Seems like that would be a more conventional loop.

demos/text_detection_demo/src/text_detection.cpp
```diff
@@ -98,7 +98,36 @@ std::vector<cv::RotatedRect> maskToBoxes(const cv::Mat &mask, float min_area, fl
...
-  }
+}
+
+std::vector<cv::RotatedRect> coordToBoxes(const float* coords,
+                                          const size_t &size_data,
+                                          float min_area, float min_height,
+                                          const cv::Size &input_shape,
+                                          const cv::Size &image_size) {
+    std::vector<cv::RotatedRect> bboxes;
+    int max_bbox_idx = size_data / 5;
```
IRDonch, line 109:
```suggestion
    int num_boxes = size_data / 5;
```

Seems like this actually isn't a valid bounding box index. It's the number of boxes.

demos/text_detection_demo/src/text_detection.cpp
```diff
@@ -198,42 +227,64 @@ std::vector<cv::RotatedRect> postProcess(const InferenceEngine::BlobMap &blobs,
...
+    	std::vector<cv::RotatedRect> rects = maskToBoxes(mask, static_cast<float>(kMinArea),
+    	                                                 static_cast<float>(kMinHeight), image_size);
+    	return rects;
+
+    } else {
+        // PostProcessing for Horizontal Text Detection model
+    	for (const auto &blob : blobs) {
+    	    if (blob.second->getTensorDesc().getDims()[1] == 5) {
+    	        kLocOutputName = blob.first;
+    	        kClsOutputName = "labels";
```
IRDonch, line 272: I just noticed that `kClsOutputName` isn't used at all in this branch. Is there even a need to set it?

models/intel/horizontal-text-detection-0001/description/horizontal-text-detection-0001.md
IRDonch: I think that's worth mentioning in the description. You basically just need to replace "\`label\`" with "0" (or whatever the class index is) in the last sentence and it'll be more descriptive.

demos/text_detection_demo/README.md
```diff
@@ -4,6 +4,7 @@ The demo shows an example of using neural networks to detect and recognize print
 
 * `text-detection-0003`, which is a detection network for finding text.
 * `text-detection-0004`, which is a lightweight detection network for finding text.
+* `horizontal-text-detection-0001`, which is a detection network that works much faster than models above, but it is applicable for funding more or less horizontal text only.
```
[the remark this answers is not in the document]
IRDonch: You only applied half of the suggestion. :-)

demos/text_detection_demo/src/text_detection.cpp
GalyaZalesskaya: This is a hack to make running the demo with other text detection models safer. In case of running unsupportable model, the script will go in the `if (kLocOutputName.empty() || kClsOutputName.empty())` condition and throw the exception.
Indeed, I don't use `kClsOutputName` in the script, but removing it will lead to simplifying the condition above to 
`if (kLocOutputName.empty())`. 
And I'm convinced that some unsupportable model may satisfy the condition, go further and crash in an unpredictable place.
I can remove it if my fears are unfounded.

Wovchena: The only requirement the else branch has is that blobs' second dimensions size is 5. So once it is satisfied you can process it and be sure you won't try reading memory that doesn't belong to the blob.

But if your intention was to allow only processing of models that has exactly the same output interface as `handwritten-score-recognition-0001` than you should have checked that the blob that was going to correspond to `kClsOutputName` has only one dimension and and the size of that dimension matches with `kLocOutputName`.

While it would be nice to have such accurate and precise check, it seems to be pretty hard to design it. My suggestion is to only verify properties that you need to process the output. So `kClsOutputName` should just be ignored the else branch.

By the way, it looks that `handwritten-score-recognition-0001` has a batch dimension in inputs but doesn't have it in outputs. That's uncommon. Is it intentional?

GalyaZalesskaya: Alright, I've removed the hardcoding of the `kClsOutputName`.
The model that I've added is horizontal-text-detection-0001 and, indeed, I've just noticed this odd thing. The inputs have the batch structure [1, 3, 704, 704], while outputs have [100, 5] dimension, where 100 - max number of prediction per image. But it is defined architecturally.

Wovchena: That probably means the model doesn't work with different batch size. If this is true, I think it's better to point out that model's property in its description.

IRDonch: For object detection models it's common to not have a separate batch dimension in the output; instead, the dimension corresponding to the number of objects is multiplied by the batch size. The same could be the case here.

GalyaZalesskaya: @Wovchena Thank you for your time. I've fixed the model's description. Do you like it or you suggest to describe it more clearly?

Wovchena: LGTM

PR #1289, opened 2020-07-06
AC docs: added note about BF16

eaidova:
In current release bfloat16 default behavior changed and it can affect accuracy, so we need aware users about it and suggest way how to avoid its usage


eaidova:
@andrew-zaytsev please take a look

eaidova:
jenkins please retry a build

PR #1299, opened 2020-07-08
downloader: implement download resumption for Google Drive files

IRDonch:
Testing showed that GDrive does support range requests, so it seems reasonable to enable it.


IRDonch:
Jenkins please retry a build

PR #1303, opened 2020-07-08
Bump dependency versions

IRDonch:
Unfortunately, pytorch/vision#2132 prevents us from moving to PyTorch 1.5, so I had to cap it at 1.4.

### DIFF
--- a/ci/requirements-ac-test.txt
+++ b/ci/requirements-ac-test.txt
@@ -1,41 +1,41 @@
 # use update-requirements.py to update this file
 
-atomicwrites==1.3.0       # via -r tools/accuracy_checker/requirements-test.in
+atomicwrites==1.4.0       # via -r tools/accuracy_checker/requirements-test.in
 attrs==19.3.0             # via pytest
 cycler==0.10.0            # via matplotlib
 decorator==4.4.2          # via networkx
 editdistance==0.5.3       # via -r tools/accuracy_checker/requirements.in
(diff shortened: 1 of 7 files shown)


IRDonch:
Jenkins please retry a build

eaidova:
@IRDonch I want to move pillow from requirements.in to requirements-core.in (reason is enable segmentation models support by default)
see: https://github.com/opencv/open_model_zoo/pull/1306
can I do it before your dependencies bumping?

IRDonch:
@eaidova Does your change conflict with mine in any way? Pillow is already in `ci/requirements-ac.txt`.

eaidova:
> @eaidova Does your change conflict with mine in any way? Pillow is already in `ci/requirements-ac.txt`.

I believe this comment will be changed if I run updating requirements script.
 `# via -r tools/accuracy_checker/requirements.in, imageio, scikit-image`
If it is not problem, then we can do not pay attention

IRDonch:
I don't think there's a need to update `ci/requirements-ac.txt` if all it does is change a comment.

vladimir-dudnik:
@IRDonch moving from ubuntu 16.04 to 18.04 in 2021.1 release it seems we do not have this limitation anymore?

IRDonch:
@vladimir-dudnik Do you mean the restriction on PyTorch < 1.5? If so, yes, that restriction was only necessary for Python 3.5 support, which will be dropped in 2021.1.

PR #1308, opened 2020-07-09
AC: Added resize postprocessor combining all resize_* postprocessors

Anna-Mironova:


tools/accuracy_checker/accuracy_checker/postprocessor/resize.py
```diff
@@ -0,0 +1,172 @@
...
+            ),
+            'size': NumberField(
+                value_type=int, optional=True, min_value=1,
+                description="Destination size for resize for both dimensions (height and width)."
+            )
+        })
+        return parameters
+
+    def configure(self):
+        if Image is None:
```
eaidova, line 62: please rebase, I make pillow official dependency today

tools/accuracy_checker/accuracy_checker/postprocessor/README.md
```diff
@@ -72,3 +72,7 @@ Accuracy Checker supports following set of postprocessors:
 * `resize_style_transfer` - resizing style transfer predicted image. Supported representations: `StyleTransferAnotation`, `StyleTransferPrediction`.
   * `dst_width` and `dst_height` - destination width and height for resizing respectively.
 * `crop_ground_truth_image` - croping ground truth image. Supported representations: `ImageInpaintingAnnotation`.
+* `resize` - resizing image or segmentation mask. Supported representations: `SegmentationAnotation`, `SegmentationPrediction`, `StyleTransferAnotation`, `StyleTransferPrediction`, `SuperResolutionAnotation`, `SuperResolutionPrediction`, `DepthEstimationAnnotation`, `DepthEstimationPrediction`.
+  * `dst_width` and `dst_height` - destination width and height for resize respectively. You can also use `size` instead in case when destination sizes are equal.
```
eaidova, line 76: it is not good idea to use image size as destination in super resolution task, you need to use annotation value size here insted for this task

tools/accuracy_checker/accuracy_checker/postprocessor/resize.py
```diff
@@ -0,0 +1,172 @@
...
+        def resize(entry, height, width):
+            return entry
+
+        @resize.register(DepthEstimationPrediction)
+        def _(entry, height, width):
+            entry.depth_map = cv2.resize(entry.depth_map, (width, height))
+
+            return entry
+
+        @resize.register(StyleTransferAnnotation)
```
eaidova, line 84: style transfer resize and super resolution resize do the same, you can just specify several decorators instead of different methods
```suggestion
        @resize.register(StyleTransferAnnotation)
        @resize.register(SuperResolutionAnnotation)
```

tools/accuracy_checker/accuracy_checker/postprocessor/resize.py
```diff
@@ -0,0 +1,172 @@
...
+            entry.mask = np.array(entry_mask)
+
+            return entry
+
+        @resize.register(SegmentationAnnotation)
+        def _(entry, height, width):
+            entry.mask = self._segm_resize(entry.mask, width, height)
+            return entry
+
+        @resize.register(SuperResolutionPrediction)
```
eaidova, line 111: my opinion that resize should be applied for annotation and prediction in all cases if we provide opprotunity to select target. It looks suspicious if user specify dst size and apply_to annotation and it do nothing

tools/accuracy_checker/accuracy_checker/postprocessor/resize.py
```diff
@@ -0,0 +1,172 @@
...
+        for prediction in predictions:
+            resize(prediction, height, width)
+
+        for annotation in annotations:
+            resize(annotation, height, width)
+
+        return annotations, predictions
+
+    @staticmethod
+    def _segm_resize(mask, width, height):
```
eaidova, line 128: Do can not just call method from old postprocessor?

Anna-Mironova: Done

tools/accuracy_checker/accuracy_checker/postprocessor/resize.py
Anna-Mironova: Done

tools/accuracy_checker/accuracy_checker/postprocessor/README.md
Anna-Mironova: Done

tools/accuracy_checker/accuracy_checker/postprocessor/resize.py
Anna-Mironova: Done

PR #1319, opened 2020-07-13
Demonstrating the use of more than 2 requests for Async inference using python deque.

anujonthemove:
Hello, 
I have extended this example to use more than 2 requests(ideally 4-8 but that I believe totally depends on the CPU) using python deque. I have added an extra parameter named `--num_requests` which takes an integer greater than 0. 

* For `--num_requests 1`, the code runs in **_sync_** mode. 
* For `--num_requests 2 or 3 or 4 or any integer` the code automatically works on **_async_** mode. 

While most of the code remains the same but I have implemented python deques to swap or rotate requests in order to be able to make it asynchronous and it works perfectly. 

I believe it is relevant to people to who are looking to extend this to using more than 2 request which was originally demonstrated in the example.


vladimir-dudnik:
@anujonthemove thanks for contribution, could you please target your PR to `develop` branch, according [CONTRIBUTING.md](https://github.com/opencv/open_model_zoo/blob/master/CONTRIBUTING.md)

anujonthemove:
> @anujonthemove thanks for contribution, could you please target your PR to `develop` branch, according [CONTRIBUTING.md](https://github.com/opencv/open_model_zoo/blob/master/CONTRIBUTING.md)

Thanks! I have moved my PR from `master` to `develop` branch.

Wovchena:
We already have similar update: https://github.com/opencv/open_model_zoo/pull/1236. I doubt we need an additional demo.

anujonthemove:
> We already have similar update: #1236. I doubt we need an additional demo.

I just checked the recent merge, it has many more features like number of threads etc. I just wrote a simpler version of the demo which is currently there on the master branch and it's easy to understand. I would leave it up to the moderators to check if this can be relevant.

Wovchena:
Your version has problems:
It doesn't match frame and corresponding inferRequest. So you draw old detection results on a newer frame.
It doesn't take into account that some inferRequest can finish earlier even if they are started later than others. And you want to keep them busy. This becomes essential when the device is "MULTI:CPU,GPU".
It doesn't make sense to have more than 2 `num_requests` without updating number of streams for IE (in case of CPU as a  device for inference).

After fixing the problems you will end up with something similar to the version from develop except for support of Tab key (which I would prefer to be removed from the develop version :) ).

anujonthemove:
> Your version has problems:
> It doesn't match frame and corresponding inferRequest. So you draw old detection results on a newer frame.
> It doesn't take into account that some inferRequest can finish earlier even if they are started later than others. And you want to keep them busy. This becomes essential when the device is "MULTI:CPU,GPU".
> It doesn't make sense to have more than 2 `num_requests` without updating number of streams for IE (in case of CPU as a device for inference).
> 
> After fixing the problems you will end up with something similar to the version from develop except for support of Tab key (which I would prefer to be removed from the develop version :) ).

Thank you for your comments, they are valuable. I will improve the code and fix the code at my end. 
This was really useful to know.

eizamaliev:
Hi @anujonthemove !
Thank you for your contribution, it is very valuable for us. But right now we are working on unifying object detection demos into one and cannot accept your PR since it produce another one :) Nevertheless, we would be proud to add your contribution into new demo. So do not stop and we will integrate your idea into the demo as soon as possible.

anujonthemove:
> Your version has problems:
> It doesn't match frame and corresponding inferRequest. So you draw old detection results on a newer frame.
> It doesn't take into account that some inferRequest can finish earlier even if they are started later than others. And you want to keep them busy. This becomes essential when the device is "MULTI:CPU,GPU".
> It doesn't make sense to have more than 2 `num_requests` without updating number of streams for IE (in case of CPU as a device for inference).
> 
> After fixing the problems you will end up with something similar to the version from develop except for support of Tab key (which I would prefer to be removed from the develop version :) ).

At the moment, I have been able to fix and check the issue of mismatching frame and corresponding inferRequest. 
Also, I was trying to understand the second issue that you mentioned `It doesn't take into account that some inferRequest can finish earlier even if they are started later than others. And you want to keep them busy. This becomes essential when the device is "MULTI:CPU,GPU".`. 

While testing for GPU, I ran into the error: 
`RuntimeError: Failed to create plugin /opt/intel/openvino_2020.3.194/deployment_tools/inference_engine/lib/intel64/libclDNNPlugin.so for device GPU
Please, check your environment
[CLDNN ERROR]. clGetPlatformIDs error -1001`. 
I did install the appropriate drivers from [this guide](https://docs.openvinotoolkit.org/latest/_docs_install_guides_installing_openvino_linux.html#additional-GPU-steps) but it is not working.

I'd be glad if you could please help me understand how do I keep those requests busy if they started later but finished earlier. I had earlier checked [this forum post](https://community.intel.com/t5/Intel-Distribution-of-OpenVINO/Question-about-Sample-Code-Python-OpenVINO-Toolkit/m-p/1140355/highlight/true#M10364) to understand how to make use of multiple requests. Are you referring to something like this by any chance? Or could you please point out appropriate resources that can help me understand this issue? 
Thanks!

anujonthemove:
> Hi @anujonthemove !
> Thank you for your contribution, it is very valuable for us. But right now we are working on unifying object detection demos into one and cannot accept your PR since it produce another one :) Nevertheless, we would be proud to add your contribution into new demo. So do not stop and we will integrate your idea into the demo as soon as possible.

Thanks @eizamaliev

PR #1324, opened 2020-07-15
AC: support detection boxes rescaling

eaidova:
added fixing boxes size on model input shape
enables ssd gluoncv models


eaidova:
Jenkins please retry a build

PR #1327, opened 2020-07-15
fix CTPN row in the table

vladimir-dudnik:
fix odd `|` character in table for CTPN row


vladimir-dudnik:
Jenkins please retry a build

PR #1329, opened 2020-07-16
Copy downloaded DLDT IR to converter.py output directory

vurusovs:
As user of converter.py, if I specify --output_dir, I'm waiting to find all DLDT IRs in it. But already downloaded IRs aren't copied to output_dir

### DIFF
--- a/tools/downloader/converter.py
+++ b/tools/downloader/converter.py
@@ -19,6 +19,7 @@ import os
 import re
 import string
 import sys
+from shutil import copytree
 
 from pathlib import Path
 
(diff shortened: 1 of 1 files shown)


vladimir-dudnik:
>  if I specify --output_dir
it should be possible to specify converter output_dir to downloader output _dir and so you should have all IRs (downloaded and converted) in one place, is not it?

IRDonch:
I don't think I like this idea. It doubles the storage requirements and increases the run time of the converter, for questionable benefits.

vurusovs:
OK, in this case I'm closing PR

PR #1330, opened 2020-07-16
PerformanceMetrics class for C++ demos

AnthonyQuantum:


demos/common/performance_metrics/metrics.cpp
```diff
@@ -0,0 +1,111 @@
...
+PerformanceMetrics::PerformanceMetrics(Duration timeWindow) {
+    // 'timeWindow' defines the length of the timespan over which the 'current fps' value is calculated
+    timeWindowSize = timeWindow;
+
+    lastMovingStatistic = Statistic();
+    currentMovingStatistic = Statistic();
+    totalStatistic = Statistic();
+}
+
+void PerformanceMetrics::update(TimePoint lastRequestStartTime,
```
eizamaliev, line 21: I suggest you to split this method into two: `update` and `drawFps`. Or even make `drawFps` separated function.

AnthonyQuantum: > I suggest you to split this method into two: `update` and `drawFps`. Or even make `drawFps` separated function.

We've discussed this earlier and the conclusion was:
> The idea behind merging methods is to prevent forgetting to call one of the methods as they are always expected to be called one after another in common use case. And to prevent calling that methods far form each other like
> ```
> performanceMetrics.update()
> postprocess_detections()
> performanceMetrics.put_moving_*()
> ```
From here https://github.com/openvinotoolkit/open_model_zoo/pull/1222#discussion_r442113416

So, I'm not sure that this method should be splitted.

demos/common/performance_metrics/metrics.cpp
```diff
@@ -0,0 +1,111 @@
...
+        currentMovingStatistic = Statistic();
+
+        lastUpdateTime = currentTime;
+    }
+
+    // Draw performance stats over frame
+    Metrics metrics = getLast();
+    std::ostringstream out;
+    if (metrics.latency == metrics.latency) {
+        out.str("");
```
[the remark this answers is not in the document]
AnthonyQuantum: Turning `out` into a field will make the class non-copy-constructible, which is not quite handy. Is it critical, in terms of run time, to avoid creating `ostringstream` each time here?

demos/object_detection_demo_ssd_async/main.cpp
```diff
@@ -561,33 +556,10 @@ int main(int argc, char *argv[]) {
...
-            endTime = (modeInfo[ExecutionMode::MIN_LATENCY].lastEndTime.time_since_epoch().count() != 0)
-                      ? modeInfo[ExecutionMode::MIN_LATENCY].lastEndTime
-                      : totalT1;
-            std::cout << "FPS: " << std::fixed << std::setprecision(1)
-                << modeInfo[ExecutionMode::MIN_LATENCY].framesCount / std::chrono::duration_cast<Sec>(
-                    endTime - modeInfo[ExecutionMode::MIN_LATENCY].lastStartTime).count() << std::endl;
-            std::cout << "Latency: " << std::fixed << std::setprecision(1)
-                << ((std::chrono::duration_cast<Ms>(modeInfo[ExecutionMode::MIN_LATENCY].latencySum).count() /
-                     modeInfo[ExecutionMode::MIN_LATENCY].framesCount)) << " ms" << std::endl;
+        for (auto & mode : modeMetrics) {
```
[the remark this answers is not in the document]
AnthonyQuantum: BTW, there are 50+ occurrences of `for (auto & ` in other demos and many times the snake_case is used in C++ demos. Maybe we can add it to the backlog and someday fix it all with a few regexps?

demos/common/performance_metrics/metrics.cpp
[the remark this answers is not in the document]
Wovchena: May be we could specify `timeWindow` as template. That should eliminate a need to define constructor for `PerformanceMetrics` and compiler will generate move constructor for you. But I think it's simpler to create `ostringstream` each time.

demos/object_detection_demo_ssd_async/main.cpp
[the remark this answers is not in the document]
Wovchena: Yes, we could add it to the backlog

demos/common/performance_metrics/CMakeLists.txt
```diff
@@ -0,0 +1,18 @@
+# Copyright (C) 2020 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+#
+
+find_package(InferenceEngine 2.0 REQUIRED)
```
Wovchena, line 5: Why do you need IE?

demos/common/performance_metrics/metrics.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+#include <iostream>
+#include <limits>
+#include <sstream>
+#include <iomanip>
+
+PerformanceMetrics::PerformanceMetrics(Duration timeWindow) {
+    // defines the length of the timespan over which the 'current fps' value is calculated
+    timeWindowSize = timeWindow;
+}
```
Wovchena, line 15:
```suggestion
// timeWindow defines the length of the timespan over which the 'current fps' value is calculated
PerformanceMetrics::PerformanceMetrics(Duration timeWindow) : timeWindowSize(timeWindow) {}
```

demos/common/performance_metrics/metrics.h
```diff
@@ -0,0 +1,60 @@
+// Copyright (C) 2020 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+//
+
+#pragma once
+
+#include <chrono>
+
+#include "../samples/ocv_common.hpp"
```
Wovchena, line 9: I'm not sure about `"`. I think folders in common are supposed to be different libraries. They are like third party libraries to each other.

So you should use `<>`. And should't use relative include paths. Modify `target_include_directories` for `performance_metrics` in cmake instead.

demos/common/performance_metrics/metrics.cpp
```diff
@@ -0,0 +1,107 @@
...
+    if (metrics.latency > std::numeric_limits<double>::epsilon()) {
+        std::cout << std::fixed << std::setprecision(1) << metrics.latency << " ms";
+    } else {
+        std::cout << "N/A";
+    }
+    std::cout << std::endl;
+
+    std::cout << "FPS: ";
+    if (metrics.fps > std::numeric_limits<double>::epsilon()) {
+        std::cout << std::fixed << std::setprecision(1) << metrics.fps;
```
Wovchena, line 102: This changes settings for `cout`. It can be unexpected. I suggest creating another `ostringstream` to build output string.

demos/common/performance_metrics/metrics.h
```diff
@@ -0,0 +1,60 @@
...
+
+class PerformanceMetrics {
+public:
+    using Clock = std::chrono::steady_clock;
+    using TimePoint = std::chrono::time_point<Clock>;
+    using Duration = Clock::duration;
+    using Us = std::chrono::microseconds;
+    using Ms = std::chrono::milliseconds;
```
Wovchena, line 17: While `Duration` should be represented as `Clock::duration`, it is better to represent `Ms` and `Sec` as `std::chrono::duration<double, ...>` because they are just helpers that simplify result conversion.

demos/common/performance_metrics/CMakeLists.txt
```diff
@@ -0,0 +1,18 @@
...
+
+set(SOURCES metrics.cpp)
+set(HEADERS metrics.h)
+
+# Create named folders for the sources within the .vcproj
+# Empty name lists them directly under the .vcproj
+source_group("src" FILES ${SOURCES})
+source_group("include" FILES ${HEADERS})
+
+add_library(performance_metrics STATIC ${SOURCES} ${HEADERS} ${InferenceEngine_INCLUDE_DIRS})
```
IRDonch, line 16: Does this need to be a separate library? IMO, it could just be part of `common`. I'm turning `common` into a real library in #1350; once that's merged, the performance metrics could just be added to it.

demos/common/performance_metrics/CMakeLists.txt
AnthonyQuantum: performance_metrics.hpp includes ocv_common.hpp, which needs ie_core.hpp.

I've moved `find_package(InferenceEngine 2.0 REQUIRED)` to CMakeLists file for the whole `common` library, since the folder structure in `common` was modified.

demos/common/src/performance_metrics.cpp
```diff
@@ -0,0 +1,112 @@
...
+
+    // Draw performance stats over frame
+    Metrics metrics = getLast();
+    std::ostringstream out;
+    
+    getLatencyMessage(out, metrics.latency, !std::isnan(metrics.latency));
+    putHighlightedText(frame, out.str(), position, cv::FONT_HERSHEY_COMPLEX, fontScale, color, thickness);
+
+    getFpsMessage(out, metrics.fps, !std::isnan(metrics.fps));
+    putHighlightedText(frame, out.str(), {position.x, position.y + 30}, cv::FONT_HERSHEY_COMPLEX, fontScale, color,
```
Wovchena, line 45: You should report on an image only if you have anything to report.

demos/common/src/performance_metrics.cpp
```diff
@@ -0,0 +1,112 @@
...
+}
+
+void PerformanceMetrics::printTotal() {
+    Metrics metrics = getTotal();
+    std::ostringstream out;
+
+    getLatencyMessage(out, metrics.latency, bool(metrics.latency > std::numeric_limits<double>::epsilon()));
+    std::cout << out.str() << std::endl;
+
+    getFpsMessage(out, metrics.fps, bool(metrics.fps > std::numeric_limits<double>::epsilon()));
```
Wovchena, line 88: Why do you check epsilon instead of nan here?

demos/common/src/performance_metrics.cpp
```diff
@@ -0,0 +1,112 @@
...
+    std::cout << out.str() << std::endl;
```
Wovchena, line 89:
```suggestion
    out << "Latancy: " << std::fixed << std::setprecision(1) << metrics.latency << " ms\nFPS: " << metrics.fps << '\n';
    std::cout << out.str();
```
`ostringstream` knows how represent `nan`s so `getLatencyMessage()` and `getFpsMessage()` can be removed.

demos/common/performance_metrics/metrics.h
Wovchena: performance_metrics are part of common now. You should use quotes.

demos/common/src/performance_metrics.cpp
```diff
@@ -0,0 +1,112 @@
...
+
+void PerformanceMetrics::update(TimePoint lastRequestStartTime,
+                                cv::Mat& frame,
+                                cv::Point position,
+                                double fontScale,
+                                cv::Scalar color,
+                                int thickness) {
+    TimePoint currentTime = Clock::now();
+
+    if (lastUpdateTime == TimePoint()) {
```
IRDonch, line 20: I don't recommend using `TimePoint()` as a sentinel value, because that's the clock's epoch, which could occur as a normal time value returned by `Clock::now()` (which is prretty unlikely, but still). I would suggest to just use a boolean flag to track whether the first frame has been processed yet.

demos/object_detection_demo_ssd_async/main.cpp
```diff
@@ -22,6 +22,7 @@
 #include <samples/args_helper.hpp>
 #include <samples/ocv_common.hpp>
 #include <samples/slog.hpp>
+#include <samples/performance_metrics.hpp>
```
IRDonch, line 25: Keep the includes sorted, please.

demos/common/include/samples/performance_metrics.hpp
```diff
@@ -0,0 +1,70 @@
...
+    using Duration = Clock::duration;
+    using Ms = std::chrono::duration<double, std::ratio<1, 1000>>;
+    using Sec = std::chrono::duration<double, std::ratio<1, 1>>;
+
+    struct Metrics {
+        double latency;
+        double fps;
+    };
+
+    PerformanceMetrics(Duration timeWindow = std::chrono::duration_cast<Duration>(Sec(1)));
```
Wovchena, line 32:
```suggestion
    PerformanceMetrics(Duration timeWindow = std::chrono::seconds(1));
```

demos/common/include/samples/performance_metrics.hpp
```diff
@@ -0,0 +1,71 @@
...
+
+        void combine(const Statistic& other) {
+            latency += other.latency;
+            period += other.period;
+            frameCount += other.frameCount;
+        }
+    };
+
+    void getLatencyMessage(std::ostringstream& out, double value, bool isAvailable);
+    void getFpsMessage(std::ostringstream& out, double value, bool isAvailable);
```
IRDonch, line 63: Undefined methods.

demos/common/include/samples/performance_metrics.hpp
```diff
@@ -0,0 +1,70 @@
...
+    PerformanceMetrics(Duration timeWindow = std::chrono::duration_cast<Duration>(Sec(1)));
+    void update(TimePoint lastRequestStartTime,
+                cv::Mat& frame,
+                cv::Point position = {15, 30},
+                double fontScale = 0.75,
+                cv::Scalar color = {200, 10, 10},
+                int thickness = 2);
+    Metrics getLast() const;
+    Metrics getTotal() const;
+    void printTotal();
```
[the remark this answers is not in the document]
AnthonyQuantum: Done

PR #1340, opened 2020-07-22
AC: fix paths conversion

eaidova:
fix for relative path in config without prefix usage

### DIFF
--- a/tools/accuracy_checker/accuracy_checker/annotation_converters/style_transfer.py
+++ b/tools/accuracy_checker/accuracy_checker/annotation_converters/style_transfer.py
@@ -14,7 +14,6 @@ See the License for the specific language governing permissions and
 limitations under the License.
 """
 
-from pathlib import Path
 from ..config import PathField
 from ..representation import StyleTransferAnnotation
 from .format_converter import BaseFormatConverter, ConverterReturn
@@ -36,7 +35,7 @@ class StyleTransferConverter(BaseFormatConverter):
         return parameters
 
     def configure(self):
-        self.image_dir = Path(self.get_value_from_config('images_dir'))
+        self.image_dir = self.get_value_from_config('images_dir')
 
     def convert(self, check_content=False, progress_callback=None, progress_interval=100, **kwargs):
         content_check_errors = [] if check_content else None
--- a/tools/accuracy_checker/accuracy_checker/annotation_converters/style_transfer.py
+++ b/tools/accuracy_checker/accuracy_checker/annotation_converters/style_transfer.py
@@ -14,6 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 """
 
+from pathlib import Path
 from ..config import PathField
 from ..representation import StyleTransferAnnotation
 from .format_converter import BaseFormatConverter, ConverterReturn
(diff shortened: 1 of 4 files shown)


tools/accuracy_checker/accuracy_checker/annotation_converters/style_transfer.py
```diff
@@ -35,7 +36,7 @@ def parameters(cls):
         return parameters
 
     def configure(self):
-        self.image_dir = self.get_value_from_config('images_dir')
+        self.image_dir = Path(self.get_value_from_config('images_dir'))
```
IRDonch, line 39: Why is this needed? I thought `images_dir` being a `PathField` meant that the value would already be a `Path`.

eaidova: you are right, I deeper investigate issue and found that type casting never performend during getting value from config. native yml read it as string. when we use path without  prefixing in the config it saved as string

tools/accuracy_checker/accuracy_checker/config/config_validator.py
```diff
@@ -152,7 +152,11 @@ def validate(self, entry, field_uri=None):
 
     @property
     def type(self):
-        return str
+        def type_def(x=None):
+            if x is None:
+                return str
+            return x
```
IRDonch, line 158: I don't understand what's going on here. This function returns `str` (a type) if `x` is `None`, but `x` (a value) otherwise. That doesn't seem right. And why is it needed at all? The logic in `utils.py` already handles the case when `x` is `None`.

eaidova: getting type from Field also used for static analisys required configuration for model in DL WorkBench. They want to get field type without providing argument. The BaseField used in cases, where several types applicable (e.g. for mean, we can provide float, list of floats for each channel or select one of precomputed values using strings), we can not know which type will have input values, but during parsing is opportunity to parsing necessary formats from string always exists.

IRDonch: I see; however, your code doesn't seem to actually support this use case: if `f` is an instance of `BaseField`, then `f.type` is `type_def`, not `str`.

IMO, a better way to do this would be to have `type` return `None` here, and add special logic for `None` both in `get_parameter_value_from_config` and in DL WB.

eaidova: ok, changes

tools/accuracy_checker/accuracy_checker/config/config_validator.py
```diff
@@ -165,7 +165,12 @@ def parameters(self):
                     parameters_dict[key] = self.__dict__[key].parameters()
                 else:
                     parameters_dict[key] = self.__dict__[key]
-            parameters_dict['type'] = type(self.type()).__name__
+            data_type = self.type
+            if data_type is not None:
+                data_type = type(data_type()).__name__  # pylint:disable=E1102
+            else:
+                data_type = str.__name__
+            parameters_dict['type'] = data_type
```
IRDonch, line 173:
```suggestion
            parameters_dict['type'] = type((self.type or str)()).__name__
```

PR #1350, opened 2020-07-23
demos: turn common into a static library

IRDonch:
See the commit messages for details.


demos/common/monitors/src/cpu_monitor.cpp
```diff
@@ -2,7 +2,8 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
-#include "cpu_monitor.h"
+#include <monitors/cpu_monitor.h>
```
Wovchena, line 5: Here, in memory_monitor.cpp, presenter.cpp and args_helper.cpp in similar places quotes should be used instead of brackets . cpu_monitor.h is a header that corresponds to the same library so it is "local" header for that .cpp file.

IRDonch: Seems like an arbitrary distinction to me, but OK.

Wovchena: There is args_helper.cpp left.

IRDonch: Oh, right. Fixed.

PR #1352, opened 2020-07-24
Dynamic Open Model Zoo as Python module

likholat:
### Adding to OpenVino
1. Create a symbolic link:
    ```bash
    ln -s /opt/intel/openvino_2021/deployment_tools/open_model_zoo/libs/python/open_model_zoo.py /opt/intel/openvino_2021/python/python3.X/openvino/open_model_zoo.py
    ```
2. Add path to downloader to python path 
    ```bash
    export PYTHONPATH=$PYTHONPATH:/opt/intel/openvino/deployment_tools/open_model_zoo/tools/downloader
    ```
3. Initialize openvino enviroment variable :
    ```bash
     source /opt/intel/openvino_2021/bin/setupvars.sh
     ```

### Model's downloading
```python
import openvino.open_model_zoo as omz

ssd = omz.Model('mobilenet-ssd')
```

```
################|| Downloading models ||################

========== Downloading /user/models/public/mobilenet-ssd/mobilenet-ssd.prototxt
... 100%, 28 KB, 38894 KB/s, 0 seconds passed

========== Downloading /user/models/public/mobilenet-ssd/mobilenet-ssd.caffemodel
... 100%, 22605 KB, 3760 KB/s, 6 seconds passed

################|| Post-processing ||################
```

### Documentation
```python
import openvino.open_model_zoo as omz

ssd = omz.Model('mobilenet-ssd', precision='FP32')

print('Model task type: ' + ssd.task_type)
print(ssd.description)
```

```
Model task type: detection​
The "mobilenet-ssd" model is a Single-Shot multibox Detection (SSD) network intended to perform object detection. This model is implemented using the Caffe* framework. For details about this model, check out the repository <https://github.com/chuanqi305/MobileNet-SSD>.​
The model input is a blob that consists of a single image of 1x3x300x300 in BGR order, also like the "densenet-121" model. The BGR mean values need to be subtracted as follows: [127.5, 127.5, 127.5] before passing the image blob into the network. In addition, values must be divided by 0.007843.​
The model output is a typical vector containing the tracked object data, as previously described.​

    License: https://raw.githubusercontent.com/chuanqi305/MobileNet-SSD/master/LICENSE
```

### Usage example
```python
import os
import numpy as np
import cv2 as cv

import openvino.open_model_zoo as omz

from openvino.inference_engine import IECore

ssd = omz.Model('mobilenet-ssd', precision='FP32')
xml_path, bin_path = ssd.xml_path, ssd.bin_path
ie = IECore()
net = ie.read_network(xml_path, bin_path)
exec_net = ie.load_network(net, 'CPU')

image = cv.imread('example.jpeg')
predictions = detect(net, image)

rows = image.shape[0]
cols = image.shape[1]

for detection in predictions.reshape(-1, 7):
    image_id = detection[0]
    if image_id < 0:
        break
 
    if detection[2] > 0.6:
        xmin = int(detection[3] * cols)
        ymin = int(detection[4] * rows)
        xmax = int(detection[5] * cols)
        ymax = int(detection[6] * rows)

        cv.rectangle(image, (xmin, ymin), (xmax, ymax), (0, 255, 0))

cv.imshow('out', image)
cv.waitKey()
```
image

- Tests
To run tests use:
```bash
source /opt/intel/openvino_2021/bin/setupvars.sh
python3 /user/open_model_zoo/tools/downloader/open_model_zoo_test.py
```


likholat:
@dkurt

tools/downloader/topologies.py
```diff
@@ -0,0 +1,70 @@
...
+        self.name = name
+        self.download_dir = download_dir
+        self.precision = precision
+        Topology.__doc__ %=  description, license_url
+
+    def get_ir(self, precision='FP32'):
+        '''
+        Creates OpenVINO Intermediate Representation (IR) network model.
+        If there is already converted model or the model has been downloaded in
+        IR - just returns the paths to the files.
```
dkurt, line 47: Please add a test for some `intel` model (which is already in IR format)

tools/downloader/topologies.py
```diff
@@ -0,0 +1,70 @@
...
+    parser = argparse.ArgumentParser()
+    args = Namespace(all=False, list=None, name=model_name, print_all=False)   
+    _topology = common.load_models_from_args(parser, args)[0]
+
+    download_dir = _cache / _topology.subdirectory
+    cache_dir = _cache / 'models_cache'
+
+    downloader.main(['--name=' + model_name, '--precisions=' + precision,
+                '--output_dir=' + str(_cache).replace('\\', '\\\\'), 
+                '--cache_dir=' + str(cache_dir).replace('\\', '\\\\')])
```
dkurt, line 25: fix alignment

tools/downloader/topologies_test.py
```diff
@@ -0,0 +1,68 @@
...
+
+    def test_get_ir(self):
+        xml_path, bin_path = self.ssd.get_ir()
+        self.assertTrue(os.path.exists(xml_path))
+        self.assertTrue(os.path.exists(bin_path))
+        self.assertEqual(os.path.basename(xml_path), 'mobilenet-ssd.xml')
+        self.assertEqual(os.path.basename(bin_path), 'mobilenet-ssd.bin')
+
+    def test_result_model(self):
+        image_path = os.path.join(os.environ.get("OPEN_MODEL_ZOO_PATH"),
```
dkurt, line 27: Let's avoid this variable - you may get script's location by `__file__` and find test image by relative path

eaidova:
Hi @likholat,

please make `develop` as target branch for your request,
thanks

tools/downloader/topologies.py
```diff
@@ -0,0 +1,66 @@
...
+import argparse
+import sys
+
+from pathlib import Path
+from argparse import Namespace
+
+import common
+import downloader
+import converter
+import info_dumper
```
dkurt, line 11: Unused import

tools/downloader/topologies.py
```diff
@@ -0,0 +1,66 @@
...
+
+_cache = Path(os.getenv('OPEN_MODEL_ZOO_CACHE_DIR', Path('.').resolve() / 'models'))
+
+def load_model(model_name, precision='FP32'):
```
dkurt, line 15: Please document this method

tools/downloader/topologies.py
```diff
@@ -0,0 +1,66 @@
...
+    parser = argparse.ArgumentParser()
+    args = Namespace(all=False, list=None, name=model_name, print_all=False)   
+    _topology = common.load_models_from_args(parser, args)[0]
```
dkurt, line 18: Inside the method you can avoid using underscore (it's used to hide some methods or variables from other scripts where `topologies.py` is imported)

tools/downloader/topologies.py
```diff
@@ -0,0 +1,66 @@
...
+
+    download_dir = _cache / _topology.subdirectory
+    cache_dir = _cache / 'models_cache'
```
dkurt, line 21: Please mode right after `_cache` definition:
```python
_cache = Path(os.getenv('OPEN_MODEL_ZOO_CACHE_DIR', Path('.').resolve() / 'models'))
_cache_dir = _cache / 'models_cache'
```

dkurt: Also, it's better to make cache hidden:
```
_cache_dir = _cache / '.models_cache'
```

tools/downloader/topologies.py
```diff
@@ -0,0 +1,81 @@
...
+from argparse import Namespace
+
+import common
+import downloader
+import converter
+
+_cache = Path(os.getenv('OPEN_MODEL_ZOO_CACHE_DIR', Path('.').resolve() / 'models'))
+_cache_dir = _cache / 'models_cache'
+
+def load_model(model_name, precision='FP32'):
```
eizamaliev, line 15: Could you add path as function argument?

eizamaliev, line 15: Also, I think `load_model` would be great as `Topology` method.

tools/downloader/topologies.py
```diff
@@ -0,0 +1,81 @@
...
+                     '--output_dir=' + str(_cache).replace('\\', '\\\\'), 
+                     '--cache_dir=' + str(_cache_dir).replace('\\', '\\\\')])
+        
+    return Topology(model_name, download_dir, precision, 
+                    topology.description, topology.license_url)
+
+class Topology:     
+    '''
+    ${0}
+
```
eizamaliev, line 47: Add information about model task type.

tools/downloader/topologies.py
dkurt: Can you show example how user's script may looks like in this way?

tools/downloader/topologies.py
```diff
@@ -39,20 +39,21 @@ def load_model(model_name, precision='FP32'):
...
+    License: %s
     '''
 
-    def __init__(self, name, download_dir, precision, description, license_url):
+    def __init__(self, name, download_dir, precision, description, license_url, task_type):
         self.name = name
         self.download_dir = download_dir
         self.precision = precision
-        Topology.__doc__.format(description, license_url)
+        Topology.__doc__ %=  description, task_type, license_url
```
dkurt, line 56: Please keep `.format` with just multiple `{}`

tools/downloader/topologies.py
dkurt: I'd like to suggest to keep this question for internal arch review process. Not hard to change but we just need to decide which version to choose.

But it's good idea to make it optional but let's check if there is a way to force method to have named arguments only so we won't meet a situation when order or set of arguments will be changed.

eizamaliev:
```
from topologies import Topology
...
ssd = Topology(name, path_to_model)
ssd.load() 
model = ssd.getIR(precision)
```
Or `load` could be inside constructor.

dkurt: The thing is that constructor is already used to define the topology (there are a lot of arguments such doc string etc.)

tools/downloader/topologies.py
```diff
@@ -0,0 +1,82 @@
...
+
+from pathlib import Path
+from argparse import Namespace
+
+import common
+import downloader
+import converter
+
+_cache = Path(os.getenv('OPEN_MODEL_ZOO_CACHE_DIR', Path('.').resolve() / 'models'))
+_cache_dir = _cache / 'models_cache'
```
dkurt, line 13: Please introduce two constants:
```
OPEN_MODEL_ZOO_MODELS_DIR -> _download_dir
OPEN_MODEL_ZOO_CACHE_DIR -> _cache_dir 
```

tools/downloader/topologies.py
```diff
@@ -9,12 +9,12 @@
 import downloader
 import converter
 
-_cache = Path(os.getenv('OPEN_MODEL_ZOO_CACHE_DIR', Path('.').resolve() / 'models'))
-_cache_dir = _cache / 'models_cache'
+_download_dir = Path(os.getenv('OPEN_MODEL_ZOO_MODELS_DIR', Path('.').resolve() / 'models'))
+_cache_dir = Path(os.getenv('OPEN_MODEL_ZOO_CACHE_DIR', _download_dir / 'models_cache'))
```
dkurt, line 13: Better to remove `model/model_` duplication and make it hidden.
```python
_cache_dir = Path(os.getenv('OPEN_MODEL_ZOO_CACHE_DIR', _download_dir / '.cache'))
```

tools/downloader/topologies.py
```diff
@@ -32,28 +32,28 @@ def load_model(model_name, precision='FP32'):
...
     topology = common.load_models_from_args(parser, args)[0]
 
-    download_dir = _cache / topology.subdirectory
+    model_dir = _download_dir / topology.subdirectory
 
     downloader.main(['--name=' + model_name, '--precisions=' + precision,
-                     '--output_dir=' + str(_cache).replace('\\', '\\\\'), 
-                     '--cache_dir=' + str(_cache_dir).replace('\\', '\\\\')])
+                     '--output_dir=' + str(_download_dir).replace('\\', '\\\\'), 
+                     '--cache_dir=' + str(_download_dir).replace('\\', '\\\\')])
```
dkurt, line 39: `_cahce_dir`?

tools/downloader/topologies.py
```diff
@@ -0,0 +1,82 @@
...
+    -------    
+    An object of the Topology class which stores information about the downloaded model.
+    '''
+    parser = argparse.ArgumentParser()
+    args = Namespace(all=False, list=None, name=model_name, print_all=False)   
+    topology = common.load_models_from_args(parser, args)[0]
+
+    model_dir = _download_dir / topology.subdirectory
+
+    downloader.main(['--name=' + model_name, '--precisions=' + precision,
```
dkurt, line 37: Please check if we can download models without FP32 precision. Maybe it make sense to make default value of precision as None

tools/downloader/topologies.py
dkurt: @likholat, we just discusses with @eizamaliev that now we have `load_model` method which directly calls `Topology` constructor. So it's a kind of redundant chain and more simple to have
```
ssd = omz.Topology('mobilenet-ssd')
```
which invocated downloader inside.

tools/downloader/topologies.py
```diff
@@ -0,0 +1,82 @@
...
+    License: {}
+    '''
+
+    def __init__(self, name, model_dir, precision, description, license_url, task_type):
+        self.name = name
+        self.model_dir = model_dir
+        self.precision = precision
+        Topology.__doc__ = Topology.__doc__.format(description, task_type, license_url)
+
+    def get_ir(self):
```
dkurt, line 58: Please make is possible to have multiple precisions:
```
ssd = omz.load_model('mobilenet-ssd')
xml32, bin32 = ssd.get_ir(precision='FP32')
xml16, bin16 = ssd.get_ir(precision='FP16')
```

tools/downloader/topologies_test.py
```diff
@@ -0,0 +1,79 @@
...
+from openvino.inference_engine import IECore
+
+class TestTopologies(unittest.TestCase):
+
+    download_path = Path(os.getenv('OPEN_MODEL_ZOO_MODELS_DIR', Path('.').resolve() / 'models'))
+    ssd = omz.load_model('mobilenet-ssd')
+
+    def test_load_model(self):
+        self.assertTrue(os.path.exists(
+            os.path.join(self.download_path, 'public', 'mobilenet-ssd', 'mobilenet-ssd.caffemodel')))
```
dkurt, line 18: I'd like to recommend to remove this test - it's not so critical which structure has downloading folder but other tests will fail if there is something wrong with it. (I just want to remove `getenv` from the test 🐤 )

tools/downloader/topologies.py
```diff
@@ -34,11 +34,12 @@ def load_model(model_name, precision='FP32'):
 
     model_dir = _download_dir / topology.subdirectory
 
-    downloader.main(['--name=' + model_name, '--precisions=' + precision,
-                     '--output_dir=' + str(_download_dir).replace('\\', '\\\\'), 
-                     '--cache_dir=' + str(_download_dir).replace('\\', '\\\\')])
+    if not precision:
```
dkurt, line 37: What if None?

tools/downloader/topologies_test.py
```diff
@@ -9,13 +9,7 @@
...
 
 class TestTopologies(unittest.TestCase):
-
-    download_path = Path(os.getenv('OPEN_MODEL_ZOO_MODELS_DIR', Path('.').resolve() / 'models'))
-    ssd = omz.load_model('mobilenet-ssd')
-
-    def test_load_model(self):
-        self.assertTrue(os.path.exists(
-            os.path.join(self.download_path, 'public', 'mobilenet-ssd', 'mobilenet-ssd.caffemodel')))
+    ssd = omz.load_model('mobilenet-ssd', 'FP32')
```
dkurt, line 12: Why we need to specify precision for non-IR model? We should do it while `get_ir` instead

tools/downloader/topologies.py
```diff
@@ -0,0 +1,103 @@
...
+
+        Returns
+        -------
+        If precision was setted in this method or not None precision was setted in __init__:
+            xml_path
+                Path to generated or downloaded .xml file
+            bin_path
+                Path to generated or downloaded .bin file
+
+        Or returns list of pairs: (xml_path, bin_path) for all precisions of target model.
```
dkurt, line 92: `get_ir` should return single model. If no precision specified for `get_ir` we should return rather precision specified during  downloading or if it also None - default Topology's precision

tools/downloader/topologies_test.py
```diff
@@ -20,13 +20,75 @@ def test_get_ir(self):
...
+        self.assertTrue(os.path.exists(bin_path_16))
+        self.assertTrue(os.path.exists(xml_path_32))
+        self.assertTrue(os.path.exists(bin_path_32))
+
+        self.assertEqual(os.path.basename(xml_path), 'mobilenet-ssd.xml')
+        self.assertEqual(os.path.basename(bin_path), 'mobilenet-ssd.bin')
+        self.assertEqual(os.path.basename(xml_path_16), 'mobilenet-ssd.xml')
+        self.assertEqual(os.path.basename(bin_path_16), 'mobilenet-ssd.bin')
+        self.assertEqual(os.path.basename(xml_path_32), 'mobilenet-ssd.xml')
+        self.assertEqual(os.path.basename(bin_path_32), 'mobilenet-ssd.bin')
```
dkurt, line 51: Please format like this:
```python
ssd = omz.load_model('mobilenet-ssd')
for p in [None, 'FP32', 'FP16']:
  xml_path, bin_path = ssd.get_ir(precision=p)
  
  self.assertTrue(os.path.exists(xml_path))
  self.assertTrue(os.path.exists(bin_path))
  // also add a check for parent folder - FP32 for None and FP32 and FP16 for FP16
```

dkurt: Even better to parametrize the test so it will accept downloading precision and `get_ir` precision and check final precision as

| ↓ load / get_ir → | None | FP32 | FP16 |
|---|---|---|---|
| None | FP32 | FP32 | FP16 |
| FP32 | FP32 | FP32 | FP16 |
| FP16 | FP16 | FP32 | FP16 |

```python
final = get_ir_precision if get_ir_precision else download_precision
```

tools/downloader/topologies.py
```diff
@@ -0,0 +1,89 @@
...
+        model_dir = _download_dir / topology.subdirectory
+
+        if precision is None:
+            downloader.main(['--name=' + model_name,
+                             '--output_dir=' + str(_download_dir).replace('\\', '\\\\'), 
+                             '--cache_dir=' + str(_cache_dir).replace('\\', '\\\\')])
+        else:
+            downloader.main(['--name=' + model_name, '--precisions=' + precision,
+                             '--output_dir=' + str(_download_dir).replace('\\', '\\\\'), 
+                             '--cache_dir=' + str(_cache_dir).replace('\\', '\\\\')])
```
dkurt, line 50: Code duplication may lead to the bugs. Please collect the flags and then call downloader:
```python
flags = ['--name=' + model_name,
         '--output_dir=' + str(_download_dir).replace('\\', '\\\\'), 
         '--cache_dir=' + str(_cache_dir).replace('\\', '\\\\')]
if precision:
  flags += ['--precisions=' + precision]

downloader.main(flags)
```

tools/downloader/topologies_test.py
```diff
@@ -0,0 +1,127 @@
...
+            xml_path, bin_path = ssd.get_ir(precision=p)
+            
+            self.assertTrue(os.path.exists(xml_path))
+            self.assertTrue(os.path.exists(bin_path))
+
+            self.assertEqual(os.path.basename(xml_path), 'mobilenet-ssd.xml')
+            self.assertEqual(os.path.basename(bin_path), 'mobilenet-ssd.bin')
+
+            self.assertEqual(os.path.basename(os.path.dirname(xml_path)), p if p else 'FP32')
+            self.assertEqual(os.path.basename(os.path.dirname(bin_path)), p if p else 'FP32')
```
dkurt, line 53: Please try to parametrize this test with download and convert precisions

likholat:
```python
import topologies as omz
face_detection = omz.Topology('face-detection-0102', precision='FP32')
xml_path, bin_path = face_detection.get_ir('FP16')
```
output:
```
################|| Downloading models ||################

========== Downloading /models/intel/face-detection-0102/FP32/face-detection-0102.xml
... 100%, 162 KB, 401520 KB/s, 0 seconds passed

========== Downloading /models/intel/face-detection-0102/FP32/face-detection-0102.bin
... 100%, 7098 KB, 4240 KB/s, 1 seconds passed

################|| Post-processing ||################

========== Skipping face-detection-0102 (no conversions defined)
```

tools/downloader/topologies_test.py
```diff
@@ -0,0 +1,99 @@
...
+            self.assertTrue(os.path.exists(xml_path))
+            self.assertTrue(os.path.exists(bin_path))
+
+            self.assertEqual(os.path.basename(xml_path), 'face-detection-0100.xml')
+            self.assertEqual(os.path.basename(bin_path), 'face-detection-0100.bin')
+
+            self.assertEqual(os.path.basename(os.path.dirname(xml_path)), p if p else 'FP32')
+            self.assertEqual(os.path.basename(os.path.dirname(bin_path)), p if p else 'FP32')
+
+    @parameterized.expand([(None, ), 'FP32', 'FP16'])
```
dkurt, line 41: Is that possible to parametrize over both precisions?

tools/downloader/topologies.py
```diff
@@ -0,0 +1,93 @@
...
+            xml_path
+                Path to generated or downloaded .xml file
+            bin_path
+                Path to generated or downloaded .bin file
+        '''
+
+        res_precision = precision if precision else 'FP32'
+        
+        if not precision and res_precision not in self.precisions:
+            res_precision = next(iter(self.precisions))
```
dkurt, line 81: For non-IR models, what is a value of `self.precisions`?

tools/downloader/topologies.py
dkurt: @likholat, please add `output_dir` argument

tools/downloader/topologies_test.py
```diff
@@ -0,0 +1,114 @@
...
+        xml_path, bin_path = ssd.get_ir(precision=pr)
+        
+        self.assertTrue(os.path.exists(xml_path))
+        self.assertTrue(os.path.exists(bin_path))
+
+        self.assertEqual(os.path.basename(xml_path), 'mobilenet-ssd.xml')
+        self.assertEqual(os.path.basename(bin_path), 'mobilenet-ssd.bin')
+
+        self.assertEqual(os.path.basename(os.path.dirname(xml_path)), pr if pr else 'FP32')
+        self.assertEqual(os.path.basename(os.path.dirname(bin_path)), pr if pr else 'FP32')
```
dkurt, line 69: Actually, this should be `pr if pr else (p if p else 'FP32')`, isn't it?

tools/downloader/topologies_test.py
```diff
@@ -0,0 +1,114 @@
...
+
+    def test_result_model(self):
+        ssd = omz.Topology('mobilenet-ssd', precision='FP32')
```
dkurt, line 72: `ssd = omz.Topology('mobilenet-ssd')`

tools/downloader/topologies.py
```diff
@@ -75,10 +75,16 @@ def get_ir(self, precision=None):
...
-        if not precision and res_precision not in self.precisions:
-            res_precision = next(iter(self.precisions))
+        if res_precision not in self.precisions:
+            raise ValueError('Incorrect precision value!\n    Current precision value is: \'' 
+                             + str(res_precision) + '\'.\n    Allowed precision values are: ' 
+                             + str(self.precisions))
+        elif self.precision and precision and precision != self.precision:
+            raise ValueError('Incorrect precision value!\n    Precision value for downloading is: \'' 
+                             + self.precision + '\'.\n    Precision value for get_ir is: \'' 
+                             + precision + '\'')
```
dkurt, line 87:
```python
if self.precision and precision and precision != self.precision:
   raise ValueError('Downloaded ({}) and requested ({}) precisions do not match'.format(self.precision, precision))

res_precision = precision if precision else self.precision
if res_precision not in self.precisions:
  raise ValueError('Incorrect precision value!\n    Current precision value is: \'' 
                             + str(res_precision) + '\'.\n    Allowed precision values are: ' 
                             + str(self.precisions))
  
res_precision = precision if precision

File: tools/downloader/topologies_test.py [hunk 28]
```diff
@@ -81,6 +84,8 @@ def test_result_model(self):
         image = np.expand_dims(image, axis=0)
         image = image.transpose((0, 3, 1, 2))
 
+        version = os.path.basename(inference_engine.__version__)
+        input_blob = next(iter(net.input_info)) if version == '4' else next(iter(net.inputs))
```
dkurt, line 88: Is there universal way to get input name? If no, please just hardcode it.

tools/downloader/topologies_test.py
```diff
@@ -0,0 +1,116 @@
...
+        self.assertTrue(os.path.exists(bin_path))
+
+        self.assertEqual(os.path.basename(xml_path), 'face-detection-0100.xml')
+        self.assertEqual(os.path.basename(bin_path), 'face-detection-0100.bin')
+
+        self.assertEqual(os.path.basename(os.path.dirname(xml_path)), p)
+        self.assertEqual(os.path.basename(os.path.dirname(bin_path)), p)
+
+    def test_load_intel_FP32(self):
+        face_detection = omz.Topology('face-detection-0100', precision='FP32')
```
dkurt, line 29: Why not to combine `test_load_intel_FP32` with `test_load_intel_without_precision`?

tools/downloader/topologies_test.py
```diff
@@ -0,0 +1,116 @@
...
+        self.assertEqual(os.path.basename(xml_path), 'face-detection-0100.xml')
+        self.assertEqual(os.path.basename(bin_path), 'face-detection-0100.bin')
+
+        self.assertEqual(os.path.basename(os.path.dirname(xml_path)), 'FP32')
+        self.assertEqual(os.path.basename(os.path.dirname(bin_path)), 'FP32')
+
+    def test_load_intel_INT8(self):
+        model_name = 'bert-large-uncased-whole-word-masking-squad-int8-0001'
+        face_detection = omz.Topology(model_name)
+        xml_path, bin_path = face_detection.get_ir('FP16-INT8')
```
dkurt, line 44: Please parametrize this test: 
```
None, None
None, FP16-INT8
FP16-INT8, None
FP16-INT8, FP16-INT8
```

likholat:
@IRDonch, @ngaloppo, Please review this PR.
This PR is a continuation of  https://github.com/openvinotoolkit/open_model_zoo/pull/304

tools/downloader/topologies_test.py
```diff
@@ -73,7 +72,7 @@ def test_result_model(self):
         image_path = os.path.join(os.path.split(os.path.dirname(os.path.realpath(__file__)))[0],
                                   'accuracy_checker', 'data', 'test_data', '1.jpg')
 
-        ie = IECore()
+        ie = inference_engine.IECore()
```
dkurt, line 64: Please use `from openvino.inference_engine import IECore` - it's convenient way.

tools/downloader/topologies.py
```diff
@@ -0,0 +1,99 @@
...
+    '''
+    {}
+    Model task type: {}
+
+    License: {}
+    '''
+
+    def __init__(self, model_name, *, precision=None, download_dir=None):
+        '''
+        Downloads target model. If the precision doesn't set, loads all the available models.
```
IRDonch, line 20:
```suggestion
        Downloads target model. If precision is None, loads all available models.
```

tools/downloader/topologies.py
```diff
@@ -0,0 +1,99 @@
...
+        If there is already downloaded model - retrieves the model from cache.
```
IRDonch, line 21:
```suggestion
        If the model has already been downloaded, retrieves the model from the cache.
```

tools/downloader/topologies.py
```diff
@@ -0,0 +1,99 @@
...
+        Creates an object which stores information about the downloaded model.
+
+        Parameters
+        ----------
+            model_name
+                Target model name.
+            precision
+                Target model precision. By default is None.
```
IRDonch, line 29: It seems short-sighted to only allow either one or all precisions to be downloaded. Why not accept a list of precisions?

tools/downloader/topologies.py
```diff
@@ -0,0 +1,99 @@
...
+            download_dir
+                Model download directory.
```
IRDonch, line 31: You need to document what the default value is (and what it means).

tools/downloader/topologies.py
```diff
@@ -0,0 +1,99 @@
...
+        '''
+        if not download_dir: 
+            download_dir = Path(Path('.').resolve() / 'models')
+        cache_dir = Path(Path(download_dir) / '.cache')
```
IRDonch, line 35:
```suggestion
        cache_dir = download_dir / '.cache'
```

tools/downloader/topologies.py
```diff
@@ -0,0 +1,99 @@
...
+
+        parser = argparse.ArgumentParser()
+        args = argparse.Namespace(all=False, list=None, name=model_name, print_all=False)   
+        topology = common.load_models_from_args(parser, args)[0]
+
+        model_dir = download_dir / topology.subdirectory
+
+        flags = ['--name=' + model_name,
+                 '--output_dir=' + str(download_dir).replace('\\', '\\\\'), 
+                 '--cache_dir=' + str(cache_dir).replace('\\', '\\\\')]
```
IRDonch, line 45: What's the purpose of all these replacements?

tools/downloader/topologies.py
```diff
@@ -0,0 +1,99 @@
...
+        if precision:
+            flags += ['--precisions=' + precision]
+        
+        downloader.main(flags)
```
IRDonch, line 49: Why do you need to use the downloader as a library? It's a tool. Why not just run it?

tools/downloader/topologies.py
```diff
@@ -0,0 +1,99 @@
...
+
+        self.name = model_name
+        self.model_dir = model_dir
+        self.download_dir = download_dir
+        self.precision = precision
+        self.precisions = topology.precisions
+        Topology.__doc__ = Topology.__doc__.format(topology.description, 
+                                                   topology.task_type, topology.license_url)
```
IRDonch, line 57: This seems questionable. There's only one `Topology.__doc__`, so if you create multiple `Topology` objects, the doc will just be set to the documentation for the most recent one. I would suggest to just add a `help_text` attribute to the object.

tools/downloader/topologies.py
```diff
@@ -0,0 +1,99 @@
+import os
+import argparse
+
+from pathlib import Path
+
+import common
+import downloader
+import converter
+
+class Topology:
```
IRDonch, line 10:
```suggestion
class Model:
```

tools/downloader/topologies.py
likholat: But then we can't use it like this:
```python
ssd = omz.Model('mobilenet-ssd', download_dir='/user/model')
```
We should do ```download_dir=Path('/user/model')```
Is it ok?

tools/downloader/topologies.py
dkurt: It goes from the closed PR. I just remember that there were some issues with Windows path so `\` should be replaced to `\\`. But we cannot write just `.replace('\', '\\')`

tools/downloader/topologies.py
dkurt: @likholat, please check once again something like this:
```
ssd = omz.Model('mobilenet-ssd')
alexnet = omz.Model('alexnet')
help(ssd)
```
what is output?

likholat: Your code is works, but it always returns documentation for the first model, we couldn't do like:
```python
ssd = omz.Model('mobilenet-ssd')
detection=omz.Model('face-detection-0100')
help(detection)
```
It returns documentation for ```mobilenet-ssd```

tools/downloader/topologies.py
likholat: But then which one precision value should be as ```self.precision```. Or for ```get_ir()``` we must always set precision value?

tools/downloader/topologies.py
dkurt: well, can it be fixed? In example by `self.__doc__` modification instead?

tools/downloader/topologies.py
eizamaliev: @likholat let FP32 be a default precision in `get_ir()` method.

likholat: @eizamaliev some intel models doesn't support ```FP32``` precision, for example ```bert-large-uncased-whole-word-masking-squad-int8-0001``` model.

tools/downloader/topologies.py
IRDonch: Yeah, fair enough. I think the version you have now is the best.

tools/downloader/topologies.py
IRDonch: I remember it from the previous PR, but I didn't get a satisfactory explanation then either. I'd like to know something more specific than "some issues".

tools/downloader/topologies.py
IRDonch: FWIW, while it _is_ possible to reassign `self.__doc__`, it will not be used by Python. Currently, using `help` on an object will show you the documentation for the object's _class_. I found some discussion about adding documentation support for objects (https://bugs.python.org/issue35449), but it looks like nothing has materialized so far.

If you ask me, I think you should just have `description` and `task_type` properties in the `Topology` class. Given that you can't get help on object instances in Python in general, I don't think we need to make it work for our class.

tools/downloader/topologies.py
```diff
@@ -0,0 +1,104 @@
...
+                                        + str(self.precision))
+        
+        if self.precision and precision and precision != self.precision[0]:
+            raise ValueError('Downloaded ({}) and requested ({}) precisions do not match'.format(self.precision[0], precision))
+
+        res_precision = precision if precision else self.precision[0]
+        if res_precision not in self.precisions:
+            raise ValueError('Incorrect precision value!\n    Current precision value is: \'' 
+                                        + str(res_precision) + '\'.\n    Allowed precision values are: ' 
+                                        + str(self.precisions))
```
IRDonch, line 92: This parameter validation code seems quite complicated, with complex, partially overlapping conditions. Wouldn't something like this be sufficient?

```python
if not precision:
    if not self.precision or len(self.precision) != 1:
        # error: ambiguous precision

    precision = self.precision[0]

if precision not in (self.precision or self.precisions):
    # error: precision wasn't downloaded
```

tools/downloader/topologies.py
```diff
@@ -0,0 +1,104 @@
...
+  
+        prefix = Path(self.model_dir) / res_precision / self.name
+
+        xml_path = str(prefix) + '.xml'
+        bin_path = str(prefix) + '.bin'
```
IRDonch, line 97: These should be `Path` objects. They are paths, so why not?

tools/downloader/topologies.py
```diff
@@ -0,0 +1,104 @@
...
+        if self.precision and precision and precision != self.precision[0]:
+            raise ValueError('Downloaded ({}) and requested ({}) precisions do not match'.format(self.precision[0], precision))
+
+        res_precision = precision if precision else self.precision[0]
+        if res_precision not in self.precisions:
+            raise ValueError('Incorrect precision value!\n    Current precision value is: \'' 
+                                        + str(res_precision) + '\'.\n    Allowed precision values are: ' 
+                                        + str(self.precisions))
+  
+        prefix = Path(self.model_dir) / res_precision / self.name
```
IRDonch, line 94:
```suggestion
        prefix = self.model_dir / res_precision / self.name
```

`model_dir` is already a `Path`.

tools/downloader/topologies_test.py
```diff
@@ -0,0 +1,103 @@
...
+import os
+import numpy as np
+import cv2 as cv
+
+import topologies as omz
+
+from pathlib import Path
+from parameterized import parameterized
+from openvino import inference_engine
+from openvino.inference_engine import IECore
```
IRDonch, line 11: Please reorder your imports to the order suggested by [PEP 8](https://www.python.org/dev/peps/pep-0008/#imports).

tools/downloader/topologies_test.py
```diff
@@ -0,0 +1,103 @@
...
+
+class TestTopologies(unittest.TestCase):
+    @parameterized.expand(['FP16', 'FP32'])
```
IRDonch, line 14: I don't think testing both of these does any good. The code path is exactly the same; you're not gaining any more coverage, but you are increasing the run time.

tools/downloader/topologies_test.py
```diff
@@ -0,0 +1,103 @@
...
+        xml_path, bin_path = face_detection.get_ir(precision=p)
+        
+        self.assertTrue(os.path.exists(xml_path))
+        self.assertTrue(os.path.exists(bin_path))
+
+        self.assertEqual(os.path.basename(xml_path), 'face-detection-0100.xml')
+        self.assertEqual(os.path.basename(bin_path), 'face-detection-0100.bin')
+
+        self.assertEqual(os.path.basename(os.path.dirname(xml_path)), p)
+        self.assertEqual(os.path.basename(os.path.dirname(bin_path)), p)
```
IRDonch, line 26: I don't think the conditions verified by these assertions are part of the library's contract. The documentation just says that the `get_ir` will give you a path to the model files; it doesn't say what the path will _be_. What you should be checking is that a) the model can be loaded into Inference Engine, and b) that it actually has the requested precision. (To be fair, I'm not sure if the latter _can_ be checked. But if you can do it, you should.)

tools/downloader/topologies_test.py
```diff
@@ -0,0 +1,103 @@
...
+
+    @parameterized.expand([[None, 'FP16-INT8'], [['FP16-INT8'], None],
+                           [['FP16-INT8'], 'FP16-INT8']])
+    def test_load_intel_INT8(self, p, pr):
```
IRDonch, line 30: These parameter names need to be more descriptive. What is `p`? What is `pr`?

tools/downloader/topologies_test.py
```diff
@@ -0,0 +1,103 @@
...
+        model_name = 'bert-large-uncased-whole-word-masking-squad-int8-0001'
```
IRDonch, line 31: This seems like a poor choice for a test case. It's one of the largest models we got, so the test will take forever to run.

tools/downloader/topologies_test.py
```diff
@@ -0,0 +1,103 @@
...
+            xmax = detection[5]
+            ymax = detection[6]
+
+        delta = 1e-05
+        self.assertEqual(label, 7)
+        self.assertAlmostEqual(confidence, 0.9891584, delta=delta)
+        self.assertAlmostEqual(xmin, 0.34685153, delta=delta)
+        self.assertAlmostEqual(ymin, 0.2038618, delta=delta)
+        self.assertAlmostEqual(xmax, 0.5851372, delta=delta)
+        self.assertAlmostEqual(ymax, 0.87824464, delta=delta)
```
IRDonch, line 100: I don't think we need this. It tests the _model_ rather than the API, and we don't need tests for models - we already have accuracy validation. The problem with it is that it makes the tests high-maintenance - if this model is removed, we'll have to redo the test to match another model. Also, the results may diverge from the reference values due to changes in IE or CPU features supported on the machine that runs them.

tools/downloader/topologies.py
```diff
@@ -0,0 +1,104 @@
+import os
```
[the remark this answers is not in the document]
dkurt: But this is API for downloader+converter. Why not `tools/downloader/open_model_zoo.py` or at least `tools/open_model_zoo.py`?

tools/downloader/topologies.py
dkurt: Why not to have 
```python
download_dir = Path(download_dir)
```

at the beginning of `__init__`? So 
```
omz.Model('modelname', download_dir='/path/to/model') 
```
will be also valid

tools/downloader/topologies.py
```diff
@@ -0,0 +1,104 @@
...
+        args = argparse.Namespace(all=False, list=None, name=model_name, print_all=False)   
+        topology = common.load_models_from_args(parser, args)[0]
+
+        model_dir = download_dir / topology.subdirectory
+
+        flags = ['--name=' + model_name,
+                 '--output_dir=' + str(download_dir).replace('\\', '\\\\'), 
+                 '--cache_dir=' + str(cache_dir).replace('\\', '\\\\')]
+        if precisions:
+            flags += ['--precisions=' + ','.join([str(elem) for elem in precisions]) ]
```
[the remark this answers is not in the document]
likholat: Now it takes a list of precision values. We should use it like:
```python
ssd = omz.Model('face-detection-0100', precisions=['FP16','FP32'])
```

tools/downloader/topologies_test.py
likholat: I can't find a way to check model precision.

tools/downloader/topologies.py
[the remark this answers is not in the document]
IRDonch: > But this is API for downloader+converter.

I disagree with this assessment. It is an API for Open Model Zoo, which _uses_ the downloader & converter to provide its functionality.

> Why not tools/downloader/open_model_zoo.py or at least tools/open_model_zoo.py?

Because it's not a tool. It's a library.

tools/downloader/topologies.py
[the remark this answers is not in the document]
IRDonch: It seems that this is no longer the case. Why did you change it to a string? Structured data should be passed via a structured type.

tools/downloader/topologies_test.py
```diff
@@ -0,0 +1,75 @@
...
+
+        input_blob = infer_request.input_blobs[next(iter(net.input_info))]
+        output_blob = infer_request.output_blobs[next(iter(net.outputs))]
+
+        self.assertEqual(input_blob.tensor_desc.dims, [1, 3, 256, 256])
+        self.assertEqual(output_blob.tensor_desc.dims, [1, 1, 200, 7])
+
+    @parameterized.expand([[None, 'FP16-INT8'], ['FP16-INT8', None],
+                           ['FP16-INT8', 'FP16-INT8']])
+    def test_load_intel_INT8(self, init_precision, get_ir_precision):
```
IRDonch, line 34: It seems like this test case is a superset of the one above - it uses a different model and a different precision, but otherwise it's equivalent to the `[None, 'FP16-INT8']` instance of this test case.

tools/downloader/topologies.py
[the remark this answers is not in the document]
likholat: I changed it to a string because for ```downloader``` we should pass ```--precision```  as comma separated string of values.

Please specify how do you want the user to pass input precision values:
1. ```ssd = omz.Model('mobilenet-ssd', precisions='FP16,FP32')```
or
2. ```ssd = omz.Model('mobilenet-ssd', precisions=['FP16', 'FP32'])```

tools/downloader/topologies.py
likholat: We needn't this replace for Windows, because user can't write like:
```python
ssd = omz.Model('mobilenet-ssd', precisions='FP16', download_dir='C:\User\model')
```
it causes
```
SyntaxError: (unicode error) 'unicodeescape' codec can't decode bytes in position 2-3: truncated \UXXXXXXXX escape
```

We should do:
```python
ssd = omz.Model('mobilenet-ssd', precisions='FP16', download_dir='C:\\User\\model')
```
or
```python
ssd = omz.Model('mobilenet-ssd', precisions='FP16', download_dir='C:/User/model')
```

dkurt:
> For topology.py, downloader.py and converter.py change imports:
import common -> from . import common
import downloader -> from . import downloader
import converter -> from . import converter

Is there a way to avoid changing these files?

tools/downloader/topologies_test.py
```diff
@@ -0,0 +1,75 @@
+import unittest
```
eizamaliev, line 1: I think there is no need in special tests, since it could be done in demo [testing](https://github.com/openvinotoolkit/open_model_zoo/blob/master/demos/tests/run_tests.py). @IRDonch, @Wovchena are you agree?

tools/downloader/open_model_zoo_test.py
```diff
@@ -0,0 +1,57 @@
+import unittest
+import os
+import numpy as np
+import cv2 as cv
+
+import open_model_zoo as omz
+
+from pathlib import Path
```
dkurt, line 8: unused import

tools/downloader/open_model_zoo_test.py
```diff
@@ -0,0 +1,57 @@
...
+from parameterized import parameterized
+from openvino import inference_engine
```
dkurt, line 10: unused import

tools/downloader/open_model_zoo.py
```diff
@@ -0,0 +1,107 @@
...
+                Target model name.
+            precisions
+                String with comma separated precisions of weights to be downloaded.
+                By default is None.
+            download_dir
+                Model download directory. By default creates a folder 'models' 
+                in current directory and downloads to it. 
+        '''
+        download_dir = Path(download_dir)
+        cache_dir = download_dir / '.cache'
```
dkurt, line 36: Let's add `cache_dir` also to the arguments.
```
def __init__(self, model_name, *, precisions=None, download_dir='models', cache_dir=None):
cache_dir = cache_dir or download_dir / '.cache'
```

tools/downloader/topologies.py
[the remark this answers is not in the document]
IRDonch: I think 2 makes the most sense.

tools/downloader/topologies_test.py
IRDonch: I don't agree. Demo tests are pretty superficial. They can only test whatever usages of the API the demos contain, and can only check whether the demo crashed or not, not any fine-grained checks. You get better coverage with dedicated tests.

IRDonch:
You seem to be missing a block of hidden comments. Make sure to expand all comments.

tools/downloader/topologies.py
likholat: Please tell me which directory should contain the ```open_model_zoo.py``` file.

tools/downloader/open_model_zoo_test.py
```diff
@@ -0,0 +1,54 @@
...
+
+        input_blob = infer_request.input_blobs[next(iter(net.input_info))]
+        output_blob = infer_request.output_blobs[next(iter(net.outputs))]
+
+        self.assertEqual(input_blob.tensor_desc.dims, [1, 3, 256, 256])
+        self.assertEqual(output_blob.tensor_desc.dims, [1, 1, 200, 7])
+
+    @parameterized.expand([[['FP16'],'FP16'], [['FP32'],'FP32'],
+                           [None, 'FP16'], [None, 'FP32'],
+                           [['FP16'], None], [['FP32'], None]])
```
dkurt, line 34: Better to keep string params but do

```python
        ssd = omz.Model('mobilenet-ssd', precisions=[init_precision])
```

likholat:
@IRDonch, please, tell me, which directory is the best place to put the ```open_model_zoo.py```?

IRDonch:
@likholat I already made a suggestion above:

> I would suggest `libs/python/open_model_zoo.py`.

libs/python/open_model_zoo.py
```diff
@@ -41,71 +41,30 @@ def __init__(self, model_name, *, precisions=None, download_dir='models', cache_
...
+        if not os.path.exists(self.xml_path) or not os.path.exists(self.bin_path):
+            converter.main(['--name=' + model_name, '--precisions=' + res_precision,
+                        '--download_dir=' + str(download_dir)])
 
-        return xml_path, bin_path
+    def description(self):
+        return self._description
+
+    def task_type(self):
+        return self._task_type
```
dkurt, line 71: I think that now it's OK to keep them as fields (just `description` and `task_type`)

druzhkov-paul:
@likholat @dkurt @IRDonch May I know why has this PR stalled?

dkurt:
@druzhkov-paul, good question. @likholat, let's revise our latest progress. For Python demos such kind of features would be very nice.

likholat:
@IRDonch @eizamaliev, please review it again

libs/python/open_model_zoo.py
```diff
@@ -0,0 +1,68 @@
+import sys
+import os
+import argparse
+
+from pathlib import Path
+
+sys.path.insert(0, str(Path(__file__).parent / '../../tools/downloader/src'))
```
dkurt, line 7: This is not reliable. `src` is internal namerelated to OMZ CI

jkamelin:
Jenkins please retry a build

jkamelin:
Jenkins please retry a build

vladimir-dudnik:
@nignatovsky please look at this, some CI issue

jkamelin:
Jenkins please retry a build

tools/model_tools/tests/open_model_zoo_test.py
```diff
@@ -0,0 +1,96 @@
...
+        ie = IECore()
+        net = ie.read_network(xml_path)
+
+        input_name = next(iter(net.input_info))
+        output_name = next(iter(net.outputs))
+
+        self.assertEqual(net.input_info[input_name].input_data.shape, [1, 1, 256, 256])
+        self.assertEqual(net.outputs[output_name].shape, [1, 2, 256, 256])
+
+    def test_load_from_prtained(self):
```
vladimir-dudnik, line 40: if that means test_load_from_pretrained, then better to name it as test_load_intel

tools/model_tools/src/openvino/model_zoo/open_model_zoo.py
```diff
@@ -0,0 +1,176 @@
...
+        self.name = name
+        self.subdirectory = subdirectory
+
+        self.net = None
+        self.exec_net = None
+        self._accuracy_config = None
+        self._model_config = None
+
+    @classmethod
+    def download_model(cls, model_name, *, precision='FP32', download_dir='models', cache_dir=None):
```
vladimir-dudnik, line 28: probably better simple `download` to avoid phrase like model.download_model

tools/model_tools/src/openvino/model_zoo/open_model_zoo.py
```diff
@@ -0,0 +1,176 @@
...
+                    self._model_config = yaml.safe_load(config_file)
+
+        return self._model_config
+
+    def inputs(self, ie=None):
+        if self.net is None:
+            try:
+                self.net = ie.read_network(self.model_path)
+            except AttributeError:
+                raise TypeError('ie argumnet must be of IECore type.')
```
eaidova, line 161:
```suggestion
                raise TypeError('ie argument must be of IECore type.')
```

tools/model_tools/src/openvino/model_zoo/open_model_zoo.py
```diff
@@ -0,0 +1,176 @@
...
+        input_blob = self.net.input_info
+
+        return input_blob
+
+    def outputs(self, ie=None):
+        if self.net is None:
+            try:
+                self.net = ie.read_network(self.model_path)
+            except AttributeError:
+                raise TypeError('ie argumnet must be of IECore type.')
```
eaidova, line 172:
```suggestion
                raise TypeError('ie argument must be of IECore type.')
```

tools/model_tools/src/openvino/model_zoo/open_model_zoo.py
jkamelin: Done

tools/model_tools/tests/open_model_zoo_test.py
jkamelin: This is to test loading from locale file. `test_load_intel` already exists for downloading intel model.

jkamelin:
Jenkins please retry a build

models/public/brain-tumor-segmentation-0001/model.yml
```diff
@@ -36,9 +36,11 @@ files:
     source:
       $type: google_drive
       id: 1T2ldCISt5N2TEUfv6ggRB1dw0cucTqT2
+input_info:
+  - name: data_crop
+    shape: [1, 4, 128, 128, 128]
+    layout: NCWHD
```
eaidova, line 42:
```suggestion
    layout: NCDHW
```

models/public/brain-tumor-segmentation-0002/model.yml
```diff
@@ -24,8 +24,11 @@ files:
     source:
       $type: google_drive
       id: 1NhH51NgKJKhu0Bax_IKQ_Byvoe2bItPX
+input_info:
+  - name: '0'
+    shape: [1, 4, 128, 128, 128]
+    layout: NCWHD
```
eaidova, line 30:
```suggestion
    layout: NCDHW
```

models/public/forward-tacotron/forward-tacotron-duration-prediction/model.yml
```diff
@@ -24,9 +24,11 @@ postprocessing:
   - $type: unpack_archive
     format: zip
     file: forward-tacotron-duration-prediction.zip
+input_info:
+  - name: input_seq
+    shape: [1, 241]
+    layout: NC
```
eaidova, line 30: it is ok to use C, but later possibly it will be changed to S (sequence len)

tools/model_tools/src/openvino/model_zoo/_configuration.py
```diff
@@ -95,6 +98,13 @@ def deserialize(cls, model, name, subdirectory, composite_model_name):
             framework = validation.validate_string_enum('"framework"', model['framework'],
                 _common.KNOWN_FRAMEWORKS.keys())
 
+            input_info = []
+            for input in model.get('input_info', []):
+                input_name = validation.validate_string('"input name"', input['name'])
+                shape = validation.validate_nonnegative_int_list('"input shape"', input.get('shape', []))
```
eaidova, line 104: why do you need this check?
-1 used for specifying dynamic shape for MO conversion, so it is valid value if we speak about shape in general understanding, if we want to have only static shapes, then yes, 1 is minimal possible value

tools/model_tools/src/openvino/model_zoo/open_model_zoo.py
```diff
@@ -0,0 +1,193 @@
...
+            self._load(ie, device)
+
+        input_names = next(iter(self.net.input_info))
+        for input_name, value in inputs.items():
+            if input_name not in input_names:
+                raise ValueError('Unknown input name {}'.format(input_name))
+
+            input_shape = self.net.input_info[input_name].input_data.shape
+            if input_shape != list(value.shape):
+                value = cv2.resize(value, input_shape)
```
eaidova, line 131: not sure that it will work for all cases, what about non-vision models?

tools/model_tools/src/openvino/model_zoo/open_model_zoo.py
```diff
@@ -0,0 +1,193 @@
...
+        model_path = str(model_path)
+
+        return cls(name, model_path, description, task_type, subdirectory)
+
+    def _load(self, ie, device):
+        self.net = ie.read_network(self.model_path)
+        self.exec_net = ie.load_network(self.net, device)
+
+    def __call__(self, inputs, ie, device='CPU'):
+        if self.exec_net is None:
```
eaidova, line 121: exec net can be exists, but loaded on another device (e.g. previously user use the same method specifying GPU as device)

tools/model_tools/src/openvino/model_zoo/open_model_zoo.py
```diff
@@ -0,0 +1,193 @@
...
+            self._load(ie, device)
+
+        input_names = next(iter(self.net.input_info))
```
eaidova, line 124: why you need this? It returns name of first input layer in the model, what about multi input networks?

tools/model_tools/src/openvino/model_zoo/open_model_zoo.py
eaidova: and BTW, even when input is image, usually input tensor is 4D (NCHW or NHWC), while opencv expects that image is 3D HWC np array

so I think we should avoid to use any preprocessing on our side and expect that user already prepare input tensor, or handle it more generally

tools/model_tools/src/openvino/model_zoo/open_model_zoo.py
```diff
@@ -0,0 +1,193 @@
...
+        return self._model_config
+
+    def inputs(self, ie=None):
+        if self.net is None:
+            try:
+                self.net = ie.read_network(self.model_path)
+            except AttributeError:
+                raise TypeError('ie argument must be of IECore type.')
+
+        input_blob = self.net.input_info
```
eaidova, line 164: it is dictionary, where can be several keys, each is description of input_blob
```suggestion
        input_blobs = self.net.input_info
```

eaidova: BTW, is there way to get not network input, but config input_info as property

tools/model_tools/src/openvino/model_zoo/open_model_zoo.py
```diff
@@ -0,0 +1,193 @@
...
+        self.description = description
+        self.task_type = task_type
+        self.name = name
+        self.subdirectory = subdirectory
+
+        self.net = None
+        self.exec_net = None
+        self._accuracy_config = None
+        self._model_config = None
+        self._model = None
```
eaidova, line 26: is it possible to provide ie into consturtor and create network, when it will be necessary as alternative of providing it into input/output getters and inference?

tools/model_tools/src/openvino/model_zoo/open_model_zoo.py
```diff
@@ -0,0 +1,193 @@
...
+            try:
+                self.net = ie.read_network(self.model_path)
+            except AttributeError:
+                raise TypeError('ie argument must be of IECore type.')
+
+        output_blob = self.net.outputs
+
+        return output_blob
+
+    def input_shape(self, name):
```
eaidova, line 179: possibly we need more meaningful name for it, preferable_input_shape pr static_input_shape. it is not obvious why we need this method now (real network shape can be different)

tools/model_tools/src/openvino/model_zoo/open_model_zoo.py
```diff
@@ -0,0 +1,193 @@
...
+        if self._model_config is None:
+            config_path = _common.MODEL_ROOT / self.subdirectory / 'model.yml'
+            if config_path.exists():
+                with config_path.open('rb') as config_file, \
+                        validation.deserialization_context('Loading config "{}"'.format(config_path)):
+                    self._model_config = yaml.safe_load(config_file)
+
+        return self._model_config
+
+    def inputs(self, ie=None):
```
eaidova, line 157: can it be property?

tools/model_tools/src/openvino/model_zoo/open_model_zoo.py
```diff
@@ -0,0 +1,193 @@
...
+            try:
+                self.net = ie.read_network(self.model_path)
+            except AttributeError:
+                raise TypeError('ie argument must be of IECore type.')
+
+        input_blob = self.net.input_info
+
+        return input_blob
+
+    def outputs(self, ie=None):
```
eaidova, line 168: can it be property?

tools/model_tools/src/openvino/model_zoo/_configuration.py
jkamelin: Removed the check

tools/model_tools/src/openvino/model_zoo/open_model_zoo.py
jkamelin: Done

tools/model_tools/src/openvino/model_zoo/open_model_zoo.py
jkamelin: Done

tools/model_tools/src/openvino/model_zoo/open_model_zoo.py
jkamelin: Updated

tools/model_tools/src/openvino/model_zoo/open_model_zoo.py
jkamelin: Renamed to `preferable_input_shape`

jkamelin:
Jenkins please retry a build

tools/model_tools/src/openvino/model_zoo/open_model_zoo.py
```diff
@@ -0,0 +1,204 @@
...
+
+    def __call__(self, inputs, device='CPU'):
+        if self.ie is None:
+            raise TypeError('ie must be of IECore type.')
+
+        self.net = self.ie.read_network(self.model_path)
+        self.exec_net = self.ie.load_network(self.net, device)
+
+        input_names = self.net.input_info.keys()
+        for input_name in inputs.keys():
```
eaidova, line 136:
```suggestion
        for input_name in inputs:
```

tools/model_tools/src/openvino/model_zoo/open_model_zoo.py
```diff
@@ -0,0 +1,204 @@
...
+            model = _configuration.load_models_from_args(parser, args, _common.MODEL_ROOT,
+                        mode=_configuration.ModelLoadingMode.ignore_composite)[0]
+        except SystemExit:
+            model = None
+
+        return model
+
+    def __call__(self, inputs, device='CPU'):
+        if self.ie is None:
+            raise TypeError('ie must be of IECore type.')
```
eaidova, line 130: message is not clear.
There are 2 possible situations:
1. self.ie is None - ie is not specified (during creating model for example
2. self.ie is provided, but it is object of wrong type (e.g. instead of openvino.inference_engine.IECore provided openvino.runtime.Core or something different)

so your exception looks misleading, if user does not specify ie at all

tools/model_tools/src/openvino/model_zoo/open_model_zoo.py
```diff
@@ -0,0 +1,204 @@
...
+
+        self.net = self.ie.read_network(self.model_path)
```
eaidova, line 132: do we really need to read network on each call?

tools/model_tools/src/openvino/model_zoo/open_model_zoo.py
```diff
@@ -0,0 +1,204 @@
...
+        self.exec_net = self.ie.load_network(self.net, device)
```
eaidova, line 133: probably, we can cache exec network, but we need properly handle device changing (store loaded device value inside class and maybe interpret call without device previously used device if model already loaded and default value if not)

tools/model_tools/src/openvino/model_zoo/open_model_zoo.py
eaidova: BTW, do you have plans to migrate on OV2.0? (it has some differences)

tools/model_tools/src/openvino/model_zoo/open_model_zoo.py
jkamelin: Done

tools/model_tools/src/openvino/model_zoo/open_model_zoo.py
jkamelin: Done

tools/model_tools/src/openvino/model_zoo/open_model_zoo.py
jkamelin: Loading updated

tools/model_tools/src/openvino/model_zoo/open_model_zoo.py
jkamelin: Messages updated

> BTW, do you have plans to migrate on OV2.0? (it has some differences)

I do. It is in progress

tools/model_tools/src/openvino/model_zoo/open_model_zoo.py
```diff
@@ -0,0 +1,217 @@
...
+        if self.net is None:
+            try:
+                self.net = self.ie.read_model(self.model_path)
+            except AttributeError:
+                raise TypeError('ie is not specified or is of the wrong type.'
+                                'Please check ie is of openvino.runtime.Core type.')
+
+        output_blob = self.net.outputs
+
+        return output_blob
```
eaidova, line 198:
```suggestion

        return self.net.outputs
```

tools/model_tools/src/openvino/model_zoo/open_model_zoo.py
```diff
@@ -0,0 +1,217 @@
...
+        self.net = self.ie.read_model(self.model_path)
+        self.exec_net = self.ie.compile_model(self.net, device)
+        self.device = device
+
+    def __call__(self, inputs, device='CPU'):
+        if self.ie is None:
+            raise TypeError('ie is not specified or is of the wrong type.'
+                            'Please check ie is of openvino.runtime.Core type.')
+
+        if self.exec_net is None or device != self.device:
```
eaidova, line 139: as you migrate on ov api 2.0, please use it terminology -
model instead of net, compiled_model instead of exec_net

tools/model_tools/src/openvino/model_zoo/open_model_zoo.py
```diff
@@ -0,0 +1,217 @@
...
+            self._load(device)
+
+        input_names = [input.get_any_name() for input in self.net.inputs]
+        infer_request = self.exec_net.create_infer_request()
```
eaidova, line 143: if it is not required for you to use the same infer request for each inference you can use
self.exec_net.infer_new_request, which combines request creation and inference inside

tools/model_tools/src/openvino/model_zoo/open_model_zoo.py
```diff
@@ -0,0 +1,217 @@
...
+
+        for input_name in inputs:
+            if input_name not in input_names:
+                raise ValueError('Unknown input name {}'.format(input_name))
+
+        output_names = [output.get_any_name() for output in self.net.outputs]
```
eaidova, line 149: in conception ov2.0, tensors can have several names or do not have at all, user can access to outputs not only by tensor name, but node name or even index, so I think this remapping can return misleading result in some cases, possibly stay output dict as is will be better, or outputs method should also return strings

tools/model_tools/src/openvino/model_zoo/open_model_zoo.py
```diff
@@ -0,0 +1,217 @@
...
+        infer_request.infer(inputs)
+        result = {name: infer_request.get_tensor(name).data for name in output_names}
+        return result
+
+    @property
+    def accuracy_checker_config(self):
+        if self._accuracy_config is None:
```
eaidova, line 156: as I can see, configs are read only properties, why do you need to cache them in class if they can not be changed?
yaml reading is not so time consuming operation for caching it

tools/model_tools/src/openvino/model_zoo/open_model_zoo.py
jkamelin: Done

tools/model_tools/src/openvino/model_zoo/open_model_zoo.py
jkamelin: Done

tools/model_tools/src/openvino/model_zoo/open_model_zoo.py
jkamelin: Done, returns dict as is

tools/model_tools/src/openvino/model_zoo/open_model_zoo.py
jkamelin: Removed config caching

> I can understand, why you need has model config in memory, it is required for access to some fields (input_shapes, layouts e.t.c.), but accuracy checker config is not so informative

There were some requests to provide a way to access info from accuracy checker configs.

vladimir-dudnik:
Jenkins please retry a build

tools/model_tools/src/openvino/model_zoo/open_model_zoo.py
```diff
@@ -0,0 +1,293 @@
...
+    def model_api_inference(self, inputs, device='CPU', model_creator=None, **kwargs):
+        self.device = device
+        if model_creator is None:
+            model_creator = self.model_creator
+
+        self._create_model_api_instance(model_creator, **kwargs)
+
+        preprocessed_inputs, meta = self.model_api.preprocess(inputs)
+        raw_results = self.model_api.infer_sync(preprocessed_inputs)
+        results = self.model_api.postprocess(raw_results, meta)
```
anzhella-pankratova, line 279:
```suggestion
        results = self.model_api(inputs)
```
Since the Python Model API wrappers have `__call__()` operator, which performs the same actions

tools/model_tools/src/openvino/model_zoo/_common.py
```diff
@@ -20,6 +20,30 @@
...
+        @classmethod
+        def create_model(cls, name=None, model_adapter=None, configuration=None):
+            return None
+
+    class Classification:
+        def __init__(self, model_adapter=None, configuration=None):
+            pass
+
+    def create_core():
+        return None
```
Wovchena, line 42: Is there any difference between `pass` and `return None`?

Wovchena (commented):
Do you know if there is a way to have `openvino.model_zoo.open_model_zoo` and `openvino.model_zoo.model_api.models` both in editable mode at the same time through `pip -e` or `PYTHONPATH`? I only found a way with `py -m pip install demos/common/python/ tools/model_tools/`, but it copies the files

Wovchena:
There are two ways of creating a namespace package. Omitting \_\_init__.py or putting `__path__ = __import__('pkgutil').extend_path(__path__, __name__)` into it. We are forced to use the second option for `openvino` namespaces after the main openvino project. To be consistent for `model_zoo` namespace we should follow the same option. @jkamelin, please, add it

tools/model_tools/src/openvino/model_zoo/_common.py
```diff
@@ -98,6 +99,31 @@ def end_session(self, category): pass
         def force_shutdown(self, timeout): pass
 
 
+try:
+    from model_api.adapters import create_core, OpenvinoAdapter
+    from model_api.models import Model, Classification
+    from model_api.pipelines import get_user_config
+except ImportError:
```
Wovchena, line 106: `model_api` becomes a part of openvino-dev just like model tools. If model_tools exists, `model_api` is also available. I think you can remove wrapping try except

tools/model_tools/tests/open_model_zoo_test.py
```diff
@@ -0,0 +1,138 @@
+import numpy as np
+import unittest
+import os
+
+import openvino.model_zoo.open_model_zoo as omz
```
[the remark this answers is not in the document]
eaidova: for me, good sounds if it will be `openvino.model_zoo` without internal namespace  or `openvino.model_zoo.models`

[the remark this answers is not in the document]
jkamelin: The proposed name was Open Model Zoo Python API, probably `openvino.model_zoo.api` is better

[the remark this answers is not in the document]
Wovchena: It clashes with modelAPI

IRDonch:
> We are forced to use the second option for openvino namespaces after the main openvino project.

FWIW, I don't think you're actually forced to use this option, because native namespace packages are compatible with `pkgutil`-style ones. `pkgutil.extend_path` will locate all directories named after the package, regardless of whether they have an `__init__.py` file or not.

Wovchena:
Yes, I realized that after I posted the message. But still, it feels reasonable to go with `__path__ = __import__('pkgutil').extend_path(__path__, __name__)` to stay aligned with root openvino init

tools/model_tools/src/openvino/model_zoo/_common.py
vladimir-dudnik: as far as I know model_api becomes part of openvino-dev at 2022.2 but not in 2022.1

cc @slyubimt is that correct?

slyubimt: model_api is not targeted for 2022.1.

vladimir-dudnik: ok, disregard my comment, this PR is targeted for 2022.2, just mistakenly thought we are some in other PR

jkamelin: Updated imports

Wovchena (changes requested):

tools/model_tools/src/openvino/model_zoo/open_model_zoo.py
```diff
@@ -0,0 +1,298 @@
...
+        name = model.name
+        subdirectory = model.subdirectory
+
+        if not os.path.exists(model_path) or not os.path.exists(bin_path):
+            omz_converter.converter(['--name=' + model_name, '--precisions=' + precision,
+                        '--download_dir=' + str(download_dir)])
+        return cls(name, model_path, description, task_type, subdirectory, model.architecture_type, ie, download_dir)
+
+    @classmethod
+    def from_pretrained(cls, model_path, *, task_type=None, ie=None):
```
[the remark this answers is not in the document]
jkamelin: The idea was to create OMZModel from local file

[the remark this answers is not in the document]
Wovchena: Who would need it? OV runtime already solves the problem of performing general inference given a model file

vladimir-dudnik:
@jkamelin there is merge conflict

PR #1362, opened 2020-07-29
Add missed `operator<<` to fix IEVersion print in demos

vurusovs:
Hello!

I found that OMZ demos doesn't print IE version information correctly (it print object address instead of human-readable message, example: https://github.com/openvinotoolkit/openvino/issues/1330#issuecomment-661587879). The rootcause is in missed `operator<<` which weren't copied from https://github.com/openvinotoolkit/openvino/blob/master/inference-engine/samples/common/samples/common.hpp#L88

This PR should fix the issue

### DIFF
--- a/demos/common/include/samples/common.hpp
+++ b/demos/common/include/samples/common.hpp
@@ -50,6 +50,22 @@ static UNUSED std::string fileNameNoExt(const std::string &filepath) {
     return filepath.substr(0, pos);
 }
 
+static UNUSED std::ostream &operator<<(std::ostream &os, const InferenceEngine::Version *version) {
+    os << "\n\tAPI version ............ ";
+    if (nullptr == version) {
+        os << "UNKNOWN";
+    } else {
+        os << version->apiVersion.major << "." << version->apiVersion.minor;
+        if (nullptr != version->buildNumber) {
+            os << "\n\t" << "Build .................. " << version->buildNumber;
+        }
+        if (nullptr != version->description) {
+            os << "\n\t" << "Description ....... " << version->description;
+        }
+    }
+    return os;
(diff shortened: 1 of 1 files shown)


demos/common/include/samples/common.hpp
```diff
@@ -50,6 +50,22 @@ static UNUSED std::string fileNameNoExt(const std::string &filepath) {
...
+        if (nullptr != version->buildNumber) {
+            os << "\n\t" << "Build .................. " << version->buildNumber;
+        }
+        if (nullptr != version->description) {
+            os << "\n\t" << "Description ....... " << version->description;
+        }
+    }
+    return os;
+}
+
```
IRDonch, line 68: There already is an output operator for IE versions, directly below this. You just need to update the demo code to dereference the version when printing it.

vurusovs: Perhaps you're right, but it will lead to some difference between OpenVINO IE samples and OMZ demos. Anyway, I think, if OMZ contains some code copied from IE samples/common, there are no limitations to extend the copied part

IRDonch: > Perhaps you're right, but it will lead to some difference between OpenVINO IE samples and OMZ demos.

It doesn't matter. There's already plenty of differences.

> Anyway, I think, if OMZ contains some code copied from IE samples/common, there are no limitations to extend the copied part

It doesn't work like that. You can't just add code because it's in IE; it has to be good.

vladimir-dudnik:
I'm going to close this, as initial issue fixed by [PR-1514](https://github.com/openvinotoolkit/open_model_zoo/pull/1514)

PR #1367, opened 2020-07-30
Added GPT2-LM-HEAD model

Anna-Mironova:


tools/accuracy_checker/accuracy_checker/adapters/nlp.py
```diff
@@ -121,7 +121,10 @@ def configure(self):
     def process(self, raw, identifiers=None, frame_meta=None):
         raw_output = self._extract_predictions(raw, frame_meta)
         result = []
-        for identifier, token_output in zip(identifiers, raw_output[self.logits_out]):
+        logits = raw_output[self.logits_out]
+        if len(logits.shape) == 4:
+            logits = logits[0]
```
eaidova, line 126: could you please provide example which shape expected and how it works?

Anna-Mironova: As a result, we need to get the last 2 dimensions in `token_output`. `logits` shape can be either 1, 1024, 50257 or 1, 1, 1024, 50257. To handle the second case, added this code

eaidova: np.squeeze(logits, axis=1), here make sence to remove second dim instead of first which contains batch

eaidova: and it is also can be done inside cycle, not before (e.g. if model will be run with batch size > 1), we will have output [42, 1, 1024, 50257]

PR #1377, opened 2020-08-04
AC: Added to_polygon function

Anna-Mironova:


tools/accuracy_checker/accuracy_checker/representation/segmentation_representation.py
```diff
@@ -17,6 +17,8 @@
 from enum import Enum
 from pathlib import Path
 from copy import deepcopy
+from skimage.measure import find_contours
+from shapely.geometry import Polygon
```
eaidova, line 21: it is not AC core requirements, it means that user may not have these libs installed, please make them optional.
BTW, regarding contours finding, the same functionality provided in opencv, probably, it will be better if you will use it (because opencv provided in the same package and do not require specific dependencies.
Also, one of the posible way of this function usage is writing to json file or providing via API to UI (written on another programming language), probably just return points for build building polygon contours will be enough.

tools/accuracy_checker/accuracy_checker/representation/segmentation_representation.py
```diff
@@ -50,7 +52,30 @@ class GTMaskLoader(Enum):
...
+                if poly.bounds:
+                    if i not in polygons:
+                        polygons[i] = [poly.bounds]
+                    else:
+                        polygons[i].append(poly.bounds)
+
+        return polygons
+
+    @staticmethod
+    def _encode_mask(mask, segmentation_colors):
```
eaidova, line 77: why this method need?

tools/accuracy_checker/accuracy_checker/representation/segmentation_representation.py
```diff
@@ -101,6 +126,21 @@ def _load_mask(self):
 
         return self._mask
 
+    @staticmethod
+    def _encode_mask(mask, segmentation_colors):
+        result = mask
+        if segmentation_colors:
+            mask = mask.astype(int)
+            num_channels = len(mask.shape)
+            encoded_mask = np.zeros((mask.shape[0], mask.shape[1]), dtype=np.int16)
```
eaidova, line 135: why int16?

eaidova:
@Anna-Mironova please add unit-tests to make sure that added functionality works correctly and cover all cases

tools/accuracy_checker/accuracy_checker/representation/segmentation_representation.py
```diff
@@ -50,7 +51,26 @@ class GTMaskLoader(Enum):
...
+
+            contours, _ = cv.findContours(binary_mask, cv.RETR_TREE, cv.CHAIN_APPROX_SIMPLE)
+
+            for contour in contours:
+                if contour.size >= 6:
+                    contour = np.squeeze(contour, axis=1)
+                    if i not in polygons:
+                        polygons[i] = [contour]
+                    else:
+                        polygons[i].append(contour)
```
eaidova, line 71: it looks like good point for starting usage defaultdict:
https://docs.python.org/2/library/collections.html#collections.defaultdict

tools/accuracy_checker/accuracy_checker/representation/segmentation_representation.py
```diff
@@ -63,14 +64,17 @@ def to_polygon(self, segmentation_colors=None):
...
-                    else:
-                        polygons[i].append(contour)
+                if contour.size < 6:
+                    continue
 
-        return polygons
+                contour = np.squeeze(contour, axis=1)
+                polygons.append((i, contour))
+
+        result = defaultdict(list)
```
eaidova, line 73: you can directly write to result, no need in polygons variable

tools/accuracy_checker/accuracy_checker/representation/segmentation_representation.py
```diff
@@ -50,7 +52,22 @@ class GTMaskLoader(Enum):
 
 
 class SegmentationRepresentation(BaseRepresentation):
-    pass
+    def to_polygon(self, segmentation_colors=None):
+        polygons = defaultdict(list)
+        mask = self._encode_mask(self.mask, segmentation_colors) if segmentation_colors else self.mask
+        if len(mask.shape) == 3:
```
eaidova, line 58: what if mask already argmaxed and have shape [h, w, 1]?

tools/accuracy_checker/accuracy_checker/representation/segmentation_representation.py
```diff
@@ -50,7 +52,22 @@ class GTMaskLoader(Enum):
 
 
 class SegmentationRepresentation(BaseRepresentation):
-    pass
+    def to_polygon(self, segmentation_colors=None):
+        polygons = defaultdict(list)
+        mask = self._encode_mask(self.mask, segmentation_colors) if segmentation_colors else self.mask
```
eaidova, line 57: who will decide when need to provide segmentation colors and then not?

tools/accuracy_checker/tests/common.py
```diff
@@ -96,6 +104,16 @@ def make_segmentation_representation(mask, ground_truth=False):
     return [SegmentationPrediction('identifier', mask)]
 
 
+def make_instance_segmentation_representation(mask, labels, ground_truth=False):
```
eaidova, line 107: looks like it require specific import, better to move closer to test where it will be used if you do not want handle case when maskUtils is None

tools/accuracy_checker/tests/test_segmentation_representation.py
```diff
@@ -0,0 +1,114 @@
...
+                np.array([[0, 1, 1, 1], [0, 0, 1, 1], [0, 0, 0, 1]])]
+        labels = [0, 1]
+        predictions = make_instance_segmentation_representation(mask, labels, False)
+        expected = {
+            0: [np.array([[1, 0], [3, 0], [3, 2]], dtype=np.int32)],
+            1: [np.array([[0, 0], [0, 2], [2, 2]], dtype=np.int32)]}
+        for prediction in predictions:
+            actual = prediction.to_polygon()
+
+        for key, _ in expected.items():
```
eaidova, line 111: why you need items() if you use only keys?

tools/accuracy_checker/tests/test_segmentation_representation.py
```diff
@@ -0,0 +1,114 @@
...
+
+    def test_to_polygon_prediction(self):
+        mask = [np.array([[1, 0, 0, 0], [1, 1, 0, 0], [1, 1, 1, 0]]),
+                np.array([[0, 1, 1, 1], [0, 0, 1, 1], [0, 0, 0, 1]])]
+        labels = [0, 1]
+        predictions = make_instance_segmentation_representation(mask, labels, False)
+        expected = {
+            0: [np.array([[1, 0], [3, 0], [3, 2]], dtype=np.int32)],
+            1: [np.array([[0, 0], [0, 2], [2, 2]], dtype=np.int32)]}
+        for prediction in predictions:
```
eaidova, line 108: why here cycle and used only last item as actual?

tools/accuracy_checker/tests/test_segmentation_representation.py
```diff
@@ -0,0 +1,114 @@
...
+
+from .common import make_segmentation_representation, make_instance_segmentation_representation
+
+
+class TestSegmentationRepresentation:
+
+    def test_to_polygon_annotation(self):
+        annotations = make_segmentation_representation(np.array([[1, 0, 0, 0], [1, 1, 0, 0], [1, 1, 1, 0]]), True)
+        expected = {
+            0: [np.array([[1, 0], [3, 0], [3, 2]], dtype=np.int32)],
```
eaidova, line 27: why int32, uint8 not enough here?

tools/accuracy_checker/accuracy_checker/representation/segmentation_representation.py
Anna-Mironova: done

tools/accuracy_checker/accuracy_checker/representation/segmentation_representation.py
Anna-Mironova: add condition

tools/accuracy_checker/tests/test_segmentation_representation.py
Anna-Mironova: replace to keys

tools/accuracy_checker/tests/test_segmentation_representation.py
Anna-Mironova: done

tools/accuracy_checker/tests/test_segmentation_representation.py
Anna-Mironova: delete type

tools/accuracy_checker/accuracy_checker/dataset.py
```diff
@@ -208,6 +208,10 @@ def set_annotation_metadata(self, annotation, image, data_source):
         segmentation_mask_source = self.config.get('segmentation_masks_source')
         annotation.set_segmentation_mask_source(segmentation_mask_source)
         annotation.set_additional_data_source(self.config.get('additional_data_source'))
+        annotation.metadata.update({
```
eaidova, line 211: please use setter like above, there is annotation representation with additional logic - AnnotationContainer where process of updating meta different from regular

eaidova: and I believe you can just put full dataset meta, who knows maybe we need it in future

tools/accuracy_checker/accuracy_checker/representation/segmentation_representation.py
```diff
@@ -201,6 +271,25 @@ def areas(self):
             areas.append(maskUtils.area(mask))
         return areas
 
+    def to_polygon(self):
+        if not self.raw_mask:
+            raise ValueError("Polygon can be found only for non-empty mask")
+
+        if not self.labels:
+            raise ValueError("Polygon can be found only for non-empty labels")
```
eaidova, line 279: I think in these case and above we can return empty dict (if you want, you can throw warning)

tools/accuracy_checker/accuracy_checker/representation/base_representation.py
```diff
@@ -46,3 +46,6 @@ def set_segmentation_mask_source(self, mask_source):
 
     def set_additional_data_source(self, source):
         self.metadata['additional_data_source'] = source
+
+    def set_dataset_metadata(self, source):
+        self.metadata.update(source)
```
eaidova, line 51: on my view better to add key `dataset_meta` and put in this field, in your solution, possible intersection and overriding original meta keys on datasets meta values

tools/accuracy_checker/tests/test_segmentation_representation.py
```diff
@@ -0,0 +1,266 @@
...
+    def test_to_polygon_prediction_with_empty_mask(self):
+        prediction = make_segmentation_representation(np.array([]), False)[0]
+
+        with pytest.warns(Warning):
+            assert len(prediction.to_polygon()) == 0
+
+
+class TestCoCoInstanceSegmentationRepresentation:
+
+    @pytest.mark.skipif(no_available_pycocotools(), reason='no installed pycocotools in the system')
```
eaidova, line 202: can we skip whole test class instead of each test?

PR #1381, opened 2020-08-04
Demo SSIM tests PoC and usage example for both yolov3 demos

AnthonyQuantum:
The testing method used to visually check whether the demos are working properly.

Demos save output frames to the file specified by the `-o` flag.

Crop is used on frames in order to remove elements (like performance metrics values and monitors) which may spoil the testing result.

The `reference_values.json` file is generated by starting the `run_tests.py` script with `--generate-reference` flag on the "gold" set of demos, which are expected to work properly and therefore used as a reference for future testing.
`reference_values.json` contains PSNR, SSIM_R, SSIM_G, and SSIM_B values for each test case.

- [x] Update READMEs
- [ ] Probably need to add checks (valid file path, etc.)


demos/object_detection_demo_yolov3_async/main.cpp
```diff
@@ -309,6 +309,8 @@ int main(int argc, char *argv[]) {
         cv::Size graphSize{frame.cols / 4, 60};
         Presenter presenter(FLAGS_u, frame.rows - graphSize.height - 10, graphSize);
 
+        cv::VideoWriter outVideo(FLAGS_o, cv::VideoWriter::fourcc('M','J','P','G'), cap->fps(), frame.size());
```
Wovchena, line 312: OpenCV reports assertion fail for me if I try to open `VideoWriter` with empty path.

You should create empty `VideoWriter`. If `FLAGS_o` is not empty open it. You should also report an error if `isOpened()` returns false after `open()` because it looks that `VideoWriter` will silently accept frames to write even if it is closed.

demos/tests/args.py
```diff
@@ -106,3 +108,14 @@ def __init__(self, file_name):
...
+
+class DemoOutputFileArg:
+    def __init__(self):
+        with tempfile.TemporaryDirectory() as temp_dir:
+            self.out_dir = Path(temp_dir) / 'demo_output'
+        if not self.out_dir.is_dir():
+            self.out_dir.mkdir(parents=True)
+
+    def resolve(self, context):
+        return str(self.out_dir / 'output.avi')
```
Wovchena, line 121:
```suggestion
        return str(self.out_dir / '%09d.png')
```
It looks that the values for similarity metrics are different for you and CI. May be saving demo output as uncompressed images will eliminate the issue.

AnthonyQuantum:
jenkins ok to test

AnthonyQuantum:
@IRDonch, could you please tell me which video codecs are installed on the CI system?

I have to use something lossless. HFYU is not supported. IYUV fails too, even though it is also used in interactive_face_detection_demo.

I suppose that codecs like this are installed apart of Ubuntu and OpenCV.

Wovchena:
@AnthonyQuantum, have you tried saving as uncompressed images: https://github.com/opencv/opencv/blob/d696d7f9eafbf458afbea58e863e0f5ebc827627/modules/videoio/include/opencv2/videoio.hpp#L888

demos/python_demos/object_detection_demo_yolov3_async/object_detection_demo_yolov3_async.py
```diff
@@ -364,8 +367,18 @@ def main():
...
 
-    presenter = monitors.Presenter(args.utilization_monitors, 55,
-        (round(cap.get(cv2.CAP_PROP_FRAME_WIDTH) / 4), round(cap.get(cv2.CAP_PROP_FRAME_HEIGHT) / 8)))
+    graph_size = (round(cap.get(cv2.CAP_PROP_FRAME_WIDTH) / 4), round(cap.get(cv2.CAP_PROP_FRAME_HEIGHT) / 8))
+    presenter = monitors.Presenter(args.utilization_monitors, int(cap.get(4)) - graph_size[1] - 10, graph_size)
+
+    out_video = cv2.VideoWriter()
+    raw_video = cv2.VideoWriter()
+    if args.output:
+        frame_size = (int(cap.get(3)), int(cap.get(4)))
```
Wovchena, line 376: You should refer to `3` and `4` via their names.

demos/python_demos/object_detection_demo_yolov3_async/object_detection_demo_yolov3_async.py
```diff
@@ -364,8 +367,18 @@ def main():
...
+        out_video.open(str(Path(args.output) / 'out_%08d.BMP'), 0, 0, frame_size)
+        raw_video.open(str(Path(args.output) / 'raw_%08d.BMP'), 0, 0, frame_size)
+
+        if not out_video.isOpened() or not raw_video.isOpened():
+            raise RuntimeError('Can\'t open VideoWriter.')
```
Wovchena, line 381:
```suggestion
            raise RuntimeError("Can't open VideoWriter.")
```

demos/tests/run_tests.py
```diff
@@ -38,12 +39,16 @@
 import sys
 import tempfile
 import timeit
-
+import cv2 as cv
+import numpy as np
```
Wovchena, line 43: Surround third party imports with empty lines.

demos/tests/run_tests.py
```diff
@@ -206,17 +218,70 @@ def option_to_args(key, value):
...
-                            print(e.output)
-                            print('Exit code:', e.returncode)
+
+                            # 'if' is dev-only, won't be needed when all demos will be compatible with ssim test
+                            if '-o' in case_args and case_args[case_args.index('-o') + 1] != '.':
+                                similarity_res = []
+
+                                out_folder = case_args[case_args.index('-o') + 1]
+                                demo_out_file = str(Path(out_folder) / 'out_%08d.BMP')
+                                demo_raw_file = str(Path(out_folder) / 'raw_%08d.BMP')
```
Wovchena, line 236: `DataPatternArg` uses `%04d` as a mask. It's good to stick to it.

Wovchena (changes requested):
@IRDonch, @eizamaliev, @fzhar, what do you think about the approach?

fzhar (changes requested):
If I understood correctly, test system compares output frame (with frames over detected objects) with original input frame. Unfortunately, this upproach is completely wrong, particularly for object detection. Imagine that output of testing demo will draw frame of the same size but in different place. If background is approximately the same, you'll get same PSNR and SSIMs. But result is actually wrong. And vise versa, if frame is shifted just for the couple of pixels, we might have significantly changed PSNR.
Besides, rectangles are to thin to change PSNR significantly. 
PSNR/SSIM comparison approach can be used only for the usecases returning image (i.e. segmentation tasks). And, we have to compare output of testing demo with output of "gold" demo (previously stored, or generated on-the -fly). 
For detection demo we should output rectangles' coordinates to compare them with gold demo. If we really want to use image output of detection demo, than, test system should first detect rectangles (by color, for instance), then calculate rectangles corners, and only then compare them with gold output data.

Several comment about general testing model:
1. It's better save raw data to the file, without any overlays (statistics, etc...) - that would allow us to avoid cropping. In some demos text area is quite large, so currently we have to crop out quite large area. Saving raw image is quite easy to implement.
2. We should have option in the test system to keep all saved frames. That is essential for later debugging. Also, it could be nice to have the option to keep only frames with metrics below thresholds.

demos/tests/run_tests.py
```diff
@@ -206,17 +218,70 @@ def option_to_args(key, value):
...
+                                while True:
+                                    out_ret, out_frame = out_cap.read()
+                                    raw_ret, raw_frame = raw_cap.read()
+
+                                    if out_ret and raw_ret:
+                                        crop = CROP_SIZE[demo.full_name]
+                                        height, width = out_frame.shape[:2]
+                                        out_frame = out_frame[crop[0] : height - crop[2], crop[3] : width - crop[1]]
+                                        raw_frame = raw_frame[crop[0] : height - crop[2], crop[3] : width - crop[1]]
+                                        similarity = list(map(lambda x: round(x, 3),
```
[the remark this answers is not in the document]
AnthonyQuantum: I think we shouldn't store such long numbers as `0.9666283495334497` in `reference_values.json`, because the method used for comparison is not so precise and thresholds are also about `10^-3`. It might confuse someone. So, I round to `0.967`.

PR #1383, opened 2020-08-04
AC: Quartznet

eaidova:


tools/accuracy_checker/accuracy_checker/evaluators/custom_evaluators/asr_encoder_decoder_evaluator.py
```diff
@@ -0,0 +1,604 @@
+"""
+Copyright (c) 2019-2020 Intel Corporation
```
Anna-Mironova, line 2: Please align copyright

eaidova: fixed

PR #1384, opened 2020-08-05
build_demos_msvc.bat: let CMake find MSBuild for us

IRDonch:
This simplifies the script and makes it easier to add support for new versions of Visual Studio. I also trust CMake more with this than I do our home-grown logic; CMake is more actively maintained, so they're more likely to handle any obscure cases.

I also took the opportunity to improve some of the error messages.

### DIFF
--- a/demos/build_demos_msvc.bat
+++ b/demos/build_demos_msvc.bat
@@ -20,8 +20,8 @@ set "ROOT_DIR=%~dp0"
 
 set "SOLUTION_DIR64=%USERPROFILE%\Documents\Intel\OpenVINO\omz_demos_build"
 
-set MSBUILD_BIN=
-set VS_PATH=
+set SUPPORTED_VS_VERSIONS=VS2015 VS2017 VS2019
+
 set VS_VERSION=
 set EXTRA_CMAKE_OPTS=
 
@@ -41,14 +41,14 @@ if not "%1" == "" (
     )
 
     if "%1"=="VS2015" (
-        set "VS_VERSION=2015"
+        set "VS_VERSION=14 2015"
     ) else if "%1"=="VS2017" (
-        set "VS_VERSION=2017"
+        set "VS_VERSION=15 2017"
     ) else if "%1"=="VS2019" (
-        set "VS_VERSION=2019"
+        set "VS_VERSION=16 2019"
     ) else (
         echo Unrecognized Visual Studio version specified: "%1"
-        echo Supported versions: VS2015, VS2017, VS2019
+        echo Supported versions: %SUPPORTED_VS_VERSIONS%
         goto errorHandling
     )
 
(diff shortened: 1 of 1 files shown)


demos/build_demos_msvc.bat
```diff
@@ -20,6 +20,8 @@ set "ROOT_DIR=%~dp0"
 
 set "SOLUTION_DIR64=%USERPROFILE%\Documents\Intel\OpenVINO\omz_demos_build"
 
+set SUPPORTED_VS_VERSIONS=VS2015 VS2017 VS2019
+
 set MSBUILD_BIN=
 set VS_PATH=
```
Wovchena, line 26:
```suggestion
```

demos/build_demos_msvc.bat
```diff
@@ -75,84 +77,54 @@ if "%PROCESSOR_ARCHITECTURE%" == "AMD64" (
...
-    if exist "!VS_PATH!\MSBuild\14.0\Bin\MSBuild.exe" (
-        set "MSBUILD_BIN=!VS_PATH!\MSBuild\14.0\Bin\MSBuild.exe"
-    )
-    if exist "!VS_PATH!\MSBuild\15.0\Bin\MSBuild.exe" (
-        set "MSBUILD_BIN=!VS_PATH!\MSBuild\15.0\Bin\MSBuild.exe"
-    )
-    if exist "!VS_PATH!\MSBuild\Current\Bin\MSBuild.exe" (
-        set "MSBUILD_BIN=!VS_PATH!\MSBuild\Current\Bin\MSBuild.exe"
+        echo vswhere.exe was not found.
+        echo Make sure Visual Studio 2017 version 15.2 or a newer version of Visual Studio
```
Wovchena, line 87: You have 2017 here because older versions don't include vswhere, right? But this error message can be misleading because it says that Visual Studio 2017+ is required while it is till possible to pass 2015 as argument.

And if vswhere appeared in 2017 you don't need to handle `2015` value when you search `VS_VERSION` with vswhere below as there will be at least the version which contains vswhere.

demos/build_demos_msvc.bat
```diff
@@ -75,84 +77,54 @@ if "%PROCESSOR_ARCHITECTURE%" == "AMD64" (
...
-            set "MSBUILD_BIN=C:\Program Files (x86)\Microsoft Visual Studio\2017\BuildTools\MSBuild\15.0\Bin\MSBuild.exe"
-            set "MSBUILD_VERSION=15 2017"
-        ) else if exist "C:\Program Files (x86)\Microsoft Visual Studio\2017\Professional\MSBuild\15.0\Bin\MSBuild.exe" (
-            set "MSBUILD_BIN=C:\Program Files (x86)\Microsoft Visual Studio\2017\Professional\MSBuild\15.0\Bin\MSBuild.exe"
-            set "MSBUILD_VERSION=15 2017"
-        ) else if exist "C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\MSBuild\15.0\Bin\MSBuild.exe" (
-            set "MSBUILD_BIN=C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\MSBuild\15.0\Bin\MSBuild.exe"
-            set "MSBUILD_VERSION=15 2017"
+    echo Searching for the latest Visual Studio...
+    for /f "usebackq tokens=*" %%i in (`"!VSWHERE!" -latest -products * -requires Microsoft.Component.MSBuild -property catalog_productLineVersion`) do (
```
Wovchena, line 93: You have `-latest` key here. I expect that it reduces the output to only one version. The loop in not needed.

IRDonch: The loop is needed, because that's the only way to get a program's output in the Windows shell.

Wovchena: Okay

demos/build_demos_msvc.bat
IRDonch: Good catch; removed.

demos/build_demos_msvc.bat
IRDonch: > You have 2017 here because older versions don't include vswhere, right? But this error message can be misleading because it says that Visual Studio 2017+ is required while it is till possible to pass 2015 as argument.

Okay, I've amended the message to make it more accurate.

> And if vswhere appeared in 2017 you don't need to handle 2015 value when you search VS_VERSION with vswhere below as there will be at least the version which contains vswhere.

vswhere appeared in 2017, but it _can_ report information about 2015, and I think that in rare cases 2015 can end up being reported as the latest (e.g. if both 2015 and 2017 are installed, but 2017 doesn't include the MSBuild component).

I should also note that from what I recall, VS2015 support is planned to be dropped in OpenVINO, so this whole point might be moot soon.

Wovchena: Seems 2015 is dropped: https://github.com/openvinotoolkit/openvino/blob/master/build-instruction.md#software-requirements-1. But may be we could keep it as it costs only few ifs for us.

PR #1390, opened 2020-08-07
Classification demo: Add demo to the global README, add a note about ILSVRC2015

AnthonyQuantum:


demos/classification_demo/README.md
```diff
@@ -90,6 +90,17 @@ It has the following format:
...
+Please note that you should use the ILSVRC 2015 annotation file (val15.txt) with the following models:
+- googlenet-v2
+- se-inception
+- se-resnet-101
+- se-resnet-152
+- se-resnet-50
+- se-resnext-101
+- se-resnext-50
+
+and the ILSVRC 2012 annotation file (val.txt) with all other models supported by the demo.
```
Wovchena, line 102: You used ground truth term before instead of annotation.

It doesn't make sense to note which file to use for the models without pointing out that it is data set specific and works only if images are taken from imageNet.

You should leave a link to the files where to get them. May be [../../../datasets.md#imagenet](../../../datasets.md#imagenet)

Shouldn't `-labels` file be different for the models as well?

AnthonyQuantum:
I guess the information about ground truth files should not be put in README because it's dataset-specific and users already know which gt file to use with which dataset.
A note about labels files should be put instead, because the id to label mapping is different for this subset of models. And labels files are not dataset-specific.

demos/classification_demo/README.md
```diff
@@ -92,7 +92,18 @@ It has the following format:
 
 Class index values must be in range from 0 to 1000. If you want to use "other" class, which is supported only by a small subset of models, specify it with -1 index.
 
-["Labels" file](./synset_words.txt) contains the list of human-readable labels, one line for each class.
+"Labels" file contains the list of human-readable labels, one line for each class.
+
+Please note that you should use [id_classes_15.txt](./id_classes_15.txt) labels file with the following models:
```
IRDonch, line 97: I think the file name should be more explicit about the file's origins, such as `imagenet_2015_classes.txt`.

IRDonch:
There's a merge conflict now; please rebase.

AnthonyQuantum:
> There's a merge conflict now; please rebase.

Done

PR #1392, opened 2020-08-07
Update accuracy for mobilenet-v2-1.0-224

Anna-Mironova:


eaidova:
FYI @kchechil @vladimir-dudnik 
updated accuracy numbers after reverification with correct preprocessing

PR #1398, opened 2020-08-07
AC: provide output layout to the adapter

eaidova:
some openvino plugins support both NCHW and NHWC execution modes. Depends from layout, output data stored in different order and can require additional postprocessing (transpsition, reshape, e. t. c.). Usually we make decision based on input shape, however it is not always matched with layout (at  least in OV), so I suggest to provide additional info about output layout.


eaidova:
Jenkins please retry a build

eaidova:
Jenkins please retry a build

PR #1401, opened 2020-08-10
smart_classroom: use images_capture

Wovchena:


demos/smart_classroom_demo/README.md
```diff
@@ -62,15 +64,15 @@ Options:
...
     -teacher_id                    Optional. ID of a teacher. You must also set a faces gallery parameter (-fg) to use it.
+    -no_show                       Optional. Do not show processed video.
     -min_ad                        Optional. Minimum action duration in seconds.
     -d_ad                          Optional. Maximum time difference between actions in seconds.
     -student_ac                    Optional. List of student actions separated by a comma.
+    -top_ac                        Optional. List of student actions (for top-k mode) separated by a comma.
     -teacher_ac                    Optional. List of teacher actions separated by a comma.
     -a_id                          Optional. Target action name.
-    -a_top                         Optional. Number of first K students. If this parameter is positive, the demo detects first K persons with the action, pointed by the parameter "a_id"
+    -a_top                         Optional. Number of first K students. If this parameter is positive, the demo detects first K persons with the action, pointed by the parameter 'top_id'
```
IRDonch, line 75: This change is actually wrong - the parameter is called `-a_id`. Maybe you could fix the help string while you're at it?

demos/smart_classroom_demo/main.cpp
```diff
@@ -760,53 +739,49 @@ int main(int argc, char* argv[]) {
...
-        bool is_monitoring_enabled = false;
-        auto prev_frame_path = cap.GetVideoPath();
+        cv::Size graphSize{static_cast<int>(frame.cols / 4), 60};
+        Presenter presenter(FLAGS_u, frame.rows - graphSize.height - 10, graphSize);
 
         cv::VideoWriter vid_writer;
         if (!FLAGS_out_v.empty()) {
             vid_writer = cv::VideoWriter(FLAGS_out_v, cv::VideoWriter::fourcc('M', 'J', 'P', 'G'),
-                                         cap.GetFPS(), Visualizer::GetOutputSize(frame.size()));
+                                         cap->fps(), Visualizer::GetOutputSize(frame.size()));
```
IRDonch, line 754: Need to be careful here. What'll this do if the source is an image/directory?

demos/smart_classroom_demo/main.cpp
```diff
@@ -1047,6 +1016,8 @@ int main(int argc, char* argv[]) {
 
                 const int start_frame = 0;
                 const int end_frame = face_obj_id_to_action_maps.size();
+                const int smooth_window_size = static_cast<int>(cap->fps() * FLAGS_d_ad);
```
IRDonch, line 1019: Same here - probably need to do something special for non-video inputs.

demos/smart_classroom_demo/README.md
Wovchena: `showUsage()` was incorrect.

demos/smart_classroom_demo/main.cpp
Wovchena: Updated

demos/smart_classroom_demo/main.cpp
Wovchena: I don't think we can do something special and useful for non-video inputs

Wovchena:
Jenkins please retry a build

PR #1410, opened 2020-08-13
AC: Added support number of used classes for ade20k

Anna-Mironova:


tools/accuracy_checker/accuracy_checker/annotation_converters/ade20k_dataset_converter.py
```diff
@@ -31,13 +31,15 @@ def parameters(cls):
             'images_dir': PathField(is_directory=True, description='Images directory'),
             'annotations_dir': PathField(is_directory=True, description='Annotation directory'),
             'object_categories_file': PathField(description='file with object categories'),
+            'num_classes': NumberField(description='Number of used classes', optional=True, value_type=int),
```
eaidova, line 34: no need to add min value?

Anna-Mironova: added

PR #1412, opened 2020-08-13
Inpainiting demo GUI and refactoring

fzhar:


demos/python_demos/image_inpainting_demo/image_inpainting_demo.py
```diff
@@ -17,6 +17,7 @@
 import cv2
 import numpy as np
 from argparse import ArgumentParser, SUPPRESS
+from inpainting_gui import InpaintingGUI
```
eizamaliev, line 20: Please, reorder import according to [PEP](https://www.python.org/dev/peps/pep-0008/#imports) (standart libs, third party lib, local lib)

demos/python_demos/image_inpainting_demo/image_inpainting_demo.py
```diff
@@ -33,30 +34,90 @@ def build_argparser():
...
+            cv2.imwrite(args.output,result)
+        if not args.no_show:
+            cv2.imshow('Image Inpainting Demo', concat_image)
+            cv2.waitKey(0)
+    else:
+        # GUI inpainting
+        gui = InpaintingGUI(args.input,args.model, args.device)
+        gui.run();
+
+def inpaintRandomHoles(args):
```
eizamaliev, line 65: Move functions (this and `createRandomMask`) upper for correct logic "definition" - "usage".

demos/python_demos/image_inpainting_demo/image_inpainting_demo.py
```diff
@@ -33,30 +34,90 @@ def build_argparser():
                       help="Optional. Specify the target device to infer on; CPU, GPU, FPGA, HDDL or MYRIAD is "
                            "acceptable. The demo will look for a suitable plugin for device specified. "
                            "Default value is CPU", default="CPU", type=str)
-    args.add_argument("-p", "--parts", help="Optional. Number of parts to draw mask.", default=8, type=int)
-    args.add_argument("-mbw", "--max_brush_width", help="Optional. Max width of brush to draw mask.", default=24, type=int)
-    args.add_argument("-ml", "--max_length", help="Optional. Max strokes length to draw mask.", default=100, type=int)
-    args.add_argument("-mv", "--max_vertex", help="Optional. Max number of vertex to draw mask.", default=20, type=int)
-    args.add_argument("--no_show", help="Optional. Don't show output", action='store_true')
+    args.add_argument("-r", "--rnd", help="Optional. Use random mask for inpainting (with parameters set by -p, -mbw, -mk and -mv).Skipped in GUI mode", action='store_true')
```
eizamaliev, line 37: Don't forget to update README.md!

demos/python_demos/image_inpainting_demo/image_inpainting_demo.py
fzhar: Done

demos/python_demos/image_inpainting_demo/image_inpainting_demo.py
fzhar: Done

demos/python_demos/image_inpainting_demo/image_inpainting_demo.py
fzhar: Done

demos/python_demos/image_inpainting_demo/image_inpainting_demo.py
```diff
@@ -33,30 +34,90 @@ def build_argparser():
...
+    args.add_argument("-p", "--parts", help="Optional. Number of parts to draw mask. Skipped in GUI mode", default=8, type=int)
+    args.add_argument("-mbw", "--max_brush_width", help="Optional. Max width of brush to draw mask. Skipped in GUI mode", default=24, type=int)
+    args.add_argument("-ml", "--max_length", help="Optional. Max strokes length to draw mask. Skipped in GUI mode", default=100, type=int)
+    args.add_argument("-mv", "--max_vertex", help="Optional. Max number of vertex to draw mask. Skipped in GUI mode", default=20, type=int)
+    args.add_argument("-mc", "--mask_color", help="Optional. Color to be treated as mask (provide 3 RGB components in range of 0...255). Default is 0 0 0. Skipped in GUI mode", default=0, type=int, nargs="+")
```
eizamaliev, line 42: Line too long. Recommended line length is 120. 
In this case good option is to use individual lines for description, e.g:
```python
    args.add_argument("-mv", "--max_vertex", default=20, type=int,
                      help="Optional. Max number of vertex to draw mask. "
                           "Skipped in GUI mode")
```

demos/python_demos/image_inpainting_demo/image_inpainting_demo.py
```diff
@@ -33,30 +34,90 @@ def build_argparser():
...
+            next_y = np.clip(next_y, 0, h - 1).astype(np.int)
+            next_x = np.clip(next_x, 0, w - 1).astype(np.int)
+            cv2.line(mask, (start_y, start_x), (next_y, next_x), 1, brush_width)
+            cv2.circle(mask, (start_y, start_x), brush_width // 2, 1)
+
+            start_y, start_x = next_y, next_x
+    return mask
+
+def inpaintRandomHoles(args):
+    if args.mask_color !=0:
```
eizamaliev, line 74:
```suggestion
    if args.mask_color != 0:
```

eizamaliev: *Hereinafter*

demos/python_demos/image_inpainting_demo/image_inpainting_demo.py
```diff
@@ -33,30 +34,90 @@ def build_argparser():
...
+        concat_image,result = inpaintRandomHoles(args)
+        if args.output!="":
+            cv2.imwrite(args.output,result)
+        if not args.no_show:
+            cv2.imshow('Image Inpainting Demo', concat_image)
+            cv2.waitKey(0)
+    else:
+        # GUI inpainting
+        gui = InpaintingGUI(args.input,args.model, args.device)
+        gui.run();
```
eizamaliev, line 120:
```suggestion
        gui.run()
```
C--

demos/python_demos/image_inpainting_demo/image_inpainting_demo.py
```diff
@@ -14,12 +14,13 @@
...
 from argparse import ArgumentParser, SUPPRESS
 
+import numpy as np
+import cv2
 from openvino.inference_engine import IECore
 
+from inpainting_gui import InpaintingGUI
 from inpainting import ImageInpainting
 
 def build_argparser():
```
eizamaliev, line 26: Use 2 blank line between functions.

demos/python_demos/image_inpainting_demo/image_inpainting_demo.py
```diff
@@ -33,30 +34,90 @@ def build_argparser():
...
+    args.add_argument("-o","--output", help="Optional. Save output to the file with provided filename. Skipped in GUI mode", default="", type=str)
+    args.add_argument("-a", "--auto", help="Optional. Use automatic (non-interactive) mode instead of GUI", action='store_true')
 
     return parser
 
-def main():
-    args = build_argparser().parse_args()
+def createRandomMask(parts, max_vertex, max_length, max_brush_width, h, w, max_angle=360):
+    mask = np.zeros((h,w,1), dtype=np.float32)
+    for part in range(parts):
```
eizamaliev, line 51:
```suggestion
    for _ in range(parts):
```
Seems unused

demos/python_demos/image_inpainting_demo/image_inpainting_demo.py
```diff
@@ -33,30 +34,90 @@ def build_argparser():
...
+        num_strokes = np.random.randint(max_vertex)
+        start_y = np.random.randint(h)
+        start_x = np.random.randint(w)
+        brush_width = 0
```
eizamaliev, line 55: Seems unused.

demos/python_demos/image_inpainting_demo/image_inpainting_demo.py
fzhar: Done

demos/python_demos/image_inpainting_demo/image_inpainting_demo.py
fzhar: Done

demos/python_demos/image_inpainting_demo/image_inpainting_demo.py
fzhar: Deleted

demos/python_demos/image_inpainting_demo/image_inpainting_demo.py
```diff
@@ -33,30 +34,98 @@ def build_argparser():
...
+    if args.auto:
+        # Command-line inpaining for just one image
+        concat_image,result = inpaintRandomHoles(args)
+        if args.output != "":
+            cv2.imwrite(args.output,result)
+        if not args.no_show:
+            cv2.imshow('Image Inpainting Demo', concat_image)
+            cv2.waitKey(0)
+    else:
+        # GUI inpainting
```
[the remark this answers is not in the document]
fzhar: those are comments for every code block of if-else. I've caged it to more human-readable form, but it should be here

demos/python_demos/image_inpainting_demo/inpainting_gui.py
```diff
@@ -0,0 +1,133 @@
...
+        self.mask = np.zeros((self.inpainter.input_height,self.inpainter.input_width,1),dtype=np.float32)
+        self.updateWindow()
+
+        if self.img is None:
+            print("Cannot load source image\n");
+            return
+
+        key= cv2.waitKey(1)
+        while key not in (27,ord('q'),ord('Q')):
+            if key==ord(" "):
```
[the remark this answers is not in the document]
fzhar: why ? space works good. I can do both, personally for me space is better

demos/python_demos/image_inpainting_demo/image_inpainting_demo.py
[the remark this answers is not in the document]
Wovchena: I can't see how you've changed it

demos/python_demos/image_inpainting_demo/inpainting_gui.py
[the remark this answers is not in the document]
Wovchena: Enter has semantics of entering the data into the system

demos/python_demos/image_inpainting_demo/image_inpainting_demo.py
```diff
@@ -33,30 +34,98 @@ def build_argparser():
...
+    args.add_argument("-p", "--parts", help="Optional. Number of parts to draw mask. Skipped in GUI mode", default=8, type=int)
+    args.add_argument("-mbw", "--max_brush_width", help="Optional. Max width of brush to draw mask. Skipped in GUI mode",
+                      default=24, type=int)
+    args.add_argument("-ml", "--max_length", help="Optional. Max strokes length to draw mask. Skipped in GUI mode",
+                      default=100, type=int)
+    args.add_argument("-mv", "--max_vertex", help="Optional. Max number of vertex to draw mask. Skipped in GUI mode",
+                      default=20, type=int)
+    args.add_argument("-mc", "--mask_color",
+                      help="Optional. Color to be treated as mask (provide 3 RGB components in range of 0...255)."
+                      " Default is 0 0 0. Skipped in GUI mode", default=0, type=int, nargs="+")
```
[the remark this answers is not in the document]
fzhar: Done, some additional code is checked

demos/python_demos/image_inpainting_demo/inpainting_gui.py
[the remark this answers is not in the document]
fzhar: Only in case of command line. I've added it anyway, path is about to be committed

demos/python_demos/image_inpainting_demo/image_inpainting_demo.py
```diff
@@ -33,30 +34,98 @@ def build_argparser():
...
+    inpainting_processor = ImageInpainting(ie, args.model, args.device)
+
+    #--- Resize to model input and generate random mask
+    img = cv2.resize(img, (inpainting_processor.input_width,inpainting_processor.input_height))
+
+    if args.rnd:
+        mask = createRandomMask(args.parts,args.max_vertex,args.max_length,args.max_brush_width,
+                                inpainting_processor.input_height,inpainting_processor.input_width)
+    else:
+        top = np.full(img.shape,[mask_color],np.uint8)
```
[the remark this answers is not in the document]
fzhar: Removed

demos/python_demos/image_inpainting_demo/inpainting_gui.py
```diff
@@ -0,0 +1,133 @@
...
+        else:
+            self.oldPoint=(-1,-1)
+
+
+    def run(self):
+        srcImg = cv2.imread(self.imgPath, cv2.IMREAD_COLOR)
+        self.img = cv2.resize(srcImg,(self.inpainter.input_width,self.inpainter.input_height))
+        self.originalImg = self.img.copy();
+        self.label = ""
+        self.mask = np.zeros((self.inpainter.input_height,self.inpainter.input_width,1),dtype=np.float32)
```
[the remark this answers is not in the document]
fzhar: Done

demos/python_demos/image_inpainting_demo/image_inpainting_demo.py
```diff
@@ -33,30 +34,103 @@ def build_argparser():
...
+    args.add_argument("-o", "--output", help="Optional. Save output to the file with provided filename. Skipped in GUI mode",
+                      default="", type=str)
+    args.add_argument("-a", "--auto", help="Optional. Use automatic (non-interactive) mode instead of GUI", action='store_true')
 
     return parser
 
-def main():
-    args = build_argparser().parse_args()
+
+def loadImage(imgPath):
```
eizamaliev, line 58:
```python
def loadImage(imgPath):
...
def build_argparser():
```
You should use only one style for whole demo. In this case, camel style is prefferable, but [PEP 8](https://www.python.org/dev/peps/pep-0008/#function-and-variable-names) insists on snake style.

demos/python_demos/image_inpainting_demo/image_inpainting_demo.py
```diff
@@ -33,30 +34,103 @@ def build_argparser():
                       help="Optional. Specify the target device to infer on; CPU, GPU, FPGA, HDDL or MYRIAD is "
                            "acceptable. The demo will look for a suitable plugin for device specified. "
                            "Default value is CPU", default="CPU", type=str)
-    args.add_argument("-p", "--parts", help="Optional. Number of parts to draw mask.", default=8, type=int)
-    args.add_argument("-mbw", "--max_brush_width", help="Optional. Max width of brush to draw mask.", default=24, type=int)
-    args.add_argument("-ml", "--max_length", help="Optional. Max strokes length to draw mask.", default=100, type=int)
-    args.add_argument("-mv", "--max_vertex", help="Optional. Max number of vertex to draw mask.", default=20, type=int)
-    args.add_argument("--no_show", help="Optional. Don't show output", action='store_true')
+    args.add_argument("-r", "--rnd",
+                      help="Optional. Use random mask for inpainting (with parameters set by -p, -mbw, -mk and -mv).Skipped in GUI mode",
```
eizamaliev, line 38: Still, too long here and in some other places. Maximum 120 symbols.

eizamaliev (changes requested):

demos/python_demos/image_inpainting_demo/image_inpainting_demo.py
```diff
@@ -33,30 +34,103 @@ def build_argparser():
...
     return parser
 
-def main():
-    args = build_argparser().parse_args()
+
+def loadImage(imgPath):
+    img = cv2.imread(imgPath, cv2.IMREAD_COLOR)
+    if img is None:
+        print(f"Cannot load image {imgPath}")
+        exit()
```
eizamaliev, line 62: Raise error instead of exiting whole program. `RuntimeError` could be ok.

demos/python_demos/image_inpainting_demo/image_inpainting_demo.py
```diff
@@ -33,30 +34,103 @@ def build_argparser():
...
+            cv2.circle(mask, (start_y, start_x), brush_width // 2, 1)
+
+            start_y, start_x = next_y, next_x
+    return mask
+
+
+def inpaintRandomHoles(args):
+    if len(args.mask_color) != 3:
+        print("Invalid mask_color is provided. Please provide 3 RGB components\n")
+        exit()
```
eizamaliev, line 93: Raise error instead of exiting whole program. In fact, you should check it in argument parser.

demos/python_demos/image_inpainting_demo/inpainting_gui.py
```diff
@@ -0,0 +1,131 @@
...
+
+import cv2
+import numpy as np
+
+from inpainting import ImageInpainting
+from openvino.inference_engine import IECore
+
+class InpaintingGUI(object):
+    def __init__(self, srcImg, modelPath, device="CPU"):
+        self.wndName="Inpainting (press H for help)"
```
eizamaliev, line 22:
```suggestion
        self.wndName="Inpainting demo (press H for help)"
```

eizamaliev (changes requested):

demos/python_demos/image_inpainting_demo/image_inpainting_demo.py
```diff
@@ -33,30 +34,103 @@ def build_argparser():
...
+        # Command-line inpaining for just one image
+        concat_image,result = inpaintRandomHoles(args)
+        if args.output != "":
+            cv2.imwrite(args.output,result)
+        if not args.no_show:
+            cv2.imshow('Image Inpainting Demo', concat_image)
+            cv2.waitKey(0)
+    else:
+        # Inpainting with GUI
+        InpaintingGUI(loadImage(args.input),args.model, args.device).run()
```
Wovchena, line 133: You should load the img in `InpaintingGUI`'s constructor like for `inpaintRandomHoles()`.

demos/python_demos/image_inpainting_demo/inpainting_gui.py
```diff
@@ -0,0 +1,131 @@
...
+
+            self.updateWindow()
+        else:
+            self.oldPoint=(-1,-1)
+
+
+    def run(self):
+        self.updateWindow()
+
+        if self.img is None:
```
Wovchena, line 58: This is no longer needed.

demos/python_demos/image_inpainting_demo/inpainting_gui.py
```diff
@@ -0,0 +1,131 @@
...
+            backbuffer = self.img.copy()
+            backbuffer[:,:,0:1][self.mask>0]=self.maskColor[0]
+            backbuffer[:,:,1:2][self.mask>0]=self.maskColor[1]
+            backbuffer[:,:,2:3][self.mask>0]=self.maskColor[2]
+
+        if self.label is not None and self.label!="":
+            lines = self.label.split("\n")
+            count = len(lines)
+            w = max(cv2.getTextSize(line, cv2.FONT_HERSHEY_COMPLEX, 0.75,1)[0][0] for line in lines)+pad*2
+            lineH = cv2.getTextSize(lines[0], cv2.FONT_HERSHEY_COMPLEX, 0.75,1)[0][1]+pad
```
Wovchena, line 125: You should be consistent about whitespaces in expressions and statements.

Wovchena (changes requested):

demos/python_demos/image_inpainting_demo/inpainting_gui.py
```diff
@@ -0,0 +1,131 @@
...
+
+
+    def run(self):
+        self.updateWindow()
+
+        if self.img is None:
+            print("Cannot load source image\n")
+            return
+
+        key= cv2.waitKey(1)
```
IRDonch, line 62: Why is there a timeout here? You only do something after the user presses a key.

demos/python_demos/image_inpainting_demo/inpainting_gui.py
```diff
@@ -0,0 +1,131 @@
...
+        while key not in (27,ord('q'),ord('Q')):
+            if key in (ord(" "),ord("\r")):
+                self.isOriginalShown = False
+                self.showInfo("Processing...")
+
+                self.img[:,:,0:1][self.mask>0]=0
+                self.img[:,:,1:2][self.mask>0]=0
+                self.img[:,:,2:3][self.mask>0]=0
```
IRDonch, line 70:
```suggestion
                self.img[self.mask>0] = 0
```

demos/python_demos/image_inpainting_demo/inpainting_gui.py
```diff
@@ -0,0 +1,131 @@
...
+            sz = cv2.getTextSize("Original", cv2.FONT_HERSHEY_COMPLEX, 0.75,1)[0]
+            imgWidth = backbuffer.shape[1]
+            labelArea = backbuffer[margin:sz[1]+pad*2+margin,imgWidth-margin-(sz[0]+pad*2):imgWidth-margin]
+            labelArea//=2
+            cv2.putText(backbuffer,"Original",(imgWidth-margin-sz[0]-pad,margin+sz[1]+pad),cv2.FONT_HERSHEY_COMPLEX, 0.75,(128,255,128))
+        else:
+            backbuffer = self.img.copy()
+            backbuffer[:,:,0:1][self.mask>0]=self.maskColor[0]
+            backbuffer[:,:,1:2][self.mask>0]=self.maskColor[1]
+            backbuffer[:,:,2:3][self.mask>0]=self.maskColor[2]
```
IRDonch, line 119:
```suggestion
            backbuffer[self.mask>0] = self.maskColor
```

demos/python_demos/image_inpainting_demo/inpainting_gui.py
```diff
@@ -0,0 +1,131 @@
...
+
+        if self.label is not None and self.label!="":
+            lines = self.label.split("\n")
+            count = len(lines)
```
IRDonch, line 123: Is this variable really necessary? It's only used once, and the expression is trivial.

demos/python_demos/image_inpainting_demo/image_inpainting_demo.py
IRDonch: > You should use only one style for whole demo.

I think in the case of Python demos we can make a stronger statement: you should use the standard Python style for the demo.

IRDonch (changes requested):

demos/python_demos/image_inpainting_demo/inpainting_gui.py
fzhar: Nope, we should just check if something is pressed, but don't wait. Anyway, I've replaced this part of code - there's another waitKey below

demos/python_demos/image_inpainting_demo/inpainting_gui.py
fzhar: won't work -  mask is [h,w,1] and img is [h,w,3]

demos/python_demos/image_inpainting_demo/inpainting_gui.py
fzhar: same as above: mask and backbuffer are of different sizes

demos/python_demos/image_inpainting_demo/inpainting_gui.py
fzhar: formulas are quite complex, I'd like to keep this variable

demos/python_demos/image_inpainting_demo/inpainting_gui.py
fzhar: applied into patch

demos/python_demos/image_inpainting_demo/image_inpainting_demo.py
fzhar: Image is loaded in main function now. Once for both parts - no matter which one is selected. Let's keep it this way

demos/python_demos/image_inpainting_demo/inpainting_gui.py
fzhar: fixed

demos/python_demos/image_inpainting_demo/inpainting_gui.py
fzhar: fixed.

demos/python_demos/image_inpainting_demo/image_inpainting_demo.py
```diff
@@ -33,30 +34,103 @@ def build_argparser():
...
                            "Default value is CPU", default="CPU", type=str)
-    args.add_argument("-p", "--parts", help="Optional. Number of parts to draw mask.", default=8, type=int)
-    args.add_argument("-mbw", "--max_brush_width", help="Optional. Max width of brush to draw mask.", default=24, type=int)
-    args.add_argument("-ml", "--max_length", help="Optional. Max strokes length to draw mask.", default=100, type=int)
-    args.add_argument("-mv", "--max_vertex", help="Optional. Max number of vertex to draw mask.", default=20, type=int)
-    args.add_argument("--no_show", help="Optional. Don't show output", action='store_true')
+    args.add_argument("-r", "--rnd",
+                      help="Optional. Use random mask for inpainting (with parameters set by -p, -mbw, -mk and -mv).Skipped in GUI mode",
+                      action='store_true')
+    args.add_argument("-p", "--parts", help="Optional. Number of parts to draw mask. Skipped in GUI mode", default=8, type=int)
```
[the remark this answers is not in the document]
fzhar: In this particuar case they are synonyms. But I agree, "Ignore" looks better here

demos/python_demos/image_inpainting_demo/image_inpainting_demo.py
```diff
@@ -33,30 +34,103 @@ def build_argparser():
...
+                      default=24, type=int)
+    args.add_argument("-ml", "--max_length", help="Optional. Max strokes length to draw mask. Skipped in GUI mode",
+                      default=100, type=int)
+    args.add_argument("-mv", "--max_vertex", help="Optional. Max number of vertex to draw mask. Skipped in GUI mode",
+                      default=20, type=int)
+    args.add_argument("-mc", "--mask_color",
+                      help="Optional. Color to be treated as mask (provide 3 RGB components in range of 0...255)."
+                      " Default is 0 0 0. Skipped in GUI mode", default=[0, 0, 0], type=int, nargs="+")
+    args.add_argument("--no_show", help="Optional. Don't show output. Skipped in GUI mode", action='store_true')
+    args.add_argument("-o", "--output", help="Optional. Save output to the file with provided filename. Skipped in GUI mode",
```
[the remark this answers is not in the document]
fzhar: When should we write it, upon quitting ? There's no need for it in GUI mode, but it is useful for automated testing.

demos/python_demos/image_inpainting_demo/image_inpainting_demo.py
```diff
@@ -33,30 +34,103 @@ def build_argparser():
...
+    args.add_argument("-mbw", "--max_brush_width", help="Optional. Max width of brush to draw mask. Skipped in GUI mode",
+                      default=24, type=int)
+    args.add_argument("-ml", "--max_length", help="Optional. Max strokes length to draw mask. Skipped in GUI mode",
+                      default=100, type=int)
+    args.add_argument("-mv", "--max_vertex", help="Optional. Max number of vertex to draw mask. Skipped in GUI mode",
+                      default=20, type=int)
+    args.add_argument("-mc", "--mask_color",
+                      help="Optional. Color to be treated as mask (provide 3 RGB components in range of 0...255)."
+                      " Default is 0 0 0. Skipped in GUI mode", default=[0, 0, 0], type=int, nargs="+")
+    args.add_argument("--no_show", help="Optional. Don't show output. Skipped in GUI mode", action='store_true')
```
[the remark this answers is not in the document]
fzhar: Why error ? It is ignored - that's sufficient for normal operation

demos/python_demos/image_inpainting_demo/image_inpainting_demo.py
```diff
@@ -33,30 +34,103 @@ def build_argparser():
...
 
     return parser
 
-def main():
-    args = build_argparser().parse_args()
+
+def loadImage(imgPath):
+    img = cv2.imread(imgPath, cv2.IMREAD_COLOR)
+    if img is None:
+        print(f"Cannot load image {imgPath}")
```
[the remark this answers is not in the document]
fzhar: OK

demos/python_demos/image_inpainting_demo/image_inpainting_demo.py
```diff
@@ -33,30 +34,103 @@ def build_argparser():
...
+    img = cv2.resize(img, (inpainting_processor.input_width,inpainting_processor.input_height))
 
-    img = cv2.imread(args.input, cv2.IMREAD_COLOR)
-    masked_image, output_image = inpainting_processor.process(img)
+    if args.rnd:
+        mask = createRandomMask(args.parts,args.max_vertex,args.max_length,args.max_brush_width,
+                                inpainting_processor.input_height,inpainting_processor.input_width)
+    else:
+        top = np.full(img.shape,mask_color,np.uint8)
+        mask = cv2.inRange(img,top,top) / 255
```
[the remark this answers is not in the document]
fzhar: Nope. It will compare RGB components separately, but we need comparison for all 3 components in one place. It can be done using additional boolean operations, but it would be similar to existing code by complexity

demos/python_demos/image_inpainting_demo/image_inpainting_demo.py
```diff
@@ -33,30 +34,103 @@ def build_argparser():
...
+    #--- Inpaint and show results
+    output_image = inpainting_processor.process(masked_image,mask)
     concat_imgs = np.hstack((masked_image, output_image))
-    cv2.putText(concat_imgs, 'summary: {:.1f} FPS'.format(
-            float(1 / inpainting_processor.infer_time)), (5, 15), cv2.FONT_HERSHEY_COMPLEX, 0.5, (0, 0, 200))
-    if not args.no_show:
-        cv2.imshow('Image Inpainting Demo', concat_imgs)
-        cv2.waitKey(0)
+    cv2.putText(concat_imgs, 'summary: {:.1f} FPS'.format(float(1 / inpainting_processor.infer_time)), (5, 15), cv2.FONT_HERSHEY_COMPLEX, 0.5, (0, 0, 200))
+    return concat_imgs,output_image
```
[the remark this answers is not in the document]
fzhar: There's overlay text as well taking local variable (delay) as argument. Too much efforts to move everything.

demos/python_demos/image_inpainting_demo/inpainting.py
```diff
@@ -40,60 +40,20 @@ def __init__(self, ie, model_path, parts, max_brush_width, max_length, max_verte
...
+
+    def process(self, src_image, mask):
+        image = np.transpose(src_image, (2, 0, 1))
         mask = np.transpose(mask, (2, 0, 1))
         image = np.expand_dims(image, axis=0)
         mask = np.expand_dims(mask, axis=0)
         output = self.infer(image, mask)
 
         output = np.transpose(output, (0, 2, 3, 1)).astype(np.uint8)
-        output[0] = cv2.cvtColor(output[0], cv2.COLOR_RGB2BGR)
```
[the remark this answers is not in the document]
fzhar: That was a bug. It was flipping colors (in original code)

demos/python_demos/image_inpainting_demo/inpainting_gui.py
```diff
@@ -0,0 +1,131 @@
...
+        self.maskColor=(255,0,0)
+        self.radius=10
+        self.oldPoint=(-1,-1)
+
+        self.inpainter = ImageInpainting(IECore(),modelPath,device)
+   
+        self.img = cv2.resize(srcImg,(self.inpainter.input_width,self.inpainter.input_height))
+        self.originalImg = self.img.copy()
+        self.label = ""
+        self.mask = np.zeros((self.inpainter.input_height,self.inpainter.input_width,1),dtype=np.float32)
```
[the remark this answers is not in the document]
fzhar: That was taken from original implementation. As I understand, inference assumes mask to be FP32

demos/python_demos/image_inpainting_demo/inpainting_gui.py
```diff
@@ -0,0 +1,131 @@
...
+
+from inpainting import ImageInpainting
+from openvino.inference_engine import IECore
+
+class InpaintingGUI(object):
+    def __init__(self, srcImg, modelPath, device="CPU"):
+        self.wndName="Inpainting (press H for help)"
+        self.maskColor=(255,0,0)
+        self.radius=10
+        self.oldPoint=(-1,-1)
```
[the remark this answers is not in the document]
fzhar: Good point. I've  changed it. It's not a dummy value BTW, it flags that there's no previously drawn point in current polygon (drawing is just started)

demos/python_demos/image_inpainting_demo/inpainting_gui.py
```diff
@@ -0,0 +1,131 @@
...
+import cv2
+import numpy as np
+
+from inpainting import ImageInpainting
+from openvino.inference_engine import IECore
+
+class InpaintingGUI(object):
+    def __init__(self, srcImg, modelPath, device="CPU"):
+        self.wndName="Inpainting (press H for help)"
+        self.maskColor=(255,0,0)
```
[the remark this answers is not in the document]
fzhar: might be changed later, so let's keep them as is

demos/python_demos/image_inpainting_demo/image_inpainting_demo.py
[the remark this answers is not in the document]
IRDonch: Well, my thought was that you could save the output after each time the image is inpainted. But on second thought, it isn't very intuitive, so never mind.

demos/python_demos/image_inpainting_demo/inpainting.py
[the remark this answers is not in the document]
IRDonch: How was it a bug? The network is documented to output images in the RGB format. Is that wrong?

demos/python_demos/image_inpainting_demo/inpainting_gui.py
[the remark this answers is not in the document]
IRDonch: Right, I forgot that the mask is used as an input to the network.

demos/python_demos/image_inpainting_demo/inpainting_gui.py
IRDonch: > Nope, we should just check if something is pressed, but don't wait.

Why?

fzhar: Because there're mouse move/press events to be processed

demos/python_demos/image_inpainting_demo/inpainting_gui.py
IRDonch: Well, okay.

demos/python_demos/image_inpainting_demo/inpainting_gui.py
IRDonch: Right, forgot about that extra dimension in the mask. Then what about `self.image[np.squeeze(self.mask, -1) > 0] = 0`?

demos/python_demos/image_inpainting_demo/inpainting_gui.py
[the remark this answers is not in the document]
IRDonch: Once you start changing them, you can turn them into variables. But right now, they are constants.

demos/python_demos/image_inpainting_demo/image_inpainting_demo.py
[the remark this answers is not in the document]
IRDonch: Fair enough.

demos/python_demos/image_inpainting_demo/image_inpainting_demo.py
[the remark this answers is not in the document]
IRDonch: A user who specifies `--no_show` will expect that the GUI will not be shown; if this is impossible, then the application should notify the user instead of silently ignoring the option.

demos/python_demos/image_inpainting_demo/inpainting_gui.py
IRDonch: Aren't those processed _in_ `waitKey`?

demos/python_demos/image_inpainting_demo/inpainting.py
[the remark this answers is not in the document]
IRDonch: But the model input is BGR.

demos/python_demos/image_inpainting_demo/image_inpainting_demo.py
```diff
@@ -33,30 +33,102 @@ def build_argparser():
...
+            next_y = np.clip(next_y, 0, h - 1).astype(np.int)
+            next_x = np.clip(next_x, 0, w - 1).astype(np.int)
+            cv2.line(mask, (start_y, start_x), (next_y, next_x), 1, brush_width)
+            cv2.circle(mask, (start_y, start_x), brush_width // 2, 1)
+
+            start_y, start_x = next_y, next_x
+    return mask
+
+
+def inpaintRandomHoles(img,args):
```
Wovchena, line 85:
```suggestion
def inpaintRandomHoles(img, args):
```

demos/python_demos/image_inpainting_demo/inpainting_gui.py
```diff
@@ -0,0 +1,123 @@
...
+        if self.isOriginalShown:
+            backbuffer = self.originalImg.copy()
+            sz = cv2.getTextSize("Original", cv2.FONT_HERSHEY_COMPLEX, 0.75, 1)[0]
+            imgWidth = backbuffer.shape[1]
+            labelArea = backbuffer[margin:sz[1]+pad*2+margin, imgWidth-margin-(sz[0]+pad*2):imgWidth-margin]
+            labelArea //= 2
+            cv2.putText(backbuffer, "Original", (imgWidth-margin-sz[0]-pad, margin+sz[1]+pad), cv2.FONT_HERSHEY_COMPLEX, 0.75, (128, 255, 128))
+        else:
+            backbuffer = self.img.copy()
+            backbuffer[np.squeeze(self.mask, -1) > 0]=self.maskColor
```
Wovchena, line 111:
```suggestion
            backbuffer[np.squeeze(self.mask, -1) > 0] = self.maskColor
```

demos/python_demos/image_inpainting_demo/image_inpainting_demo.py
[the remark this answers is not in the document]
Wovchena: > let's keep these parameters as optional

Applying @IRDonch's suggestion keeps `--no_show` optional.

> It's noted in help that they will be ignored in GUI mode. I don't want to confuse users with extra error messages indeed.

That requires careful help reading. Error notification is more explicit and less confusing as it detects a case containing two opposite settings: gui mode and disabling that gui.

demos/python_demos/image_inpainting_demo/inpainting.py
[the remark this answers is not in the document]
fzhar: Model input and output are now BGR. Additional conversions are removed. This change is implemented here: https://github.com/openvinotoolkit/open_model_zoo/pull/1499

demos/python_demos/image_inpainting_demo/image_inpainting_demo.py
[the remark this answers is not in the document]
fzhar: done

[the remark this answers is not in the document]
IRDonch: OK, but you need to update the parameter description to reflect the new semantics.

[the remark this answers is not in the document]
fzhar: Parameter description is updated, but it is not shown here for some GitHub reason. Now it says "Cannot be used in GUI mode"

models/public/gmcnn-places2-tf/freeze_model.patch
```diff
@@ -35,7 +35,7 @@ index 0000000..c3fabef
 +
 +        output = model.evaluate(input_image, input_mask, config)
 +        output = (output + 1) * 127.5
-+        output = tf.minimum(tf.maximum(output[:, :, :, ::-1], 0), 255)
++        output = tf.minimum(tf.maximum(output, 0), 255)
 +
 +        vars_list = tf.get_collection(tf.GraphKeys.GLOBAL_VARIABLES)
 +        assign_ops = [tf.assign(x, tf.train.load_variable(input_checkpoints, x.name)) for x in vars_list]
```
IRDonch, line 41: Somehow changes from #1499 ended up here. Please remove them.

fzhar: Fixed. Branches were occasionally mixed, sorry for that.

IRDonch (approved):

PR #1415, opened 2020-08-17
gaze_estimation: add images_capture

Wovchena:


IRDonch:
The test seems to be frozen; could you investigate?

Wovchena:
Fixed

demos/gaze_estimation_demo/src/eye_state_estimator.cpp
```diff
@@ -50,41 +50,43 @@ void EyeStateEstimator::rotateImageAroundCenter(const cv::Mat& srcImage,
...
-
-    cv::Mat leftEyeImageRotated, rightEyeImageRotated;
-    rotateImageAroundCenter(leftEyeImage, leftEyeImageRotated, roll);
-    rotateImageAroundCenter(rightEyeImage, rightEyeImageRotated, roll);
-    leftEyeImage = leftEyeImageRotated;
-    rightEyeImage = rightEyeImageRotated;
-
+    auto leftEyeBoundingBox = createEyeBoundingBox(outputResults.faceLandmarks[0], outputResults.faceLandmarks[1]);
     outputResults.leftEyeBoundingBox = leftEyeBoundingBox;
+    if (outputResults.faceLandmarks[0] != outputResults.faceLandmarks[1]) {
```
IRDonch, line 59: Maybe it would be better to check `leftEyeBoundingBox` here instead of the original landmarks? It is `leftEyeBoundingBox` that's used as the ROI.

Wovchena: Ok, but don't see a difference as there are comments pointing out the linkage.

IRDonch: The more obvious the connection, the better.

Wovchena: Well, `outputResults.faceLandmarks[0] != outputResults.faceLandmarks[1]` suggests there is a connection, while checking `leftEyeBoundingBox` doesn't :)

IRDonch: I mean a connection between the condition in the `if` statement and the body. :-)

PR #1423, opened 2020-08-17
IE Config helper for demos

AnthonyQuantum:
- Accept device names written in lowercase (ex. -d cpu)
- Move all code for config setting to the common function (at least for classification demo and object detection ssd demo)
- Force Myriad to use 1 stream in demos which don't support multiple streams

- Test everything on Myriad (can't do this right now)

@Wovchena, could you please take a look at the general idea?


demos/common/src/ie_config_helper.cpp
```diff
@@ -0,0 +1,95 @@
...
+#include "samples/ie_config_helper.hpp"
+
+void configureInferenceEngine(Core& ie,
+                              std::string& deviceString,
+                              std::string& deviceInfo,
+                              const std::string& lString,
+                              const std::string& cString,
+                              bool pc) {
+    for (char& ch : deviceString) {
+        ch = std::toupper(ch);
```
Wovchena, line 14: This can be tricky. In case of multiple sticks device names start to contain lower case. I don't know if IE accepts that ids capitalized. It can probably be tested on https://devcloud.intel.com/edge/

demos/common/src/ie_config_helper.cpp
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    std::stringstream strBuffer;
```
Wovchena, line 17: I don't think we really need the rest of the function:

It doesn't seem to be useful to print the version: we didn't even notice that some demos print just a ptr: https://github.com/openvinotoolkit/open_model_zoo/pull/1362 But version printing spams in terminal. Even if demos will keep printing the version, it's not about setting up ie.

There are no models left in omz that require CPU or GPU extensions.

`-pc` is usually for profiling a net. I would leave it to `benchmark_app`

demos/common/src/ie_config_helper.cpp
```diff
@@ -0,0 +1,95 @@
...
+    /** Per layer metrics **/
+    if (pc) {
+        ie.SetConfig({ { PluginConfigParams::KEY_PERF_COUNT, PluginConfigParams::YES } });
+    }
+}
+
+std::map<std::string, std::string> createConfig(const std::string& deviceString,
+                                                const std::string& nstreamsString,
+                                                int nthreads,
+                                                bool minLatency) {
```
Wovchena, line 41: If the option is kept, it should also affect MYRIAD.

demos/common/src/ie_config_helper.cpp
```diff
@@ -0,0 +1,95 @@
...
+                                                                 ? std::to_string(deviceNstreams.at(device))
+                                                                 : CONFIG_VALUE(GPU_THROUGHPUT_AUTO)) });
+
+            if (deviceString.find("MULTI") != std::string::npos && devices.find("CPU") != devices.end()) {
+                // multi-device execution with the CPU + GPU performs best with GPU throttling hint,
+                // which releases another CPU thread (that is otherwise used by the GPU driver for active polling)
+                config.insert({ CLDNN_CONFIG_KEY(PLUGIN_THROTTLE), "1" });
+            }
+        } else if (device == "MYRIAD") {
+            if (!nstreamsString.empty()) {
```
Wovchena, line 88: You should check `deviceNstreams` instead

demos/common/src/ie_config_helper.cpp
```diff
@@ -0,0 +1,95 @@
...
+                config.insert({ CONFIG_KEY(CPU_THROUGHPUT_STREAMS), "1" });
```
Wovchena, line 89: I think this is the option you need https://github.com/openvinotoolkit/openvino/blob/f17262e05f448bc69b2162c137ce0736d3e48cfb/inference-engine/include/vpu/myriad_config.hpp#L60

You have
> Force Myriad to use 1 stream in demos which don't support multiple streams

in the description. But apply the config to the one that does support multiple streams.

There could be two different functions:
* the smart one for classification demo and object detection ssd demo
* naive for simple demos using only one stream

AnthonyQuantum: I've tried this option `VPU_MYRIAD_CONFIG_KEY(MYRIAD_THROUGHPUT_STREAMS)`, but the build fails because `MYRIAD_THROUGHPUT_STREAMS` was not declared.

I use recent Openvino version (built from Github repo yesterday) and include `vpu/myriad_config.hpp`. There is an assumption that this thing was not fully implemented yet (the commit https://github.com/openvinotoolkit/openvino/commit/66aa119b11ba69977b37d25872951a9e17a3565c was merged only a few days ago). Maybe I should open an issue.

Wovchena: > `MYRIAD_THROUGHPUT_STREAMS` was not declared

How is this possible its right in `vpu/myriad_config.hpp`.

According to the commit you pointed `ie::MYRIAD_THROUGHPUT_STREAMS` is supposed to be used directly instead of `VPU_MYRIAD_CONFIG_KEY(THROUGHPUT_STREAMS)` (or `VPU_MYRIAD_CONFIG_KEY(MYRIAD_THROUGHPUT_STREAMS)` which I can't find to be mentioned in the diff at all). But it's better to ask the author.

demos/common/src/ie_config_helper.cpp
```diff
@@ -3,36 +3,49 @@
...
+#include <vpu/myriad_config.hpp>
+
+void formatDeviceString(std::string& deviceString) {
+    bool preserveCase = false;
 
-    std::stringstream strBuffer;
-    strBuffer << ie.GetVersions(deviceString);
-    deviceInfo = strBuffer.str();
+    for (size_t i = 0; i < deviceString.size(); ++i) {
+        // These two conditions handle the special case of MYRIAD device names in "ma1234" format, where letters should
```
AnthonyQuantum, line 15: https://devcloud.intel.com/edge/ does not report NCS names in `ie.GetAvailableDevices()`, but I found a name like this `ma2485` in pbsnodes and it looks similar to NCS names that I've seen before while working with real NSCes.

demos/common/src/ie_config_helper.cpp
```diff
@@ -84,9 +97,15 @@ std::map<std::string, std::string> createConfig(const std::string& deviceString,
...
-                config.insert({ CONFIG_KEY(CPU_THROUGHPUT_STREAMS), "1" });
+        } else if (device == "MYRIAD" || device.find("ma") == 0) {
+            if (minLatency) {
+                config.insert({ InferenceEngine::MYRIAD_THROUGHPUT_STREAMS, "1" });
+                continue;
+            }
+
+            if (deviceNstreams.count(device) > 0) {
+                config.insert({ InferenceEngine::MYRIAD_THROUGHPUT_STREAMS,
+                                std::to_string(deviceNstreams.at(device)) });
```
AnthonyQuantum, line 108: There's no MYRIAD_THROUGHPUT_AUTO, so I don't insert anything

AnthonyQuantum:
@Wovchena 
I should add the same set of functions (or a wrapper) for Python demos, right?
Now configs are used in 3d_segmentaton, action_recognition, and object_detection Python demos.

demos/common/src/ie_config_helper.cpp
```diff
@@ -0,0 +1,114 @@
...
+            preserveCase = false;
+        }
+
+        if (!preserveCase) {
+            deviceString[i] = std::toupper(deviceString[i]);
+        }
+    }
+}
+
+std::map<std::string, std::string> createSimpleConfig(const std::string& deviceString) {
```
Wovchena, line 30: The idea I had behind introducing the simple config is that demos will use it.

This config was meant to be used with demos that have only one stream. `CONFIG_VALUE(CPU_THROUGHPUT_AUTO)` may be greater than 1. For CPU and GPU the default value is already 1, it may be better to add `{{InferenceEngine::MYRIAD_THROUGHPUT_STREAMS, "1"}}` to network loading unconditionally in each one stream demo.

demos/common/src/ie_config_helper.cpp
Wovchena: `ie.GetAvailableDevices()` reports NCS names when there are more than one of them. You should try a set up with more than one stick. The name you've found looks correct, but it would be great to know if IE can take that name capitalized.

demos/common/src/ie_config_helper.cpp
```diff
@@ -0,0 +1,114 @@
...
+            config.insert({ CONFIG_KEY(GPU_THROUGHPUT_STREAMS), (deviceNstreams.count(device) > 0
+                                                                 ? std::to_string(deviceNstreams.at(device))
+                                                                 : CONFIG_VALUE(GPU_THROUGHPUT_AUTO)) });
+
+            if (deviceString.find("MULTI") != std::string::npos && devices.find("CPU") != devices.end()) {
+                // multi-device execution with the CPU + GPU performs best with GPU throttling hint,
+                // which releases another CPU thread (that is otherwise used by the GPU driver for active polling)
+                config.insert({ CLDNN_CONFIG_KEY(PLUGIN_THROTTLE), "1" });
+            }
+        } else if (device == "MYRIAD" || device.find("ma") == 0) {
```
Wovchena, line 100: In case of `MYRIAD` the device name always starts with `MYRIAD` as far as I know. It just has gets suffix if there are multiple sticks. You can search for string that starts with `MYRIAD`.

demos/common/src/ie_config_helper.cpp
Wovchena: Ok

demos/object_detection_demo_ssd_async/main.cpp
```diff
@@ -105,68 +106,15 @@ int main(int argc, char *argv[]) {
...
-                                                                  : CONFIG_VALUE(CPU_THROUGHPUT_AUTO)) });
-
-                minLatencyConfig.insert({ CONFIG_KEY(CPU_THROUGHPUT_STREAMS), "1" });
-            } else if (device == "GPU") {
-                userSpecifiedConfig.insert({ CONFIG_KEY(GPU_THROUGHPUT_STREAMS),
-                                (deviceNstreams.count(device) > 0 ? std::to_string(deviceNstreams.at(device))
-                                                                  : CONFIG_VALUE(GPU_THROUGHPUT_AUTO)) });
+        std::string deviceInfo;
+        std::string deviceString = FLAGS_d;
+        formatDeviceString(deviceString);
```
Wovchena, line 111: You will need to copy the string for both demos. I suggest returning a copy from `formatDeviceString()`.

Wovchena (commented):
> I should add the same set of functions (or a wrapper) for Python demos, right?
Now configs are used in 3d_segmentaton, action_recognition, and object_detection Python demos.

Yes, but 3d_segmentaton and action_recognition aren't special: they just set `CPU_THREADS_NUM` which already is equal to a number of threads by default and `VPU_HW_STAGES_OPTIMIZATION` is deprecated and its equivalent is enabled by default.

demos/object_detection_demo_ssd_async/main.cpp
AnthonyQuantum: I also use `deviceString` in `ie.LoadNetwork()`, so I think It's better to do the formatting task only once and then just make string copies when needed.

demos/common/src/ie_config_helper.cpp
AnthonyQuantum: "one stream demo" is a demo in which the user can't specify the number of streams on startup? Therefore, everything except for classification, object detection ssd and security barrier is considered "one stream demo"?

Wovchena: Seems to be so, but may be there are other demos utilizing more than one InferRequest for a given net and having default(2) number of streams for VPU can be beneficial for them.

demos/object_detection_demo_ssd_async/main.cpp
Wovchena: `deviceString` is not modified before `ie.LoadNetwork()`, you would reuse what `formatDeviceString()` returned.

demos/common/src/ie_config_helper.cpp
AnthonyQuantum: > There are no models left in omz that require CPU or GPU extensions.
> -pc is usually for profiling a net. I would leave it to benchmark_app

Then I can remove `-l`, `-c` and `-pc` from all demos, right?

AnthonyQuantum:
@IRDonch, could you please update the OpenVINO version in CI?
This PR uses `MYRIAD_THROUGHPUT_STREAMS`, which was added on Aug 17 here https://github.com/openvinotoolkit/openvino/commit/66aa119b11ba69977b37d25872951a9e17a3565c .

IRDonch:
OK, I'll do that when I can.

vladimir-dudnik:
Jenkins please retry a build

vladimir-dudnik:
@AnthonyQuantum 01:22:45    c:\jenkins\workspace\omz\precommit\demos_windows10\src\demos\common\include\samples\ie_config_helper.hpp(16): fatal error C1083: Cannot open include file: 'vpu/myriad_config.hpp': No such file or directory

AnthonyQuantum:
> @AnthonyQuantum 01:22:45 c:\jenkins\workspace\omz\precommit\demos_windows10\src\demos\common\include\samples\ie_config_helper.hpp(16): fatal error C1083: Cannot open include file: 'vpu/myriad_config.hpp': No such file or directory

@vladimir-dudnik, it's probably because Windows 10 test machine has older version of OpenVINO. `vpu/myriad_config.hpp` was added after 2020.4. Ubuntu test shows no issues related to `vpu/myriad_config.hpp`, because it has 2021.1 installed.

I've checked the Windows test log file, but there's no OpenVINO version specified. So, I'm not sure which version is installed there.

vladimir-dudnik:
Jenkins please retry a build

vladimir-dudnik:
Jenkins please retry a build

PR #1427, opened 2020-08-18
AC: enabled suffixes ignoring in sr converter

eaidova:
and handled scenario when lr suffix is empty string


tools/accuracy_checker/accuracy_checker/annotation_converters/super_resolution_converter.py
```diff
@@ -111,51 +113,111 @@ def configure(self):
             raise ConfigError('provided not existing loader')
         self.generate_upsample = self.get_value_from_config('generate_upsample')
         self.upsample_factor = self.get_value_from_config('upsample_factor')
+        self.relative_dir = ''
+        if self.ignore_suffixes:
```
Anna-Mironova, line 117: Please add check for hr_dir existence

eaidova: added

Anna-Mironova (approved):

PR #1430, opened 2020-08-19
Fix WER metric, rename greedy_wer to wer, add CER metric for speech recognition

AlexeyKruglov:
The existing implementation of "wer" metric computed CER metric instead, and incorrectly averaged it over samples.  The existing greedy_wer metric was the correct WER metric, consistent with the other implementations (for example, Mozilla DeepSpeech).


eaidova:
Jenkins ok to test

eaidova:
@AlexeyKruglov thank you, please also update info about metrics in readme: https://github.com/openvinotoolkit/open_model_zoo/blob/develop/tools/accuracy_checker/accuracy_checker/metrics/README.md

eaidova (approved):

eaidova:
@mzuevx do you agree with this changes?

mzuevx:
@eaidova, yes, let's try it. I'll fix tests-private configs after merging this PR

PR #1444, opened 2020-08-21
AC: removed tqdm and added sort by images size for ms_coco converter

Anna-Mironova:


tools/accuracy_checker/accuracy_checker/annotation_converters/ms_coco.py
```diff
@@ -104,8 +103,16 @@ def convert(self, check_content=False, progress_callback=None, progress_interval
...
-        if self.sort_annotations:
+        if self.sort_annotations and self.sort_annotations_by_size:
+            image_shape = np.array([value[2] for value in image_ids])
+            image_ids = list(np.array(image_ids)[np.lexsort((np.array(image_ids)[:, 0],
+                                                             image_shape[:, 1], image_shape[:, 0]))])
+        if self.sort_annotations and not self.sort_annotations_by_size:
             image_ids.sort(key=lambda value: value[0])
+        if self.sort_annotations_by_size and not self.sort_annotations:
+            image_shape = np.array([value[2] for value in image_ids])
+            image_ids = list(np.array(image_ids)[np.lexsort((image_shape[:, 1], image_shape[:, 0]))])
```
eaidova, line 114: not clear which difference between first if statement and last

tools/accuracy_checker/accuracy_checker/annotation_converters/ms_coco.py
```diff
@@ -105,7 +105,11 @@ def convert(self, check_content=False, progress_callback=None, progress_interval
         image_ids = [(image['id'], image['file_name'], np.array([image['height'], image['width'], 3]))
                      for image in image_info]
         if self.sort_annotations:
-            image_ids.sort(key=lambda value: value[0])
+            if self.sort_key == 'image_id':
+                image_ids.sort(key=lambda value: value[0])
+            else:
+                image_ids.sort(key=lambda value: tuple(value[2]))
```
eaidova, line 111: on my view better to have dict with lambda functions as values and key is sort_key parameters. It will be easy extend in future if we will need to add any other sort keys

Anna-Mironova: Done

eaidova (approved):

PR #1454, opened 2020-08-25
Create human_pose_estimation_2d_demo.py

Chen-MingChang:
Many customers have requirements for the python version of human pose estimation demonstration, the demo can provide them for easy deployment.
Based on "human_pose_estimation_3d_demo.py" to modify it to 2d human pose estimation, it can be executed using "human-pose-estimation-0001", which is a bottom-up model not suitable to "single_human_pose_estimation_demo".


eizamaliev (changes requested):
Add README.md

playboy007gy:
> Add README.md

Done

Daniil-Osokin:
@playboy007gy Will [this](https://github.com/openvinotoolkit/training_extensions/blob/develop/pytorch_toolkit/human_pose_estimation/demo.py) demo work?

playboy007gy:
> @ playboy007gy請問[這個](https://github.com/openvinotoolkit/training_extensions/blob/develop/pytorch_toolkit/human_pose_estimation/demo.py)演示的工作？

playboy007gy:
> @playboy007gy Will [this](https://github.com/openvinotoolkit/training_extensions/blob/develop/pytorch_toolkit/human_pose_estimation/demo.py) demo work?

It can work by easily modifying the output layer name from 'modules/inference_engine.py'

Need to replace 'Mconv7_stage2_L1' to 'Conv_149', and 'Mconv7_stage2_L2' to 'Conv_146''.

Daniil-Osokin:
So it is possibly better to extend (or just use for the demo purposes) that code rather copy-paste one from 3D pipeline if you need pure Python demo.

playboy007gy:
> So it is possibly better to extend (or just use for the demo purposes) that code rather copy-paste one from 3D pipeline if you need pure Python demo.

Sorry for not sure what you are referring to...
I guess you mean to extend the demo, let it recognize the name of the output node without setting it. Right?
If so, I can try to modify the demo.

Daniil-Osokin:
I mean that there is a pure Python demo already. Why just not to use it? Demo in Python will be quite slow anyway, due to results post-processing. And there is fast demo, which already in C++.

playboy007gy:
> I mean that there is a pure Python demo already. Why just not to use it? Demo in Python will be quite slow anyway, due to results post-processing. And there is fast demo, which already in C++.



> I mean that there is a pure Python demo already. Why just not to use it? Demo in Python will be quite slow anyway, due to results post-processing. And there is fast demo, which already in C++.

Oh, maybe I didn't express it well
The python demo in the [link ](https://github.com/openvinotoolkit/training_extensions/tree/develop/pytorch_toolkit/human_pose_estimation) is for pytorch model, not for IR.
The purpose of this pull is that openVINO currently lacks a python demo for bottom-up model to estimate 2D human poses.
Although C++ is faster than python in terms of execution speed, many customers still ask python demo for easier deployment.
For this reason, I had made this demo and wanted to provide it to people who have the same needs.

Daniil-Osokin:
@playboy007gy Hi! Recently guys added [Python demo for EfficientHRNet human pose estimation](https://github.com/openvinotoolkit/open_model_zoo/pull/1718) (bottom-up) and also generalized it to support the `human-pose-estimation-001` model. Looks like it resolves this.

vladimir-dudnik:
@Chen-MingChang I'll close this PR as desired functionality has been implemented in different PR

PR #1458, opened 2020-08-26
AC: support input data stored in pickle

eaidova:


vladimir-dudnik:
@eaidova some checks failed, is that because of this warning?
12:20:40  accuracy_checker/representation/regression_representation.py:122: [C0326(bad-whitespace), ] Exactly one space required after assignment
12:20:40              data_source =self.metadata['data_source']

vladimir-dudnik (approved):

PR #1461, opened 2020-08-26
VoronovaIntern

VoronovaIntern:


vladimir-dudnik:
@VoronovaIntern would be nice to add to PR description what it is about. Also, please pay attention to merge conflicts. Seems you need to rebase your branch to the latest develop

demos/object_detection_demo_ssd_async/include/yolov3.hpp
```diff
@@ -0,0 +1,58 @@
+#pragma once
```
vladimir-dudnik, line 1: if we can handle not only tolo-v3 but also other variants it would be better to name file simple yolo.hpp

IRDonch:
@Wovchena @eizamaliev Am I understanding correctly that this is obsoleted by #1639?

Wovchena:
#1639 and @ivanvikhrev's work will cover that.

IRDonch:
Okay, then I'm closing it.

PR #1463, opened 2020-08-27
demos: treat warnings as errors for WIN32 by default

Wovchena:


IRDonch:
The warning fix seems fine, but I'm against enabling warnings-as-errors by default. A future version of the Microsoft compiler could emit new warnings, thus making already-released code fail to build.

Wovchena:
But for non windows branch we have the similar option set unconditionally.

Wovchena:
It IE deprecation warnings trigger errors

IRDonch:
> But for non windows branch we have the similar option set unconditionally.

Well, I'm against that too. :-)

PR #1464, opened 2020-08-27
AC: Added print info about annotation conversion

Anna-Mironova:


tools/accuracy_checker/accuracy_checker/dataset.py
```diff
@@ -94,6 +105,10 @@ def _load_annotation(self):
             meta_name = self._config.get('dataset_meta')
             if meta_name:
                 meta_name = Path(meta_name)
+                print_info("{dataset_name} dataset metadata will be saved to {file}".format(
+                    dataset_name=self._config['name'], file=meta_name))
```
eaidova, line 109: Do you check path representation, when you print it?
If I not mistaken, it will be printed like PosixPath('path')

Anna-Mironova: I have checked it. Path will be printed as str without object specification

eaidova (approved):

PR #1472, opened 2020-08-27
Mzuevx ac criteo block

mzuevx:
Work around too slow accuracy tests with DLRM models on Criteo datasets.

Key problem: prepare batch data for model inputs from record-oriented annotations consumes much more time than batch inference.
Solution: make processing batch-oriented 

Changes: 
1. Add "block" parameter to criteo_kaggle_dac converter to produce batch-oriented annotations
2. Add "block" parameter to numpy_reader to allow reading full batch input
3. Modification of DLSDKLauncher._data_to_blob() to correctly handle 3D data to 2D input layout
4. Add "block" parameter to ClassificationAdapter to allow whole batch output be transformed into single prediction object
5. Add "match" parameter to "accuracy" metric to allow sklearn.metric.accuracy_score in the context of AC "accuracy" metric 

Performance benefit from these changes is about 40x times

### DIFF
--- a/tools/accuracy_checker/accuracy_checker/adapters/classification.py
+++ b/tools/accuracy_checker/accuracy_checker/adapters/classification.py
@@ -40,7 +40,7 @@ class ClassificationAdapter(Adapter):
             'block': BoolField(
                 optional=True, default=False, description="process whole batch as a single data block"
             ),
-            'classification_output': StringField(optional=True, description='otarget output layer name')
+            'classification_output': StringField(optional=True, description='target output layer name')
         })
 
         return parameters
--- a/tools/accuracy_checker/accuracy_checker/adapters/classification.py
+++ b/tools/accuracy_checker/accuracy_checker/adapters/classification.py
@@ -37,6 +37,9 @@ class ClassificationAdapter(Adapter):
             'argmax_output': BoolField(
                 optional=True, default=False, description="identifier that model output is ArgMax layer"
             ),
+            'block': BoolField(
+                optional=True, default=False, description="process whole batch as a single data block"
+            ),
             'classification_output': StringField(optional=True, description='otarget output layer name')
         })
 
@@ -44,6 +47,7 @@ class ClassificationAdapter(Adapter):
 
     def configure(self):
         self.argmax_output = self.get_value_from_config('argmax_output')
+        self.block = self.get_value_from_config('block')
         self.classification_out = self.get_value_from_config('classification_output')
 
     def process(self, raw, identifiers, frame_meta):
@@ -66,12 +70,21 @@ class ClassificationAdapter(Adapter):
         prediction = np.reshape(prediction, (prediction.shape[0], -1))
 
         result = []
-        for identifier, output in zip(identifiers, prediction):
+        if self.block:
             if self.argmax_output:
-                single_prediction = ArgMaxClassificationPrediction(identifier, output[0])
+                single_prediction = ArgMaxClassificationPrediction(identifiers[0], prediction)
             else:
-                single_prediction = ClassificationPrediction(identifier, output)
+                single_prediction = ClassificationPrediction(identifiers[0], prediction)
+
             result.append(single_prediction)
+
+        else:
+            for identifier, output in zip(identifiers, prediction):
+                if self.argmax_output:
+                    single_prediction = ArgMaxClassificationPrediction(identifier, output[0])
+                else:
+                    single_prediction = ClassificationPrediction(identifier, output)
+                result.append(single_prediction)
 
         return result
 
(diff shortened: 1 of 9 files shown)


eaidova:
jenkins ok to test

tools/accuracy_checker/accuracy_checker/metrics/classification.py
```diff
@@ -23,9 +23,10 @@
     ArgMaxClassificationPrediction
 )
 
-from ..config import NumberField, StringField, ConfigError
+from ..config import NumberField, StringField, ConfigError, BoolField
 from .metric import PerImageEvaluationMetric
 from .average_meter import AverageMeter
+from sklearn.metrics import accuracy_score
```
eaidova, line 29: please make sklearn optional or do not use it at all for such simple metric.
it is problematic dependency for some platforms

tools/accuracy_checker/accuracy_checker/adapters/classification.py
```diff
@@ -37,13 +37,17 @@ def parameters(cls):
             'argmax_output': BoolField(
                 optional=True, default=False, description="identifier that model output is ArgMax layer"
             ),
+            'block': BoolField(
```
eaidova, line 40: please document all added parameters in readme

tools/accuracy_checker/accuracy_checker/metrics/classification.py
mzuevx: fixed

tools/accuracy_checker/accuracy_checker/metrics/classification.py
```diff
@@ -46,27 +50,43 @@ def parameters(cls):
...
-            )
+            ),
+            'match': BoolField(optional=True, default=False)
         })
 
         return parameters
 
     def configure(self):
         self.top_k = self.get_value_from_config('top_k')
+        self.match = self.get_value_from_config('match')
```
eaidova, line 61: please add check that accuracy_score is not None in case when match=False

tools/accuracy_checker/accuracy_checker/metrics/classification.py
```diff
@@ -46,27 +50,43 @@ def parameters(cls):
...
+        else:
+            self.accuracy = []
 
     def update(self, annotation, prediction):
-        return self.accuracy.update(annotation.label, prediction.top_k(self.top_k))
+        if not self.match:
+            return self.accuracy.update(annotation.label, prediction.top_k(self.top_k))
+        else:
+            accuracy = accuracy_score(annotation.label, prediction.label)
+            self.accuracy.append(accuracy)
```
eaidova, line 76: we need to return metric for single sample in quantization case
```suggestion
            self.accuracy.append(accuracy)
            return accuracy
```

tools/accuracy_checker/accuracy_checker/data_readers/data_reader.py
```diff
@@ -396,11 +401,14 @@ def read(self, data_id):
...
-                recno = int(recno)
-                start, _ = Path(data_id).name.split('.')
-                start = int(start)
-                return data[key[0]][recno - start, :]
+                if self.block:
+                    return data[key[0]]
+                else:
+                    recno = field_id.split('_')[-1]
+                    recno = int(recno)
+                    start, _ = Path(data_id).name.split('.')
```
eaidova, line 409: it is not safe for case when file have several suffixes
```suggestion
                    start = Path(data_id).name.split('.')[0]
```

tools/accuracy_checker/accuracy_checker/annotation_converters/criteo_kaggle_dac.py
```diff
@@ -104,14 +107,24 @@ def convert(self, check_content=False, **kwargs):
...
+                        "lS_o_{}{}{}".format(i, self.separator, c_file),
                     ],
-                    y[j, ...]
+                    y[i:i+self.batch, ...]
                 ))
+            else:
+                for j in range(i, i + self.batch):
+                    annotations.append(ClassificationAnnotation(
+                        [
+                            "input.1_{}{}{}".format(j, self.separator, c_file),
```
eaidova, line 123: I right understand it is model specific input name? Or is it common?

mzuevx: Yes, model specific name. Are you want to map this in converter parameters?

eaidova: on my view better keep it neutral in the identifier and map in config using value for specific input layer.
Who knows, maybe in future we will get new model with different input name

mzuevx: Ok

mzuevx: Fixed

tools/accuracy_checker/accuracy_checker/metrics/classification.py
mzuevx: fixed

tools/accuracy_checker/accuracy_checker/adapters/classification.py
mzuevx: Fixed

eaidova (approved):

PR #1474, opened 2020-08-28
security_barrier_camera: check every lpr output dimension

Wovchena:
That avoids a mismatch when the model wrapper have an implicit assumption which dimension is of non 1 size.

Every LPR model output shape is [1, 88, 1, 1] now.
But the implemented approach relaxes the requirement to have a specific order of dimensions.


Wovchena:
@vladimir-dudnik, @eaidova

IRDonch (approved):

IRDonch:
Jenkins please retry a build

PR #1477, opened 2020-08-28
AC: fix instance segmentation to polygon dumping

eaidova:
1. fix for labels stored in numpy array
2. fix for annotation decoding mask (original format stored in raw_mask is not applicable for decoding)


tools/accuracy_checker/accuracy_checker/representation/segmentation_representation.py
```diff
@@ -279,12 +279,12 @@ def to_polygon(self):
             warnings.warn("Polygon can be found only for non-empty mask")
```
Anna-Mironova, line 279: Please change raw_mask to mask in check

Anna-Mironova (approved):

PR #1478, opened 2020-08-28
AC: tf import localization

eaidova:
tf is trouble maker due to specific requirements on protobuf version and issues with its search.
I suggest to import it only if user want use it directly, it reduces troubles connected with tf configuration from user side


vladimir-dudnik (approved):

ngaloppo:
Same issue exists for MO on Catalina. Can we reduce scope on MO as well?

PR #1492, opened 2020-09-01
AC: Added UnsupportedPackage class to raise exceptions in the same format

Anna-Mironova:


tools/accuracy_checker/tests/test_segmentation_representation.py
```diff
@@ -17,12 +17,13 @@
 import numpy as np
 import pytest
 
+from .common import make_segmentation_representation, make_instance_segmentation_representation
+from ..accuracy_checker.utils import UnsupportedPackage
```
eaidova, line 21:
```suggestion
from accuracy_checker.utils import UnsupportedPackage
```

tools/accuracy_checker/accuracy_checker/utils.py
```diff
@@ -798,3 +798,14 @@ def eof(fd):
         fd.seek(next_position)
     fd.close()
     return mdict
+
+
+class UnsupportedPackage():
```
eaidova, line 803:
```suggestion
class UnsupportedPackage:
```

eaidova (approved):

PR #1496, opened 2020-09-01
Add a script to check release readiness

IRDonch:


IRDonch:
CC @vladimir-dudnik

eizamaliev (commented):
I think you should also find `accuracy-check.yml` files.

IRDonch:
> I think you should also find `accuracy-check.yml` files.

I'm not sure this check belongs here. Nothing prevents people from adding `accuracy-check.yml` alongside the model description, so we can enforce that at pull request review time (and potentially add a check to CI). This script is specifically for things we _can't_ enforce at pull request review time.

eizamaliev (approved):

IRDonch:
I'm merging this to `release`, since that's where it's most useful (and it won't cause any regressions, because the `ci` folder isn't included in packages).

PR #1510, opened 2020-09-07
AC: support intermediate metric results logging

eaidova:


tools/accuracy_checker/accuracy_checker/main.py
```diff
@@ -269,6 +269,8 @@ def build_arguments_parser():
         '--profiler_logs_dir', required=False, type=partial(get_path, is_directory=True), default=Path.cwd()
     )
     parser.add_argument('--profile_report_type', required=False, choices=['csv', 'json'], default='csv')
+    parser.add_argument('--intermediate_metrics_results', required=False, default=False, type=cast_to_bool)
+    parser.add_argument('--metrics_interval', required=False, default=1000, type=int)
```
eizamaliev, line 273: Please, could you add some info about it to README.md?

eizamaliev (changes requested):

tools/accuracy_checker/accuracy_checker/main.py
eaidova: yes, but before, I want simply pass precommit check for my code, do not modify readme twice if I want to change something!!!

eaidova: @eizamaliev this feature targeted for OV 2021.2, so I do not hurry to update docs

PR #1556, opened 2020-09-16
universal cmake changes

shubha-ramani:
This patch is to allow seamless building of demos between several different versions of OpenVino to include 1) open source OpenVINO 2) Released OpenVINO 3) cross-compiled OpenVINO 4) internal gitlab OpenVINO

For instance for 2020.4
export ngraph_DIR=/opt/intel/openvino_2020.4.287/deployment_tools/ngraph/cmake/
export TBB_DIR=/opt/intel/openvino_2020.4.287/deployment_tools/inference_engine/external/tbb/cmake/
export OpenCV_DIR=/opt/intel/openvino_2020.4.287/opencv/cmake
export InferenceEngine_DIR=/opt/intel/openvino_2020.4.287/deployment_tools/inference_engine/share

for opensource openVINO:
export InferenceEngine_DIR=/home/shubha/openvino/build
export ngraph_DIR=/home/shubha/openvino/build/ngraph
export TBB_DIR=/home/shubha/openvino/inference-engine/temp/tbb/cmake/
export OpenCV_DIR=/home/shubha/openvino/inference-engine/temp/opencv_4.3.0_ubuntu18/opencv/cmake

cd open_model_zoo/demos
mkdir build
cd build
cmake ..
make

*This recipe works across all types of OpenVINO builds

### DIFF
--- a/demos/CMakeLists.txt
+++ b/demos/CMakeLists.txt
@@ -176,8 +176,8 @@ macro(ie_add_sample)
         target_include_directories(${IE_SAMPLE_NAME} PRIVATE ${IE_SAMPLE_INCLUDE_DIRECTORIES})
     endif()
 
-    target_link_libraries(${IE_SAMPLE_NAME} PRIVATE ${OpenCV_LIBRARIES} ${InferenceEngine_LIBRARIES}
-                                                    ${IE_SAMPLE_DEPENDENCIES} common gflags)
+    target_link_libraries(${IE_SAMPLE_NAME} PRIVATE ${TBB_IMPORTED_TARGETS} ${OpenCV_LIBRARIES} ${InferenceEngine_LIBRARIES}
+                                                    ngraph::ngraph ${IE_SAMPLE_DEPENDENCIES} common gflags)
 
     if(UNIX)
         target_link_libraries(${IE_SAMPLE_NAME} PRIVATE pthread)
@@ -206,6 +206,7 @@ else()
 endif()
 
 find_package(ngraph REQUIRED)
+find_package(TBB REQUIRED tbb)
 
 # collect all samples subdirectories
 file(GLOB samples_dirs RELATIVE ${CMAKE_CURRENT_SOURCE_DIR} *)
--- a/demos/multi_channel/object_detection_demo_yolov3/CMakeLists.txt
+++ b/demos/multi_channel/object_detection_demo_yolov3/CMakeLists.txt
@@ -45,7 +45,7 @@ add_executable(${TARGET_NAME} ${MAIN_SRC} ${MAIN_HEADERS})
 set_target_properties(${TARGET_NAME} PROPERTIES COMPILE_PDB_NAME ${TARGET_NAME})
 
 target_link_libraries(${TARGET_NAME} PRIVATE
-    ${InferenceEngine_LIBRARIES} gflags ${OpenCV_LIBRARIES} ngraph::ngraph monitors multi_channel_common)
+    ${TBB_IMPORTED_TARGETS} ${InferenceEngine_LIBRARIES} ngraph::ngraph gflags ${OpenCV_LIBRARIES} ngraph::ngraph monitors multi_channel_common)
 
 if(COMMAND add_cpplint_target)
     add_cpplint_target(${TARGET_NAME}_cpplint FOR_TARGETS ${TARGET_NAME})
--- a/demos/multi_channel/common/CMakeLists.txt
+++ b/demos/multi_channel/common/CMakeLists.txt
@@ -106,5 +106,5 @@ target_include_directories(${TARGET_NAME} PUBLIC "${CMAKE_CURRENT_SOURCE_DIR}")
 find_package(Threads REQUIRED)
 
 target_link_libraries(${TARGET_NAME}
-    PRIVATE ${InferenceEngine_LIBRARIES} gflags ${OpenCV_LIBRARIES}
+    PRIVATE ${TBB_IMPORTED_TARGETS} ${InferenceEngine_LIBRARIES} ngraph::ngraph gflags ${OpenCV_LIBRARIES}
     PUBLIC Threads::Threads common)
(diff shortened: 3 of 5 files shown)


Wovchena (changes requested):
Demos don't directly depend on ngraph (in most cases) or tbb. These libs are IE dependencies. That means to me it's IE responsibility to find packages if it needs them and to link ngraph and tbb with `${InferenceEngine_LIBRARIES}`. Additionally I can't see IE samples using similar to that patch approach.

As for helping IE to find the libs, I agree that paths to libs should be provided via env vars or cmd arguments for cmake.

demos/multi_channel/object_detection_demo_yolov3/CMakeLists.txt
```diff
@@ -59,7 +59,7 @@ if(MULTICHANNEL_DEMO_USE_TBB)
 endif()
 
 target_link_libraries(${TARGET_NAME}
-    ${InferenceEngine_LIBRARIES} gflags ${OpenCV_LIBRARIES} ngraph::ngraph multi_channel_common)
+        ${TBB_IMPORTED_TARGETS} ${InferenceEngine_LIBRARIES} ngraph::ngraph gflags ${OpenCV_LIBRARIES} multi_channel_common)
```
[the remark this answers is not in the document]
shubha-ramani: Dear @Wovchena  in an ideal world you're correct. This should be fixed in dldt. However this recipe is highly reliable and works every time and we need an urgent solution ASAP. Do you accept the patch ? Please let me know definitively.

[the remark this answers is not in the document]
Wovchena: @ilya-lavrenov, @asomsiko, @ilyachur, can you confirm the problem can't be resolved on IE side quickly?


As for the problem definition, there are some setups when demos build fails because of missing tbb and ngraph. And simple export of `InferenceEngine_DIR`, `TBB_DIR`, `OpenCV_DIR` and `ngraph_DIR` doesn't work without cmake modifications. @shubha-ramani, am I correct?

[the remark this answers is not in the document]
shubha-ramani: @Wovchena  you are correct. With this pull request, all user has to do is set those 4 environment variables you mention - and the demos build will **_succeed every time._** And this is true regardless of whether the openvino build is open source, gitlab, released openvino, cross-compiled or whatever. That is the beauty and elegance of this pull-request.

demos/CMakeLists.txt
```diff
@@ -206,6 +206,7 @@ else()
 endif()
 
 find_package(ngraph REQUIRED)
+find_package(TBB REQUIRED tbb)
```
asomsiko, line 209: @shubha-ramani cold you explain the purpose of that line?

We should be careful to use TBB from IE and not from Yocto. Can you confirm that TBB is taken from IE?

DaliborSegan-Intel: @asomsiko, IE depends on TBB. User should pick right version of TBB by setting TBB_DIR env variable. If the user wants to cross compile samples then they should set TBB_DIR to point to Yocto TBB. In case of IA compilation TBB_DIR should be set to point to TBB version which comes with OpenVINO or just omitted.

demos/CMakeLists.txt
```diff
@@ -176,8 +176,8 @@ macro(ie_add_sample)
         target_include_directories(${IE_SAMPLE_NAME} PRIVATE ${IE_SAMPLE_INCLUDE_DIRECTORIES})
     endif()
 
-    target_link_libraries(${IE_SAMPLE_NAME} PRIVATE ${OpenCV_LIBRARIES} ${InferenceEngine_LIBRARIES}
-                                                    ${IE_SAMPLE_DEPENDENCIES} common gflags)
+    target_link_libraries(${IE_SAMPLE_NAME} PRIVATE ${TBB_IMPORTED_TARGETS} ${OpenCV_LIBRARIES} ${InferenceEngine_LIBRARIES}
+                                                    ngraph::ngraph ${IE_SAMPLE_DEPENDENCIES} common gflags)
```
ilyachur, line 180: Why do we need `nGraph` dependency for sample?
Do samples work with nGraph function and it causes this dependency?

demos/multi_channel/object_detection_demo_yolov3/CMakeLists.txt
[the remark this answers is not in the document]
ilyachur: I don't understand at the current moment why do you need TBB and nGraph. Demos don't touch it, Inference Engine has a private dependency on these components. In general if you use the right environment all should work.
Am I wrong?

demos/CMakeLists.txt
shubha-ramani: @ilyachur The simple answer is, if you don't link ngraph sample will not compile (link). And answer to your question here : "I don't understand at the current moment why do you need TBB and nGraph. Demos don't touch it, Inference Engine has a private dependency on these components. In general if you use the right environment all should work.
Am I wrong?" Respectfully @ilyachur , yes you are wrong. I can prove it to you if you set up some time with me. Or you can try it yourself. Try to build demo against **_a released version of OpenVINO_** (for instance 2020.4) and do not run setupvars.sh. Try building the demos now **_without running setupvars.sh_** .  Try again in the oecore (cross-compiled) environment. Does it work ? You will get errors with ngraph and/or TBB.

ilya-lavrenov: @shubha-ramani who said that we must support scenario when setupvars.sh is not used?

ilyachur: As I said, if you use the right environment all should work. But if you don't use **setupvars.sh** you can have incorrect path. And this causes your issues, because nGraph or TBB cannot be found in the environment path.

vladimir-dudnik:
Jenkins please retry a build

ilya-lavrenov:
I don't think that we need to merge these changes until a proper way is found. This looks like a WA

vladimir-dudnik:
@shubha-ramani I've rebased your branch to the latest state of develop branch, pls check that your proposed changed has not been lost. We have just merged [PR-1567 Simplified multi channel demos cmake](https://github.com/openvinotoolkit/open_model_zoo/pull/1567), which cause conflicts in your PR.

IRDonch:
From what I recall of the internal discussion about this, these changes should no longer be necessary because of a fix made in OpenVINO. Am I correct? If so, please close the PR.

IRDonch:
Closing due to inactivity.

PR #1576, opened 2020-09-21
save mutating dict while iterating over it

morkovka1337:
This patch fixes unsafe mutating of the dictionary while iterating over it. Previous version of this code lead to incorrect conversion of the keys (not all keys were converted from str to int).

### DIFF
--- a/tools/accuracy_checker/accuracy_checker/annotation_converters/im2latex.py
+++ b/tools/accuracy_checker/accuracy_checker/annotation_converters/im2latex.py
@@ -40,7 +40,8 @@ def read_vocab(vocab_path):
     elif vocab_path.suffix == '.json':
         with open(vocab_path, "r") as f:
             vocab_dict = json.load(f)
-            vocab_dict['id2sign'] = {int(k): v for k, v in vocab_dict['id2sign'].items()}
+            id2sign = {int(k): v for k, v in vocab_dict['id2sign'].items()}
+            vocab_dict['id2sign'] = id2sign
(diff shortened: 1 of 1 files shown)


vladimir-dudnik:
Jenkins please retry a build

tools/accuracy_checker/accuracy_checker/annotation_converters/im2latex.py
```diff
@@ -40,9 +40,7 @@ def read_vocab(vocab_path):
     elif vocab_path.suffix == '.json':
         with open(vocab_path, "r") as f:
             vocab_dict = json.load(f)
-            for k, v in vocab_dict['id2sign'].items():
-                del vocab_dict['id2sign'][k]
-                vocab_dict['id2sign'][int(k)] = v
+            vocab_dict['id2sign'] = {int(k): v for k, v in vocab_dict['id2sign'].items()}
```
eaidova, line 43: do you can guarantee that keys are always can be casted to the int?

eaidova: you still modify vocab during iteration
```suggestion
            id2sign = {int(k): v for k, v in vocab_dict['id2sign'].items()}
            vocab_dict['id2sign'] = id2sign
```
or provide json reading scheme which will cast it automatically

eaidova (approved):

eaidova:
Jenkins ok to test

PR #1586, opened 2020-09-22
AC: Added preprocessors rgb/bgr_to_ycrcb and postprocessor restoring ycrcb and converting prediction to rgb/bgr

Anna-Mironova:


tools/accuracy_checker/accuracy_checker/adapters/README.md
```diff
@@ -153,6 +153,12 @@ AccuracyChecker supports following set of adapters:
   * `u_output` - U channel output layer.
   * `v_output` - V channel output layer.
   * `target_color` - taret color space for super resolution image - `bgr` and `rgb` are supported. (Optional, default `bgr`).
+* `super_resolution_gray` - converts output of super resolution model, which return output in GRAY format, to `SuperResolutionPrediction`.
```
eaidova, line 156: not sure that it is correct name, if model works with gray scale image, then it should return gray scale image, not BGR and not RGB

eaidova: Are you sure that you do every thing right?
according to description:

> Postprocessing involves converting the array of pixels into an image that is scaled to a higher resolution. The color channels (Cb, Cr) are also scaled to a higher resolution using bicubic interpolation. Then the color channels are combined and converted back to RGB format, producing the final output image.

It means that you use the first channel from YCbCr image as prediction and rescale other channels from the YCbCr image to new size

tools/accuracy_checker/accuracy_checker/adapters/image_processing.py
```diff
@@ -213,3 +213,26 @@ def process(self, raw, identifiers=None, frame_meta=None):
             results.append(SuperResolutionPrediction(identifier, sr_img))
 
         return results
+
+
+class SuperResolutionGray(ImageProcessingAdapter):
+    __provider__ = 'super_resolution_gray'
+    prediction_types = (SuperResolutionPrediction, )
```
eaidova, line 220: do we actually need specific  adapter or usual super_resolution can work here?

tools/accuracy_checker/accuracy_checker/postprocessor/super_resolution_image_recovery.py
```diff
@@ -0,0 +1,66 @@
...
+
+    annotation_types = (SuperResolutionAnnotation, )
+    prediction_types = (SuperResolutionPrediction, )
+
+    @classmethod
+    def parameters(cls):
+        parameters = super().parameters()
+        parameters.update({
+            'target_color': StringField(optional=True, choices=['bgr', 'rgb'], default='rgb'),
+            'input_size': NumberField(
```
eaidova, line 37: what if they are not equal?

eaidova: BTW, you already have info about input shape in image_meta, so you can extract it from it withoud additional parameters.

tools/accuracy_checker/accuracy_checker/adapters/image_processing.py
Anna-Mironova: No, model works correct with super_resolution adapter. New adapter has been removed

tools/accuracy_checker/accuracy_checker/postprocessor/super_resolution_image_recovery.py
Anna-Mironova: fixed. we need to get not size of image, but size of input of model

tools/accuracy_checker/accuracy_checker/postprocessor/super_resolution_image_recovery.py
```diff
@@ -0,0 +1,72 @@
...
+            'size': NumberField(
+                value_type=int, optional=True, min_value=1,
+                description="Size of model input for both dimensions (height and width) before recovering"
+            )
+        })
+        return parameters
+
+    def configure(self):
+        self.color = cv2.COLOR_YCrCb2BGR if self.get_value_from_config('target_color') == 'bgr' else cv2.COLOR_YCrCb2RGB
+        self.input_height, self.input_width = get_size_from_config(self.config, allow_none=True)
```
eaidova, line 53: you allow none but the code below will be crashed in case if self.input_width=None or self.input_height=None

Anna-Mironova: fixed

eaidova (approved):

PR #1588, opened 2020-09-22
update OMZ models for 2021.1 RC

vladimir-dudnik:
update OMZ models links for 2021.1 RC


eaidova (approved):

eaidova:
@vladimir-dudnik pipeline failed with downloading  FP16-INT8/yolo-v2-tiny-vehicle-detection-0001.bin
22:39:06  ########## Error: Downloaded file is shorter (1086428355 B) than expected (1340567776 B)

is it expected?

IRDonch:
@eaidova That's not the cause of the failure; it's a recoverable error that was resolved by resuming the download. The message was printed at the end because Jenkins can't tell which parts of stderr come after which parts of stdout.

PR #1594, opened 2020-09-24
Command line  and mask processign are changed for image_inpainting_demo.

fzhar:


vladimir-dudnik (commented):
for -am/-ar options it may make sense to avoid ambiguations by having priority of one option higher then other, so if both specified (by mistake) only one will be taken and other ignored (optionally with warning message)

fzhar:
> for -am/-ar options it may make sense to avoid ambiguations by having priority of one option higher then other, so if both specified (by mistake) only one will be taken and other ignored (optionally with warning message)

If both options are specified, demo will show error message with exact explanation : "-am and -ar cannot be used together". Prioritizing them might make user confused (as both options are generally of the same priority it would be confusing - why one of them is made of higher priority ?)

vladimir-dudnik:
both -ar and -am optional. what definitely will confuse user is runtime error while using optional keys.

Wovchena (changes requested):
> both -ar and -am optional. what definitely will confuse user is runtime error while using optional keys.

It's possible to get a runtime error while using optional keys in general. For example `-am 255 255` even without `-ar`.
It's clearer to report an error because it's less probable it can be missed.

demos/python_demos/image_inpainting_demo/image_inpainting_demo.py
```diff
@@ -117,9 +117,13 @@ def main():
         print("Cannot load image " + args.input)
         return -1
 
-    if args.auto:
+    if (not args.auto_mask is None) and args.auto_rnd:
```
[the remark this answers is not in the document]
eizamaliev: May be 
```
if args.auto_mask and args.auto_rnd:
```
more suitable here?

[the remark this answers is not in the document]
fzhar: agree

Wovchena (approved):

eizamaliev (approved):

vladimir-dudnik (approved):

demos/python_demos/image_inpainting_demo/image_inpainting_demo.py
```diff
@@ -82,24 +81,25 @@ def create_random_mask(parts, max_vertex, max_length, max_brush_width, h, w, max
...
 
-def inpaint_random_holes(img, args):
-    mask_color = args.mask_color[::-1] # argument comes in RGB mode, but we will use BGR notation below
-
+def inpaint_auto(img, args):
     ie = IECore()
 
     inpainting_processor = ImageInpainting(ie, args.model, args.device)
 
     #--- Resize to model input and generate random mask
```
IRDonch, line 89: This comment is not quite correct: 1) the order of operations is now reversed; 2) the mask is not necessarily random.

demos/python_demos/image_inpainting_demo/image_inpainting_demo.py
```diff
@@ -46,13 +42,16 @@ def build_arg_parser():
...
-                      help="Optional. Color to be treated as mask (provide 3 RGB components in range of 0...255)."
-                      " Default is 0 0 0. Ignored in GUI mode", default=[0, 0, 0], type=int, nargs=3)
     args.add_argument("--no_show", help="Optional. Don't show output. Cannot be used in GUI mode", action='store_true')
     args.add_argument("-o", "--output", help="Optional. Save output to the file with provided filename."
                       " Ignored in GUI mode", default="", type=str)
-    args.add_argument("-a", "--auto", help="Optional. Use automatic (non-interactive) mode instead of GUI",
+    args.add_argument("-am", "--auto_mask", help="Optional. Use automatic (non-interactive) mode with color mask."
+                      "Provide color to be treated as mask (3 RGB components in range of 0...255). "
+                      "Cannot be used together with -ar.",
+                      default=None, type=int, nargs=3)
```
IRDonch, line 51: I would suggest `metavar='C'` or something similar, otherwise the usage message looks weird.

demos/python_demos/image_inpainting_demo/image_inpainting_demo.py
```diff
@@ -46,13 +42,16 @@ def build_arg_parser():
...
+    args.add_argument("-ar", "--auto_rnd",
+                      help="Optional. Use automatic (non-interactive) mode with random mask for inpainting"
+                      " (with parameters set by -p, -mbw, -mk and -mv). Cannot be used together with -am.",
```
IRDonch, line 54: The names of these options could be improved. Currently they contrast "mask" against "random", which is a false comparison, as both modes involve masks. IMO, `--mask_color` already had a good name, and `--rnd` just needed to be renamed to `--mask_random`. But if you'd like to highlight that these options enable automatic mode, you could name them `--auto_mask_color` and `--auto_mask_random`.

demos/python_demos/image_inpainting_demo/README.md
```diff
@@ -12,11 +12,12 @@ This demo can work in 2 modes:
 Running the application with the `-h` option yields the following usage message:
 
 ```
-usage: image_inpainting_demo.py [-h] -m MODEL [-i INPUT] [-d DEVICE] [-r]
+usage: image_inpainting_demo.py [-h] -m MODEL [-i INPUT] [-d DEVICE]
```
IRDonch, line 15: There is explanatory text above this that should be updated with the new option names.

IRDonch (changes requested):

demos/python_demos/image_inpainting_demo/image_inpainting_demo.py
fzhar: Corrected

demos/python_demos/image_inpainting_demo/image_inpainting_demo.py
fzhar: Done

demos/python_demos/image_inpainting_demo/image_inpainting_demo.py
IRDonch: The short names also need to be updated; right now they're formed inconsistently: **a**uto\_**m**ask\_color **a**uto\_mask\_**r**andom.

fzhar: whatever... -am is renamed to -ac for consistency

IRDonch (approved):

PR #1596, opened 2020-09-24
fix cvs 39310

vladimir-dudnik:
avoid division to zero


vladimir-dudnik:
there is better solution

PR #1606, opened 2020-09-28
super_resolution: update

Wovchena:
Don't fix input names. This can be handy for srgan model, which is
not added, but has AC support.

Resize input image instead of skipping it.

Set batchSize equal to inputImages.size().

While I was sorting out all that problems I replaced
setBatchSize() with reshape(). reshape() is more stable usually so I
kept this change.

### DIFF
--- a/demos/super_resolution_demo/main.cpp
+++ b/demos/super_resolution_demo/main.cpp
@@ -92,35 +92,49 @@ int main(int argc, char *argv[]) {
         slog::info << "Preparing input blobs" << slog::endl;
 
         /** Taking information about all topology inputs **/
-        InputsDataMap inputInfo(network.getInputsInfo());
+        ICNNNetwork::InputShapes inputShapes(network.getInputShapes());
 
-        if (inputInfo.size() != 1 && inputInfo.size() != 2)
+        if (inputShapes.size() != 1 && inputShapes.size() != 2)
             throw std::logic_error("The demo supports topologies with 1 or 2 inputs only");
-
-        const std::string lrInputBlobName = "0";
+        std::string lrInputBlobName = inputShapes.begin()->first;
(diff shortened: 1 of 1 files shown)


demos/super_resolution_demo/main.cpp
```diff
@@ -92,44 +92,62 @@ int main(int argc, char *argv[]) {
...
-        const std::string lrInputBlobName = "0";
+        std::string lrInputBlobName = inputShapes.begin()->first;
+        const SizeVector &lrShape = inputShapes[lrInputBlobName];
+        // A model like single-image-super-resolution-???? may take bicubic interpolation of the input image as the
+        // second input
+        std::string bicInputBlobName;
+        if (inputShapes.size() == 2) {
+            bicInputBlobName = (++inputShapes.begin())->first;
+            const SizeVector &bicShape = inputShapes[bicInputBlobName];
+            if (lrShape.size() != 4 || bicShape.size() != 4) {
```
IRDonch, line 107: The `lrShape.size()` check should be outside of the `if`, otherwise it'll be skipped when there's only one input.

demos/super_resolution_demo/main.cpp
```diff
@@ -92,44 +92,62 @@ int main(int argc, char *argv[]) {
...
+                throw std::logic_error("Number of dimensions for both inputs must be 4");
+            }
+            if (lrShape[2] >= bicShape[2] && lrShape[3] >= bicShape[3]) {
+                lrInputBlobName.swap(bicInputBlobName);
```
IRDonch, line 111: You should also swap `lrShape` and `bicShape`, because `lrShape` is used below and it needs to correspond to `lrInputBlobName`.

IRDonch (changes requested):

IRDonch (approved):

PR #1610, opened 2020-09-29
Add metrics for GAN models

alexmasterdi:
This PR provides realizations of IS(Inception score) and FID(Frechet Inception Distance) metrics.

### DIFF
--- a/tools/accuracy_checker/accuracy_checker/metrics/gan_metrics.py
+++ b/tools/accuracy_checker/accuracy_checker/metrics/gan_metrics.py
@@ -1,12 +1,9 @@
 """
 Copyright (c) 2018-2020 Intel Corporation
-
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
-
       http://www.apache.org/licenses/LICENSE-2.0
-
 Unless required by applicable law or agreed to in writing, software
 distributed under the License is distributed on an "AS IS" BASIS,
 WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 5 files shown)


eaidova:
Jenkins ok to test

tools/accuracy_checker/accuracy_checker/metrics/gan_metrics.py
```diff
@@ -0,0 +1,106 @@
...
+    ImageProcessingAnnotation,
+    ImageProcessingPrediction,
+)
+
+from .metric import FullDatasetEvaluationMetric
+from ..config import NumberField
+
+
+class BaseGanMetric(FullDatasetEvaluationMetric):
+    annotation_types = (ImageProcessingAnnotation, )
```
eaidova, line 29: if my understanding correct, annotations and predictions will be result produced by classification model (probs in case of score and just raw features in frechet distance case), is it true?
ImageProcessingPrediction store GAN image result.

My suggestion to use separated representations for this task created by evaluator (e.g. named RawTensorAnnotation (for classification model inference result on original image), RawTensorPrediction(for classification model inference result on generated image)).

alexmasterdi: Yes, it is true. I created RawTensorRepresentation(annotation and prediction), but it used as container to storing results of classification model (so representation used directly, without adapter and annotation converted).

eaidova: I think it is not problem for us, because these entities will work in the specific context (evaluator). It is great example for usage custom evaluation, where pipeline steps have difference from default defined.

eaidova (approved):

PR #1618, opened 2020-10-01
add serialize capability to source

SDxKeeper:


tools/downloader/common.py
```diff
@@ -286,6 +286,9 @@ class FileSourceHttp(FileSource):
     def __init__(self, url):
         self.url = url
 
+    def __str__(self):
+        return str(self.url)
```
IRDonch, line 290: I don't think that's the right string representation for it. A source is not just an URL, it's an object. I think a representation like `FileSourceHttp(url='...')` would be more appropriate.

IRDonch: It _is_ suitable for `__repr__`, but it is suitable for `__str__` as well. The URL by itself is not a sufficient representation, because a `FileSource` object is more than just an URL.

SDxKeeper: I can add dedicated url method instead of usage of __str__.

IRDonch: Why would we need a dedicated method for this? Nothing calls it.

SDxKeeper: OneZoo does call this method during indexing.

IRDonch: I don't want to add methods for use by internal tools. This is an open source project, and this would just be unused code from an outsider's point of view.

PR #1635, opened 2020-10-07
hbonet-*: remove the output name from the AC configs

IRDonch:


eaidova:
@IRDonch the output was pinned in the config due to case, where model have fake outputs, if now they contains only one output, it makes sense to remove `classification_out` from configs at all

eaidova:
also if we still need to preserve names in config, they should be compatible with internal repo (configs used in quantization), so models should be updated too

IRDonch:
There's only one output, so I removed the names.

eaidova (approved):

PR #1484, opened 2020-08-31
Add semantic-segmentation model HRNetV2

alexmasterdi:
At this moment model based on opset11 from onnx.

PR #1638, opened 2020-10-07
Raise the ONNX opset version used for converting PyTorch models to 11

IRDonch:
This has a few beneficial effects:

* it removes conversion warnings that used to be printed for several models about how certain layers don't work the same way in ONNX until opset 11;

* it allows us to remove one of the regex replacements in the midasnet model, which, I believe, was originally added to work around ONNX limitations;

* it allows us to add future models which require a newer opset.

I have run an accuracy check on all affected models (on CPU), and didn't see any degradation. midasnet actually improved, going from an RMSE of 0.08 to 0.07.

### DIFF
--- a/models/public/midasnet/model.yml
+++ b/models/public/midasnet/model.yml
@@ -43,10 +43,6 @@ framework: pytorch
 postprocessing:
   - $type: regex_replace
     file: models/blocks.py
-    pattern: 'align_corners=True'
-    replacement: 'align_corners=False'
-  - $type: regex_replace
-    file: models/blocks.py
     pattern: 'import torch.nn as nn'
     replacement: 'import torch.nn as nn\nimport torchvision.models as models'
   - $type: regex_replace
--- a/tools/downloader/pytorch_to_onnx.py
+++ b/tools/downloader/pytorch_to_onnx.py
@@ -98,7 +98,7 @@ def convert_to_onnx(model, input_shape, output_file, input_names, output_names):
     model.eval()
     dummy_input = torch.randn(input_shape)
     model(dummy_input)
-    torch.onnx.export(model, dummy_input, str(output_file), verbose=False, opset_version=9,
+    torch.onnx.export(model, dummy_input, str(output_file), verbose=False, opset_version=11,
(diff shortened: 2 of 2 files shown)


IRDonch:
CC @eaidova @vladimir-dudnik

eizamaliev:
Could be implemented in [PR 1484](https://github.com/openvinotoolkit/open_model_zoo/pull/1484). I think you could leave changes only in `midasnet/model.yml`.

eaidova:
@IRDonch according to MO plans, opset11 is not fully supported in 2021.1, so it is kind of risk taking for new models raise opset11 without opportunity to override version

IRDonch:
@eizamaliev I don't agree. This is a global change and it doesn't belong in the HRNet pull request.

@eaidova That's a fair point, but we _can_ add an override if we need to. I just don't see a reason to do so now when all current models work with opset 11.

eaidova (approved):

eizamaliev:
@IRDonch are you sure opset change won't lead to model accuracy metric changes?

eizamaliev (approved):

IRDonch:
@eizamaliev I have compared my accuracy results with the listed reference metrics, and I don't think there were any cases where the difference was more than 0.1%.

Also, I compared the XML files from before and after the changes, and the only models for which they are not identical are 1) models that previously had ONNX conversion warnings (`hbonet-*`, `midasnet`, `single-human-pose-estimation-0001`), and 2) `resnest-50-pytorch`.

IRDonch:
FWIW, I took another look at `resnest-50-pytorch`, and while the IR changed, I think it's semantically equivalent. The only differences look like this:

```
                <layer id="###" name="xxx" type="AvgPool" version="opset1">
-                       <data exclude-pad="true" kernel="2,2" pads_begin="0,0" pads_end="1,1" rounding_type="floor" strides="2,2"/>
+                       <data exclude-pad="true" kernel="2,2" pads_begin="0,0" pads_end="0,0" rounding_type="ceil" strides="2,2"/>
```

So it uses rounding up instead of padding to achieve the same output shape.

PR #1646, opened 2020-10-12
AC: Fix several spelling errors

IRDonch:
A few of the things with errors in them weren't needed at all, so I just removed them.

One of these changes should actually reenable a test that has been accidentally disabled this whole time.

### DIFF
--- a/tools/accuracy_checker/accuracy_checker/launcher/dlsdk_launcher_readme.md
+++ b/tools/accuracy_checker/accuracy_checker/launcher/dlsdk_launcher_readme.md
@@ -87,5 +87,5 @@ launchers:
       batch: 4
     mo_flags:
       - reverse_input_channels
-    cpu_extensions: cpu_extentions_avx512.so
+    cpu_extensions: custom_cpu_extensions.so
 ```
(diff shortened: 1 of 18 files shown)


tools/accuracy_checker/accuracy_checker/launcher/dlsdk_launcher_readme.md
```diff
@@ -87,5 +87,4 @@ launchers:
       batch: 4
     mo_flags:
       - reverse_input_channels
-    cpu_extensions: cpu_extentions_avx512.so
```
eaidova, line 90: we not remove functional opportunity to set extension, so I do not think that we should remove it from docs sample (as I know, it is still used for example for running on arm cpu)

eaidova (approved):

tools/accuracy_checker/accuracy_checker/launcher/dlsdk_launcher_readme.md
IRDonch: I suppose I could keep the key, but what do I use for the value? The `cpu_extensions_avx512` library doesn't exist anymore.

eaidova: cpu_extention.so or custom_cpu_extension.so will be ok (it is just abstract name used in example purposes, for running with extension, of course, user should compile them on own side)

IRDonch: Alright, updated.

eaidova:
Jenkins please retry a build

PR #1649, opened 2020-10-12
Add a script to do basic checks

IRDonch:
The idea is that these should be reasonably quick checks that we can do for every pull request.

I disable the yamllint newline check, because I want the basic checks to be runnable on Windows, and this check fails on Windows (because Git uses native line endings when checking files out).


IRDonch:
This is going to be a draft for now, because:

1. there are currently loads of issues in the codebase of the kind that this script is supposed to detect; and
1. CI support is not ready yet.

However, I'd like to first get feedback on whether you agree with the idea in general, and with the specific checks that I implemented.

FYI, I'd also like to add some other checks later:

* maximum path length check;
* running Flake8 on all Python sources;

I'm skipping these for now to avoid spending too much time at once.

eaidova (commented):
Why can not to use yamlint config from repo root directly in AC precommit?

yamllint -c <path_to_config> is supported, so we do not need extra link.

The second question, do we really need additional check in AC precommit pipeline after these changes? Now it checks links in accruacy_checker/configs directory, which contains the links on files which should be already checked in your check, if my understanding correct?

IRDonch:
@eaidova My plan is to remove the yamllint check from the AC job in favor of this check. The symlink is there to enable a smooth transition.

eaidova (approved):

Wovchena:
git diff has `--check` option. May be it's better to use it to catch trailing whitespaces.
How do IE and MO perform such checks? We could align with them or get some ideas.

IRDonch:
@Wovchena

> git diff has --check option. May be it's better to use it to catch trailing whitespaces.

Yeah, fair point. I originally avoided `git diff --check` because I didn't want to re-read every file yet again, but it only takes a negligible amount of time and offers more checks, so we might as well use it. Reworked.

(sadly, `git diff --check` can't ensure that every file ends with a newline, so I still have to do that check myself)

> How do IE and MO perform such checks? We could align with them or get some ideas.

A quick check shows that they have a lot of trailing whitespace in their code, so they probably don't. 🙂

Wovchena (approved):

eizamaliev (approved):

IRDonch:
Updated to include a `requirements.txt` file (compiled from `requirements.in`).

IRDonch:
Jenkins please retry a build

PR #1657, opened 2020-10-14
Support per-model pre-convert scripts and use it for yolo-v2-tiny-tf

IRDonch:
See the commit messages for a detailed explanation.


models/public/yolo-v2-tiny-tf/accuracy-check.yml
```diff
@@ -5,7 +5,7 @@ models:
       - framework: tf
         tags:
           - FP32
-        model:   yolo-v2-tiny.pb
+        model: yolov2-tiny.pb
```
eizamaliev, line 8: I don't see any use in such renaming since you can specify the output name in the script below.

eizamaliev (changes requested):

tools/downloader/converter.py
```diff
@@ -25,6 +25,25 @@
 import common
 
 
+def run_pre_convert(reporter, model, output_dir, args):
+    script = common.MODEL_ROOT / model.subdirectory / 'pre-convert.py'
+    if not script.exists():
```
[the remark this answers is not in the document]
IRDonch: I think that would be less elegant - you need the script path both to check that it exists and to run it, so I would have to either duplicate the path building or to add an extra parameter.

models/public/yolo-v2-tiny-tf/accuracy-check.yml
IRDonch: Well, I originally changed the name to align it with the name of the weights file. But I suppose that is a gratuitous change. Reverted.

tools/downloader/converter.py
```diff
@@ -25,6 +25,25 @@
 import common
 
 
+def run_pre_convert(reporter, model, output_dir, args):
```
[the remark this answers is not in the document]
IRDonch: Hmm... I don't think "custom convert script" really reflects the semantics here - that would imply that the script replaces the entire conversion sequence with custom code, which isn't the case.

I named it "pre-convert script", because it runs before the standard conversion process. I think `convert_to_supported_format` is a decent proposal, but a) it's a bit unwieldy, and b) the script doesn't necessarily have to do that. I think it's possible that in the future we might use it for other purposes, e.g. to edit a file that's already in a supported format, or to create some auxiliary file that's needed for conversion.

eizamaliev:
@IRDonch what do you think about adding path to pre-convert script into `model.yml`? It could free us from defining script name in the code, also shows model transformation path in one place. It even detects absent of pre-convert script when it necessary.

IRDonch:
> @IRDonch what do you think about adding path to pre-convert script into `model.yml`?

I considered it when I was designing this feature and reject it. The path to the pre-convert script doesn't need to be customizable, much like the names `model.yml`, `accuracy-check.yml` or `quantization.yml`.

> It could free us from defining script name in the code

And condemn us to defining it in every `model.yml` that needs it. 🙂 I think that's a tradeoff for the worse.

> also shows model transformation path in one place.

Sure, that's kind of nice. But I don't think it's enough justification by itself.

> It even detects absent of pre-convert script when it necessary.

We can already detect the absence of a pre-convert script: if it's absent, the model converter will fail.

eizamaliev:
Good point (something like that I was waiting for 🙂)
A little remark to 
> It even detects absent of pre-convert script when it necessary.
We can already detect the absence of a pre-convert script: if it's absent, the model converter will fail.

Current script seeks file `pre-convert.py` and if doesn't find it just skips this step. On contrary, explicitly defined script won't be missed.

But seems to me, that is not enough to persuade you, so just keep in mind.

models/public/yolo-v2-tiny-tf/pre-convert.py
```diff
@@ -0,0 +1,44 @@
...
+import argparse
+import subprocess
+import sys
+
+from pathlib import Path
+
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument('download_dir', type=Path)
+    parser.add_argument('convert_dir', type=Path)
```
eizamaliev, line 26: These arguments are not `download_dir` and `convert_dir` (well, the last one may be) from very this script perspective

eizamaliev (changes requested):

models/public/yolo-v2-tiny-tf/pre-convert.py
IRDonch: What do you mean? `download_dir` is the directory with the downloaded files and `convert_dir` is the directory to put conversion results in.

IRDonch:
> Current script seeks file pre-convert.py and if doesn't find it just skips this step. On contrary, explicitly defined script won't be missed.

Yeah, but then the problem shifts to "what if the developer forgets to define the script"? Ultimately, the developer still has to communicate in _some_ way that a pre-convert script exists.

models/public/yolo-v2-tiny-tf/pre-convert.py
eizamaliev: Script knows nothing about downloading. It is awaiting "model in - model out" parameters

IRDonch: I'm not sure I agree, but it doesn't really matter. Renamed to `input_dir`/`output_dir`, respectively.

eizamaliev (approved):

PR #1671, opened 2020-10-19
AC: update affinity setting via ngraph

eaidova:
to do:

- [x]  resolve question about memory deallocation (for network and function)


tools/accuracy_checker/accuracy_checker/launcher/dlsdk_launcher.py
```diff
@@ -301,6 +306,31 @@ def _set_affinity(self, affinity_map_path):
         for layer in custom_affinity:
             if layer not in automatic_affinity:
                 raise ConfigError('Layer \'{layer}\' is not present in network'.format(layer=layer))
+        if isinstance(ng, UnsupportedPackage):
+            if not hasattr(self.network, 'layers'):
+                ng.raise_error('affinity setting')
+            self._set_affinity_via_layers(custom_affinity, automatic_affinity)
+            return
+        self._set_affinity_ng(custom_affinity, automatic_affinity)
```
IRDonch, line 314: Could you invert the logic here to check for `.layers` first, and to only use ngraph if it's not available? That would make the code compatible with both the last release of OpenVINO and the most recent development build (which doesn't have `.layers` anymore). Then you could merge this PR while CI still uses the release build of OpenVINO.

tools/accuracy_checker/tests/test_dlsdk_launcher.py
```diff
@@ -169,12 +180,15 @@ def test_dlsdk_onnx_import(self, data_dir, models_dir):
         dlsdk_test_model.predict([{'data': input_blob.astype(np.float32)}], [image.metadata])
         assert dlsdk_test_model.output_blob == 'fc3'
 
-
-
+@pytest.mark.skipif(ng is None and not has_layers(), reason='no functionality to set affinity')
```
IRDonch, line 183: How could this happen? Shouldn't at least one of these always be present?

eaidova: if we use package, then yes both them present, if we build OV from source and follow original build instruction, then we do not have ngraph python, because it has own instruction how it should be installed (people who does not know about it or do not want work install it, can skip this).

IRDonch: OK.

IRDonch (approved):

PR #1672, opened 2020-10-20
Model descrtiption fix

eizamaliev:
Fixes description `model.yml` according to description from `README.md`.


IRDonch:
For public models this is fine, but I think you should leave the Intel models alone. Their configs are regenerated every time we publish a new version of the models, and these changes will be lost the next time that happens.

(I used to painstakingly remove all description changes from the pull requests that update the configs, but with how many models we have now, that approach is just not tenable anymore.)

IRDonch:
Obsoleted by #2290.

PR #1674, opened 2020-10-20
AC: Added custom evaluator for two-stream I3D model

Anna-Mironova:


tools/accuracy_checker/accuracy_checker/annotation_converters/action_recognition.py
```diff
@@ -67,11 +70,14 @@ def configure(self):
         self.subset = self.get_value_from_config('subset')
         self.dataset_meta = self.get_value_from_config('dataset_meta_file')
         self.numpy_input = self.get_value_from_config('numpy_input')
+        self.image_input = self.get_value_from_config('image_input')
```
eaidova, line 73: do you have check that at least one input provided

tools/accuracy_checker/accuracy_checker/annotation_converters/action_recognition.py
```diff
@@ -83,85 +89,126 @@ def convert(self, check_content=False, progress_callback=None, progress_interval
...
 
-    @staticmethod
-    def get_clips(video, clips_per_video, clip_duration, temporal_stride=1, file_ext='jpg'):
-        shift = int(file_ext == 'npy')
-        num_frames = video['n_frames'] - shift
-        clip_duration *= temporal_stride
-
-        if clips_per_video == 0:
-            step = clip_duration
+    def get_ext_and_dir(self):
```
eaidova, line 122: these conditions can not be true in the same time
it means that you can just call return after check in all cases
```suggestion
    def get_ext_and_dir(self):
        if self.numpy_input and not self.image_input:
            return ['npy'], [self.data_dir / self.numpy_subdir if self.numpy_subdir else self.data_dir]
        if self.image_input and not self.numpy_input:
            return ['jpg'], [self.data_dir / self.image_subdir if self.image_subdir else self.data_dir]
        return ['jpg', 'npy'], [self.data_dir / self.image_subdir, self.data_dir / self.numpy_subdir]


            
        
```

eaidova: BTW, do you check that in the last case you have both subdirs?

tools/accuracy_checker/accuracy_checker/evaluators/custom_evaluators/i3d_evaluator.py
```diff
@@ -0,0 +1,411 @@
...
+                raise ConfigError('configuration for flow/rgb does not exist')
+
+        flow_model = I3DFlowModel(
+            network_info.get('flow', {}), launcher, data_source, delayed_model_loading
+        )
+        rgb_model = I3DRGBModel(
+            network_info.get('rgb', {}), launcher, data_source, delayed_model_loading
+        )
+        if rgb_model.output_blob != flow_model.output_blob:
+            warnings.warn("Outputs for rgb and flow models have different names. "
```
eaidova, line 85: it will be nice if you provide output names in error

tools/accuracy_checker/accuracy_checker/evaluators/custom_evaluators/i3d_evaluator.py
```diff
@@ -0,0 +1,411 @@
...
+        identifiers = dataset.annotation_reader.identifiers
+
+        return annotation, identifiers
+
+    @staticmethod
+    def combine_predictions(output_rgb, output_flow):
+        output = {}
+        for key_rgb, key_flow in zip(output_rgb.keys(), output_flow.keys()):
+            if key_rgb != key_flow:
+                warnings.warn("Outputs for rgb and flow models have different names. "
```
eaidova, line 102: you already have the same warning above, maybe only one enough?

tools/accuracy_checker/accuracy_checker/evaluators/custom_evaluators/i3d_evaluator.py
```diff
@@ -0,0 +1,411 @@
...
+                              "Output name of rgb model will be used in combined output")
+            data_rgb = np.asarray(output_rgb[key_rgb])
+            data_flow = np.asarray(output_flow[key_flow])
+            result_data = (data_rgb + data_flow) / 2
```
eaidova, line 106: also I think check that they have the same dims required

tools/accuracy_checker/accuracy_checker/evaluators/custom_evaluators/i3d_evaluator.py
```diff
@@ -0,0 +1,411 @@
...
+
+class I3DRGBModel(BaseModel):
+    def __init__(self, network_info, launcher, data_source, delayed_model_loading=False):
+        self.net_type = 'rgb'
+        super().__init__(network_info, launcher, data_source, delayed_model_loading)
+
+    def prepare_data(self, data):
+        video = Path(data[0].video)
+        frames_identifiers = [video / frame for frame in data[0].frames]
+        prepared_data = self.reader(frames_identifiers)
```
eaidova, line 386: why does this part can not be inside the reader. If I right understand, you just repack clip identifier logic?

tools/accuracy_checker/dataset_definitions.yml
```diff
@@ -762,6 +764,7 @@ datasets:
       clip_duration: 79
       temporal_stride: 1
       num_samples: 400
+      image_input: True
```
eaidova, line 767: possibly it is not so clear, but maybe better create parameter for getting 2 streams instead of changing converter logic for working with images?
Current solution is not backward compatible and may affect users.

tools/accuracy_checker/accuracy_checker/annotation_converters/action_recognition.py
Anna-Mironova: there is no check in current realization. if provided both inputs or no one, two-stream will be processed
there is no check in current implementation. if both inputs are provided or no one the rgb and flow will be processed

eaidova (approved):

PR #1681, opened 2020-10-21
Add the possibility to set multiple inputs for PyTorch models

alexmasterdi:


eaidova:
Jenkins ok to test

tools/downloader/pytorch_to_onnx.py
```diff
@@ -92,14 +102,14 @@ def load_model(model_name, weights, model_path, module_name, model_params):
...
     """Convert PyTorch model to ONNX and check the resulting onnx model"""
 
     output_file.parent.mkdir(parents=True, exist_ok=True)
     model.eval()
-    dummy_input = torch.randn(input_shape)
-    model(dummy_input)
-    torch.onnx.export(model, dummy_input, str(output_file), verbose=False, opset_version=11,
+    dummy_inputs = [torch.randn(input_shape) for input_shape in input_shapes]
+    model(*dummy_inputs)
+    torch.onnx.export(model, tuple(dummy_inputs), str(output_file), verbose=False, opset_version=11,
```
IRDonch, line 112:
```suggestion
    dummy_inputs = tuple(torch.randn(input_shape) for input_shape in input_shapes)
    model(*dummy_inputs)
    torch.onnx.export(model, dummy_inputs, str(output_file), verbose=False, opset_version=11,
```

tools/downloader/pytorch_to_onnx.py
```diff
@@ -8,19 +8,29 @@
 import torch
 import torch.onnx
 
+import numpy as np
+import ast
```
IRDonch, line 12: `ast` should be in the top block of imports, since it is a builtin module.

tools/downloader/pytorch_to_onnx.py
```diff
@@ -8,19 +8,29 @@
...
+
 
 def positive_int_arg(values):
     """Check positive integer type for input argument"""
+    inputs = ast.literal_eval(values)
+    number_of_inputs = np.ndim(inputs)
```
IRDonch, line 18: This is wrong. First of all, `ndim` does not give you the number of inputs, but the nesting level. `ndim([[1], [2], [3]])` is 2, yet that list corresponds to 3 inputs.

But more importantly, `ndim` won't even give you the expected result in some cases. `ndim([[1,2,3], [1,2]])` is 1.

There's no need to involve NumPy at all here. What you need to do is:

* verify that `inputs` is a sequence;
* if all elements of `inputs` are integers, wrap it in a tuple;
* for each element of `inputs`:
   * verify it's a sequence;
   * verify that every subelement is a positive integer.

You don't need a `result` variable, because the function can only return one value, `inputs`.

IRDonch (changes requested):

tools/downloader/pytorch_to_onnx.py
```diff
@@ -1,6 +1,7 @@
 import argparse
 import importlib
 import sys
+import ast
```
IRDonch, line 4: Keep the alphabetical order, please.

tools/downloader/pytorch_to_onnx.py
```diff
@@ -9,19 +10,28 @@
 import torch.onnx
 
 
+def is_sequence(element):
+    return isinstance(element, (list, tuple))
```
IRDonch, line 14: You should use `collections.abc.Sequence` for this.

tools/downloader/pytorch_to_onnx.py
```diff
@@ -9,19 +10,28 @@
...
-        try:
-            ivalue = int(value)
-            if ivalue < 0:
-                raise argparse.ArgumentTypeError('Argument must be a positive integer')
-            result.append(ivalue)
-        except Exception as exc:
-            print(exc)
-            sys.exit('Invalid value for input argument: {!r}, a positive integer is expected'.format(value))
-    return result
+    inputs = ast.literal_eval(values)
```
IRDonch, line 19:
```suggestion
    shapes = ast.literal_eval(values)
```

tools/downloader/pytorch_to_onnx.py
```diff
@@ -9,19 +10,28 @@
...
+    if not is_sequence(inputs):
+        sys.exit('Model inputs must be a sequence')
```
IRDonch, line 21:
```suggestion
        raise argparse.ArgumentTypeError('{!r}: must be a sequence'.format(inputs))
```

You're supposed to raise an `ArgumentTypeError` if the argument is invalid, not exit immediately. (Yes, I know this function used `sys.exit` before, but since you're revamping it, you might as well fix it.)

tools/downloader/pytorch_to_onnx.py
```diff
@@ -9,19 +10,28 @@
 import torch.onnx
 
 
+def is_sequence(element):
+    return isinstance(element, (list, tuple))
+
+
 def positive_int_arg(values):
     """Check positive integer type for input argument"""
```
IRDonch, line 18:
```suggestion
def shapes_arg(values):
    """Checks that the argument represents a tensor shape or a sequence of tensor shapes"""
```

This function name was already wrong before, but now it's just grotesquely wrong. Might as well fix it. :-)

tools/downloader/pytorch_to_onnx.py
```diff
@@ -44,8 +54,8 @@ def parse_args():
                         help='Model to convert. May be class name or name of constructor function')
     parser.add_argument('--weights', type=str,
                         help='Path to the weights in PyTorch\'s format')
-    parser.add_argument('--input-shape', metavar='INPUT_DIM', type=positive_int_arg, required=True,
-                        help='Shape of the input blob')
+    parser.add_argument('--input-shapes', metavar='INPUT_DIM', type=positive_int_arg, required=True,
+                        help='Shapes of the input blobs')
```
eizamaliev, line 58: Please, add example to description.

eizamaliev (changes requested):

tools/downloader/pytorch_to_onnx.py
alexmasterdi: I specially don't use `collections.abc.Sequence`, because string is sequence too. So if you use `--input-shape="[1,1,256,256],[1,2,256,256],[1,1,256,256]"` , shapes will be interpreted as sequence and integer error will be raised, but situation is not related to integers. For this reason I established sequence as (tuple, list).

IRDonch: OK, good point. (To be fair, a string would get rejected by the other checks, but it's good to fail early.)

tools/downloader/pytorch_to_onnx.py
```diff
@@ -9,19 +10,28 @@
...
+        raise argparse.ArgumentTypeError('{!r}: must be a sequence'.format(shapes))
+    if not all(is_sequence(shape) for shape in shapes):
+        shapes = (shapes, )
+    for shape in shapes:
+        if not is_sequence(shape):
+            raise argparse.ArgumentTypeError('{!r}: must be a sequence'.format(shape))
+        for value in shape:
+            try:
+                if type(value) is not int or value < 0:
+                    raise argparse.ArgumentTypeError('Argument must be a positive integer')
```
IRDonch, line 30:
```suggestion
                if not isinstance(value, int) or value < 0:
                    raise argparse.ArgumentTypeError('{!r}: must be a positive integer'.format(value))
```

tools/downloader/pytorch_to_onnx.py
```diff
@@ -9,19 +10,28 @@
...
+            except Exception as exc:
+                print(exc)
+                sys.exit('Invalid value for input argument: {!r}, a positive integer is expected'.format(value))
```
IRDonch, line 33: This `except` block is useless.

IRDonch (changes requested):

IRDonch (approved):

eizamaliev (approved):

PR #1684, opened 2020-10-22
AC: Fix metrics profiling in case of multiple metrics

Anna-Mironova:


tools/accuracy_checker/accuracy_checker/metrics/metric_profiler/base_profiler.py
```diff
@@ -60,7 +61,7 @@ def __init__(self, dump_iterations=100, report_type='csv'):
         self.report_file = '{}.{}'.format(self.__provider__, report_type)
         self.out_dir = Path()
         self.dump_iterations = dump_iterations
-        self.storage = []
+        self.storage = defaultdict(dict)
```
eaidova, line 64: please use ordereddict , it is important to dump results with keeping order

tools/accuracy_checker/accuracy_checker/metrics/metric_profiler/base_profiler.py
```diff
@@ -74,24 +75,21 @@ def update(self, *args, **kwargs):
         profiling_data = self.generate_profiling_data(*args, **kwargs)
         self._last_profile = profiling_data
         if isinstance(profiling_data, list):
```
eaidova, line 77: I believe the list cases, report also contains list of records (e.g. each line represents the bounding box in detection report)

tools/accuracy_checker/accuracy_checker/metrics/metric_profiler/object_detection_metric_profiler.py
```diff
@@ -81,20 +81,19 @@ def generate_json_report(self, identifier, metric_result, metric_name):
 
     def per_box_result(self, identifier, metric_result):
         per_box_results = []
-        for label, per_class_result in enumerate(metric_result):
+        for label, per_class_result in metric_result.items():
```
eaidova, line 84: why this behavior changed?
Not all detection metrics return data in dictionary format. some of them return list. Also not always label here is the same like in annotation, so approach with label_id is better

tools/accuracy_checker/accuracy_checker/metrics/metric_profiler/base_profiler.py
```diff
@@ -119,13 +134,15 @@ def write_json_result(self):
...
                 },
-                'report': self.storage,
+                'report': list(self.storage.values()),
                 'report_type': self.__provider__,
                 'dataset_meta': self.dataset_meta
             }
         with open(str(out_path), 'w') as f:
             json.dump(out_dict, f)
 
+        self.storage = OrderedDict()
```
eaidova, line 144: possibly, it makes sense to create method for reset storage. if in future storage is changed, it will be easier maintain

eaidova (approved):

PR #1685, opened 2020-10-22
demos: Fix an issue breaking build with Clang

IRDonch:
Clang is more strict than GCC (and Visual C++) when it comes to looking up unqualified names in templates:

https://clang.llvm.org/compatibility.html#dep_lookup

What this means for us is that `slog::LogStream::operator<<` fails to find `operator<<(ostream&, ie::Version&)` if the latter isn't defined before the former is defined (i.e. if `slog.hpp` is included before `common.hpp`).

While we could work around this by moving includes around, I think that's too brittle. The correctness of the code shouldn't depend on what order headers are included in. So instead, I implemented a more fundamental solution, to wrap `ie::Version` in a wrapper class and add an `operator<<` for that class. This makes the code quite ugly, but also quite correct.

I also implemented a similar treatment for the other `operator<<`, the one that was defined for `map<string, ie::Version>`. That wasn't strictly necessary, because for some reason even the demos that normally used `slog`, only printed the version maps to `cout`. However, that seems inexplicably inconsistent, so I modified those demos to use `slog` instead.


eizamaliev (approved):
This looks less clear, but let's sacrifice some clarity to CLang :)
*More wrappers for wrapper god*

IRDonch:
> This looks less clear, but let's sacrifice some clarity to CLang :)

To be clear, Clang is just following the standard. It's the other compilers that are too lax. :)

PR #1700, opened 2020-10-27
use ngraph instead of layers in yolo python demo

eaidova:


eizamaliev (approved):

IRDonch:
I just enabled whitespace checks, so I trimmed trailing whitespace in your branch in order to merge it quicker.

PR #1714, opened 2020-10-30
Move converters from other repository

mzuevx:
Add converters for MPII and Cluttered_MNIST datasets


tools/accuracy_checker/accuracy_checker/annotation_converters/cluttered_mnist.py
```diff
@@ -0,0 +1,46 @@
+
```
eaidova, line 1: please add license headers

eaidova:
Jenkins ok to test

eaidova (approved):

PR #1731, opened 2020-11-02
Add command line input option to specifiy model root path

szabi-luxonis:
Added CLI option to specify path to a custom model zoo, for example if the user has a model collection in `~/mymodels/...`
Usage example:
`./downloader.py --model_root ~/mymodels/ ...`

### DIFF
--- a/tools/downloader/common.py
+++ b/tools/downloader/common.py
@@ -551,6 +551,9 @@ def load_models(args):
     models = []
     model_names = set()
 
+    if args.model_root is not None:
+        MODEL_ROOT = Path(args.model_root)
+
     for config_path in sorted(MODEL_ROOT.glob('**/model.yml')):
         subdirectory = config_path.parent.relative_to(MODEL_ROOT)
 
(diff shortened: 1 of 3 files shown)


IRDonch:
What scenario is this meant to support? The downloader isn't really supposed to be used with a "custom model zoo", it is bound to OMZ.

szabi-luxonis:
> What scenario is this meant to support? The downloader isn't really supposed to be used with a "custom model zoo", it is bound to OMZ.

We are using the modified model compiler/converter [here](https://github.com/luxonis/depthai/tree/develop) for our custom model zoo (models that we support) where we have models from OMZ and also our custom models : tiny-yolo-v3, yolo-v3 etc.

From user point of view with these changes they can have their own model zoo, with custom models and use the downloader/converter/myriad compiler to do the conversion automatically, the same way openvino does.

IRDonch:
Sorry, but this isn't really a supported scenario. The model config format is not considered stable, and it's subject to change in the future. Therefore we can't promise that the downloader will remain compatible with a custom zoo.

PR #1733, opened 2020-11-03
Add CMake option to specify demos to build

ivanvikhrev:
Add  `-DDEMOS=<list of demos"` option to CMake file and build scripts. It allows to specify one or more demos to build.

### DIFF
--- a/demos/CMakeLists.txt
+++ b/demos/CMakeLists.txt
@@ -121,7 +121,7 @@ function(add_samples_to_build)
                 if (BUILD_SAMPLE_NAME)
                     list(FIND BUILD_SAMPLE_NAME ${dir} index)
                 endif()
-                if ((index EQUAL -1 OR NOT dir IN_LIST DEMOS_LIST) AND NOT DEMOS STREQUAL "ALL")
+                if (index EQUAL -1 OR (NOT dir IN_LIST DEMOS_LIST AND NOT DEMOS STREQUAL "ALL"))
                     message(STATUS "${dir} SKIPPED")
                 else()
                     # Include subdirectory to the project.
--- a/demos/CMakeLists.txt
+++ b/demos/CMakeLists.txt
@@ -8,7 +8,7 @@ project(Demos)
 
 option(ENABLE_PYTHON "Whether to build extension modules for Python demos" OFF)
 
-set(DEMOS "ALL" CACHE STRING "Specify demo to build using comma(ex. -DSPECIFIC_DEMOS classification_demo, text_detection_demo), Default is ALL")
+set(DEMOS "ALL" CACHE STRING "Specify demo to build using comma(ex. -DDEMOS=classification_demo,text_detection_demo). Default is ALL")
 string(REPLACE "," ";" DEMOS_LIST ${DEMOS})
 
 if (CMAKE_BUILD_TYPE STREQUAL "")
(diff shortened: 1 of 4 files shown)


demos/build_demos_msvc.bat
```diff
@@ -34,6 +34,14 @@ if not "%1" == "" (
...
         goto argParse
     )
+    rem to build more than one specific demo use quotation marks,
+    rem list the necessary demos comma-separated and without spaces,
+    rem ex. -DDEMOS="classification_demo,segmentation_demo"
+    if "%1" == "-DDEMOS" (
+        set EXTRA_CMAKE_OPTS=%EXTRA_CMAKE_OPTS% %1=%2
+        shift & shift
+        goto argParse
+    )
```
Wovchena, line 44: CMake creates a target per each demo. You should run cmake like `cmake --build . --config Release --target object_detection_demo_ssd_async segmentation_demo` instead of avoiding adding subdirectory.

Wovchena (changes requested):

demos/CMakeLists.txt
```diff
@@ -8,6 +8,9 @@ project(Demos)
 
 option(ENABLE_PYTHON "Whether to build extension modules for Python demos" OFF)
 
+set(DEMOS "ALL" CACHE STRING "Specify demo to build using comma(ex. -DDEMOS=classification_demo,text_detection_demo). Default is ALL")
```
IRDonch, line 11: I don't understand, what's the purpose of this option?

ivanvikhrev:
Cmake with version lower than 3.15 doesn't support multiple targets for build, it accepts only single target . Should we increase minimum required version from 3.9 to 3.15 for this feature?

demos/build_demos_msvc.bat
ivanvikhrev: Fixed

demos/CMakeLists.txt
ivanvikhrev: Reworked as @Wovchena suggested.

demos/build_demos.sh
```diff
@@ -26,12 +26,18 @@ error() {
...
 
 for opt in "$@"; do
     case "$opt" in
     -DENABLE_PYTHON=*)
         extra_cmake_opts+=("$opt")
         ;;
+    --target=*)
+        tmp="${opt%\"}"
+        tmp="${tmp#\"}"
+        extra_cmake_build_opts+=("${tmp//=/ }")
```
Wovchena, line 39: I suggest to rename `extra_cmake_build_opts`  as `targets`

demos/build_demos.sh
```diff
@@ -82,6 +88,7 @@ fi
 mkdir -p "$build_dir"
 
 (cd "$build_dir" && cmake -DCMAKE_BUILD_TYPE=Release "${extra_cmake_opts[@]}" "$DEMOS_PATH")
-cmake --build "$build_dir" -- "$NUM_THREADS"
+c="cmake --build $build_dir ${extra_cmake_build_opts[@]} -- $NUM_THREADS"
+eval "$c"
```
Wovchena, line 92: Why not
```suggestion
cmake --build $build_dir ${extra_cmake_build_opts[@]} -- $NUM_THREADS
```
?

ivanvikhrev: Well, I tried next variant:  
`cmake --build "$build_dir" "${extra_cmake_build_opts[@]}" -- "$NUM_THREADS`"
and it doesn't work, probably because "${extra_cmake_build_opts[@]}" contains spaces inside and bash can't resolve it without eval command.
But without quotes all is ok, I'll change it.

Wovchena: May be the quotes were useful. I didn't notice you removed them :)

ivikhrev:
Continue work in PR #

PR #1739, opened 2020-11-03
Serialize http and google_drive file sources

SDxKeeper:


IRDonch:
Is this another variation on #1618? I already explained that I don't want to add code that is only used by internal tools.

IRDonch:
Closing due to inactivity.

PR #1746, opened 2020-11-05
AC: Add detection labels shift postprocessing for Efficientdet

jkamelin:
converted Efficientdet model returns class labels shifted by 1 (starting from 0, where 0 is person label)

### DIFF
--- a/tools/accuracy_checker/accuracy_checker/postprocessor/shift.py
+++ b/tools/accuracy_checker/accuracy_checker/postprocessor/shift.py
@@ -16,7 +16,7 @@ limitations under the License.
 
 import numpy as np
 from ..config import NumberField
-from .postprocessor import PostprocessorWithSpecificTargets
+from .postprocessor import PostprocessorWithSpecificTargets, Postprocessor
 from ..representation import SegmentationAnnotation, SegmentationPrediction, DetectionPrediction, DetectionAnnotation
 
 
@@ -57,9 +57,9 @@ class Shift(PostprocessorWithSpecificTargets):
         return annotation, prediction
 
 
-class ShiftLabels(PostprocessorWithSpecificTargets):
+class ShiftLabels(Postprocessor):
     """
-    Shift predicted detection labels or annotation detection labels.
+    Shift predicted detection labels.
     """
 
     __provider__ = 'shift_labels'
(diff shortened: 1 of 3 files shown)


tools/accuracy_checker/accuracy_checker/postprocessor/shift.py
```diff
@@ -55,3 +55,34 @@ def process_image(self, annotation, prediction):
...
+        self.offset = self.get_value_from_config('offset')
+
+    def process_image(self, annotation, prediction):
+        for prediction_ in prediction:
+            prediction_.labels += self.offset
+
+        for annotation_ in annotation:
+            annotation_.labels += self.offset
+
+        return annotation, prediction
```
eaidova, line 88: @jkamelin could you please describe why these changes need?
Is it impossible to align labels during annotation conversion?

eaidova:
@jkamelin please add your postrocessor to readme

tools/accuracy_checker/accuracy_checker/postprocessor/shift.py
jkamelin: These changes were made for uniformity, for this specific model, shift is only needed for predictions, in this case code for annotations can be removed. Do you suggest to perform align not in postprocessor, but during annotation conversion?

eaidova: no, i think for now it is ok. If we change logic in converter, it will be not backward compatible

tools/accuracy_checker/accuracy_checker/postprocessor/shift.py
```diff
@@ -55,3 +55,31 @@ def process_image(self, annotation, prediction):
...
+    __provider__ = 'shift_labels'
+
+    prediction_types = (DetectionPrediction, )
+    annotation_types = (DetectionAnnotation, )
+
+    @classmethod
+    def parameters(cls):
+        parameters = super().parameters()
+        parameters.update({
+            'offset': NumberField(value_type=int, optional=True, default=0, description="Value for shift.")
```
eaidova, line 74: I think we do not need default value here. with 0 it has not effect.

eaidova (approved):

PR #1757, opened 2020-11-06
demos: simplify the specification of output directories

IRDonch:
We can mostly use the same logic for Windows and non-Windows, thanks to generator expressions.

We can also remove some variables completely:

* `CMAKE_LIBRARY_PATH` is a completely irrelevant varialble.
* `LIBRARY_OUTPUT_PATH` is obsolete.
* `LIBRARY_OUTPUT_DIRECTORY` isn't even a builtin CMake variable.

Also, remove the obsolete `IE_MAIN_SAMPLES_DIR` variable.


Wovchena (approved):
BTW, do you have any idea why we append `lib` to output dirs?

IRDonch:
> BTW, do you have any idea why we append `lib` to output dirs?

Nope. Although IMO it's useful, since it helps keep the main output directory clean.

PR #1760, opened 2020-11-09
added labels info

Ilya-Krylov:


snosov1:
Good God! Some models have target ID as 0 and some have as 1 for no comprehensible reason. I guess, that'll do it for now, but, hopefully, this'll somehow streamline going forward.

BTW, is there a way to make such meta info a part of IR?

snosov1 (approved):

Ilya-Krylov:
@snosov1 
1. mmdetection1.x models are exported (and trained) so that background class is 0, so target class is 1
2. mmdetection2.x models
   * SSD with detection_output are trained (and exported) so that background class is *n_labels*, so target class is 0
   * FCOS, ATSS and maybe other exported with *boxes* and *labels* tensors ... I don't know why target class is 1. mmdetection posted that they changed background class for models to be *n_labels*, maybe this is a matter of onnx export.

@druzhkov-paul FYI

vladimir-dudnik (approved):

PR #1766, opened 2020-11-10
speech_recognition_demo: build the extension module using CMake

IRDonch:
It is very rude to globally install a Python package from a CMake build. Beside that, it's also slow, because it's rebuild on every CMake run and the build process can't be parallelized. So instead of that, just build it as a normal target the same way we already do it for `pose_extractor` (with the added complication of needing to copy the Python files).

Delete `ThreadPool/example.cpp` to avoid it getting picked up by globs (and also because we simply don't need it).

I fixed all GCC warnings in non-generated code. There are also some warnings from MSVC, but I think that they can wait, since we don't currently build with `/WX`.


IRDonch:
CC @vladimir-dudnik @AlexeyKruglov

Wovchena (approved):

PR #1769, opened 2020-11-10
mask_rcnn_demo: skip zero-size boxes

IRDonch:
These aren't normally supposed to occur, but there's a currently a bug in OpenVINO master where the GPU plugin produces invalid boxes when inferring with `mask_rcnn_resnet101_atrous_coco`. This change works around that (and, in general, makes the demo more resilient to misbehaving networks).


Wovchena:
`box_width` and `box_height` are already clipped. Your check just verifies they aren't 0. I suggest to write exactly that: `box_width != 0` or you should have filtered such boxes before clipping.

IRDonch:
I think it would make the most sense to just remove the clipping. Clipping from below is redundant with the checks I added, while clipping from above is redundant, because the corner coordinates are clipped already, so the box can't be bigger than the image.

Wovchena:
Yes, this will go.

Wovchena (approved):

Wovchena (changes requested):
Wait, we don't need clipping at all

IRDonch:
Oops, my bad. I wrote a correct commit message, but I forgot to actually do what was in it. :-)

Wovchena (approved):

PR #1770, opened 2020-11-10
demos: reorder ImagesCapture

Wovchena:


Wovchena:
VideoCapture prints warnings when it can't open a folder. I think it's worse to have these warnings or have a code suppressing them than handle minor corner cases.

PR #1772, opened 2020-11-11
speech_recognition_demo: fix various warnings in ctcdecode-numpy

IRDonch:
* uninitialized class members;
* useless calls to "empty";
* some narrowing conversions.


Wovchena (approved):
I'm not getting what `trie_chars_.empty();` or `clear()` are for as `add_words()` is only called after `dictionary` reset. It's hard to deduce the intention. So let it be.

IRDonch:
> I'm not getting what `trie_chars_.empty();` or `clear()` are for as `add_words()` is only called after `dictionary` reset. It's hard to deduce the intention. So let it be.

The function description says "Fill (**replace**) prefix set", so I think it's reasonable to assume that the function is supposed to fully rewrite the state of the object. In fact, I went ahead and made it clear `trie_weights_` too.

Wovchena:
Yep, it was confusing to see `add` in function name. I would expect `set` or something like this.

PR #1542, opened 2020-09-11
Add CoCosNet model

alexmasterdi:
TODO list:
- [x] Add accuracy check file
- [x] Create model.yml to load model and load reference dict
- [x] Create description for model
- [x] Add demo

PR #1774, opened 2020-11-11
Add monolithic CoCosNet model

akorobeinikov:


eaidova:
@alexmasterdi does it mean that previous request #1542  should be closed?

models/public/cocosnet/model.yml
```diff
@@ -0,0 +1,170 @@
...
+    file: models/networks/correspondence.py
+    pattern: '/ self.opt.down\)'
+    replacement: '// self.opt.down)'
+  - $type: regex_replace
+    file: options/base_options.py
+    pattern: 'import data'
+    replacement: ''
+  - $type: regex_replace
+    file: options/base_options.py
+    pattern: '# modify dataset-related parser options'
```
eaidova, line 129: what the reason to change comments?

eaidova:
Jenkins ok to test

models/public/cocosnet/cocosnet.md
```diff
@@ -0,0 +1,93 @@
...
+### Output
+
+Image, name: "exemplar_based_output", shape: [1x3x256x256] - A result (generated) image based on exemplar in the format [BxCxHxW],
+   where:
+    - B - batch size
+    - C - number of channels
+    - H - image height
+    - W - image width
+
+    Output color order is RGB.
```
eaidova, line 64: what happen? input is BGR, output is RGB

demos/python_demos/cocosnet_demo/cocosnet_demo/postprocessing.py
```diff
@@ -0,0 +1,34 @@
...
+import numpy as np
+import cv2
+import os
+
+
+def postprocess(out):
+    out = np.squeeze(out)
+    out = np.transpose(out, (1, 2, 0))
+    result = np.uint8(127.5 * out + 127.5)
+    result = cv2.cvtColor(result, cv2.COLOR_BGR2RGB)
```
eaidova, line 23: why do you convert output to RGB?
possibly do you mean
```suggestion
    result = cv2.cvtColor(result, cv2.COLOR_RGB2BGR)
```
I know that it is the same op, but it is clearer for people who does not know all details

tools/accuracy_checker/accuracy_checker/annotation_converters/ade20k_image_translation.py
```diff
@@ -0,0 +1,67 @@
...
+    def convert(self, check_content=False, progress_callback=None, progress_interval=100, **kwargs):
+        content_check_errors = [] if check_content else None
+        annotations = []
+        ref_dict = self.get_ref()
+        for key, value in ref_dict.items():
+            input_mask_filename = "annotations/validation/" + key.replace('.jpg', '.png')
+            reference_image_filename = "images/training/" + value
+            reference_mask_filename = "annotations/training/" + value.replace('.jpg', '.png')
+            identifier = [str(input_mask_filename), str(reference_image_filename), str(reference_mask_filename)]
+            annotation = ImageProcessingAnnotation(identifier, self.image_dir / "validation" / key)
```
eaidova, line 55: please do not use Path objects inside annotation. They are not os agnostic after serialization
```suggestion
            annotation = ImageProcessingAnnotation(identifier,  "validation/{}".format(key))
```
prefix path can be set via config using additional_data_source key if it is different from `data_source`

tools/accuracy_checker/accuracy_checker/annotation_converters/ade20k_image_translation.py
```diff
@@ -0,0 +1,67 @@
...
+                description="Path to directory with masks."
+            ),
+            'reference_file': PathField(
+                optional=False, description="Path to .txt file with pairs (validation:train)."
+            )
+        })
+        return parameters
+
+    def configure(self):
+        self.image_dir = self.get_value_from_config('images_dir')
```
eaidova, line 42: after my suggestion, this parameter will not used

tools/accuracy_checker/accuracy_checker/annotation_converters/ade20k_image_translation.py
```diff
@@ -0,0 +1,67 @@
...
+        self.annotation_dir = self.get_value_from_config('annotations_dir')
+        self.reference_dict = self.get_value_from_config('reference_file')
+
+    def convert(self, check_content=False, progress_callback=None, progress_interval=100, **kwargs):
+        content_check_errors = [] if check_content else None
```
eaidova, line 47: I do not see any code related with content existence check inside converter, do you plan to handle it?
If not, then I think you can directly set None in ConverterReturn tuple

tools/accuracy_checker/accuracy_checker/evaluators/custom_evaluators/cocosnet_evaluator.py
```diff
@@ -0,0 +1,477 @@
...
+    def set_input_and_output(self):
+        has_info = hasattr(self.exec_network, 'input_info')
+        input_info = self.exec_network.input_info if has_info else self.exec_network.inputs
+        self.input_blob = next(iter(input_info))
+        self.output_blob = list(self.exec_network.outputs.keys())
+        self.number_of_metrics = len(self.output_blob)
+
+    def fit_to_input(self, input_data):
+        input_data = cv2.cvtColor(input_data, cv2.COLOR_RGB2BGR)
+        input_data = cv2.resize(input_data, dsize=(299, 299))
```
eaidova, line 467: input data size is model specific parameter, so I think it will be better if you will use height and width from network input instead of hard coded value (majority imagenet trained models has input 224x224)

tools/accuracy_checker/dataset_definitions.yml
```diff
@@ -1064,3 +1064,22 @@ datasets:
...
+      converter: ade20k_image_translation
+      images_dir: ADEChallengeData2016/images
+      annotations_dir: ADEChallengeData2016/annotations
+      reference_file: cocosnet_ade20k_ref_test.txt
+    data_source: ADEChallengeData2016
+    metrics:
+      - type: psnr
+        scale_border: 0
+        presenter: print_vector
+      - type: ssim
```
eaidova, line 1079: psnr and ssim do not use classification model. Is there way to use custom evaluator without it, if user wants only calculate them?

demos/python_demos/cocosnet_demo/cocosnet_demo/preprocessing.py
```diff
@@ -0,0 +1,44 @@
...
+ See the License for the specific language governing permissions and
+ limitations under the License.
+"""
+
+import numpy as np
+import cv2
+
+
+def scatter(source):
+    out = np.zeros((1, 151, 256, 256))
```
IRDonch, line 19: All of these hardcoded numbers should be replaced with dimensions queried from the model passed by the user.

models/public/cocosnet/cocosnet.md
```diff
@@ -0,0 +1,93 @@
...
+## Use Case and High-Level Description
+
+Cross-domain correspondence network is a exemplar-based image translation model, consisting of correspondence and translation parts. Model was pre-trained on ADE20k dataset.
+For details see [paper](https://arxiv.org/pdf/2004.05571) and [repository](https://github.com/microsoft/CoCosNet).
+
+## Specification
+
+| Metric                          | Value                                     |
+|---------------------------------|-------------------------------------------|
+| Type                            | Image_translation                         |
```
IRDonch, line 12:
```suggestion
| Type                            | Image translation                         |
```

models/public/cocosnet/cocosnet.md
```diff
@@ -0,0 +1,93 @@
...
+| GFLOPs                          | 1080.7032                                 |
+| MParams                         | 167.9141                                  |
+| Source framework                | PyTorch\*                                 |
+
+## Accuracy
+
+Metrics was calculated between generated images by model and real validation images from ADE20k dataset.
```
IRDonch, line 19:
```suggestion
Metrics were calculated between generated images by model and real validation images from ADE20k dataset.
```

models/public/cocosnet/cocosnet.md
```diff
@@ -0,0 +1,93 @@
...
+    - C - number of channels
+    - H - image height
+    - W - image width
+
+    Output color order is RGB.
+
+## Legal Information
+
+The original model is distributed under the following
+[license](https://github.com/microsoft/CoCosNet/blob/master/LICENSE):
```
IRDonch, line 69: You should also list the license of the Synchronized-BatchNorm-PyTorch project that's used as the dependency.

models/public/cocosnet/model.yml
```diff
@@ -0,0 +1,170 @@
...
+# limitations under the License.
+
+description: >-
+  Cross-domain correspondence network is a exemplar-based image translation model,
+  consisting of correspondence and translation parts. Model was pre-trained on ADE20k
+  dataset. For details see paper <https://arxiv.org/pdf/2004.05571> and repository
+  <https://github.com/microsoft/CoCosNet>.
+task_type: image_translation
+files:
+  - name: util/__init__.py
```
IRDonch, line 22: Please place all source files in a subdirectory to avoid clutter in the download directory.

models/public/cocosnet/model.yml
```diff
@@ -0,0 +1,170 @@
...
+    size: 387069636
+    sha256: 9a77b52d0c34a279d6b4e7eff4bd00077543f7ffc2806b5b2a32222cfbccc655
+    source:
+      $type: google_drive
+      id: 1zVG4SnR1uI8bC8fagZrgxNYCI5GRmJuR
+postprocessing:
+  - $type: regex_replace
+    file: models/networks/normalization.py
+    pattern: 'x.var\(dim=1, keepdim=True\).add\(epsilon\).sqrt\(\)'
+    replacement: 'x.std(dim=1, keepdim=True)'
```
IRDonch, line 118: That seems odd, what about the epsilon?

models/public/cocosnet/model.yml
```diff
@@ -0,0 +1,170 @@
...
+    pattern: 'parser = dataset_option_setter\(parser, self.isTrain\)'
+    replacement: "'''"
+  - $type: regex_replace
+    file: options/base_options.py
+    pattern: '# set gpu ids'
+    replacement: 'opt.gpu_ids = "-1"'
+  - $type: regex_replace
+    file: options/base_options.py
+    pattern: 'opt = parser.parse_args'
+    replacement: 'opt, _ = parser.parse_known_args'
```
IRDonch, line 142: All this option-related hackery seems pretty hairy. Maybe it would be easier, instead of using `TestOptions`, to create a `types.SimpleNamespace` with all the options and pass that to the model factory instead?

models/public/cocosnet/model.yml
```diff
@@ -0,0 +1,170 @@
...
+  - $type: regex_replace
+    file: util/util.py
+    pattern: '# returns a configuration for creating a generator'
+    replacement: "'''"
+  - $type: regex_replace
+    file: util/util.py
+    pattern: 'return mask'
+    replacement: "'''"
```
IRDonch, line 150: This seems excessive; it should be sufficient (and more self-explanatory) to just comment out the line with `loadmat`.

IRDonch (changes requested):

models/public/cocosnet/cocosnet.md
alexmasterdi: Because using flag `reverse_input_channels`.

models/public/cocosnet/model.yml
```diff
@@ -0,0 +1,130 @@
...
+    pattern: 'x.var\(dim=1, keepdim=True\).add\(epsilon\).sqrt\(\)'
+    replacement: 'x.std(dim=1, keepdim=True)'
+  - $type: regex_replace
+    file: model_files/models/networks/__init__.py
+    pattern: 'from models.networks.ContextualLoss import \*'
+    replacement: '# from models.networks.ContextualLoss import *'
+  - $type: regex_replace
+    file: model_files/models/networks/__init__.py
+    pattern: 'from models.networks.loss import \*'
+    replacement: '# from models.networks.loss import *'
```
IRDonch, line 102:
```suggestion
  - $type: regex_replace
    file: model_files/models/networks/__init__.py
    pattern: 'from models\.networks\.(ContextualLoss|loss) '
    replacement: '# \g<0>'
```

models/public/cocosnet/model.py
```diff
@@ -0,0 +1,87 @@
...
+            pass
+        for child in item.children():
+            remove_all_spectral_norm(child)
+    if isinstance(item, nn.ModuleList):
+        for module in item:
+            remove_all_spectral_norm(module)
+    if isinstance(item, nn.Sequential):
+        modules = item.children()
+        for module in modules:
+            remove_all_spectral_norm(module)
```
IRDonch, line 35: `Sequential` is a subtype of `Module`, so this case is redundant.

I suspect the `ModuleList` case might be redundant too.

tools/accuracy_checker/dataset_definitions.yml
alexmasterdi: I moved metrics to accuracy config file. Now users are able to calculate psnr and ssim metrics without calculating IS and FID metrics.

models/public/cocosnet/model.yml
alexmasterdi: I can add epsilon, but accuracy did not change when replacing `var` (using epsilon) with `std`. Epsilon is too small.

demos/python_demos/image_translation_demo/image_translation_demo/models.py
```diff
@@ -0,0 +1,66 @@
...
+        self.input_size = self.network.input_info[self.input_name].input_data.shape
+        self.output_size = self.exec_net.outputs[self.output_name].shape
+
+    def infer(self, input):
+        input_data = {self.input_name: input}
+        result = self.exec_net.infer(input_data)
+        return result[self.output_name]
+
+
+class SegmentationModel(SingleInputModel):
```
eaidova, line 63: here only one instance of SingleInputModel
probably it makes sence to merge SegmentationModel with SingleInputModel

tools/accuracy_checker/accuracy_checker/evaluators/custom_evaluators/cocosnet_evaluator.py
```diff
@@ -0,0 +1,483 @@
...
+                verification_network['model'] = model_args[1] if len(model_args) > 1 else None
+                verification_network['_model_is_blob'] = models_is_blob
+            network_info.update({
+                'cocosnet_network': cocosnet_network,
+                'verification_network': verification_network
+            })
+            if not contains_all(network_info, ['cocosnet_network', 'verification_network']):
+                raise ConfigError('configuration for cocosnet_network/verification_network does not exist')
+        gan_model = CocosnetModel(network_info.get('cocosnet_network', {}), launcher, delayed_model_loading)
+        if network_info.get('verification_network', {})['model']:
```
eaidova, line 87: you got key error in case of empty verificatuon_network dict

tools/accuracy_checker/accuracy_checker/evaluators/custom_evaluators/cocosnet_evaluator.py
```diff
@@ -0,0 +1,483 @@
...
+                cocosnet_network['model'] = model_args[0]
+                cocosnet_network['_model_is_blob'] = models_is_blob
+            if 'model' not in verification_network and model_args:
+                verification_network['model'] = model_args[1] if len(model_args) > 1 else None
+                verification_network['_model_is_blob'] = models_is_blob
+            network_info.update({
+                'cocosnet_network': cocosnet_network,
+                'verification_network': verification_network
+            })
+            if not contains_all(network_info, ['cocosnet_network', 'verification_network']):
```
eaidova, line 84: this condition contradicts with the next (you raise error if verification_network is not provided, while below you accept emptiness of this field)

tools/accuracy_checker/accuracy_checker/evaluators/custom_evaluators/cocosnet_evaluator.py
```diff
@@ -0,0 +1,483 @@
...
+            self._part_by_name[network_dict['name']].load_network(network_dict, launcher)
+
+    def load_network(self, network_list, launcher):
+        for network_dict in network_list:
+            self._part_by_name[network_dict['name']].load_network(network_dict['model'], launcher)
+
+    def get_network(self):
+        return [
+            {'name': 'gan_network', 'model': self.test_model.network},
+            {'name': 'verification_network', 'model': self.check_model.network}
```
eaidova, line 264: what if check_model is None?

tools/accuracy_checker/accuracy_checker/evaluators/custom_evaluators/cocosnet_evaluator.py
```diff
@@ -0,0 +1,483 @@
...
+            if output_callback:
+                output_callback(
+                    predictions,
+                    metrics_result=metrics_result,
+                    element_identifiers=batch_identifiers,
+                    dataset_indices=batch_input_ids
+                )
+
+            if self.metric_executor.need_store_predictions:
+                self._annotations.extend(gan_annotations)
```
eaidova, line 175: probably need to save separated list of metrics and annotations/predictions with usage verification network and without it (in both cases you can have metric s=which required save predictions)

tools/accuracy_checker/accuracy_checker/evaluators/custom_evaluators/cocosnet_evaluator.py
alexmasterdi: Now this function considers only existing models

eaidova (approved):

models/public/cocosnet/model.yml
IRDonch: I think it's better to not make gratuitous changes to the model. Just replace `x.var(...)` with `(\g<0> ** 2)` and add a comment to explain that it's to support PyTorch < 1.7 (to remind us to remove it later).

demos/python_demos/image_translation_demo/README.md
```diff
@@ -0,0 +1,110 @@
+# Image Translation Demo
+
+This demo application demonstrates an example of using neural networks to synthesize a photo-realistic image based on exemplar image. You can use the following models with the demo:
+
+* `cocosnet`
```
IRDonch, line 5: Suitable segmentation models should also be listed here.

IRDonch (changes requested):

demos/python_demos/image_translation_demo/image_translation_demo.py
```diff
@@ -0,0 +1,135 @@
...
+                           "acceptable. Default value is CPU",
+                      default="CPU", type=str)
+    return parser
+
+
+def get_files(path):
+    if path is None:
+        return []
+    if os.path.isdir(path):
+        file_paths = [os.path.join(path, file) for file in os.listdir(path) if os.path.isfile(path + file)]
```
IRDonch, line 64:
```suggestion
        file_paths = [os.path.join(path, file) for file in os.listdir(path) if os.path.isfile(os.path.join(path, file))]
```

vladimir-dudnik:
Jenkins please retry a build

akorobeinikov:
Jenkins please retry a build

IRDonch (approved):

IRDonch:
Oh, one other thing - please add a link to the demo to `demos/README.md`.

Wovchena (changes requested):

demos/python_demos/image_translation_demo/image_translation_demo.py
```diff
@@ -0,0 +1,135 @@
...
+    seg_model = SegmentationModel(ie_core, args.segmentation_model,
+                                  args.segmentation_model.replace(".xml", ".bin"),
+                                  args.device) if args.segmentation_model else None
+
+    log.info("Preparing input data")
+    input_data = []
+    assert args.reference_images and ((args.input_semantics and args.reference_semantics)
+                                      or (args.input_images and args.segmentation_model)
+                                      ), "Not enough data to do inference"
+    use_seg = not (args.reference_images and args.input_semantics and args.reference_semantics)
```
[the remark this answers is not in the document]
akorobeinikov: fixed, but using | instead ^ in assert. If you will use ^, you get error "Don't know where to get data" when all parameters are set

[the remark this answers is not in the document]
Wovchena: That's the point. If everything is set you will just abandon two of args. Instead of error you could print a warning, but demo shouldn't silently make a guess.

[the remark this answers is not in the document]
akorobeinikov: Ok, done

Wovchena (approved):

PR #1784, opened 2020-11-13
Convert ctdet_coco_dlav0_* from the original files

IRDonch:
Turns out we don't actually need any of the CenterNet requirements to convert these models, and in fact, only need 3 files from the source repository.

I have not retained the change to `dlav0.py` that was included in `pytorch-onnx.patch`, because I believe we no longer need it. The only effect of that change was to ensure deterministic order of model outputs, but we now depend on Python 3.6+, in which dictionaries are deterministic already.

Since I was modifying these models anyway, I took the opportunity to give their outputs human-friendly names.

Other than the change in names, the IRs generated by the new configs are identical.

One big downside of this change is that the PyTorch weights file is strangely three times as big as the ONNX file. I guess you win some, you lose some.


IRDonch:
CC @vladimir-dudnik @eaidova 

I have made this a draft for now, because I'm not sure what to do with `ctdet_coco_dlav0_512`. It would have the exact same conversion recipe, except for the input shape options. I don't think it makes sense to keep two models with nearly identical recipes, so maybe we should remove one?

vladimir-dudnik:
@IRDonch there were requests from platforms teams for both these models

vladimir-dudnik (approved):

vladimir-dudnik:
@IRDonch we may merge this I think, if you do not have anything in plans to further changes

jkamelin (approved):

IRDonch:
@vladimir-dudnik I don't plan to make _changes_, but I would at least need to copy this patch to the other CenterNet model. It would be silly to only implement this to one of them.

vladimir-dudnik:
Jenkins please retry a build

IRDonch:
Updated the other CenterNet model too. I think we should hold this change off until the next release, though; there's no urgency.

PR #1794, opened 2020-11-16
googlenet-v4-tf: convert from original files by using a pre-convert script

IRDonch:
The TensorFlow models in the `research/slim` directory have been updated to no longer use `tf.contrib.slim`, which has been removed in TensorFlow 2.0. Instead, they use TF-Slim, which is a separate library (presumably based on the original `tf.contrib.slim` codebase). What this means for us is that it's now possible to load these models in either TF 1.x or 2.x, and so we can use a pre-convert script to enable the conversion of such models from their original source files, without hosting intermediate conversion artifacts.

We don't even need to require TF-Slim as a dependency. It is a pure Python library, so we can just download it the same way we download source files.

It would be simpler (and likely faster) to not freeze the graph in the pre-convert script, and instead just save the unfrozen graph and pass it along with the checkpoint file to MO. The only reason I kept the freezing is to enable accuracy checks for the original model, since AFAICS, Accuracy Checker doesn't support separate checkpoint files.

The IRs produced by the new conversion recipe are identical to the ones produced before, except for layer names.


IRDonch:
CC @vladimir-dudnik

jkamelin (approved):

jkamelin:
We, probably, can use the same approach for other slim models

IRDonch:
> We, probably, can use the same approach for other slim models (googlenet-v1-tf, v2-tf)

Yeah, absolutely. I even edited the pre-convert script so that we would only need to change one line for each model. :-)

PR #1797, opened 2020-11-16
googlenet-v1-tf: convert from original files by using a pre-convert script

IRDonch:
Pretty much the same thing as ba2ddb13, but for `googlenet-v1-tf`.

This doesn't change the XML file in the resulting IR (apart from layer names), but the BIN file does change, and I don't know why. However, I retested the accuracy of the IR (on CPU with FP32 precision), and it exactly matches what's listed in the documentation, so I assume the differences aren't significant.


jkamelin (approved):

IRDonch:
I think I figured it out.

For clarity, I will use `old.bin` to mean the `FP32/googlenet-v1-tf.bin` file that you will get if you use the model converter on the current OMZ `develop` branch, and `new.bin` to mean that same file that you will get with this pull request's branch.

Above I wrote that following the documented conversion recipe and then running MO gets you `old.bin`. That is actually 
wrong, because I made a mistake in my experiment. You actually get `new.bin`. So my conversion script does _not_ deviate from the original recipe.

So if following the documented recipe results in `new.bin`, why do you get `old.bin` with the `develop` branch? Well, it turns out that that frozen graph file that we host on `download.01.org` was _not_ created with the documented recipe. I wasn't able to reproduce that file exactly, but I got very close by using TensorFlow 1.3 and tensorflow/models@f87a58cd. If you use those versions to freeze the graph, _then_ you get `old.bin`. If you use TensorFlow 1.4 (or newer) with the same commit of the models, then you get `new.bin`.

What's the difference, then? I did a cursory comparison of the TF 1.3 and 1.4 versions of the frozen graph, and the main difference seems to be that the latter uses `FusedBatchNorm` operations, while the former does not. (`FusedBatchNorm` was enabled in tensorflow/tensorflow@349188152.) My guess is that TensorFlow preprocesses the weights in some way when fusing batch normalization operations, and that causes the eventual difference in the `.bin` files; although I don't know TensorFlow in enough depth to say for certain that that is the cause.

Bottom line is that the discrepancy seems to be the result of an optimization made in TensorFlow after our frozen graph file was created. Since it doesn't affect the accuracy metrics, it should be harmless.

PR #1798, opened 2020-11-17
enable model/yolov4 sample

942860759:
enable model/yolov4 sample


IRDonch:
Looks like a duplicate of #1799.

PR #1804, opened 2020-11-17
googlenet-v2-tf: convert from original files by using a pre-convert script

IRDonch:
Nearly identical to 6f24c05a, except with v2 instead of v1.

Like with v1, this change doesn't affect the output XML files (except for layer names), but does affect output BIN files. However, the accuracy metrics (on CPU with FP32 precision) still exactly match the documentation.

Since 6f24c05a, I have investigated this further, and determined that the change in BIN files happens due to a change in TensorFlow since we originally created the frozen graph that was published to download.01.org. The change happened between TensorFlow 1.3 and 1.4, and is likely to be tensorflow/tensorflow@3491881. It appears to be harmless.


jkamelin (approved):

IRDonch:
Jenkins please retry a build

PR #1809, opened 2020-11-19
person-vehicle-bike-detection-crossroad-0078

vladimir-dudnik:
update model accuracy ref value for map@person metric. originally reported value can't be reproduced


vladimir-dudnik:
Jenkins please retry a build

Ilya-Krylov (approved):

PR #1812, opened 2020-11-19
Corner crop of ground truth image

mzuevx:
Corner crop of ground truth image can be used for predictable crop of reference images  in accuracy checking pipelines of super resolution models


eaidova:
Jenkins ok to test

eaidova (approved):

PR #1815, opened 2020-11-20
Add generalized reader of images and video for python demos

anzhella-pankratova:
The reader is located in python_demos/common/images_capture.py and can take an image, folder with images or video as input data. Also, reading data in a loop mode is supported.


demos/python_demos/common/images_capture.py
```diff
@@ -0,0 +1,114 @@
...
+import copy
+
+import cv2
+
+
+class InvalidInput(Exception):
+    pass
+
+
+class ImagesCapture():
```
Wovchena, line 11:
```suggestion
class ImagesCapture:
```

demos/python_demos/common/images_capture.py
```diff
@@ -0,0 +1,114 @@
...
+
+    def __init__(self, loop):
+        self.loop = loop
+        self.isVideo = False
```
Wovchena, line 15: Demos could test if `openImagesCapture()` returned an instance of `VideoReader` instead.

demos/python_demos/common/images_capture.py
```diff
@@ -0,0 +1,114 @@
...
+
+    def read():
+        pass
```
Wovchena, line 18:
```suggestion
        raise NotImplementedError
```

demos/python_demos/common/images_capture.py
```diff
@@ -0,0 +1,114 @@
...
+
+
+class ImageReader(ImagesCapture):
+
+    def __init__(self, input, loop):
+        super().__init__(loop)
+        self.image = cv2.imread(input, cv2.IMREAD_COLOR)
+        self.canRead = True
```
Wovchena, line 26: Follow snake_case style

demos/python_demos/common/images_capture.py
```diff
@@ -0,0 +1,114 @@
...
+                return copy.deepcopy(image)
+
+            if self.loop:
+                self.imageId = 0
+                continue
+
+            return None
+
+
+class VideoReader(ImagesCapture):
```
Wovchena, line 73: Align it and `DirReader` with C++ version: name, what they do in init and read

demos/python_demos/common/images_capture.py
```diff
@@ -0,0 +1,114 @@
...
+                return None
+            self.cap.set(cv2.CAP_PROP_POS_FRAMES, 0)
+            status, frame = self.cap.read()
+            if not status:
+                return None
+        return frame
+
+
+def openImagesCapture(input, loop, cameraResolution=(1280, 720)):
+    input = input[0]
```
Wovchena, line 101:
```suggestion
```

demos/python_demos/human_pose_estimation_3d_demo/human_pose_estimation_3d_demo.py
```diff
@@ -63,6 +63,7 @@ def rotate_poses(poses_3d, R, t):
                       type=str, default=None)
     args.add_argument('--fx', type=np.float32, default=-1, help='Optional. Camera focal length.')
     args.add_argument('--no_show', help='Optional. Do not display output.', action='store_true')
+    args.add_argument('--loop', default=False, action='store_true', help='Optional. Run the input data in a loop.')
```
Wovchena, line 66:
```suggestion
    args.add_argument('-loop', '--loop', default=False, action='store_true', help='Optional. Run the input data in a loop.')
```
Put this argument after `-i`.

demos/python_demos/human_pose_estimation_3d_demo/human_pose_estimation_3d_demo.py
```diff
@@ -98,12 +99,14 @@ def rotate_poses(poses_3d, R, t):
...
     presenter = monitors.Presenter(args.utilization_monitors, 0)
-    for frame in frame_provider:
+
+    cap = input.read()    
+    while cap is not None:
         current_time = cv2.getTickCount()
-        input_scale = base_height / frame.shape[0]
-        scaled_img = cv2.resize(frame, dsize=None, fx=input_scale, fy=input_scale)
+        input_scale = base_height / cap.shape[0]
+        scaled_img = cv2.resize(cap, dsize=None, fx=input_scale, fy=input_scale)
```
Wovchena, line 107: `cap` usually refers to the source of frames but not to the actual frame.

Wovchena (changes requested):
Remove `nargs='+'` from `args.add_argument('-i', '--input',...)`, replace `default=''` with `required=True`.

Remove
```py
    if args.input == '':
        raise ValueError('Please, provide input data.')
```
from `human_pose_estimation_3d_demo.py`

Align help messages with C++. Take object_detection_demo_yolov3_async as an example, but keep `--input` for python version as alternative to `-i`.

It's usually better to base your work on this repo `develop` branch because `master` can be outdated.

demos/python_demos/common/images_capture.py
anzhella-pankratova: Done

demos/python_demos/common/images_capture.py
anzhella-pankratova: Done

demos/python_demos/common/images_capture.py
```diff
@@ -0,0 +1,114 @@
...
+        self.names = sorted(os.listdir(self.dir))
+        if not self.names:
+            raise InvalidInput
+        self.file_id = 0
+        read_images = 0
+        for name in self.names:
+            filename = os.path.join(self.dir, name)
+            image = cv2.imread(filename, cv2.IMREAD_COLOR)
+            if image is not None:
+                read_images += 1
```
Wovchena, line 54: This counter is not needed. Return instead. The idea is to accept the input once you can make sure there is at least one image.

demos/python_demos/common/images_capture.py
```diff
@@ -0,0 +1,114 @@
...
+            raise RuntimeError("Can't read the first image from {} dir".format(self.dir))
+
+    def read(self):
+        while True:
+            while self.file_id < len(self.names):
+                filename = os.path.join(self.dir, self.names[self.file_id])
+                image = cv2.imread(filename, cv2.IMREAD_COLOR)
+                self.file_id += 1
+                if image is not None:
+                    return copy.deepcopy(image)
```
Wovchena, line 65:
```suggestion
                    return image
```

demos/python_demos/common/images_capture.py
```diff
@@ -0,0 +1,114 @@
...
+            if self.loop:
```
Wovchena, line 66: You don't want to pass that `if` more than once per `read()` call, otherwise you may stuck in infinite `while True` outer loop.

demos/python_demos/common/images_capture.py
```diff
@@ -0,0 +1,114 @@
...
+        super().__init__(loop)
+        self.cap = cv2.VideoCapture()
+        try:
+            self.cap.open(int(input))
+            self.cap.set(cv2.CAP_PROP_BUFFERSIZE, 1)
+            self.cap.set(cv2.CAP_PROP_FRAME_WIDTH, camera_resolution[0])
+            self.cap.set(cv2.CAP_PROP_FRAME_HEIGHT, camera_resolution[1])
+            self.cap.set(cv2.CAP_PROP_AUTOFOCUS, 1)
+            self.cap.set(cv2.CAP_PROP_FOURCC, cv2.VideoWriter_fourcc('M', 'J', 'P', 'G'))
+        except:
```
Wovchena, line 84: The idea is to test whether `input` can be casted to `int`. So here you should catch a specific exception which is risen when the cast fails.

demos/python_demos/common/images_capture.py
```diff
@@ -0,0 +1,114 @@
...
+            self.cap.open(input)
+        if not self.cap.isOpened():
```
Wovchena, line 86: `VideoCapture.open()` returns `True` if it has been successfully opened. Check that instead of another call to `isOpened()`.

demos/python_demos/human_pose_estimation_3d_demo/human_pose_estimation_3d_demo.py
```diff
@@ -51,7 +51,9 @@ def rotate_poses(poses_3d, R, t):
                       type=str, required=True)
     args.add_argument('-i', '--input',
                       help='Required. Path to input image, images, video file or camera id.',
-                      nargs='+', default='')
+                      required=True)
+    args.add_argument('-loop', '--loop', default=False, action='store_true',
+                      help='Optional. Run the input data in a loop.')
```
Wovchena, line 56: Align help messages with C++. Take object_detection_demo_yolov3_async as an example, but keep --input for python version as alternative to -i.

Don't forget to update `README.md` after.

Wovchena (changes requested):
The test has hang.

demos/python_demos/common/images_capture.py
anzhella-pankratova: done

demos/python_demos/common/images_capture.py
anzhella-pankratova: done

demos/python_demos/common/images_capture.py
anzhella-pankratova: done

demos/python_demos/common/images_capture.py
anzhella-pankratova: done

demos/python_demos/common/images_capture.py
anzhella-pankratova: done

demos/python_demos/human_pose_estimation_3d_demo/human_pose_estimation_3d_demo.py
anzhella-pankratova: done

Wovchena: You should refer to VideoCapture as `cv2.VideoCapture` in python.

Wovchena:
It looks the test hang again.

demos/python_demos/human_pose_estimation_3d_demo/human_pose_estimation_3d_demo.py
Wovchena: Why parentheses after cv2.VideoCapture?

demos/python_demos/human_pose_estimation_3d_demo/human_pose_estimation_3d_demo.py
```diff
@@ -130,6 +135,7 @@ def rotate_poses(poses_3d, R, t):
         cv2.putText(frame, 'FPS: {}'.format(int(1 / mean_time * 10) / 10),
                     (40, 80), cv2.FONT_HERSHEY_COMPLEX, 1, (0, 0, 255))
         if args.no_show:
+            frame = cap.read()
```
Wovchena, line 138: I suggest to move the part working with windows under `if not args.no_show` instead, so you won't need to have `frame = cap.read()` twice.

anzhella-pankratova: Done

demos/python_demos/human_pose_estimation_3d_demo/human_pose_estimation_3d_demo.py
anzhella-pankratova: I deleted it

Wovchena (approved):

demos/python_demos/human_pose_estimation_3d_demo/human_pose_estimation_3d_demo.py
```diff
@@ -87,8 +87,12 @@ def rotate_poses(poses_3d, R, t):
     R = np.array(extrinsics['R'], dtype=np.float32)
     t = np.array(extrinsics['t'], dtype=np.float32)
 
-    frame_provider = InputReader(args.input)
-    is_video = frame_provider.is_video
+    cap = open_images_capture(args.input, args.loop)
+    is_video = isinstance(cap, VideoCapWrapper)
```
eizamaliev, line 91: I reccomend add field `type` (or any) to wrapper classes and use `is_video = cap.type == VIDEO` in app level.

demos/python_demos/common/images_capture.py
```diff
@@ -0,0 +1,116 @@
...
+        return None
+
+
+class DirReader(ImagesCapture):
+
+    def __init__(self, input, loop):
+        super().__init__(loop)
+        self.dir = input
+        if not os.path.isdir(self.dir):
+            raise InvalidInput
```
eizamaliev, line 44: Add some message to error which can explain it.

demos/python_demos/common/images_capture.py
```diff
@@ -0,0 +1,116 @@
...
+            if not self.loop:
+                return None
+            self.cap.set(cv2.CAP_PROP_POS_FRAMES, 0)
+            status, image = self.cap.read()
+            if not status:
+                return None
+        return image
+
+
+def open_images_capture(input, loop, camera_resolution=(1280, 720)):
```
eizamaliev, line 103: It is better to use `camera_resolution=None` here and use default settings in `VideoCapWrapper`.
In other case please explain why you choose `1280, 720`.

eizamaliev (changes requested):

demos/python_demos/human_pose_estimation_3d_demo/human_pose_estimation_3d_demo.py
Wovchena: Python doesn't need that, `isinstance()` does exactly that.

demos/python_demos/common/images_capture.py
Wovchena: This exception is used only for `open_images_capture()` no one is expected to see that exception.

demos/python_demos/common/images_capture.py
Wovchena: Setting specific resolution makes the result less dependent on environment. The default value follows C++ implementation. There is no specific reason to have exactly that value.

demos/python_demos/common/images_capture.py
anzhella-pankratova: If the input cannot be opened (and InvalidInput catch it), the RuntimeError with message will appear at the end of `open_images_capture()`

demos/python_demos/human_pose_estimation_3d_demo/human_pose_estimation_3d_demo.py
anzhella-pankratova: Done

demos/python_demos/common/images_capture.py
anzhella-pankratova: Done

demos/python_demos/common/images_capture.py
anzhella-pankratova: We didn't invent common solution, hence there is no changes for now.

demos/python_demos/common/images_capture.py
```diff
@@ -0,0 +1,166 @@
...
+    errors = {'InvalidInput': [], 'OpenError': []}
+    for idx, reader in enumerate((DirReader, ImreadWrapper, VideoCapWrapper)):
+        try:
+            if idx != 2:
+                return reader(input, loop)
+            return reader(input, loop, camera_resolution)
+        except (InvalidInput, OpenError) as e:
+            errors[str(e)].append(e.message)
```
Wovchena, line 161: Use `type(e)` as a key after updating `errors = {InvalidInput: [], OpenError: []}`. And you can append exceptions themselves.
`Exception` is derived from `BaseException` which is already good at handling `str()`.  So you don't need to define `__str__()`, print stored exceptions.

demos/python_demos/common/images_capture.py
```diff
@@ -0,0 +1,166 @@
...
+    if not errors['OpenError']:
+        print(*errors['InvalidInput'], sep='\n')
+    else:
+        print(*errors['OpenError'], sep='\n')
```
Wovchena, line 165: Print to stderr.

Wovchena (changes requested):

demos/python_demos/common/images_capture.py
```diff
@@ -0,0 +1,157 @@
...
+
+    def __init__(self, message):
+        InvalidInput.message = message
+
+
+class OpenError(Exception):
+    message = ''
+
+    def __init__(self, message):
+        OpenError.message = message
```
Wovchena, line 19: Here and in `InvalidInput` `message` should be instance attribute, but not class attribute.

anzhella-pankratova: Done

Wovchena (approved):

demos/python_demos/human_pose_estimation_3d_demo/human_pose_estimation_3d_demo.py
IRDonch: Do we need to do anything like that in the first place? Why not use the same logic in all cases?

demos/python_demos/common/images_capture.py
```diff
@@ -0,0 +1,155 @@
...
+        except (InvalidInput, OpenError) as e:
+            errors[type(e)].append(e.message)
+    try:
+        return VideoCapWrapper(input, loop, camera_resolution)
+    except (InvalidInput, OpenError) as e:
+        errors[type(e)].append(e.message)
+    if not errors[OpenError]:
+        print(*errors[InvalidInput], file=sys.stderr, sep='\n')
+    else:
+        print(*errors[OpenError], file=sys.stderr, sep='\n')
```
IRDonch, line 154: This all seems really involved. Why not follow the C++ implementation? Try each implementation ignoring `InvalidError`, and if all of them fail, throw an exception.

anzhella-pankratova: Firstly, the python version had followed the c++ implementation, but I changed it according to Eduard request about errors https://github.com/openvinotoolkit/open_model_zoo/pull/1815#discussion_r528633557.

eizamaliev (approved):

IRDonch (approved):

IRDonch:
Jenkins please retry a build

demos/python_demos/common/images_capture.py
```diff
@@ -0,0 +1,156 @@
...
+        return None
+
+    def fps(self):
+        return 1.0
+
+    def get_type(self):
+        return 'DIR'
+
+
+class VideoCapWrapper(ImagesCapture):
```
fzhar, line 103: Please split this class into 2 separate classes. At next step we will implement check for stream data depletion (as separate PR I presume), so we'll definetely have to doit  in separate class for video reader. So, please split this class into VideoFileReader and CameraReader clsasses.

fzhar (changes requested):

demos/python_demos/common/images_capture.py
IRDonch: Wouldn't it make sense to wait until there's actually a difference between the two cases before splitting the class?

fzhar: It's better make consistant sloution from the beginning - other cases follow one case -one class paradigm. Well, if it really would take much time to implement, I'd agree to postpone this, but I believe this is not the case.

anzhella-pankratova: It is easy to implement, but the difference between two classes will be only in `__init__`, so the `self.read()` in class `CameraReader` looks a bit redundant. Also, we can check stream data depletion (when the `self.read()` returns None) only if the class `VideoCapWrapper` has a video type, otherwise raise some exception about interruption of camera capture.

fzhar: Well, init is 1/2 of whole class code - so it's quite enough. Also, there're questions about read function - do we really need to self.cap.set(cv2.CAP_PROP_POS_FRAMES, 0) for camera ? loop won't work there, so it could be erased too. So, there would be quite different code, particulalry after implementing check for the end of stream

anzhella-pankratova: I agree with you about camera's loop, so I will split it into two classes.

fzhar (approved):

demos/python_demos/common/images_capture.py
```diff
@@ -0,0 +1,176 @@
...
+    def fps(self):
+        return self.cap.get(cv2.CAP_PROP_FPS)
+
+    def get_type(self):
+        return 'VIDEO'
+
+
+class CameraCapWrapper(ImagesCapture):
+
+    def __init__(self, input, loop, camera_resolution):
```
Wovchena, line 132: > loop won't work there, so it could be erased too.

Init shouldn't take loop now. You need to either initialize `ImagesCapture`'s loop with `False` or remove this field from the base class.

anzhella-pankratova: Remove it.

Wovchena (approved):

demos/python_demos/common/images_capture.py
```diff
@@ -0,0 +1,175 @@
...
+
+    def __init__(self, message):
+        self.message = message
+
+
+class ImagesCapture:
+
+    def __init__(self):
+        raise NotImplementedError
+
```
IRDonch, line 24:
```suggestion
```

`__init__` shouldn't raise an exception, because there's no reason to require subclasses to implement it (a subclass could have no `__init__` and still be a valid implementation). Just remove it.

anzhella-pankratova: Done

PR #1828, opened 2020-11-24
downloader: cap the PyTorch version at 1.6 to work around a change in behavior

IRDonch:


IRDonch:
CC @vladimir-dudnik

vladimir-dudnik (approved):

IRDonch:
I also bumped the minimal version, since it turns out we need it for CoCosNet.

PR #1829, opened 2020-11-24
update ref for license-plate-recognition-barrier-0007

vladimir-dudnik:
update ref values for actually used synthetic dataset

### DIFF
--- a/models/public/license-plate-recognition-barrier-0007/license-plate-recognition-barrier-0007.md
+++ b/models/public/license-plate-recognition-barrier-0007/license-plate-recognition-barrier-0007.md
@@ -7,7 +7,7 @@ Chinese license plates in traffic.
 
 ## Validation Dataset - Internal
 
-1165 Chinese plates from different provinces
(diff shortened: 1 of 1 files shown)


snosov1:
Not sure I see the point here. For one thing, synthetic dataset was used for training (in that sense we do something "anti-data-scientific" by validating and training on the same set). For the second thing, it seems unnecessarily big for validation purposes. 

What was wrong with the original set of 3K images?

eaidova (approved):

Anna-Mironova (approved):

vladimir-dudnik:
@nosov that aligns ref metric value (used in validation) with actual state. We may return back to this after release, if you want.

snosov1:
I see that it has to match validation. My question is why do you validate on the synthetic set (vs 3K real world images)

vladimir-dudnik:
I think it is because you guys define validation procedure in this way, is not it? We do not invent the way how to validate models.

snosov1:
So, it was 3K and *we* changed it to 300K?

vladimir-dudnik:
I think it was started from 300K when model go to OMZ.

PR #1830, opened 2020-11-25
Aa/melgan pipeline

a-a-egorovich:
MelGAN with Upsamlpe layers instead of TransposedConvolution1d + ForwardTacotron


models/intel/text-to-speech-en-0001/accuracy-check.yml
```diff
@@ -0,0 +1,31 @@
+evaluations:
+  - name: text-to-speech-en-0001
```
vladimir-dudnik, line 2: model has different name. you need either rename model to this name or refer right model name there

vladimir-dudnik (changes requested):

a-a-egorovich:
Not relevant

PR #1831, opened 2020-11-25
downloader: add a separate requirements file for TensorFlow model conversion

IRDonch:
There are two reasons why such a file is needed:

* The EfficientDet models are only convertible with TF 2.3+, while Model Optimizer requires 1.15+. This, if the user just uses the requirements from Model Optimizer, they can install an incompatible version of TF.

* TensorFlow is incompatible with h5py 3.x, but it doesn't say so in its requirements. If the user just uses the requirements from MO, they may end up installing an incompatible version of h5py, which will break the conversion of YOLO models. This should be fixed in TensorFlow 2.4, but it has not yet been released. Meanwhile, we can add a workaround via this file.

### DIFF
--- a/ci/update-requirements.py
+++ b/ci/update-requirements.py
@@ -69,7 +69,7 @@ def main():
         'tools/accuracy_checker/requirements-core.in')
     pc('ci/requirements-check-basics.txt', 'ci/requirements-check-basics.in')
     pc('ci/requirements-conversion.txt',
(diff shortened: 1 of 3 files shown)


vladimir-dudnik (approved):

IRDonch:
Jenkins please retry a build

jkamelin (approved):

PR #1834, opened 2020-11-25
AC: enable non-inferenece based metrics for colorization

eaidova:
it is difficult to quantize colorization models due to composite config struture and metrics based on inference classification model.
Now standard image processing metrics (PSNR, SSIM, MAE can be calculated on colorization models)


vladimir-dudnik (approved):
do we need to adopt REF values in model description?

eaidova:
@vladimir-dudnik, I think for this release no, I just enable this feature for internal experiments, if experiments will be successful, then I update configs and ref values publicly

vladimir-dudnik:
ok, then

PR #1863, opened 2020-12-01
Add missing function

eizamaliev:


demos/python_demos/common/models/ssd.py
```diff
@@ -58,6 +58,15 @@ def __call__(self, outputs):
 
     class BoxesLabelsParser:
         def __init__(self, input_size, all_outputs, labels_layer='labels', default_label=1):
+            def find_layer_by_name(name, layers):
```
IRDonch, line 61: This is the same function as in `MultipleOutputParser`. Couldn't you turn it into a global function?

PR #1874, opened 2020-12-02
flake8: add a check for unused imports

IRDonch:
Fix all issues found, either by removing the import, or by re-exporting it through `__all__`.


eaidova (approved):

demos/python_demos/common/models/__init__.py
```diff
@@ -15,8 +15,16 @@
...
+from .ssd import SSD
+from .yolo import YOLO
+
+__all__ = [
+    'CenterNet',
+    'FaceBoxes',
+    'RetinaFace',
+    'SSD',
+    'YOLO',
+]
```
Wovchena, line 30: This is not needed:
> [It also includes any submodules of the package that were explicitly loaded by previous import statements.](https://docs.python.org/3/tutorial/modules.html#importing-from-a-package)

IRDonch: I'm aware, but the problem is that flake8 will consider the imports unused if the imported names aren't... well... _used_.

Wovchena (approved):

PR #1581, opened 2020-09-21
Use unified performance metrics classes in all demos

AnthonyQuantum:
## C++ demos
- [x] classification_demo
- [ ] crossroad_camera_demo
- [ ] gaze_estimation_demo
- [ ] human_pose_estimation_demo
- [ ] interactive_face_detection_demo
- ~~mask_rcnn_demo~~
- [ ] multi_channel (x3)
- ~~object_detection_demo_faster_rcnn~~
- [x] object_detection_demo_ssd_async
- [ ] object_detection_demo_yolov3_async
- [ ] pedestrian_tracker_demo
- [ ] security_barrier_camera_demo
- [ ] segmentation_demo
- [ ] smart_classroom_demo
- ~~super_resolution_demo~~
- [ ] text_detection_demo

## Python demos
- ~~3d_segmentation_demo~~
- [ ] action_recognition
- [ ] asl_recognition_demo
- ~~? bert_question_answering_demo~~
- ~~colorization_demo~~
- ~~? formula_recognition_demo~~
- [ ] handwritten_japanese_recognition_demo
- [ ] human_pose_estimation_3d_demo
- [ ] image_inpainting_demo
- ~~? image_retrieval_demo~~
- [ ] instance_segmentation_demo
- ~~monodepth_demo~~
- [ ] multi_camera_multi_target_tracking
- [ ] object_detection_demo_centernet
- [ ] object_detection_demo_faceboxes
- [ ] object_detection_demo_retinaface
- [ ] object_detection_demo_ssd_async
- [x] object_detection_demo_yolov3_async
- ~~segmentation_demo~~
- [ ] single_human_pose_estimation_demo
- [ ] text_spotting_demo
- [ ] whiteboard_inpainting_demo

---

- [x] - done
- [ ] - to do
- ~~demo~~ - doesn't need metrics
- ~~? demo~~ - up for discussion

## Last update of the list: Sep 3, 2020

PR #1876, opened 2020-12-02
Use perf metrics in classification_demo

Wovchena:


Wovchena:
Closes https://github.com/openvinotoolkit/open_model_zoo/pull/1581

demos/classification_demo/grid_mat.hpp
```diff
@@ -29,6 +31,7 @@ struct LabeledImage {
 class GridMat {
 public:
     cv::Mat outImg;
+    PerformanceMetrics performanceMetrics;
```
IRDonch, line 34: Should this really be here? The performance metrics don't really have anything to do with grid-ness; maybe it would be better to have it as local variable of `main` instead?

demos/classification_demo/grid_mat.hpp
```diff
@@ -57,32 +60,26 @@ class GridMat {
         testMessageSize = cv::getTextSize(testMessage, fontType, fontScale, thickness, &baseline);
     }
 
-    void textUpdate(double avgFPS, double avgLatency, double accuracy,
-                    bool isFpsTest, bool showAccuracy,
+    void textUpdate(PerformanceMetrics::TimePoint lastRequestStartTime,
+                    cv::Mat& frame,
```
IRDonch, line 64: What's the purpose of this parameter? You always pass `outImg` as the value anyway.

demos/classification_demo/grid_mat.hpp
Wovchena: Moved

demos/classification_demo/grid_mat.hpp
Wovchena: Good note

IRDonch (approved):

PR #1881, opened 2020-12-03
Evaluator-level support for model output feedback to next step input

mzuevx:
Attempt to implement feedback from model output to next step input.
Based on MultiFrameInputIdentifiers, replaces last frame from frame sequence with data provided from model output or initializer.
Model output can be selected between inference, adapter and postprocessor stages,
Allows to process feedback data on the same pipeline as ordinal inputs.

### DIFF
--- a/tools/accuracy_checker/accuracy_checker/evaluators/model_evaluator.py
+++ b/tools/accuracy_checker/accuracy_checker/evaluators/model_evaluator.py
@@ -49,6 +49,9 @@ class ModelEvaluator(BaseEvaluator):
         self._annotations = []
         self._predictions = []
         self._metrics_results = []
+        self._inference_stage_buffer = {}
+        self._adapter_stage_buffer = {}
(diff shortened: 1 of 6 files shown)


tools/accuracy_checker/accuracy_checker/evaluators/model_evaluator.py
```diff
@@ -49,6 +49,9 @@ def __init__(
         self._annotations = []
         self._predictions = []
         self._metrics_results = []
+        self._inference_stage_buffer = {}
+        self._adapter_stage_buffer = {}
+        self._postprocessor_stage_buffer = {}
```
eaidova, line 54: I do not think that it should be in evaluator

mzuevx:
Functionality implemented in custom evaluator, see https://github.com/openvinotoolkit/open_model_zoo/pull/2099

PR #1898, opened 2020-12-07
machine_translation_demo: fix minor issues

IRDonch:


IRDonch:
Apparently `from_file` is only available since tokenizers 0.9.0 (and actually works only since 0.9.3). I'll revive this PR after I do the usual post-release dependency bump.

PR #1903, opened 2020-12-10
flake8: enable all remaining pyflakes checks

IRDonch:
Fix all existing issues, which are all star imports.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -15,18 +15,8 @@ select=
     E7,
     # parse errors
     E9,
-    # unused imports
-    F401,
-    # formatting errors
-    F5,
-    # semantic errors
-    F6,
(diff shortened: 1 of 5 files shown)


eaidova:
@IRDonch anyway I plan to remove useless parameters serialization script (nobody use it and there is no plan to use it), so you can remove it and related tests as part of this request

IRDonch:
@eaidova If you're going to remove it, then I'd rather just wait for that and then rebase on top. :-)

eaidova:
> @eaidova If you're going to remove it, then I'd rather just wait for that and then rebase on top. :-)

ok, removed

Wovchena (approved):

PR #1908, opened 2020-12-11
Replace storage.openvinotoolkit.org links with download.01.org links

IRDonch:
This creates an alternate version of the model configs for users who can't access storage.openvinotoolkit.org.


IRDonch:
CC @yssemaev @moslex

vladimir-dudnik (approved):

PR #1916, opened 2020-12-14
object_detection_demo models.lst

vladimir-dudnik:
extend object_detection_demo models.lst with ssd models


Wovchena (changes requested):
Keep the lists sorted alphabetically inside of each group.

vladimir-dudnik:
put items in alhabet order

Wovchena (approved):

PR #1918, opened 2020-12-15
demos: fix/disable MSVC warnings

Wovchena:
The compiler asks to have `/EHsc` if cmake is run with
`-DCMAKE_CXX_FLAGS=/WX`.
`WIN32` is not defined.
`fileNameNoExt()` is unused from anywhere.
Disable warnings for `ctcdecode_numpy_impl` for MSVC as well as for GCC
like compilers.


demos/CMakeLists.txt
```diff
@@ -45,7 +45,8 @@ endif()
 
 if(MSVC)
     add_definitions(-D_CRT_SECURE_NO_WARNINGS -DSCL_SECURE_NO_WARNINGS)
-    add_compile_options(/wd4251 /wd4275 /wd4267) #disable some warnings
+    add_compile_options(/wd4251 /wd4275 /wd4267  # disable some warnings
+                        /EHsc)  # Enable standard C++ stack unwinding, assume functions with extern "C" never throw
```
IRDonch, line 49: You should probably explicitly add `/W3` too, since it will also be omitted if the user overrides `CMAKE_CXX_FLAGS`.

demos/python_demos/speech_recognition_demo/ctcdecode-numpy/CMakeLists.txt
```diff
@@ -34,6 +34,8 @@ endforeach()
 # SWIG-generated code causes some warnings; disable them.
 if(COMPILER_IS_GCC_LIKE)
     target_compile_options(${target_name} PRIVATE -Wno-narrowing)
+elseif(MSVC)
+    target_compile_options(${target_name} PRIVATE /W0)
```
IRDonch, line 38: Surely it's not necessary to disable _all_ warnings? I don't think this library is that hopeless. :-)

IRDonch (changes requested):

demos/python_demos/speech_recognition_demo/ctcdecode-numpy/CMakeLists.txt
Wovchena: Yes, but now I think that we shouldn't introduce differences with the original code without a strong reason so it would be easier to compare them if the original one introduces something what we want.

IRDonch: What do you mean by the original code?

Wovchena: The one these files were copied from.

demos/python_demos/speech_recognition_demo/ctcdecode-numpy/CMakeLists.txt
```diff
@@ -34,6 +34,12 @@ endforeach()
 # SWIG-generated code causes some warnings; disable them.
 if(COMPILER_IS_GCC_LIKE)
     target_compile_options(${target_name} PRIVATE -Wno-narrowing)
+elseif(MSVC)
+    target_compile_options(${target_name} PRIVATE
+        /wd4244  # Disable conversion warning =
+        /wd4838  # Disable conversion warning in aggregate or list initialization
+        /wd4834  # Discarding return value of function with 'nodiscard' attribute
```
IRDonch, line 41: I don't think this one should be disabled, because it's a real issue. :-) It's caused by an incorrect usage of `std::unique`, which should be fixed.

Wovchena: It doesn't seem to be strong enough reason. It works after all … somehow :)

demos/python_demos/speech_recognition_demo/ctcdecode-numpy/CMakeLists.txt
IRDonch: I should note that not all of these files were copied; several of them are original to us, including `word_prefix_set.cpp`, where that 4834 warning occurs.

demos/python_demos/speech_recognition_demo/ctcdecode-numpy/CMakeLists.txt
IRDonch: It always works, until it doesn't. Maybe in our usage all words are already unique, but they might not be in the future. Why not fix it?

demos/python_demos/speech_recognition_demo/ctcdecode-numpy/CMakeLists.txt
Wovchena: Fixed

demos/python_demos/speech_recognition_demo/ctcdecode-numpy/ctcdecode_numpy/word_prefix_set.cpp
```diff
@@ -29,15 +29,16 @@ size_t WordPrefixSet::add_words(const std::vector<std::vector<int> >& words) {
...
+  auto last_it = std::unique(word_ptrs.begin(), word_ptrs.end(), int_word_equal);
 
   trie_chars_.clear();
   trie_starts_.clear();
   trie_weights_.clear();
 
   // Go through the sorted list, and update children lists at every layer of trie.
   // The new child may only be the last child in each list.
-  for (auto word : word_ptrs) {
+  for (auto word_it = word_ptrs.begin(); word_it != last_it; word_it++) {
```
IRDonch, line 40: This is insufficient; you also need to replace `word_ptrs.size()` with `last_it - word_ptrs.begin()` below.

(It might be simpler to do `word_ptrs.erase(last_it, word_ptrs.end());` instead.)

IRDonch (approved):

IRDonch:
Jenkins please retry a build

PR #1921, opened 2020-12-15
New command line parser

fzhar:
Main purpose: remove dependency of gflags, keep all settings in one object instead of using macros. Curently it is integrated into object_detection_demo only - as test. If it will be approved, it will integrated into other demos also
Pros (comparing to currently existing code): single storage for all settings, can be passed to functions as single parameter, supports long and short names (as in python), description is bound to parameter (instead of separate variables) and we can adjust generated description text as we want to. Also, it supports multiple values for each parameter and parses them automatically.
Cons: command line format change for some parameters (due to -- prefix for long names, this can be changed). Arguments are converted on-the-fly from string to proper type - that may cause some slowdown if used in cycle (very low probability).

### DIFF
--- a/demos/common/include/samples/params_parser.h
+++ b/demos/common/include/samples/params_parser.h
@@ -12,7 +12,8 @@ struct Param {
            std::string defVal,
            std::string description,
            bool isRequired,
-           int expectedArgumentsNumber) {
+           int expectedArgumentsNumber,
+           bool unsplittable) {
         this->name = name;
         this->shortName = shortName;
         values.push_back(defVal);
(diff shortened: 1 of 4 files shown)


vladimir-dudnik:
@fzhar would be nice to put in PR description what it is about, what issues solves (if any), what functionality enables, what is the purpose of that change

IRDonch:
FYI, I'm probably not going to have time to review this before my vacation.

vladimir-dudnik (changes requested):
would be nice to put in PR description what it is about, what issues solves (if any), what functionality enables, what is the purpose of that change

PR #1939, opened 2020-12-18
C++ demos: add .gif files to README

anzhella-pankratova:
add animated GIF files to show how demo output may look like.

- [x] crossroad_barrier_demo
- [x] gaze_estimation_demo
- [x] inteactive_face_detection_demo
- [x] object_detection_demo
- [x] pedestrian_tracker_demo
- [x] security_barrier_camera_demo
- [x] segmentation_demo
- [x] smart_classroom_demo
- [x] python action_recognition
- [x] python whiteboard_inpainting


demos/crossroad_camera_demo/README.md
```diff
@@ -1,5 +1,7 @@
 # Crossroad Camera C++ Demo
 
+![](./crossroad_camera.gif)
```
Wovchena, line 3: The usage is valid, but the source video doesn't seem to be representative enough.
`security_barrier_camera.gif` is also run on a bit different video: the demo name suggests that the camera is placed at level of barrier, but not above of passing cars.

@vladimir-dudnik should the gifs be kept?

demos/interactive_face_detection_demo/README.md
```diff
@@ -1,5 +1,7 @@
 # Interactive Face Detection C++ Demo
 
+![](./interactive_face_detection.gif)
```
Wovchena, line 3: I think it's better to use all other nets the demo supports like emotions to show all capabilities.

Wovchena (changes requested):

demos/crossroad_camera_demo/README.md
vladimir-dudnik: I think we can use a few subsequent frames from internal validation set, just resize it to lower resolution.
@yssemaev hope there is no legal issues with this, as it should not fall into dataset redistribution case. It is just screenshots of demo execution.

demos/interactive_face_detection_demo/README.md
anzhella-pankratova: done.

vladimir-dudnik:
@andrew-zaytsev FYI. we are going to add animated GIF to OMZ demo descriptions

demos/crossroad_camera_demo/README.md
yssemaev: Hi, I have no objections if the dataset(s) was approved for Publishing and Distribution cases.

Wovchena (approved):

vladimir-dudnik (approved):

PR #1941, opened 2020-12-18
Remove labels list adjustment in object detection models (SSD and Yolo).

fzhar:
Algorithm for labels list adjustment (adding class for background) is removed. Check for labels list to correspond to number of output classes is removed.
If there's no label to some particular ID it will be shown as Label #N

### DIFF
--- a/demos/object_detection_demo/README.md
+++ b/demos/object_detection_demo/README.md
@@ -96,7 +96,7 @@ To run the demo, you can use public or pre-trained models. To download the pre-t
 
 > **NOTE**: Before running the demo with a trained model, make sure the model is converted to the Inference Engine format (\*.xml + \*.bin) using the [Model Optimizer tool](https://docs.openvinotoolkit.org/latest/_docs_MO_DG_Deep_Learning_Model_Optimizer_DevGuide.html).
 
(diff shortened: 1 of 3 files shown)


demos/object_detection_demo/README.md
```diff
@@ -96,6 +96,8 @@ To run the demo, you can use public or pre-trained models. To download the pre-t
 
 > **NOTE**: Before running the demo with a trained model, make sure the model is converted to the Inference Engine format (\*.xml + \*.bin) using the [Model Optimizer tool](https://docs.openvinotoolkit.org/latest/_docs_MO_DG_Deep_Learning_Model_Optimizer_DevGuide.html).
 
+If labels file is used, it should correspond to model output. Demo suggests labels listed in the file to be indexed from 0, one line - one label (i.e. very first line contains label for ID 0). Note that s	ome models may return labels IDs in range 1..N, in this case label file should contain "background" label at the very first line.
```
vladimir-dudnik, line 99:
```suggestion
If labels file is used, it should correspond to model output. Demo suggests labels listed in the file to be indexed from 0, one line - one label (i.e. very first line contains label for ID 0). Note that some models may return labels IDs in range 1..N, in this case label file should contain "background" label at the very first line.
```

vladimir-dudnik (commented):
typo

vladimir-dudnik (approved):

PR #1945, opened 2020-12-22
images_capture.py: reading of corrupted frames in video files

anzhella-pankratova:


demos/python_demos/common/images_capture.py
```diff
@@ -103,18 +103,30 @@ def __init__(self, input, loop):
...
+                status, image = self.cap.read()
+                if not status:
+                    return None
+            else:
+                for _ in range(self.corrupted_frames_threshold-1):
+                    self.actual_frames_count += 1
+                    status, image = self.cap.read()
+                    if status:
+                        return image
+                raise RuntimeError("The video contains corrupted frames")
```
Wovchena, line 128: Do you have an example of such video to prove the issue may happen?

anzhella-pankratova: @fzhar, could you provide an example please?

PR #1949, opened 2020-12-22
object detection demo models

vladimir-dudnik:
add efficientdets and faceboxes to models.lst and test cases

### DIFF
--- a/demos/tests/cases.py
+++ b/demos/tests/cases.py
@@ -202,6 +202,10 @@ NATIVE_DEMOS = [
             '-i': DataPatternArg('object-detection-demo')}),
         [
             *combine_cases(
+                TestCase(options={'--architecture_type': 'faceboxes',
+                                  '-m': ModelArg('faceboxes-pytorch')})
+            ),
+            *combine_cases(
                 TestCase(options={'-at': 'ssd'}),
                 single_option_cases('-m',
                     ModelArg('efficientdet-d0-tf'),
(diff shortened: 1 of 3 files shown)


demos/tests/cases.py
```diff
@@ -237,8 +240,13 @@ def single_option_cases(key, *args):
             *combine_cases(
                 TestCase(options={'-at': 'yolo'}),
                 single_option_cases('-m',
+                    ModelArg('person-vehicle-bike-detection-crossroad-yolov3-1020'),
                     ModelArg('yolo-v3-tf'),
                     ModelArg('yolo-v3-tiny-tf'))),
+            *combine_cases(
+                TestCase(options={'-at': 'faceboxes'}),
+                single_option_cases('-m',
+                    ModelArg('faceboxes-pytorch'))),
```
Wovchena, line 249: There are no other `faceboxes` models. You should merge `-at` and `-m` here.

vladimir-dudnik: ok, changed. also put options in alphabet order

Wovchena: Wrapping with `*combine_cases()` is not needed: product of one element is just that element.

Wovchena: And as the test says there is no `--architecture_type` for C++ version, only `-at`.

vladimir-dudnik: fixed

Wovchena: There is `combine_cases()` left for python

Wovchena (approved):

PR #1951, opened 2020-12-23
Python demos: add capability to record output video in a file

anzhella-pankratova:
This PR is intended to add demo capability to write output, which usually shown on screen to video file, using motion jpeg as an encoder.

- [x] action_recognition
- [x] colorization
- [x] human_pose_3d_estimation
- [x] human_pose_estimation
- [x] image_retrieval  
- [x] instance_segmentation
- [x] object_detection
- [x] segmentation
- [x] single_human_pose_estimation
- [x] whiteboard_inpainting

### DIFF
--- a/demos/python_demos/action_recognition/README.md
+++ b/demos/python_demos/action_recognition/README.md
@@ -36,6 +36,7 @@ Running the application with the `-h` option yields the following usage message:
 
 ```
 usage: action_recognition.py [-h] -i INPUT [--loop] [-o OUTPUT]
+                             [-limit OUTPUT_LIMIT]
                              -m_en M_ENCODER [-m_de M_DECODER]
                              [-l CPU_EXTENSION] [-d DEVICE]
                              [-lb LABELS] [--no_show] [-s LABEL_SMOOTHING]
@@ -50,6 +51,9 @@ Options:
   --loop                Optional. Enable reading the input in a loop.
   -o OUTPUT, --output OUTPUT
                         Optional. Name of output to save.
+  -limit OUTPUT_LIMIT, --output_limit OUTPUT_LIMIT
+                        Optional. Number of frames to store in output.
+                        If -1 is set, all frames will be stored.
   -m_en M_ENCODER, --m_encoder M_ENCODER
                         Required. Path to encoder model.
   -m_de M_DECODER, --m_decoder M_DECODER
--- a/demos/python_demos/action_recognition/README.md
+++ b/demos/python_demos/action_recognition/README.md
@@ -46,7 +46,7 @@ Options:
   -h, --help            Show this help message and exit.
   -i INPUT, --input INPUT
                         Required. An input to process. The input must be a single image,
-                        a folder of images or anything that cv2.VideoCapture can process.
+                        a folder of images, video file or camera id.
   --loop                Optional. Enable reading the input in a loop.
   -o OUTPUT, --output OUTPUT
                         Optional. Name of output to save.
--- a/demos/python_demos/action_recognition/README.md
+++ b/demos/python_demos/action_recognition/README.md
@@ -35,8 +35,9 @@ You can change the value of `num_requests` in `action_recognition.py` to find an
 Running the application with the `-h` option yields the following usage message:
 
 ```
-usage: action_recognition.py [-h] -i INPUT -m_en M_ENCODER [-m_de M_DECODER]
-                             [-l CPU_EXTENSION] [-d DEVICE] [--fps FPS]
+usage: action_recognition.py [-h] -i INPUT [--loop] [-o OUTPUT]
+                             -m_en M_ENCODER [-m_de M_DECODER]
+                             [-l CPU_EXTENSION] [-d DEVICE]
                              [-lb LABELS] [--no_show] [-s LABEL_SMOOTHING]
                              [--seq DECODER_SEQ_SIZE]
                              [-u UTILIZATION_MONITORS]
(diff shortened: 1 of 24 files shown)


Wovchena (changes requested):

demos/python_demos/colorization_demo/colorization_demo.py
```diff
@@ -43,6 +43,8 @@ def build_arg():
                               'a folder of images or anything that cv2.VideoCapture can process.')
     in_args.add_argument('--loop', default=False, action='store_true',
                          help='Optional. Enable reading the input in a loop.')
+    in_args.add_argument('-o', '--output_video', required=False,
+                         help='Optional. Path to an output video file.')
```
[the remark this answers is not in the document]
anzhella-pankratova: done.

demos/python_demos/colorization_demo/colorization_demo.py
```diff
@@ -83,6 +85,13 @@ def build_arg():
     graph_size = (imshow_size[0] // 2, imshow_size[1] // 4)
     presenter = monitors.Presenter(args.utilization_monitors, imshow_size[1] * 2 - graph_size[1], graph_size)
 
+    fps = cap.fps()
+    if args.output_video:
+        fourcc = cv.VideoWriter_fourcc(*'MJPG')
+        output_video = cv.VideoWriter(args.output_video, fourcc, fps, (1280, 960))
```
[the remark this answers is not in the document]
anzhella-pankratova: done.

demos/python_demos/colorization_demo/colorization_demo.py
```diff
@@ -134,4 +147,8 @@ def build_arg():
                 break
             presenter.handleKey(key)
         original_frame = cap.read()
+
+        if output_video is not None:
+            output_video.release()
```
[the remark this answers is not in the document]
anzhella-pankratova: done.

demos/python_demos/object_detection_demo/object_detection_demo.py
```diff
@@ -258,6 +254,10 @@ def main():
             presenter.drawGraphs(frame)
             frame = draw_detections(frame, objects, palette, model.labels, args.prob_threshold)
             metrics.update(start_time, frame)
+
+            if video_writer.isOpened():
+                video_writer.write(frame)
```
Wovchena, line 259: There is the second part of pipeline below: `# Process completed requests` which also should save frames. The same is for segmentation.

anzhella-pankratova: done.

Wovchena (approved):

demos/python_demos/action_recognition/README.md
```diff
@@ -45,32 +46,33 @@ Options:
   -h, --help            Show this help message and exit.
   -i INPUT, --input INPUT
                         Required. An input to process. The input must be a single image,
-                        a folder of images or anything that cv2.VideoCapture can process
-  --loop                Optional. Enable reading the input in a loop
+                        a folder of images or anything that cv2.VideoCapture can process.
```
vladimir-dudnik, line 49:
```suggestion
                        a folder of images, video file or camera id.
```

vladimir-dudnik: We need take into account that users might be not an OpenCV experts and so, not aware of what cv2.VideoCapture can process and how to supply input for VideoCapture. All the details should not be put into command line option description, seems that might be a separate common section in each demo readme. Need think on this. For this particular case, I think we may just change phrase a bit.

demos/python_demos/action_recognition/README.md
```diff
@@ -45,32 +46,33 @@ Options:
...
+  --loop                Optional. Enable reading the input in a loop.
+  -o OUTPUT, --output OUTPUT
```
vladimir-dudnik, line 51: Do we define somewhere what will happens in case of endless input, like input from camera or when '-loop' option is active? We might need some safety guard for that potentially dangerous option. Like, by default store only first 100 or 1000 frames and with adding additional parameter, '-output_limit' allow to set -1 to store all the input frames and positive number to specify how many frames to store (or even the range of frames to store, like 100,-1 - to store from 100-th frame up to the end of input stream.

vladimir-dudnik (changes requested):
option to store video with detections is useful, but need some guard from endless input

demos/python_demos/action_recognition/README.md
Wovchena: I assume it will fill all available disk space with output. But that's what one requests by running a demo for long enough. And it doesn't seem to be dangerous. At least no more dangerous then fully taking any other system resource, RAM for example, which we don't handle explicitly as well. My opinion is that we shouldn't extend functionality without actual need.

Wovchena (changes requested):

demos/python_demos/action_recognition/README.md
vladimir-dudnik: this is dangerous for system stability and such action should requested by user explicitly rather then be side effect. this is an actual need.

Wovchena: I still can't see how it differs from RAM, but ok.

demos/python_demos/action_recognition/README.md
anzhella-pankratova: done.

demos/python_demos/action_recognition/README.md
```diff
@@ -45,32 +46,33 @@ Options:
...
+                        Optional. Path to file with label names.
+  --no_show             Optional. Don't show output.
   -s LABEL_SMOOTHING, --smooth LABEL_SMOOTHING
                         Optional. Number of frames used for output label
-                        smoothing
+                        smoothing.
   --seq DECODER_SEQ_SIZE
                         Optional. Length of sequence that decoder takes as
-                        input
+                        input.
```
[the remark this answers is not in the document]
anzhella-pankratova: done.

demos/python_demos/action_recognition/action_recognition_demo/timer.py
```diff
@@ -60,6 +60,8 @@ def time_section(self):
         self.tock()
 
     def __repr__(self):
+        if not self._times:
+            return "{:.2f}ms (+/-: {:.2f}) {:.2f}fps".format(0, 0, 0)
```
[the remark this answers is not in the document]
anzhella-pankratova: done.

Wovchena (approved):

vladimir-dudnik (approved):

demos/python_demos/action_recognition/action_recognition.py
```diff
@@ -41,6 +41,9 @@ def build_argparser():
                       help='Optional. Enable reading the input in a loop.')
     args.add_argument('-o', '--output', required=False,
                       help='Optional. Name of output to save.')
+    args.add_argument('-limit', '--output_limit', required=False, default=1000, type=int,
+                      help='Optional. Number of frames to store in output. '
+                           'If -1 is set, all frames will be stored.')
```
Wovchena, line 46: Can you change from first conditional to zero? :) That's not just possible that if -1 is set then all frames may be are to be stored. That's a rule you describe.

anzhella-pankratova: Done.

Wovchena (approved):

PR #1974, opened 2021-01-11
Fix bug of colorization demo for siggraph model

akorobeinikov:


demos/python_demos/colorization_demo/colorization_demo.py
```diff
@@ -98,7 +98,16 @@ def build_arg():
...
+
+        if len(load_net.input_info.keys()) == 1:
+            res = exec_net.infer(inputs={input_blob: [img_l_rs]})
+        else:
+            input_blob, user_ab_input, user_map = load_net.input_info.keys()
+            res = exec_net.infer(inputs={
+                input_blob: [img_l_rs],
+                user_ab_input: np.zeros((2, *img_l_rs.shape)),
+                user_map: [np.zeros(img_l_rs.shape)]
+            })
```
eaidova, line 110: could you please make consistent inputs filling format (everywhere np.array or everywhere starts with list)?

eaidova: Also if demo supports several input files processing, then better to create feed_dict with initiall input data and then chenge only input_blob content

eaidova:
@akorobeinikov I am right understand, that the same fixes required for AC?

demos/python_demos/colorization_demo/colorization_demo.py
akorobeinikov: Done

eaidova (approved):

PR #1980, opened 2021-01-13
Updated accuracy value for midasnet model

Anna-Mironova:


IRDonch:
Where did you get this value? When I ran the accuracy check on this model, I just got 0.07.

Anna-Mironova:
> Where did you get this value? When I ran the accuracy check on this model, I just got 0.07.

This value can be obtained by running AC with `--ignore_result_formatting True` argument

IRDonch (approved):

PR #1981, opened 2021-01-13
Update model weights of midasnet

dvdhfnr:
We updated our model weights and would like to push those weights to the model zoo.

Done:
- updated model weights of midasnet (gdrive, same file id: https://drive.google.com/file/d/1Jf7qRG9N8IW8CaZ7gPisO5RtlLl63mNA)
- updated rmse, gflops, mparams for new weights
- updated sha256 and size of new model.pt

**TODO**: Please update the weights stored at https://download.01.org/opencv/public_models/032020/midasnet/model-001.pt with the new weights at https://drive.google.com/file/d/1Jf7qRG9N8IW8CaZ7gPisO5RtlLl63mNA

(I do not have access to this folder. As a reminder: You used this storage since a direct download from gdrive via the download script caused problems due to the file size.)

Thanks!


models/public/midasnet/model.yml
```diff
@@ -36,9 +36,9 @@ files:
     sha256: ba87812d0b5b247966471511d2d1a47fd6f492afb372bc49af6de3a46f4cf6c7
     source: https://raw.githubusercontent.com/intel-isl/MiDaS/ffb70fd13361434114383ce7eb898c2f5bec6176/models/blocks.py
   - name: model.pt
-    size: 422397766
-    sha256: 5237e007f53f05d0a66f40590abfead08dbb662629089e3462daf741ac9932d6
-    source: https://download.01.org/opencv/public_models/032020/midasnet/model-001.pt
+    size: 422509849
+    sha256: 933c46ec14e4041122852d8d0acf822ff9f4b94fd07f0ff42fa593f135d57cab
+    source: TODO PLEASE UPDATE https://download.01.org/opencv/public_models/032020/midasnet/model-001.pt WITH NEW VERSION OF https://drive.google.com/file/d/1Jf7qRG9N8IW8CaZ7gPisO5RtlLl63mNA
```
eizamaliev, line 41: You can use Google drive as file location. You just need replace this line with
```suggestion
    source:
      $type: google_drive
      id: 1Jf7qRG9N8IW8CaZ7gPisO5RtlLl63mNA
```

dvdhfnr: We had this previously and this created problems; see the note in [midasnet.md](https://github.com/openvinotoolkit/open_model_zoo/blob/617f34c7f05026a4523e1e10d519fb96cddb806c/models/public/midasnet/midasnet.md) as well as [this commit](https://github.com/openvinotoolkit/open_model_zoo/commit/1bc82be124dcd5e24822620e95b63a93b0ee5a81#diff-a331ee1f9b92e94f27c206566b4c8dd59f7ef238a3a7bb09425a5d8d6a804ae7).

The file size is quite big (> 400 MB). This causes the gdrive to (sometimes) ask for "Do you really want to download" or sth. like this, and this did not work with the download script for some users (including me).

IRDonch:
I'm a bit confused about this new model. I see that in the original repository there also is a new version of the model, but that is not the same file as the one. Meanwhile, this file doesn't seem to be referenced in that repository anywhere. So where does it come from, and why aren't you using the original file from the repository?

models/public/midasnet/midasnet.md
```diff
@@ -25,15 +25,15 @@ See [here](https://github.com/intel-isl/MiDaS)
 | Metric            | Value         |
 |-------------------|---------------|
 | Type              | Monodepth     |
-| GFLOPs            | 207.4915      |
-| MParams           | 104.0814      |
+| GFLOPs            | 207.25144     |
+| MParams           | 104.081       |
```
IRDonch, line 29: How could GFLOPS/MParams change if only the weights are different?

models/public/midasnet/midasnet.md
```diff
@@ -25,15 +25,15 @@ See [here](https://github.com/intel-isl/MiDaS)
...
 | Source framework  | PyTorch\*     |
 
 ## Accuracy
 
-| Metric | Value |
-| ------ | ----- |
-| rmse   | 7.5878|
+| Metric | Value   |
+| ------ | ------- |
+| rmse   | 0.07071 |
```
IRDonch, line 36: Note that we have recalculated the accuracy of the current model, and it's actually _better_ than this new value. So is the model even worth replacing?

models/public/midasnet/midasnet.md
dvdhfnr: I just checked it again with the current DL Workbench version and could those results.

models/public/midasnet/midasnet.md
dvdhfnr: This model improves upon the results of the old model in our large-scale evaluation.

models/public/midasnet/model.yml
dvdhfnr: We use the link to github now.

models/public/midasnet/midasnet.md
IRDonch: OK.

IRDonch:
Jenkins: test this please

IRDonch (approved):

PR #1996, opened 2021-01-15
Create data folder for dataset definitions and labels

jkamelin:


eaidova:
Jenkins please retry a build

eaidova:
@jkamelin, looks like demo tests should be also updated

Wovchena:
This may be counterintuitive: labels are in a single folder now, but https://github.com/openvinotoolkit/open_model_zoo/blob/develop/models/intel/bert-large-uncased-whole-word-masking-squad-0001/vocab.txt for bert models or https://github.com/openvinotoolkit/open_model_zoo/blob/develop/demos/python_demos/speech_recognition_demo/default_alphabet_example.conf are still separate.

I think @eaidova mentioned that before, but I don't know, if that layout is intentional thus pointing that out.

data/datasets_labels/cityscapes19.labels
```diff
@@ -0,0 +1,20 @@
+road
```
IRDonch, line 1: I have two suggestions for this directory:

1. I think we should adjust the terminology. In machine learning, labels are output variables corresponding to the data instances, and that's not what we have here. (For example, the _labels_ for ImageNet 2012 are in `val.txt`.) These are human-readable representations of classes represented by those labels, so I think this directory should be called `dataset_classes`.

2. The names of the files should be made uniform. I think just `<dataset>.txt` (or `<dataset>.json`, although IMO demos that take classes in JSON format should be updated to use the text format instead) makes the most sense.

demos/tests/run_tests.py
```diff
@@ -167,7 +167,7 @@ def main():
 
             with temp_dir_as_path() as temp_dir:
                 arg_context = ArgContext(
-                    source_dir=demos_dir / demo.subdirectory,
+                    source_dir=omz_dir / 'data/datasets_labels',
```
IRDonch, line 170: That's not right. `source_dir` should point to the demo's source directory (hence the name).

IRDonch (changes requested):

IRDonch:
GitHub refuses to let me comment on the `tools/accuracy_checker/dataset_definitions.yml` link, so I'll ask here: do we actually need it? Does anything depend on the old path?

eaidova:
> GitHub refuses to let me comment on the `tools/accuracy_checker/dataset_definitions.yml` link, so I'll ask here: do we actually need it? Does anything depend on the old path?

yes, dl workbench uses it and pot tests internally too, we need to give them time for migration

IRDonch:
> > GitHub refuses to let me comment on the `tools/accuracy_checker/dataset_definitions.yml` link, so I'll ask here: do we actually need it? Does anything depend on the old path?
> 
> yes, dl workbench uses it and pot tests internally too, we need to give them time for migration

Got it, thanks. In that case, the creation of the link should be done in a separate commit, so that Git can follow the history of the original file.

data/datasets_labels/cityscapes19.labels
vladimir-dudnik: not sure we need deep folder structure (although I'm ok with this).

@IRDonch support uniform naming, but I think replacing json with txt files probably better to do in separate PR after we finish with this one.

jkamelin:
> > > GitHub refuses to let me comment on the `tools/accuracy_checker/dataset_definitions.yml` link, so I'll ask here: do we actually need it? Does anything depend on the old path?
> > 
> > 
> > yes, dl workbench uses it and pot tests internally too, we need to give them time for migration
> 
> Got it, thanks. In that case, the creation of the link should be done in a separate commit, so that Git can follow the history of the original file.

Done

demos/tests/run_tests.py
jkamelin: done

data/datasets_labels/cityscapes19.labels
jkamelin: done

demos/python_demos/gesture_recognition_demo/README.md
```diff
@@ -77,7 +77,7 @@ python gesture_recognition_demo.py \
 -m_a /home/user/asl-recognition-0004.xml \
 -m_d /home/user/person-detection-asl-0001.xml \
 -i 0 \
--c ./msasl100-classes.json
+-c <omz_dir>/data/datasets_classes/msasl100-classes.json
```
Wovchena, line 80: There are other links below, though they point to nonexistent `./classes.json`. Can you update them please.

Wovchena (changes requested):
There are more of them :)
https://github.com/openvinotoolkit/open_model_zoo/blob/develop/models/intel/formula-recognition-medium-scan-0001/vocab.json
https://github.com/openvinotoolkit/open_model_zoo/blob/develop/models/intel/formula-recognition-polynomials-handwritten-0001/vocab.json
https://github.com/openvinotoolkit/open_model_zoo/tree/develop/demos/python_demos/handwritten_text_recognition_demo/data
https://github.com/openvinotoolkit/open_model_zoo/blob/develop/demos/python_demos/speech_recognition_demo/default_alphabet_example.conf

data/datasets_labels/cityscapes19.labels
IRDonch: > support uniform naming, but I think replacing json with txt files probably better to do in separate PR after we finish with this one.

Yes, I definitely agree.

jkamelin:
> There are more of them :)
> https://github.com/openvinotoolkit/open_model_zoo/blob/develop/models/intel/formula-recognition-medium-scan-0001/vocab.json
> https://github.com/openvinotoolkit/open_model_zoo/blob/develop/models/intel/formula-recognition-polynomials-handwritten-0001/vocab.json
> https://github.com/openvinotoolkit/open_model_zoo/tree/develop/demos/python_demos/handwritten_text_recognition_demo/data
> https://github.com/openvinotoolkit/open_model_zoo/blob/develop/demos/python_demos/speech_recognition_demo/default_alphabet_example.conf

Should https://github.com/openvinotoolkit/open_model_zoo/blob/develop/demos/python_demos/speech_recognition_demo/default_alphabet_example.conf also be moved? It looks like an example not a dataset

demos/python_demos/gesture_recognition_demo/README.md
jkamelin: done

IRDonch (changes requested):

IRDonch:
> > > > GitHub refuses to let me comment on the `tools/accuracy_checker/dataset_definitions.yml` link, so I'll ask here: do we actually need it? Does anything depend on the old path?
> > > 
> > > 
> > > yes, dl workbench uses it and pot tests internally too, we need to give them time for migration
> > 
> > 
> > Got it, thanks. In that case, the creation of the link should be done in a separate commit, so that Git can follow the history of the original file.
> 
> Done

Not done; you still have the move and the symlink creation in the same commit.

Wovchena (approved):
> Should https://github.com/openvinotoolkit/open_model_zoo/blob/develop/demos/python_demos/speech_recognition_demo/default_alphabet_example.conf also be moved? It looks like an example not a dataset

No, it shouldn't. You are right.

demos/tests/args.py
```diff
@@ -108,3 +108,11 @@ def __init__(self, file_name):
 
     def resolve(self, context):
         return str(context.source_dir / self.file_name)
+
+
+class DatasetClassesFileArg:
```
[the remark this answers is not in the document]
jkamelin: done

demos/tests/cases.py
```diff
@@ -17,7 +17,7 @@
 import sys
 
 from args import (
-    DataDirectoryArg, DataDirectoryOrigFileNamesArg, DataPatternArg, DemoFileArg,
```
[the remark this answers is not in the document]
jkamelin: done

IRDonch:
I have thought about the vocab files, and IMO, they shouldn't be moved, or at least they shouldn't be moved like this:

* The vocabularies are by no means descriptions of classes in the datasets (which don't _have_ classes to begin with). More broadly, they aren't properties of the dataset at all; they are, essentially, _models_ trained on the dataset.
* The file you called `squad_v1.1_vocab.txt` wasn't trained on SQuAD. "im2latex" isn't a dataset at all.

I think the way these files are stored right now is acceptable (even if the duplication is annoying). We could potentially remove the duplication by packaging them as standalone models, but that would be a lot of effort (as they are conceptually different, not being deep learning-based), and it just doesn't seem worthwhile.

Wovchena:
> The vocabularies are by no means descriptions of classes in the datasets

The vocabularies are derived from datasets: they contain tokens that were in a used dataset.

> More broadly, they aren't properties of the dataset at all; they are, essentially, models trained on the dataset.

A neural net or a language model can be trained, but a mapping from id to token isn't usually trained. I agree that vocab files are closer to a model than to a dataset, but the same is about classification or detection tasks: we use classes files to interpret what model predicts in first place and then as a next step it is possible to map values with ground truth. Or in reverse order: encode ground truth with classes to **_match the model_** output.

> (which don't have classes to begin with)

Reformulating the task as classification problem at a given timestamp makes classes appear.

> I think the way these files are stored right now is acceptable

I still believe this may feel counterintuitive to store classes in a single folder and to store vocabs, which are any way similar conceptually to classes, in a different place.

Though it all don't help to make a decision :)

IRDonch:
> The vocabularies are derived from datasets: they contain tokens that were in a used dataset.

The dataset in question doesn't consist of tokens - it consists of text. The set of tokens used is the result of the tokenizer training algorithm - it's not intrinsic to the dataset. A different tokenizer (or the same tokenizer with different parameters) would yield a different set of tokens.

> I agree that vocab files are closer to a model than to a dataset, but the same is about classification or detection tasks: we use classes files to interpret what model predicts in first place and then as a next step it is possible to map values with ground truth.

I think the core difference is that classes are intrinsic to the dataset and aren't dependent on the training process.

> Reformulating the task as classification problem at a given timestamp makes classes appear.

What timestamp? We're talking about text processing. :-)

> I still believe this may feel counterintuitive to store classes in a single folder and to store vocabs, which are any way similar conceptually to classes, in a different place.

I can understand that, and I'm starting to feel like we should add a symlink to the proper class file to each model subdirectory (where it's relevant). I think that's an effort for another PR, though.

vladimir-dudnik (approved):

jkamelin:
> > The vocabularies are derived from datasets: they contain tokens that were in a used dataset.
> 
> The dataset in question doesn't consist of tokens - it consists of text. The set of tokens used is the result of the tokenizer training algorithm - it's not intrinsic to the dataset. A different tokenizer (or the same tokenizer with different parameters) would yield a different set of tokens.
> 
> > I agree that vocab files are closer to a model than to a dataset, but the same is about classification or detection tasks: we use classes files to interpret what model predicts in first place and then as a next step it is possible to map values with ground truth.
> 
> I think the core difference is that classes are intrinsic to the dataset and aren't dependent on the training process.
> 
> > Reformulating the task as classification problem at a given timestamp makes classes appear.
> 
> What timestamp? We're talking about text processing. :-)
> 
> > I still believe this may feel counterintuitive to store classes in a single folder and to store vocabs, which are any way similar conceptually to classes, in a different place.
> 
> I can understand that, and I'm starting to feel like we should add a symlink to the proper class file to each model subdirectory (where it's relevant). I think that's an effort for another PR, though.

Moved vocab files back to models directories

IRDonch:
Looks mostly good, but a few more changes are needed:

* `datasets_classes` should be renamed to `dataset_classes`;
* `classes` and `char_list` should be removed from file names;
* `tools/accuracy_checker/configs/README.md` links to `datasets.md` and should be updated.

IRDonch:
You commit history is still wrong when it comes to dataset_definitions.yml: you have the creation at the new path in one commit, and the deletion and the symlink creation in the other commit. The creation and the deletion need to be in the same commit, and the symlink creation in a different commit.

IRDonch (approved):

IRDonch:
@eaidova Any comments?

eaidova (approved):

PR #2014, opened 2021-01-22
Rename some demo directories to improve consistency

IRDonch:
All demo directories should end with `_demo` (possibly followed by `_<model>`).


Wovchena (approved):
I would prefer removing `_demo` suffix from everywhere instead of adding it, but ok :)

eizamaliev (approved):

PR #1334, opened 2020-07-17
add labels for doxygen

ntyukaev:

PR #2023, opened 2021-01-26
Add a script to prepare documentation for inclusion into OpenVINO docs

IRDonch:
The layout file and page IDs generated by the script were previously maintained by the OpenVINO documentation team. Having them generated automatically lifts that burden off their shoulders and also makes it possible for us to autogenerate other aspects of the documentation in the future.

In the generated layout file, the titles of some pages are overridden. This is so that they match the titles in the table of contents from the previous releases of OpenVINO. Ideally, we should align the titles in the layout with the titles in the pages themselves, but I'm not yet sure which ones are better, so I'm deferring that for later.

About the requirements file: I used a pretty generic name for it (requirements-documentation), because I expect that other
documentation-related scripts (e.g. a model index checker) might be added later. I'll add a version with pinned dependencies
(`requirements-documentation.txt`) when I integrate this script into CI.

### DIFF
--- a/ci/prepare-documentation.py
+++ b/ci/prepare-documentation.py
@@ -1,0 +1,304 @@
+#!/usr/bin/env python3
+
+# Copyright (c) 2021 Intel Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+'''
+This script prepares the OMZ documentation for inclusion into the OpenVINO
+documentation. This currently involves two things:
+
+* Inserting Doxygen page IDs into all pages. Without them, Doxygen autogenerates
+  ID based on the Markdown file paths. The advantages of explicit IDs are that:
+
+  a) the Doxygen page URLs look nicer;
+  b) if we decide to change the layout of the documentation files, we can
+     do so without affecting the page IDs (by updating this script so that it
+     assigns the same IDs as before), thus not breaking any documentation links.
+
+* Generating a Doxygen layout file.
+
+The script also runs some basic checks on the documentation contents.
+
+Install dependencies from requirements-documentation.in to use this script.
+'''
+
+import argparse
+import logging
+import re
+import shutil
+import urllib.parse
+import urllib.request
+import xml.etree.ElementTree as ET
+
+from pathlib import Path
+
+import mistune
+import yaml
+
+OMZ_ROOT = Path(__file__).resolve().parents[1]
+
+# For most task types, a simple transformation yields a human-readable
+# English description, but for a few types it's useful to override it
+# to improve consistency and grammar.
+HUMAN_READABLE_TASK_TYPES = {
+    'detection': 'Object Detection',
+    'object_attributes': 'Object Attribute Estimation',
+    'text_to_speech': 'Text-to-speech',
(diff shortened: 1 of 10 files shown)


ci/prepare-documentation.py
```diff
@@ -0,0 +1,304 @@
...
+import urllib.parse
+import urllib.request
+import xml.etree.ElementTree as ET
+
+from pathlib import Path
+
+import mistune
+import yaml
+
+OMZ_ROOT = Path(__file__).resolve().parent.parent
```
eaidova, line 49:
```suggestion
OMZ_ROOT = Path(__file__).resolve().parents[1]
```

IRDonch:
CC @azaytsev1980 @ntyukaev 

Obsoletes #1334.

The generated layout file is based on https://github.com/ntyukaev/open_model_zoo/blob/releases/2021/2/omz_docs.xml, although it's not a 1:1 copy.

@eaidova I used some heuristics to generate IDs for Accuracy Checker documentation pages, but the results still aren't stellar (e.g. `omz_tools_accuracy_checker_accuracy_checker_launcher_pytorch_launcher`). Any ideas on how to make them better? Maybe we could move all AC docs to a separate `docs` directory, so that we can just base the page ID on the file name?

eaidova:
> @eaidova I used some heuristics to generate IDs for Accuracy Checker documentation pages, but the results still aren't stellar (e.g. `omz_tools_accuracy_checker_accuracy_checker_launcher_pytorch_launcher`). Any ideas on how to make them better? Maybe we could move all AC docs to a separate `docs` directory, so that we can just base the page ID on the file name?

@IRDonch Do you suggest to move them in the repo or during docs generation?
I believe the second way will not work because docs can be linked with each other.
The first way, maybe inconvenient for user to have info in the separated directory (it will be hidden from view, e.g. sample should be executed in the specific directory, moving its instruction may lead to incorrect result or if we move instruction how to install, it  may not be obvious to search in docs for newcomers), but we can discuss internally this way. It may be useful for idea extending AC docs with troubleshooting guide and tool extending guide, which will be ready soon. 

If we can continue with heuristics:
I believe we ca remove repeated part (accuracy_checker) from index. All directoires contains only one readme file, except launchers. In the same time, there is also difference in file naming in this directory and filename includes related subject, possibly we can also remove parent directory for them.
To sum up, your provided example will be looks like: 
`omz_tools_accuracy_checker_pytorch_launcher`
any other file out of this directory (e.g. readers)
omz_tools_accuracy_checker_data_readers

IRDonch:
> Do you suggest to move them in the repo or during docs generation?

The former.

> The first way, maybe inconvenient for user to have info in the separated directory (it will be hidden from view, e.g. sample should be executed in the specific directory, moving its instruction may lead to incorrect result or if we move instruction how to install, it may not be obvious to search in docs for newcomers), but we can discuss internally this way.

🤷 I think having a `docs` directory with all the docs is quite ordinary. Also, you need to keep in mind that these docs are published on `docs.openvinotoolkit.org`, where the location of the original Markdown file is invisible. So the docs have to be clear about which directory commands must be executed in, anyway.

> I believe we ca remove repeated part (accuracy_checker) from index. All directoires contains only one readme file, except launchers. In the same time, there is also difference in file naming in this directory and filename includes related subject, possibly we can also remove parent directory for them.

Am I understanding correctly that you're proposing to take the currently-implemented ID scheme, but strip all intermediate directories from it? If so, that's fine by me, as long as you can commit to not create ID conflicts (by having multiple directories with the same name). This would also make it easy to later migrate to having a dedicated `docs` directory if we so choose.

ci/prepare-documentation.py
IRDonch: Applied.

IRDonch:
> Am I understanding correctly that you're proposing to take the currently-implemented ID scheme, but strip all intermediate directories from it?

I have updated the branch to use this ID scheme.

eaidova (approved):
yes, this what I mean. I can guarantee that no directories with the same name in AC.

Regarding docs directory, it makes sense, possibly later, we can migrate on such structure

eizamaliev (approved):

PR #2037, opened 2021-01-29
regnetx-3.2gf: use the weights file that we download

IRDonch:
Previously, it wasn't used in any way; the model constructor just downloaded the weights anew.


IRDonch:
CC @akorobeinikov

eaidova (approved):

PR #2039, opened 2021-01-29
speech_recognition_demo: remove unnecessary code from ctcdecode_numpy

IRDonch:
This has two main parts:

* Removing unnecessary includes and directives from `decoders.i`, so that we don't generate bindings that are never used.

* Removing the `numpy_beam_decode_no_lm` function, which is useless: it's equivalent to `numpy_beam_decode` with `scorer` = `nullptr`, so in Python one can just call `numpy_beam_decode` with `scorer=None`.

Altogether, this removes over 3000 lines of code.


Wovchena (approved):
Is there any reason other than speeding up the compilation for this?

IRDonch:
> Is there any reason other than speeding up the compilation for this?

Less code, fewer problems. 🤷 

(But also, I was looking into eliminating the narrowing conversion warnings from the generated code, and it'll be easier to do the less code there is.)

Wovchena:
I think it's better to focus on new code rather then on one we already have. May be setting stricter complication keys for all demos or enabling code style checks / static analyzers is a better task so we could simplify code reviews.

PR #2049, opened 2021-02-02
AC: supports DETR

eaidova:


tools/accuracy_checker/accuracy_checker/adapters/__init__.py
```diff
@@ -52,7 +52,8 @@
     FaceDetectionAdapter,
     FaceDetectionRefinementAdapter,
     FasterRCNNONNX,
-    TwoStageDetector
+    TwoStageDetector,
+    DETRAdapter
```
Anna-Mironova, line 56: Maybe we need to add DETRAdapter to __all__ list

eaidova: already fixed

Anna-Mironova (approved):

PR #2051, opened 2021-02-03
AC: supports YoloV5

akorobeinikov:


eaidova (approved):

tools/accuracy_checker/accuracy_checker/adapters/yolo.py
```diff
@@ -563,3 +563,91 @@ def postprocess_boxes(self, pred_bbox, org_img_shape, input_size):
...
+
+
+class YoloV5Adapter(YoloV3Adapter):
+    __provider__ = 'yolo_v5'
+
+    @classmethod
+    def parameters(cls):
+        params = super().parameters()
+        params.update({
+            'transpose': ListField(optional=True, description="Transpose output tensor to specified format.")
```
eaidova, line 575: Do we really need additional adapter for yolo v5?
Or this parameter can be integrated in yolo v3 adapter directly?

akorobeinikov: Yes, it can. If it be better, I'll integrate this parameter directly to yolo v3.

eaidova: I think, have separated adapter is ok, but potentially transpose can be also usefull for other models (I see simmilar behaviour with yolo v3 in tf2).
So if it is possible, let's integrate transpose to v3 and keep this adapter too (if It makes sense)

akorobeinikov: Ok

eaidova (approved):

PR #2054, opened 2021-02-03
speech_recognition: support QuartzNet

Wovchena:


Wovchena:
@eaidova, this `audio_to_melspectrum()` gives 4.193928788800974 WER. I don't remember what the value was for you, just informing.

demos/speech_recognition_demo/python/speech_recognition_demo.py
```diff
@@ -27,26 +26,24 @@ def build_argparser():
                              "The sample will look for a suitable IE plugin for this device. (default is CPU)")
     parser.add_argument('-m', '--model', type=str, metavar="FILENAME", required=True,
                         help="Path to an .xml file with a trained model (required)")
+    parser.add_argument('-m_decoder', '--model_decoder', type=str, metavar="FILENAME",
+                        help="Optional. Path to an .xml QuartzNet decoder. Makes treat --model as QuartzNet encoder")
     parser.add_argument('-L', '--lm', type=str, metavar="FILENAME",
-                        help="path to language model file (optional)")
-    parser.add_argument('-p', '--profile', type=str, metavar="NAME", required=True,
+                        help="path to language model file (optional). Ignored if --model_decoder is set")
```
IRDonch, line 32: You shouldn't ignore it; you should make sure it's _not_ specified if it's not appropriate for the current model.

demos/speech_recognition_demo/python/speech_recognition_demo.py
```diff
@@ -56,55 +53,66 @@ def get_profile(profile_name):
...
-    print("Overall time: {} s".format(timeit.default_timer() - start_time))
-
-    print("\nTranscription and confidence score:")
-    for candidate in transcription:
-        print(
-            "{}\t{}".format(
-                candidate['conf'],
-                candidate['text'],
+        character_probs = stt.extract_per_frame_probs(audio_features)
+    print("Infernce time: {} s".format(timer.elapsed))
```
IRDonch, line 93:
```suggestion
    print("Inference time: {} s".format(timer.elapsed))
```

demos/speech_recognition_demo/python/utils/quartz_net.py
```diff
@@ -0,0 +1,77 @@
...
+class QuartzNet:
+    pad_to = 16
+    alphabet = " abcdefghijklmnopqrstuvwxyz'"
+
+    def __init__(self, encoder_path, decoder_path, input_shape, device):
+        assert not input_shape[2] % self.pad_to, f"{self.pad_to} must be a divisor of input_shape's third dimension"
+        self.ie = IECore()
+        encoder_net = self.ie.read_network(encoder_path)
+        if len(encoder_net.input_info) != 1:
+            raise RuntimeError('QuratzNet encoder must have one input')
```
IRDonch, line 16:
```suggestion
            raise RuntimeError('QuartzNet encoder must have one input')
```

IRDonch (changes requested):
Where's the model itself?

Wovchena:
> Where's the model itself?

It's not there yet. I used one from benchmark.

demos/speech_recognition_demo/python/speech_recognition_demo.py
```diff
@@ -25,28 +24,35 @@ def build_argparser():
...
-                        help="Beam width for beam search in CTC decoder (default 500)")
-    parser.add_argument('-c', '--max-candidates', type=int, default=1, metavar="N",
-                        help="Show top N (or less) candidates (default 1)")
-
-    parser.add_argument('-l', '--cpu_extension', type=str, metavar="FILENAME",
-                        help="Optional. Required for CPU custom layers. "
-                             "MKLDNN (CPU)-targeted custom layers. Absolute path to a shared library with the"
-                             " kernels implementations.")
+
+    architecture_type_subparsers = parser.add_subparsers(dest='architecture_type', required=True,
```
IRDonch, line 28: You've probably seen me say this in the MelGAN demo PR, but I don't think subcommands are the best way to implement this. Especially since we already do have a pattern for selecting the architecture type - an `-architecture_type` option.

IRDonch (changes requested):

demos/speech_recognition_demo/python/speech_recognition_demo.py
Wovchena: `-architecture_type` is redundant there because it would follow `bool(args.model_decoder)`. Updated

IRDonch (approved):
LGTM, but I'd like to see the model itself before merging this. It seems to me that it should be possible to merge the two parts of the model into one, and if we do that, it will affect the demo.

demos/README.md
```diff
@@ -43,6 +43,7 @@ The Open Model Zoo includes the following demos:
 - [Pedestrian Tracker C++ Demo](./pedestrian_tracker_demo/cpp/README.md) - Demo application for pedestrian tracking scenario.
 - [Security Barrier Camera C++ Demo](./security_barrier_camera_demo/cpp/README.md) - Vehicle Detection followed by the Vehicle Attributes and License-Plate Recognition, supports images/video and camera inputs.
 - [Speech Recognition Python\* Demo](./speech_recognition_demo/python/README.md) - Speech recognition demo: takes audio file with an English phrase on input, and converts it into text.
+- [Speech Recognition Offline Python\* Demo](./speech_recognition_offline_demo/python/README.md) - Offline speech recognition demo: takes a whole audio file with an English phrase on input, and converts it into text.
```
IRDonch, line 46: IMO, this doesn't sufficiently explain the difference between online and offline. Maybe the description of the online demo should be amended to clarify that it does the conversion chunk-by-chunk.

demos/speech_recognition_offline_demo/python/README.md
```diff
@@ -0,0 +1,39 @@
+# Speech Recognition Offline Demo
+
+This demo demonstrates Automatic Speech Recognition (ASR) with pretrained QuartzNet model.
+
+## How It Works
+
+After computing audio features, running a neural network to get haracter probabilities, and CTC greedy decoding, the demo prints the decoded text together.
```
IRDonch, line 7:
```suggestion
After computing audio features, running a neural network to get character probabilities, and CTC greedy decoding, the demo prints the decoded text together.
```

Also, together... with what?

demos/speech_recognition_offline_demo/python/speech_recognition_offline_demo.py
```diff
@@ -0,0 +1,113 @@
...
+            raise RuntimeError("QuartzNet input second dimension can't be reshaped")
+        if encoder_input_shape[2] % self.pad_to:
+            raise RuntimeError(f'{self.pad_to} must be a divisor of QuartzNet input third dimension')
+        if len(net.outputs) != 1:
+            raise RuntimeError('QuartzNet must have one output')
+        decoder_output_shape = next(iter(net.outputs.values())).shape
+        if len(decoder_output_shape) != 3:
+            raise RuntimeError('QuartzNet decoder output must be 3-dimensional')
+        if decoder_output_shape[2] != len(self.alphabet) + 1:  # +1 for blank char
+            raise RuntimeError('QuartzNet decoder second dimension must be similar to a length of an alphabet assumed')
```
IRDonch, line 50: This doesn't mean anything to the user. What is "an alphabet assumed" and what does "similar" mean?

demos/speech_recognition_offline_demo/python/speech_recognition_offline_demo.py
```diff
@@ -0,0 +1,113 @@
...
+                        help="Optional. Specify the target device to infer on, for example: CPU, GPU, FPGA, HDDL, MYRIAD or HETERO. "
+                             "The sample will look for a suitable IE plugin for this device. Default value is CPU.")
+    return parser
+
+
+def main():
+    args = build_argparser().parse_args()
+
+    with wave.open(args.input, 'rb') as wave_read:
+        channel_num, sample_width, sampling_rate, pcm_length, compression_type, _ = wave_read.getparams()
```
IRDonch, line 100: The demo description claims that only 16 KHz audio is supported, but there's no check on the sample rate here.

models/public/quartznet-15x5-en/model.yml
```diff
@@ -80,6 +80,10 @@ postprocessing:
     file: models/nemo/collections/asr/__init__.py
     pattern: 'JasperEncoder\,'
     replacement: 'JasperEncoder)'
+  - $type: regex_replace  # Windows build lacks np.float128
```
IRDonch, line 83: Windows build of what?

IRDonch (changes requested):

demos/README.md
Wovchena: Currently `speech_recognition_demo` processes a whole file. Your statement will become true after https://github.com/openvinotoolkit/open_model_zoo/pull/2108 merge. I think it should be rephrased along with `speech_recognition_demo` renaming and be done in https://github.com/openvinotoolkit/open_model_zoo/pull/2108.

IRDonch: Okay, fair enough.

demos/speech_recognition_offline_demo/python/speech_recognition_offline_demo.py
IRDonch: The new version is better, but the user still wouldn't know what the expected value is. Why not just use the number?

Wovchena: Updated

IRDonch (approved):

eaidova:
Jenkins please retry a build

demos/speech_recognition_offline_demo/python/speech_recognition_offline_demo.py
```diff
@@ -0,0 +1,123 @@
...
+
+# Workaround to import librosa on Linux without installed libsndfile.so
+try:
+    import librosa
+except OSError:
+    import sys
+    import types
+    sys.modules['soundfile'] = types.ModuleType('fake_soundfile')
+    import librosa
+    del sys.modules['soundfile']
```
IRDonch, line 26: I don't think you're accomplishing anything by deleting it. If librosa imports it, then it will retain a reference to it, so the module object will stick around anyway. You might as well keep the program state consistent by retaining it in `sys.modules`.

Wovchena: Removed

PR #2055, opened 2021-02-03
ci/prepare-documentation.py: remove the hardcoded task type list

IRDonch:
I originally added the list so that if a model with a new task type is added, the prepare-documentation script crashes and so the contributor has to consider what the human-readable description of the task type should be.

However, given that:

1. for almost all task types, the default string transformation yields a good description, and

2. the cost of getting the description wrong is just an awkwardly-named section in the documentation TOC,

I don't think that the cost of keeping an extra copy of the task type list is justified. Instead we should just instruct contributors to name task types so that the default transformation gives a good result, or, if that's not possible, to add an override.

### DIFF
--- a/ci/prepare-documentation.py
+++ b/ci/prepare-documentation.py
@@ -52,36 +52,15 @@ OMZ_ROOT = Path(__file__).resolve().parents[1]
 
 XML_ID_ATTRIBUTE = '{http://www.w3.org/XML/1998/namespace}id'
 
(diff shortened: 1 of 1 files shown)


IRDonch:
Jenkins please retry a build

IRDonch:
@eizamaliev Ping?

eizamaliev (approved):

PR #2059, opened 2021-02-04
Support PyTorch1.7 conversion for CoCosNet and MidasNet models

akorobeinikov:


models/public/midasnet/model.yml
```diff
@@ -49,6 +49,11 @@ postprocessing:
     file: models/blocks.py
     pattern: 'torch.hub.load\("facebookresearch/WSL-Images", "resnext101_32x8d_wsl"\)'
     replacement: 'models.resnext101_32x8d()'
+  # Replace Squeeze with Gather operation for support PyTorch >= 1.7
```
IRDonch, line 52: It's just 1.7, AFAIK. They fixed the Squeeze export in 1.8.

models/public/cocosnet/model.yml
```diff
@@ -97,11 +97,11 @@ postprocessing:
     file: model_files/models/networks/correspondence.py
     pattern: '/ self.opt.down\)'
     replacement: '// self.opt.down)'
-  # MO can only convert Squeeze with fixed dimensions
+  # Replace Squeeze with Gather operation for support PyTorch >= 1.7
```
IRDonch, line 100: I don't think you should just replace the comment like that; the original comment explains _why_ Squeeze is bad in this case.

IRDonch (commented):
I guess this works, but are we in a rush to support 1.7? Personally, I would rather wait for 1.8, because then we can actually _remove_ one of these workarounds.

eaidova:
@akorobeinikov I believe we can close this request as pytorch1.8 solves all issues

PR #2065, opened 2021-02-05
AC: fix annotation reloading

eaidova:


tools/accuracy_checker/accuracy_checker/evaluators/quantization_model_evaluator.py
```diff
@@ -194,7 +194,7 @@ def register_dumped_annotations(self):
         if not self._dumped_annotations:
             return
         meta = Dataset.load_meta(self.dataset.dataset_config)
-        self.dataset.annotation_provider = AnnotationProvider(self._dumped_annotations, meta)
+        self.dataset.set_annotation(annotation, meta)
```
Anna-Mironova, line 197: may be something like this:

```suggestion
        self.dataset.set_annotation(self._dumped_annotations, meta)
```

Anna-Mironova (approved):

PR #2070, opened 2021-02-07
add aclnet-int8 model

vladimir-dudnik:
aclnet quantized and finetuned with NNCF

- [x] download link
- [x] changes in converter to get model IR in FP32-INT8/FP16-INT8 folders (this model should not have FP32/FP16 folders and IRs)
- [x] models GFlops and MParams metrics


vladimir-dudnik:
CC @sathishkuttan

tools/downloader/converter.py
```diff
@@ -160,10 +160,11 @@ def convert(reporter, model):
             for arg in model.mo_args]
 
         for model_precision in sorted(model_precisions):
+            dir_precision = model_precision + '-INT8' if model.quantized else model_precision
```
IRDonch, line 163: This is not the right place for this. `model_precision` should already equal `FP32-INT8`/`FP16-INT8`. What you need to do is change the code in `common.py` that sets `model.precisions` so that it works appropriately for quantized models. That way:

* info dumper's output will be correct;
* Converter options like `--precisions=FP32-INT8` will work correctly.

IRDonch (changes requested):

tools/downloader/converter.py
vladimir-dudnik: model_precision is also used as MO parameter '--data_type={}'.format(model_precision), and have to be FP32/16, other variants MO do not accept

IRDonch: You can strip off the `-INT8` part when running MO.

vladimir-dudnik: what is the difference between adding `-INT8` to folder name and stripping it out of MO parameter?

IRDonch: The difference is that the latter preserves the property that a converted model is placed in the subdirectory with the name of the target precision. Running the converter with `--precision=FP32` should place the model in the `FP32` subdirectory, while running it with `--precision=FP32-INT8` should place it into `FP32-INT8`.

vladimir-dudnik: ok, changed

vladimir-dudnik:
Jenkins please retry a build

tools/downloader/common.py
```diff
@@ -537,6 +538,8 @@ def deserialize(cls, model, name, subdirectory):
 
                 precisions = set(files_per_precision.keys())
 
+            quantized = model.get('quantized', None)
```
IRDonch, line 541: Please validate the field: it should be either `"INT8"` or `None`. Also, check that if this is specified, then `model_optimizer_args` is also specified.

tools/downloader/common.py
```diff
@@ -454,18 +454,19 @@ def apply(self, reporter, output_dir):
...
 class Model:
-    def __init__(self, name, subdirectory, files, postprocessing, mo_args, quantizable, framework,
+    def __init__(self, name, subdirectory, files, postprocessing, mo_args, quantized, quantizable, framework,
                  description, license_url, precisions, task_type, conversion_to_onnx_args):
         self.name = name
         self.subdirectory = subdirectory
         self.files = files
         self.postprocessing = postprocessing
         self.mo_args = mo_args
+        self.quantized = quantized
```
IRDonch, line 464: I don't think this field needs to be saved; nothing uses it afterwards.

tools/downloader/common.py
```diff
@@ -454,18 +454,19 @@ def apply(self, reporter, output_dir):
...
         self.quantizable = quantizable
         self.framework = framework
         self.description = description
         self.license_url = license_url
-        self.precisions = precisions
+        self.precisions = {'FP32-INT8', 'FP16-INT8'} if self.quantized else precisions
```
IRDonch, line 469: This logic should be in `deserialize`, not here. Also, I would suggest to make it more flexible by using, e.g. `precisions = {f'{precision}-{quantized}' for precision in precisions}`.

IRDonch (changes requested):

tools/downloader/converter.py
```diff
@@ -160,9 +161,10 @@ def convert(reporter, model):
             for arg in model.mo_args]
 
         for model_precision in sorted(model_precisions):
+            data_type = functools.reduce(lambda a, b: a.replace(b, ''), '-INT8', model_precision)
```
IRDonch, line 164:
```suggestion
            data_type = model_precision.split('-')[0]
```

tools/downloader/common.py
vladimir-dudnik: validation added (also move getting values a bit above)

tools/downloader/common.py
vladimir-dudnik: removed

tools/downloader/common.py
vladimir-dudnik: logic moved to deserialize. did not get how to make it more flexible (getting error: local variable 'precisions' referenced before assignment)

tools/downloader/common.py
```diff
@@ -454,7 +454,7 @@ def apply(self, reporter, output_dir):
 Postproc.types['unpack_archive'] = PostprocUnpackArchive
 
 class Model:
-    def __init__(self, name, subdirectory, files, postprocessing, mo_args, quantizable, framework,
+    def __init__(self, name, subdirectory, files, postprocessing, mo_args, quantized, quantizable, framework,
```
IRDonch, line 457: Unused parameter.

tools/downloader/common.py
```diff
@@ -507,11 +507,14 @@ def deserialize(cls, model, name, subdirectory):
                 if conversion_to_onnx_args:
                     raise DeserializationError('Conversion to ONNX not supported for "{}" framework'.format(framework))
 
+            quantized = model.get('quantized', None)
+            if quantized is not None and quantized != 'INT8':
+                raise DeserializationError('"quantized": expected a INT8, got {!r}'.format(quantized))
```
IRDonch, line 512:
```suggestion
                raise DeserializationError('"quantized": expected "INT8", got {!r}'.format(quantized))
```

tools/downloader/common.py
```diff
@@ -507,11 +507,14 @@ def deserialize(cls, model, name, subdirectory):
...
+
             if 'model_optimizer_args' in model:
                 mo_args = [validate_string('"model_optimizer_args" #{}'.format(i), arg)
                     for i, arg in enumerate(model['model_optimizer_args'])]
-
-                precisions = {'FP16', 'FP32'}
+                precisions = {'FP16-INT8', 'FP32-INT8'} if quantized is not None else {'FP16', 'FP32'}
```
IRDonch, line 517:
```suggestion
                precisions = {f'FP16-{quantized}', f'FP32-{quantized}'} if quantized is not None else {'FP16', 'FP32'}
```

IRDonch (changes requested):

models/public/aclnet-int8/labels.txt
```diff
@@ -0,0 +1,53 @@
+Dog
```
eaidova, line 1: I think labels should be places in the same directory with other labels (data directory)

tools/downloader/common.py
vladimir-dudnik: right, removed

models/public/aclnet-int8/labels.txt
vladimir-dudnik: agree, moved to data subfolder

models/public/aclnet-int8/aclnet-int8.md
```diff
@@ -0,0 +1,67 @@
...
+
+- `N` - batch size
+- `C` - channel
+- `L` - number of PCM samples (minimum value is 16000)
+
+## Output
+
+### Original Model
+
+Sound classifier (see [labels](./labels.txt)), name - `486`, shape - `1,53`, output data format is `N,C` where:
```
eaidova, line 53:
```suggestion
Sound classifier (see [labels](../../../data/dataset_classes/aclnet_classes.txt)), name - `486`, shape - `1,53`, output data format is `N,C` where:
```

models/public/aclnet-int8/aclnet-int8.md
```diff
@@ -0,0 +1,67 @@
...
+
+- `N` - batch size
+- `C` - Predicted softmax scores for each class in [0, 1] range
+
+### Converted Model
+
+Sound classifier (see [labels](./labels.txt)), name - `486`, shape - `1,53`, output data format is `N,C` where:
```
eaidova, line 60: and here also link should be updated

data/dataset_classes/aclnet_classes.txt
```diff
@@ -0,0 +1,53 @@
+Dog
```
IRDonch, line 1: Should just be called `aclnet.txt`.

models/public/aclnet-int8/aclnet-int8.md
IRDonch: I would actually suggest to write `OMZ_ROOT/data/dataset_classes/...`. Links to non-Markdown files don't work in the published documentation.

models/public/aclnet-int8/aclnet-int8.md
vladimir-dudnik: uodated

data/dataset_classes/aclnet_classes.txt
vladimir-dudnik: renamed

models/public/aclnet-int8/aclnet-int8.md
vladimir-dudnik: this should be done globally for OMZ, so better to do in separate PR

eaidova (approved):

IRDonch (approved):

PR #2074, opened 2021-02-08
AC: allow use set_partial_shape for resolving input shapes

eaidova:


eaidova:
Jenkins please retry a build

PR #2088, opened 2021-02-11
image_retrieval, place_recognition demos: fix the opening of VideoWriter

anzhella-pankratova:
+Deleted unused code in `segmentation.py`.

### DIFF
--- a/demos/image_retrieval_demo/python/image_retrieval_demo.py
+++ b/demos/image_retrieval_demo/python/image_retrieval_demo.py
@@ -134,7 +134,6 @@ def main():
 
     frames_processed = 0
     presenter = monitors.Presenter(args.utilization_monitors, 0)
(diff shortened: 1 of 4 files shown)


demos/image_retrieval_demo/python/image_retrieval_demo.py
```diff
@@ -134,7 +134,6 @@ def main():
 
     frames_processed = 0
     presenter = monitors.Presenter(args.utilization_monitors, 0)
-    video_writer = cv2.VideoWriter()
```
Wovchena, line 137: Why move this? For `place_recognition_demo.py` it stays there.

anzhella-pankratova: Updated.

Wovchena (approved):

PR #2093, opened 2021-02-12
Object Detection Demo Jupyter Notebook

helena-intel:
Jupyter Notebook for Async Object Detection. I took the Python demo and made it into a notebook that allows a user to select a model and a video (or upload a video of their own), and set settings for `max_num_requests`, `num_threads` and `num_streams`. 

The notebook downloads the model, optionally converts it, does inference, and shows the results. It shows "live" results of a selected model on one video, and on three frames of multiple selected models (to quickly compare results). By default currently only Intel models are used because they do not require using the model converter, so this notebook can be used with the pip version of OpenVINO. I can update this when the model optimizer becomes part of the pip distribution.

The target audience for this notebook includes less experienced developers. The goal of this specific notebook is to demonstrate Object Detection with the Async API. I will also make more introductionary notebooks for using OpenVINO/OMZ in general.

### Try without installing

You can try it here: https://mybinder.org/v2/gh/helena-intel/open_model_zoo/object-detection-demo-notebook-binder?filepath=%2Fdemos%2Fobject_detection_demo%2Fjupyter-python%2Fobject_detection_demo.ipynb There is also a version where the code is hidden: https://mybinder.org/v2/gh/helena-intel/open_model_zoo/detection-async-notebook?urlpath=voila%2Frender%2Fobject_detection_demo.ipynb  Note that performance may be slow and it may randomly crash. 

### README/Usage notes

The README instructions need to be updated, but I wanted to wait for the updated pip install method to do that. If you already have OpenVINO installed you can run the notebook from a terminal/cmd.exe where you have set the PATH/LD_LIBRARY_PATH or run setupvars. Then install the requirements in the demos/object_detection_demo/jupyter-python folder and run the notebook, with jupyter notebook or jupyter lab. Note: the notebook calls subprocess to run the model optimizer. If you use a virtualenv, this only works if your environment variables are set globally. If you set them in the shell before running the notebook, subprocess doesn't see them. With the new pip release this should no longer be an issue (otherwise I can change the subprocess call). 

Tested on Windows and Linux (Ubuntu 18). 

### Notes and questions

- I am relatively new to OpenVINO and OMZ and am not sure if I follow best practices. All feedback is welcome!
- I would like to include a bit more information on the options for `num_threads`, `num_streams `and `max_num_requests` but I don't know enough about this to write this.
- I got the intro text from the Python demo, but the target audience for the notebook demo's is a bit different and the intro text could be a bit more catchy. Suggestions are welcome.
- I put some code in a `detection_utils` file and some directly in the notebook. I tried to find a balance, so that the important code for this demo is still in the notebook, but that it is not too much code. I am not sure if I found the right balance.
- I used ipywidgets so that there are dropdowns to select a model/video and buttons to start inferencing. I find it useful for this demo because there are so many models to choose from, and it allows for really quick testing/comparing of detection models. It also makes it possible to make an "app version" (using voila, to quickly show results to others for example). But it does add a bit of new complexity. 
- I made a models-subset.lst with a mostly random selection of models. I found that useful for testing, for example on Binder, and for slow computers/connections. If we want to keep that, I would welcome a selection of models that is less random. 
- I download the models by default to `$HOME/open_model_zoo_models` and set the cache to `$HOME/open_model_zoo_cache`.  I keep the converted models in the same directory as the downloaded models. I am happy to change these directories to something else, I do not know what is best practice. I would like to have a standard directories for this that always work, so that I can use them for further OMZ notebooks. 
- New Python packages in requirements.txt are voila, ipywidgets and jupyterlab. They are all BSD licensed.

### Screenshots of the completed notebook and app-version

omz_object_detection_demo_notebook
omz_object_detection_demo_notebook_voila


vladimir-dudnik:
@helena-intel that basically work for me (tested on Windows). Although get some questions:

- downloading and converting all OMZ models is costly process, it make take hours. It would be much easier if by starting notebook we can pass as a parameter folders for downloaded models and downloader cache, so if customer already downloaded models when was playing with other demos, no need to manually edit notebook. Currently it silently start model downloading to predefined place. (do not know if notebooks provide such capabilities, to start with params).
- it seems there is no need to install openvino python module, just initialize openvino environment by running setupvars.sh/.bat from openvino installed location, just like it is needed for all other demos. This script will initialize both PYTHONPATH and PATH environment vars for console it was called from.

vladimir-dudnik:
Jenkins please retry a build

helena-intel:
> * downloading and converting all OMZ models is costly process, it make take hours. It would be much easier if by starting notebook we can pass as a parameter folders for downloaded models and downloader cache, so if customer already downloaded models when was playing with other demos, no need to manually edit notebook. Currently it silently start model downloading to predefined place. (do not know if notebooks provide such capabilities, to start with params).

If you change the base_model_dir and omz_cache_dir in the Settings cell (at the top) it will not redownload everything and just use the models that are there. But as far as I know there is no way to set this from the command line when you start Jupyter. I plan to reuse this directory across all demo's (and am happy to change to other directory names) so then a user only needs to download once. I already made a subset of models to download by default. I can make that a lot smaller. I can also highlight the setting in the README and make it a highlighted alert box in the notebook itself. 

> * it seems there is no need to install openvino python module, just initialize openvino environment by running setupvars.sh/.bat from openvino installed location, just like it is needed for all other demos. This script will initialize both PYTHONPATH and PATH environment vars for console it was called from.

Correct. The instructions were for people who are completely new to this and do not have OpenVINO installed yet. I will update the instructions, also to account for the new pip method.  I can add an "if you have already installed OpenVINO" part to the instructions (where I can also point to the directory settings). I would like to still keep this method to install everything including openvino with `pip install -r requirements.txt` though (and not requiring the full install of OpenVINO and running setupvars). That makes it much easier to get started with this, also during live demos/webinars.

vladimir-dudnik:
@helena-intel please note pre-commit style check failed:
13:49:31  demos/object_detection_demo/jupyter-python/README.md:32: trailing whitespace.
13:49:31  +2. Install Python 
13:49:31  demos/object_detection_demo/jupyter-python/README.md:43: trailing whitespace.
13:49:31  +    - NOTE: the PATH is remembered as long as you do not close the terminal application. If you open a new terminal and want to run the notebooks again, you have to set the PATH again. 
13:49:31  demos/object_detection_demo/jupyter-python/README.md:48: trailing whitespace.
13:49:31  + 
13:49:31  demos/object_detection_demo/jupyter-python/README.md:70: trailing whitespace.
13:49:31  +steps under *Configure the Model Optimizer*. 
13:49:31  demos/object_detection_demo/jupyter-python/README.md:71: new blank line at EOF.
13:49:31  demos/object_detection_demo/jupyter-python/detection_utils.py:55: trailing whitespace.
13:49:31  +    
13:49:31  demos/object_detection_demo/jupyter-python/detection_utils.py:56: trailing whitespace.
13:49:31  +    
13:49:32  running yamllint...
13:49:37  running flake8...
13:49:45  ./demos/object_detection_demo/jupyter-python/detection_utils.py:13:1: F401 'monitors' imported but unused
13:49:45  demos/object_detection_demo/jupyter-python/detection_utils.py:117: last line doesn't end with a newline character

vladimir-dudnik:
Jenkins please retry a build

helena-intel:
> @helena-intel please note pre-commit style check failed:
> 13:49:31 demos/object_detection_demo/jupyter-python/README.md:32: trailing whitespace.

Apologies, I fixed that, and found the `check-basics.py` script to prevent this in the future.

vladimir-dudnik:
Jenkins please retry a build

vladimir-dudnik:
Jenkins please retry a build

helena-intel:
I added a NotebookDemo to the demo tests and a test case for this object detection notebook. The test converts the notebook to a Python script and runs it with default arguments. For now, the test case only tests that the notebook runs without exceptions (so it catches import errors etc.) not if there is actually useful output. I will update the test to a more useful one if this is the right approach.

vladimir-dudnik:
Jenkins please retry a build

demos/tests/cases.py
```diff
@@ -68,6 +70,33 @@ def fixed_args(self, source_dir, build_dir):
...
+        super().__init__(name, implementation, device_keys, test_cases)
+        self._exec_name = self._exec_name.replace('_jupyter-python', '')
+
+    def fixed_args(self, source_dir, build_dir):
+        sys.path.append(os.path.join(os.path.dirname(os.path.abspath(os.curdir)), "common", "python"))
+
+        notebook_file = str(source_dir / self.subdirectory / (self._exec_name + '.ipynb'))
+        python_file = notebook_file.replace('ipynb', 'py')
+        python_test_file = python_file.replace('.py', '_test.py')
+        subprocess.run([sys.executable.replace('python', 'jupyter'), 'nbconvert', '--to', 'python', notebook_file])
```
vladimir-dudnik, line 87:
```suggestion
        subprocess.run(['jupyter', 'nbconvert', '--to', 'python', notebook_file])
```

vladimir-dudnik: seems with this change test should run in CI

vladimir-dudnik:
Jenkins please retry a build

vladimir-dudnik:
@helena-intel we need to update CI requirements in order to make this work.

cc @nignatovsky @IRDonch

vladimir-dudnik:
Jenkins please retry a build

vladimir-dudnik:
@nignatovsky please check your modification of requirements for CI. On Linux it gives an error:
ERROR: Could not find a version that satisfies the requirement pywin32==300

demos/object_detection_demo/jupyter-python/detection_utils.py
```diff
@@ -0,0 +1,116 @@
...
+            cv2.rectangle(frame, (xmin, ymin), (xmax, ymax), color, 2)
+            cv2.putText(
+                frame,
+                "{} {:.1%}".format(det_label, detection.score),
+                (xmin, ymin - 7),
+                cv2.FONT_HERSHEY_COMPLEX,
+                0.6,
+                color,
+                1,
+            )
```
Fedor-Z, line 99: Probably we should better put this function into notebook to show that we actually draw rectangles and other overlays basing on numeric results received from openVino.

demos/object_detection_demo/jupyter-python/object_detection_demo.ipynb
```diff
@@ -0,0 +1,750 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "technical-grocery",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import time\n",
+    "start = time.perf_counter()\n",
+    "uploader = widgets.FileUpload(multiple=False)\n",
+    "uploader"
```
Fedor-Z, line 598: Please move embedding of sample_video and uploader down to show_results_on_model and show_inference_mulitple_models.
I.e. delete ",sample_video" and ",uploader" here
and add something like this:
    videos = sample_video,
    custom_video = uploader
to show_results_on_model and show_inference_mulitple_models functions as arguments.
You'll probably have to modify get_input_filename() function as well to pass instances of these 2 widgets as arguments.
Main goal is to keep all widgets in one place - where actually inference is done and results are shown.

demos/object_detection_demo/jupyter-python/object_detection_demo.ipynb
```diff
@@ -0,0 +1,750 @@
...
+   "cell_type": "markdown",
+   "id": "powerful-valuable",
+   "metadata": {},
+   "source": [
+    "## Download Models and convert them to IR format\n",
+    "\n",
+    "The [Model Downloader](https://github.com/openvinotoolkit/open_model_zoo/blob/master/tools/downloader/README.md) downloads models from the Open Model Zoo. Models that are not in OpenVINO IR format are converted to this format by the Model Converter. \n",
+    "\n",
+    "A subset of [Open Model Zoo](https://github.com/openvinotoolkit/open_model_zoo/) models that are compatible with this demo are listed in the file `models_file` (default \"models.lst\") in the same folder as this notebook. By default these models are downloaded, with the `--list=models_file` argument for the Model Downloader. You can choose to download a specific model by using `--name=model_name` instead of `--list=models.lst`. If you already have downloaded Open Zoo Models, you can set the `base_model_dir` variable in the *Settings* cell to the folder that contains your models (this should be a folder with subfolders `intel` and `public`) and set `DOWNLOAD_MODELS` to `False`.\n",
+    "\n",
```
Fedor-Z, line 141:
```suggestion
    "\n",
    "Note that [Model Downloader](https://github.com/openvinotoolkit/open_model_zoo/blob/master/tools/downloader/README.md) and Model Converter are standalone tools and can be executed using command line. Models have to be downloaded only once. Thus the code below (calling those tools) is usually not included into real applications as they assume that model is already downloaded and converted (or provided in application package)."
    "\n",
```

Fedor-Z (changes requested):

demos/object_detection_demo/jupyter-python/object_detection_demo.ipynb
helena-intel: That is nicer indeed, I agree that the input video and especially the second inference function are too far away. I find that if I add them with the inference it adds confusion though. It is not clear that the uploader widget is optional. 

I could also delete the uploader widget, and include video's that are uploaded to the notebook in (what is now the) the sample video dropdown. I would add a line on how to do that, so that adds a bit of complexity, but it reduces UI clutter a lot and makes the code easier to read (also getting rid of the `get_input_filename` function that is a bit of a clutch). Downside is that it doesn't work in the "app version" with Voila, but that is not a priority. What do you think? Other suggestions welcome too.

demos/object_detection_demo/jupyter-python/detection_utils.py
helena-intel: I'll change that, thanks!

vladimir-dudnik:
Jenkins please retry a build

demos/tests/cases.py
```diff
@@ -68,6 +70,33 @@ def fixed_args(self, source_dir, build_dir):
...
+class NotebookDemo(Demo):
+    """
+    Convert notebook to Python script and runs it
+    """
+    def __init__(self, name, implementation='jupyter-python', device_keys=None, test_cases=None):
+        super().__init__(name, implementation, device_keys, test_cases)
+        self._exec_name = self._exec_name.replace('_jupyter-python', '')
+
+    def fixed_args(self, source_dir, build_dir):
+        sys.path.append(os.path.join(os.path.dirname(os.path.abspath(os.curdir)), "common", "python"))
```
nignatovsky, line 82:
```suggestion
        demo_env = {**os.environ,
                   'PATH': f"{os.environ['PATH']}{os.pathsep}"
                           f"{os.path.dirname(sys.executable)}",
                   'PYTHONPATH': f"{os.environ['PYTHONPATH']}{os.pathsep}"
                                 f"{os.path.join(os.path.dirname(os.path.abspath(os.curdir)), 'common', 'python')}"}
```
Possible fix for `FileNotFoundError: [Errno 2] No such file or directory: 'jupyter': 'jupyter'`

demos/tests/cases.py
```diff
@@ -68,6 +70,33 @@ def fixed_args(self, source_dir, build_dir):
...
+
+        notebook_file = str(source_dir / self.subdirectory / (self._exec_name + '.ipynb'))
+        python_file = notebook_file.replace('ipynb', 'py')
+        python_test_file = python_file.replace('.py', '_test.py')
+        subprocess.run(['jupyter', 'nbconvert', '--to', 'python', notebook_file])
```
nignatovsky, line 87:
```suggestion
        subprocess.run(['jupyter', 'nbconvert', '--to', 'python', notebook_file], env=demo_env)
```

demos/object_detection_demo/jupyter-python/object_detection_demo.ipynb
helena-intel: I like the first part, that is a good clarification. Would you be okay with leaving out this part "Thus the code below (calling those tools) is usually not included into real applications as they assume that model is already downloaded and converted (or provided in application package"? A notebook is always a bit different than a "real application". I also show a message that "DOWNLOAD_MODELS" can be set to False once it has been run.

vladimir-dudnik:
Jenkins please retry a build

demos/object_detection_demo/jupyter-python/object_detection_demo.ipynb
```diff
@@ -0,0 +1,750 @@
...
+    "            PRECISION,\n",
+    "            \"--list\",\n",
+    "            models_file,\n",
+    "            \"--jobs\",\n",
+    "            \"4\",\n",
+    "        ],\n",
+    "        stdout=subprocess.PIPE,\n",
+    "        stderr=subprocess.PIPE,\n",
+    "        shell=False,\n",
+    "        text=True,\n",
```
nignatovsky, line 173:
```suggestion
    "        universal_newlines=True,\n",
```

demos/object_detection_demo/jupyter-python/object_detection_demo.ipynb
```diff
@@ -0,0 +1,750 @@
...
+    "            PRECISION,\n",
+    "            \"--mo\",\n",
+    "            model_optimizer_path,\n",
+    "            \"--jobs\",\n",
+    "            \"4\",\n",
+    "        ],\n",
+    "        stdout=subprocess.PIPE,\n",
+    "        stderr=subprocess.PIPE,\n",
+    "        shell=False,\n",
+    "        text=True,\n",
```
nignatovsky, line 220:
```suggestion
    "        universal_newlines=True,\n",
```

demos/object_detection_demo/jupyter-python/object_detection_demo.ipynb
```diff
@@ -0,0 +1,750 @@
...
+    "info_result = subprocess.run(\n",
+    "    [\n",
+    "        sys.executable,\n",
+    "        info_command,\n",
+    "        \"--list\",\n",
+    "        models_file,\n",
+    "    ],\n",
+    "    stdout=subprocess.PIPE,\n",
+    "    stderr=subprocess.PIPE,\n",
+    "    capture_output=False,\n",
```
nignatovsky, line 260:
```suggestion
```

demos/object_detection_demo/jupyter-python/object_detection_demo.ipynb
```diff
@@ -0,0 +1,750 @@
...
+    "    shell=False,\n",
+    "    text=True,\n",
```
nignatovsky, line 262:
```suggestion
    "    universal_newlines=True,\n",
```

nignatovsky (commented):
Fixes for `TypeError: __init__() got an unexpected keyword argument 'text' / 'capture_output'` (These parameters don't exist in `subprocess` lib for Python 3.6, so I replaced them with correct ones). 
We use Python 3.6 in validation on purpose, it is the lowest python version, supported by OpenVINO.

vladimir-dudnik:
Jenkins please retry a build

vladimir-dudnik:
Jenkins please retry a build

vladimir-dudnik:
@helena-intel demo now pass the tests. please unset Draft status if you are ready to merge this

vladimir-dudnik (approved):

demos/object_detection_demo/jupyter-python/object_detection_demo.ipynb
fzhar: We can just put label "Optional" near Upload.
Another option is to add labels, something like:
Select preloaded video from dropdown: [videos list]
- or - 
Upload your own video: [Upload button]

Not sure if it would be easy to do using currently used functions/decorators.

BTW, in case of standalone application current version is fine, but in case of notebook videos list and upload  are too far from actual inference

demos/object_detection_demo/jupyter-python/object_detection_demo.ipynb
fzhar: We can rephrase second part. I just wanted to emphasize that in most application that part of python code would not be used because model is either already converted or downloader/converter are used as standalone tools. That would give customers a hint that this part of code should not be copied directly to their apps, as other parts. 
But if we cannot rephrase it - I'm OK with leaving it out.

demos/object_detection_demo/jupyter-python/object_detection_demo.ipynb
helena-intel: I changed this in the most recent commit. I now only show the dropdown, which now shows sample video's and user's own video's. There is a small drawback that the user now has to use Jupyter's upload function instead of a widget, but I think it looks much clearer now and there is no awkward `get_input_filename` function anymore.

helena-intel:
@vladimir-dudnik , I changed the widgets for selecting a video. Previous version had two widgets: one for uploading your own video, one for downloading a sample video. I now made one dropdown widget that supports both sample video's and user's own video's. The video dropdown is now also shown closer to the inference, below the model selection dropdown. See screenshot. 
image
I also updated the README and description in the notebook. 

I'm moving this from draft to regular PR. I would still like to update the instructions for model converter to use `pip install openvino-dev`, but I'm not sure yet how exactly that will work. Can we change these minor documentation improvements also after the PR is merged?

vladimir-dudnik (approved):

Fedor-Z (approved):

nignatovsky (approved):

demos/object_detection_demo/jupyter-python/requirements.txt
```diff
@@ -0,0 +1,10 @@
+requests
+jupyterlab
+ipywidgets
+voila
+matplotlib
+openvino
+opencv-python-headless
```
IRDonch, line 7: Why are these included here? We should be using the OpenVINO and OpenCV libraries from the OpenVINO toolkit package we're in, not from PyPI.

demos/object_detection_demo/jupyter-python/README.md
```diff
@@ -0,0 +1,77 @@
+# Object Detection Async Demo
```
IRDonch, line 1:
```suggestion
# Object Detection Jupyter Notebook Demo
```

demos/object_detection_demo/jupyter-python/detection_utils.py
```diff
@@ -0,0 +1,88 @@
+import random
```
IRDonch, line 1: The name of the directory should be `jupyter_python` (with an underscore). That would be consistent with `cpp_gapi`, which we have elsewhere.

demos/object_detection_demo/jupyter-python/requirements.txt
```diff
@@ -0,0 +1,10 @@
+requests
+jupyterlab
+ipywidgets
+voila
```
IRDonch, line 4: This doesn't seem to actually be used anywhere?

IRDonch (changes requested):

demos/object_detection_demo/jupyter-python/requirements.txt
helena-intel: We wanted these notebooks to work as standalone demo's, with instructions that work for people who are completely new to OpenVINO. I can remove openvino from the requirements.txt and explain say in the README to install OpenVINO first. OpenCV is not installed by the PyPI distribution of OpenVINO (as far as I know), so I think that should remain in the requirements for the notebook.

demos/object_detection_demo/jupyter-python/requirements.txt
helena-intel: jupyterlab ipywidgets and voila are not used in the notebook, but they are needed to run the notebook (voila only for the dashboard mode version). Requests is for model downloader. With this requirements.txt the user can immediately run the notebook without needing to first look how to install OpenVINO and the Open Model Zoo requirements. If a user already has this installed in their environment nothing will happen.

demos/object_detection_demo/jupyter-python/requirements.txt
```diff
@@ -0,0 +1,10 @@
...
+matplotlib
+openvino
+opencv-python-headless
+pyyaml
+ipython==7.10
+numpy==1.19
```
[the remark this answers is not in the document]
helena-intel: ipython gives an annoying irrelevant warning message in the notebook with their most recent version. 7.10 does not give that warning. With numpy I ran into issues because OpenVINO installed I think 1.16 and that did not work with another package. I read that there would be standardization of numpy across OpenVINO tools though. If `pip install openvino` will install numpy 1.18 or 1.19 I think I can avoid this version pin.

demos/object_detection_demo/jupyter-python/detection_utils.py
helena-intel: Changed to `jupyter_python`

vladimir-dudnik:
Jenkins please retry a build

demos/object_detection_demo/jupyter-python/requirements.txt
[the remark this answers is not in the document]
IRDonch: > ipython gives an annoying irrelevant warning message in the notebook with their most recent version. 7.10 does not give that warning.

This doesn't seem like a good enough reason. Generally, any pinned package is a liability in that it may stop working on a newer version of Python, be incompatible with the user's other installed packages or contain security vulnerabilities. A warning is comparatively harmless.

> With numpy I ran into issues because OpenVINO installed I think 1.16 and that did not work with another package.

I don't see why it would do that. Looking at the wheel, it requires `numpy (>=1.16.3)`, so it would install the latest version by default. Moreover, if it really does require 1.16, then pinning it to 1.19 would just create an inconsistent environment, which isn't ideal.

demos/object_detection_demo/jupyter-python/requirements.txt
IRDonch: > We wanted these notebooks to work as standalone demo's, with instructions that work for people who are completely new to OpenVINO.

I can understand that, but if you're including them in OMZ, they are de facto not standalone demos - they are a part of OMZ, which is in itself part of the OpenVINO toolkit. It doesn't make sense to install OpenVINO with pip if you found this demo inside the OpenVINO toolkit.

It's not unreasonable to support the standalone scenario, but it should be understood that it's not the only possibility.

> I can remove openvino from the requirements.txt and explain say in the README to install OpenVINO first.

That would be reasonable.

> OpenCV is not installed by the PyPI distribution of OpenVINO (as far as I know), so I think that should remain in the requirements for the notebook.

OpenCV is in the same situation as OpenVINO - it only needs to be installed if we're not in the OpenVINO toolkit.

demos/object_detection_demo/jupyter-python/requirements.txt
IRDonch: > jupyterlab ipywidgets and voila are not used in the notebook, but they are needed to run the notebook (voila only for the dashboard mode version).

OK.

> Requests is for model downloader.

In that case I would suggest to not refer to it directly, but to include the MD requirements file with `-r`.

demos/object_detection_demo/jupyter-python/requirements.txt
helena-intel: @dkurt or @mryzhov Installing OpenVINO with `pip install openvino` or `pip install openvino-dev` from the testing wheels did not install OpenCV. Has that changed and is there a way to get an OpenVINO environment with OpenCV installed with the pip packages?

demos/object_detection_demo/jupyter-python/requirements.txt
helena-intel: Ill change that, thanks!

demos/object_detection_demo/jupyter-python/requirements.txt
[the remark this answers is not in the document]
helena-intel: After installing openvino with `pip install openvino`, if you install the OMZ requirements-pytorch.in, to be able to convert PyTorch models with Model Converter, it installs onnx. This requires a higher Numpy version than 1.16.3, and pip installs Numpy 20 then. See screenshot. 
image

It is my understanding that Numpy versions would be standardized across OpenVINO tools to either 1.18 or 1.19. If that is the case then I can remove this pinning. @mryzhov will the final version of the PyPI packages have a different numpy version requirement?

If Numpy 20 doesn't cause problems I can also remove the version pinning.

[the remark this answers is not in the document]
helena-intel: We noticed when people tested this that they were a bit alarmed by warnings that were not related to OpenVINO, like this one. Inexperienced devs will not recognize it as a harmless warning. I'll remove the pinning for this notebook though. I hope that ipython will fix this in a future version.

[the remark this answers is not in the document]
IRDonch: > After installing openvino with pip install openvino, if you install the OMZ requirements-pytorch.in, to be able to convert PyTorch models with Model Converter, it installs onnx. This requires a higher Numpy version than 1.16.3, and pip installs Numpy 20 then.

What I don't understand is why 1.16.3 was installed to begin with. As I've said, `openvino` only has a lower bound on the version of NumPy (at least in the last release).

demos/tests/cases.py
```diff
@@ -68,6 +71,59 @@ def fixed_args(self, source_dir, build_dir):
...
+
+    def __init__(self, name, implementation='jupyter_python', device_keys=None, test_cases=None, test_code=None):
+        super().__init__(name, implementation, device_keys, test_cases)
+        self._exec_name = self._exec_name.replace('_jupyter_python', '')
+        self.test_code = test_code
+
+    def fixed_args(self, source_dir, build_dir):
+        demo_env = {**os.environ,
+                   'PATH': f"{os.environ['PATH']}{os.pathsep}"
+                           f"{os.path.dirname(sys.executable)}",
```
IRDonch, line 90: This isn't reliable, because the directory with the Jupyter entrypoint is not necessarily the same as the directory with the Python executable (for example, the Jupyter package could be installed into the per-user directory). It's better to run Jupyter via `[sys.executable, '-m', 'jupyter']` instead.

demos/tests/cases.py
```diff
@@ -68,6 +71,59 @@ def fixed_args(self, source_dir, build_dir):
...
+                   'PYTHONPATH': f"{os.environ['PYTHONPATH']}{os.pathsep}"
+                                 f"{os.path.join(os.path.dirname(os.path.abspath(os.curdir)), 'common', 'python')}"}
```
IRDonch, line 92: It's unclear to me why this is necessary. Extracting the script out of the notebook shouldn't require importing modules, right?

demos/tests/cases.py
```diff
@@ -68,6 +71,59 @@ def fixed_args(self, source_dir, build_dir):
...
+
+        notebook_file = str(source_dir / self.subdirectory / (self._exec_name + '.ipynb'))
+        python_file = notebook_file.replace('ipynb', 'py')
```
IRDonch, line 95: You should keep `notebook_file` a `Path` and then use `Path` methods here, e.g. `python_file = notebook_file.with_suffix('.py')`. That will ensure that you don't perform unexpected replacements (e.g. if `ipynb` is present somewhere else in the path).

demos/tests/cases.py
```diff
@@ -68,6 +71,59 @@ def fixed_args(self, source_dir, build_dir):
...
+        python_test_file = python_file.replace('.py', '_test.py')
+        subprocess.run(['jupyter', 'nbconvert', '--to', 'python', notebook_file], env=demo_env)
+
+        with open(python_file, 'r') as python_script:
+            # Change the working directory to the directory that contains the notebook
+            changedir = f"import os\nos.chdir(r'{os.path.dirname(python_file)}')\n"
+            # Add some lines to the test script to make it accept command line arguments
+            content = changedir + python_script.read()
```
IRDonch, line 103: Assuming you change `python_file` to a `Path`, this could be replaced with just `content = changedir + python_file.read_text()` and the `with` statement could be removed. (Similarly, the other `with` statement could be replaced with a call to `write_text`.)

demos/tests/cases.py
```diff
@@ -68,6 +71,59 @@ def fixed_args(self, source_dir, build_dir):
...
+                                 f"{os.path.join(os.path.dirname(os.path.abspath(os.curdir)), 'common', 'python')}"}
+
+        notebook_file = str(source_dir / self.subdirectory / (self._exec_name + '.ipynb'))
+        python_file = notebook_file.replace('ipynb', 'py')
+        python_test_file = python_file.replace('.py', '_test.py')
+        subprocess.run(['jupyter', 'nbconvert', '--to', 'python', notebook_file], env=demo_env)
+
+        with open(python_file, 'r') as python_script:
+            # Change the working directory to the directory that contains the notebook
+            changedir = f"import os\nos.chdir(r'{os.path.dirname(python_file)}')\n"
```
IRDonch, line 101: Why is this needed? The script ought to work from any directory.

demos/tests/cases.py
```diff
@@ -68,6 +71,59 @@ def fixed_args(self, source_dir, build_dir):
...
+                parser.add_argument('-m')
+                parser.add_argument('-i')
+                return parser.parse_args()
+
+            args = parse_args()
+            """)
+            # Change settings for test script
+            content += '\nbase_model_dir=dirname(dirname(dirname(dirname(args.m))))'
+            # Add a line that executes the function that is tested
+            content += f'\n{self.test_code}'
```
IRDonch, line 119: Seems to me like these could all be merged into the preceding string literal, to form one complete template into which `self.test_code` and the contents of the original file are merged.

demos/tests/cases.py
```diff
@@ -726,4 +782,18 @@ def single_option_cases(key, *args):
     )),
 ]
 
-DEMOS = NATIVE_DEMOS + PYTHON_DEMOS
+
+NOTEBOOK_DEMOS = [
+   NotebookDemo(name='object_detection_demo', device_keys=[],
+   test_code = 'JUMP_FRAMES=200\nresult=get_results_for_model(os.path.basename(args.m)[:-4], args.i, 4, 4, 4)\nassert len(result[0]) > 2',
```
IRDonch, line 788: This would be more readable as a multiline string.

IRDonch (changes requested):

demos/tests/cases.py
helena-intel: Thanks, I'll change this.

demos/tests/cases.py
helena-intel: This was necessary to get the tests to run (locally and in CI). I made some changes since then. The test passes on my local computer without this now. I'll remove this line.

demos/tests/cases.py
helena-intel: When the notebook is run, the current directory is the notebook directory, and imports from that directory work. This makes the test script do the same thing.

demos/object_detection_demo/jupyter-python/requirements.txt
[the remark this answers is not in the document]
helena-intel: I'll delete the numpy version pinning from the notebook. The `openvino-dev` pip package installed 1.16.3 (see screenshot above). The latest testing wheel (2774) installs 1.16.6 and pins the version to this. There's still a conflict error message from pip though, if you `pip install -r requirements-tensorflow.in` (in model downloader directory) it installs numpy 1.19.5.

```
pkginfo -f requires_dist openvino_dev-2021.3.0-2774-py3-none-any.whl
requires_dist: ['defusedxml (>=0.5.0)', 'scipy (==1.5.4)', 'jstyleson (==0.0.2)', 'numpy (==1.16.6)', 'addict (==2.2.1)', 'pandas (==0.24.2)', 'hyperopt (==0.1.2)', 'networkx (==2.2)', 'tqdm (==4.31.1)', 'texttable (==1.6.3)', 'py-cpuinfo (!=5.0,!=6.0)', 'PyYAML (>=5.4.1)', 'pillow (>=8.1.0)', 'scikit-image', 'scikit-learn', 'yamlloader', 'shapely', 'nibabel', 'pydicom', 'sentencepiece', 'tokenizers', 'editdistance', 'parasail', 'fast-ctc-decode', 'rawpy', 'nltk', 'opencv-python', 'openvino (==2021.3.0)']`
``

**Comment by helena-intel** [comment]
@IRDonch thanks for your review and suggestions! I have implemented most of your suggestions now, and marked those conversations as resolved.

**Comment by vladimir-dudnik** [comment]
Jenkins please retry a build

**Comment by vladimir-dudnik** [comment]
@helena-intel please take a look at pre-commit failures:

12:55:42  Fixed arguments: /home/jenkins/workspace/omz/precommit/demos_ubuntu18/venv/bin/python /home/jenkins/workspace/omz/precommit/demos_ubuntu18/src/demos/object_detection_demo/jupyter_python/object_detection_demo_test.py
12:55:42  
12:55:42  Test case #0/CPU: -i /mnt/datasets/Image_Retrieval/d0c460d0-4d75-4315-98a8-a0116d3dfb81.dav -m /tmp/tmpl63ylz3v/models/public/yolo-v3-tiny-tf/FP16/yolo-v3-tiny-tf.xml
12:55:42  
12:55:42  /home/jenkins/workspace/omz/precommit/demos_ubuntu18/venv/bin/python: can't open file '/home/jenkins/workspace/omz/precommit/demos_ubuntu18/src/demos/object_detection_demo/jupyter_python/object_detection_demo_test.py': [Errno 2] No such file or directory

**Comment by vladimir-dudnik** [comment]
@helena-intel please take a look at pre-commit failures:

12:55:42  Fixed arguments: /home/jenkins/workspace/omz/precommit/demos_ubuntu18/venv/bin/python /home/jenkins/workspace/omz/precommit/demos_ubuntu18/src/demos/object_detection_demo/jupyter_python/object_detection_demo_test.py
12:55:42  
12:55:42  Test case #0/CPU: -i /mnt/datasets/Image_Retrieval/d0c460d0-4d75-4315-98a8-a0116d3dfb81.dav -m /tmp/tmpl63ylz3v/models/public/yolo-v3-tiny-tf/FP16/yolo-v3-tiny-tf.xml
12:55:42  
12:55:42  /home/jenkins/workspace/omz/precommit/demos_ubuntu18/venv/bin/python: can't open file '/home/jenkins/workspace/omz/precommit/demos_ubuntu18/src/demos/object_detection_demo/jupyter_python/object_detection_demo_test.py': [Errno 2] No such file or directory

File: demos/object_detection_demo/jupyter-python/requirements.txt (hunk 11 above)
Comment (helena-intel) [reply demos/object_detection_demo/jupyter-python/requirements.txt]: I tested the very last openvino-dev package and saw that opencv is installed there, so I can remove it from the requirements of the notebook. @mryzhov could you confirm if opencv will always be included in openvino-dev? I am worried that if I assume requirements of other packages, the other packages may change and then my code breaks. But since OMZ is part of OpenVINO, if we know this will always be installed that is not a concern in this case.

**Comment by helena-intel** [comment]
> @helena-intel please take a look at pre-commit failures:
> [...]
> 12:55:42 /home/jenkins/workspace/omz/precommit/demos_ubuntu18/venv/bin/python: can't open file '/home/jenkins/workspace/omz/precommit/demos_ubuntu18/src/demos/object_detection_demo/jupyter_python/object_detection_demo_test.py': [Errno 2] No such file or directory

@vladimir-dudnik After reading the change requests, and changing the command to run jupyter, I had hoped that the PATH and PYTHONPATH variables were unnecessary now, so I removed them. The test did pass on my local machine. I'll add the PATH variable back. That should hopefully fix this. Is there a way for me to check the CI pipeline without making a commit and update the PR?

**Comment by vladimir-dudnik** [comment]
Jenkins please retry a build

**Comment by vladimir-dudnik** [comment]
@helena-intel without making commit you can only test this locally, by making the same environment and call run_tests.py

**Comment by vladimir-dudnik** [comment]
Jenkins please retry a build

**Comment by vladimir-dudnik** [comment]
Jenkins please retry a build

File: demos/object_detection_demo/jupyter-python/requirements.txt (hunk 11 above)
Comment (mryzhov) [reply demos/object_detection_demo/jupyter-python/requirements.txt]: Yes, the openvino-dev wheel requires opencv as dependence

**Comment by vladimir-dudnik** [comment]
Jenkins please retry a build

**Comment by vladimir-dudnik** [comment]
@helena-intel still can't find script file

17:00:20  /home/jenkins/workspace/omz/precommit/demos_ubuntu18/venv/bin/python: can't open file '/home/jenkins/workspace/omz/precommit/demos_ubuntu18/src/demos/object_detection_demo/jupyter_python/object_detection_demo_test.py': [Errno 2] No such file or directory

**Comment by helena-intel** [comment]
@vladimir-dudnik I think the recent commit doesn't fix it. Working on it now.

File: demos/tests/cases.py (hunk 20 above)
Comment (IRDonch) [reply demos/tests/cases.py]: > When the notebook is run, the current directory is the notebook directory

Not necessarily, the user could run it from any directory.

> and imports from that directory work

Is it really required for the current directory to be the notebook directory for the imports to work?

File: demos/tests/cases.py [hunk 23]
```diff
@@ -68,6 +71,60 @@ def fixed_args(self, source_dir, build_dir):
...
+            parser = argparse.ArgumentParser()
+            parser.add_argument('-m')
+            parser.add_argument('-i')
+            return parser.parse_args()
+
+        args = parse_args()
+        # Change settings for test script
+        content += '\nbase_model_dir=dirname(dirname(dirname(dirname(args.m))))'
+        # Add a line that executes the function that is tested
+        content += f'\n{self.test_code}'
```
IRDonch, line 120: You need to remove the assignments for this to work.

helena-intel: Thanks, I noticed that too, I'll fix it today.

helena-intel: @vladimir-dudnik @IRDonch I hope I fixed the tests in the CI. I kept the PATH and PYTHONPATH statements, even though I think they're unnecessary now, but I really want the CI tests to pass again first. Doublechecked on my local computer that run_tests.py works correctly.

helena-intel: @IRDonch "Is it really required for the current directory to be the notebook directory for the imports to work?" (I cannot reply there) Yes. I can also solve it in some other way by adding a line to change the PYTHONPATH there, but I don't know if that's better.

IRDonch: > I hope I fixed the tests in the CI. I kept the PATH and PYTHONPATH statements, even though I think they're unnecessary now, but I really want the CI tests to pass again first.

Well, CI did pass (and I don't think the failures were related to that change in the first place), so I suggest you remove them again.

> "Is it really required for the current directory to be the notebook directory for the imports to work?" (I cannot reply there) Yes.

But why? Now that I think about it, you're not even running a notebook; you're running a normal Python script. And in Python, the script directory is automatically added to the module search path. So why do you need to change the directory.

demos/requirements.txt
```diff
@@ -8,3 +8,6 @@ tokenizers
 tensorboard
 tensorboardX
 tqdm
+jupyterlab
+ipywidgets
+opencv-python-headless
```
IRDonch, line 13: Please remove OpenCV from here, as well. Also, keep the list in alphabetic order.

demos/tests/cases.py
IRDonch: OK, I see what's going on. You used `os.curdir` in the notebook to find the OMZ root. You should be using `__file__` instead - that way, you can run the notebook from any directory.

helena-intel: That would solve this (and other issues) indeed, but `__file__`  is not defined in notebooks. I spend some time looking into this when I first made this notebook and there doesn't seem to be a solution that always works in the way it does for regular scripts (I tried quite a few things I found on Stack Overflow). Suggestions welcome.

IRDonch: I see. That is quite annoying.

Well, in that case your solution is the best option you have. I would only request you to merge the `changedir` variable into the template, which would make it more readable.

helena-intel: Thanks, I will do that, and remove the demo_env with the PATH variables. They were initially added with another change request, when the tests also would not run in the CI, but I think they are no longer necessary after the recent changes/new way to invoke Jupyter.

demos/object_detection_demo/jupyter_python/object_detection_demo.ipynb
```diff
@@ -0,0 +1,754 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "hungarian-opposition",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "model_names = [model[\"name\"] for model in info if \"intel\" in model[\"subdirectory\"]]\n",
```
IRDonch, line 293: The condition here is unreliable, because the string "intel" could just be a part of the model name or subdirectory. Moreover, it's indirect. You'd be better off just checking whether the framework is "dldt".

demos/object_detection_demo/jupyter_python/object_detection_demo.ipynb
```diff
@@ -0,0 +1,754 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "engaging-personal",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "model_architectures = {}\n",
+    "modellist = open(models_file).read().splitlines()\n",
```
IRDonch, line 326: Use `with` to ensure the file is closed.

demos/object_detection_demo/jupyter_python/object_detection_demo.ipynb
```diff
@@ -0,0 +1,754 @@
...
+    "\n",
+    "for line in modellist[1:]:\n",
+    "    if line.startswith(\"# For\"):\n",
+    "        _, architecture = line.split(\"=\")\n",
+    "    else:\n",
+    "        model_architectures[line] = architecture\n",
+    "        for modelname in model_names:\n",
+    "            modelpattern = re.search(line.replace(\"?\", \"[0-9]\"), modelname)\n",
```
IRDonch, line 334: Don't try to reimplement the `fnmatch` algorithm - just use `fnmatch`.

demos/object_detection_demo/jupyter_python/object_detection_demo.ipynb
```diff
@@ -0,0 +1,754 @@
...
+   "execution_count": null,
+   "id": "civilian-ballot",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "def make_result_frames(inference_result, has_landmarks):\n",
+    "    \"\"\" \"\n",
+    "    Draw boxes on frames from inference results and return the list of frames.\n",
+    "    \"\"\"\n",
+    "    framelist = list()\n",
```
IRDonch, line 487:
```suggestion
    "    framelist = []\n",
```

demos/object_detection_demo/jupyter_python/object_detection_demo.ipynb
```diff
@@ -0,0 +1,754 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "tired-malaysia",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "sample_video_base_url = \"https://github.com/intel-iot-devkit/sample-videos/raw/master\"\n",
+    "sample_video_filenames = open(\"sample_videos.lst\").read().splitlines()\n",
```
IRDonch, line 604: Another unclosed file here.

demos/object_detection_demo/jupyter_python/object_detection_demo.ipynb
```diff
@@ -0,0 +1,754 @@
...
+    "all_videos = glob.glob(os.path.join(video_directory, \"*mp4\")) + glob.glob(os.path.join(video_directory, \"*avi\"))\n",
```
IRDonch, line 605:
```suggestion
    "all_videos = glob.glob(os.path.join(video_directory, \"*.mp4\")) + glob.glob(os.path.join(video_directory, \"*.avi\"))\n",
```

demos/object_detection_demo/jupyter_python/detection_utils.py
```diff
@@ -0,0 +1,88 @@
...
+
+def download_video(path: str, video_directory: str) -> np.ndarray:
+    """
+    Download a video from `path` and save it to `video_directory`. `video_directory` will be created if it does not exist
+    yet.
+    """
+
+    # Set User-Agent to Mozilla because some websites block requests with User-Agent Python
+    request = urllib.request.Request(path, headers={"User-Agent": "Mozilla/5.0"})
+    response = urllib.request.urlopen(request)
```
IRDonch, line 84: Use `with` to close the response object.

demos/object_detection_demo/jupyter_python/detection_utils.py
```diff
@@ -0,0 +1,88 @@
...
+    data = response.read()
```
IRDonch, line 85: It seems dangerous to load the whole file into memory at once. It could be multiple gigabytes in size. I would suggest to download it in chunks.

IRDonch (changes requested):

demos/object_detection_demo/jupyter_python/object_detection_demo.ipynb
```diff
@@ -0,0 +1,754 @@
...
+    "base_model_dir = os.path.expanduser(\"~/open_model_zoo_models\")\n",
+    "# Directory for Open Model Zoo cache files. Caching speeds up subsequent downloads.\n",
+    "omz_cache_dir = os.path.expanduser(\"~/open_model_zoo_cache\")\n",
+    "models_file = \"models-subset.lst\"  # models-subset.lst contains a subset of supported models.\n",
+    "# models_file = 'models.lst'  # models.lst contains all supported models.\n",
+    "video_directory = \"videos\"\n",
+    "\n",
+    "## Model settings\n",
+    "DEVICE = \"CPU\"\n",
+    "PRECISION = \"FP16\"\n",
```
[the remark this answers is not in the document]
helena-intel: For the notebooks we only want to test that they run, not how well they perform. When I started with OpenVINO I was recommended to use FP16 for CPU since the results are the same as FP32 but it results in smaller models. Can change to FP32 if that is better.

demos/object_detection_demo/jupyter_python/object_detection_demo.ipynb
helena-intel: The file itself is not assigned to a variable, only the result of read().splitlines(), so the file object itself should be discarded immediately and not remain in memory. No problem to change it to use with though.

demos/object_detection_demo/jupyter_python/detection_utils.py
helena-intel: I'm afraid the notebook will have more problems if people start using multi-gigabyte video's... (The instructions say to use small video's too). I'll change it though, it is good to err on the safe side.

demos/object_detection_demo/jupyter_python/object_detection_demo.ipynb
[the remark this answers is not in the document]
IRDonch: I guess it makes sense for the notebook itself; however, tests should be changed to use FP32 models. The reason for that is that the test runner script will download every precision (of every model) that is used in at least one test case, so testing with FP16 models will make it download a huge number of unused FP16 models.

demos/object_detection_demo/jupyter_python/object_detection_demo.ipynb
```diff
@@ -0,0 +1,754 @@
...
+    "\n",
+    "## Demo settings\n",
+    "DOWNLOAD_MODELS = True  # Use Model Downloader to download models from Open Model Zoo\n",
```
IRDonch, line 111: Doesn't this mean that the notebook will download the models even when run as part of the test suite? That would be undesirable. Can you make it so that it just uses the models already downloaded by the test runner script?

demos/object_detection_demo/jupyter_python/object_detection_demo.ipynb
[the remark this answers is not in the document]
helena-intel: Done.

demos/object_detection_demo/jupyter_python/object_detection_demo.ipynb
helena-intel: Done.

demos/object_detection_demo/jupyter_python/object_detection_demo.ipynb
helena-intel: Done.

vladimir-dudnik:
Jenkins please retry a build

demos/object_detection_demo/jupyter_python/object_detection_demo.ipynb
IRDonch: AFAIK, there is no guarantee that Python will perform garbage collection when you expect it to. If it works correctly, it's only because the current version of the implementation happens to behave that way. This is why you get a `ResourceWarning` when you don't close a file (if you enable this warning).

demos/requirements.txt
IRDonch: You also need to regenerate `ci/requirements-demos.txt` from the new version of this file.

demos/tests/cases.py
```diff
@@ -68,6 +71,58 @@ def fixed_args(self, source_dir, build_dir):
...
+        python_file = notebook_file.with_suffix('.py')
+        python_test_file = python_file.with_name(python_file.stem + '_test.py')
+        subprocess.run([sys.executable, '-m', 'jupyter', 'nbconvert', '--to', 'python', str(notebook_file)])
+
+        original_content = python_file.read_text()
+        original_content = original_content.replace("DOWNLOAD_MODELS = True", "DOWNLOAD_MODELS = False")
+        # Add `test_code` to the test script and make it accept command line arguments
+        content = f"""
+import os
+os.chdir(r'{os.path.dirname(python_file)}')
```
IRDonch, line 98:
```suggestion
os.chdir({os.path.dirname(python_file)!r})
```

Let Python escape any special characters for you. :-)

demos/tests/cases.py
```diff
@@ -68,6 +71,58 @@ def fixed_args(self, source_dir, build_dir):
...
+    parser.add_argument('-m')
+    parser.add_argument('-i')
+    return parser.parse_args()
+
+args = parse_args()
+
+# Change settings for test script
+base_model_dir=dirname(dirname(dirname(dirname(args.m))))
+
+PRECISION = 'FP32'
```
IRDonch, line 116: Seems like this ought to be part of the test code.

vladimir-dudnik:
we will close this PR in favour of new [repo](https://github.com/openvinotoolkit/openvino_notebooks) especially created to develop and distribute OpenVINO jupyter notebook demos.

PR #2103, opened 2021-02-16
 py object detection demo: model type name retina changed to retinaface in args 

ivikhrev:
Model type retina could be a little confusing because we have retinanet model and it differs from retinaface models, so it's would be good to use more specific and clear name - retinaface.

### DIFF
--- a/demos/object_detection_demo/python/README.md
+++ b/demos/object_detection_demo/python/README.md
@@ -51,7 +51,7 @@ python3 object_detection_demo.py -h
 The command yields the following usage message:
 ```
 usage: object_detection_demo.py [-h] -m MODEL -at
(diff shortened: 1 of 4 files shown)


vladimir-dudnik (approved):
that's right, also should be updated in C++ demo (might be a separate PR if it is convenient)

ivikhrev:
> that's right, also should be updated in C++ demo (might be a separate PR if it is convenient)

in C++ demo it was initially done [right](https://github.com/openvinotoolkit/open_model_zoo/blob/09b4369ecef48c0bf75524e19d6cbae4541eed4b/demos/object_detection_demo/cpp/main.cpp#L52)

PR #2108, opened 2021-02-16
Add online ASR mode to Speech Recognition Demo

AlexeyKruglov:
This commit doesn't change quality metrics (they're exactly the same) and speed, but adds a new mode to the demo (and a new interface to the underlying ctcdecode_numpy package). "Online mode" is a mode with lower speech recognition latency: in "offline mode" recognition result was available only after processing whole file.


vladimir-dudnik:
Jenkins please retry a build

vladimir-dudnik:
@AlexeyKruglov please fix failed pre-commit checks

AlexeyKruglov:
That issue of WER 6.89% instead of 6.13% is not related to this PR (WER is equal with and without this PR) -- I checked that it is caused by the changes to audio_preprocessing.py in commit 28e5a72 on 29 Jan 2021 . With file audio_preprocessing.py before 28e5a72 (from `28e5a72^`) AC gives the same recognition results as in the last release. So this WER issue shall be fixed separately from this PR.

AlexeyKruglov:
@vladimir-dudnik Please restart the pipeline.

vladimir-dudnik:
Jenkins please retry a build

AlexeyKruglov:
@vladimir-dudnik Please restart the pipeline one more time. Missed one line.

vladimir-dudnik:
Jenkins please retry a build

Wovchena (changes requested):
That's it for now :)

demos/speech_recognition_demo/python/speech_recognition_demo.py
```diff
@@ -55,56 +65,69 @@ def get_profile(profile_name):
...
+        iter_wrapper = tqdm if not args.online else (lambda x: x)
+        for audio_iter in iter_wrapper(range(0, pcm_length, args.block_size)):
+            audio_block = np.frombuffer(wave_read.readframes(args.block_size * channel_num), dtype=np.int16).reshape((-1, channel_num))
+            if audio_block.shape[0] == 0:
+                break
+            audio_pos += audio_block.shape[0]
+            #
+            # It is possible to call stt.recognize_audio() for the whole audio files instead like this:
+            #   transcription1 = stt.recognize_audio(whole_audio1, sampling_rate)
+            #   transcription2 = stt.recognize_audio(whole_audio2, sampling_rate)
```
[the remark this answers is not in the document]
AlexeyKruglov: I think it was not obvious that one `DeepSpeechSeqPipeline` object (`stt`) can be used for multiple inferences. The demo only shows inference for one file (one utterance), but in real life users will need more than one. Initialization of this object is slow, so if a user mistakenly thinks that he needs to create a new `stt` for every utterance, it will hit performance hard.

demos/speech_recognition_demo/python/ctcdecode-numpy/ctcdecode_numpy/__init__.py
```diff
@@ -1,29 +1,100 @@
...
+    """
+    Non-batched sequential CTC + n-gram LM beam search decoder
+    """
+    def __init__(self, alphabet, beam_size, max_candidates=None,
+            cutoff_prob=1.0, cutoff_top_n=40,
+            scorer_lm_fname=None, alpha=0.75, beta=1.85,
+            text_decoder=None):
+        self.alphabet = alphabet
+        if text_decoder is None:
+            text_decoder = lambda x: x  # noqa: E731
```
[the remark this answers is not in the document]
AlexeyKruglov: I'd like to keep default values for `max_candidates` and `text_decoder` and keep them `None`:

1) `__init__.py` is an interface to ctcdecode-numpy package, that can be reused without our demo in case they'd like to reuse some of our code in their POC/product. Both these features (limiting # of candidates and decoding text from letter indices into strings) are optional, and None value corresponds to just not using them (and same about LM: not providing it on initialization). Without default, they'd have to decline each of the things they don't need explicitly.

2) About `None` vs. `lambda x: x`. I prefer to use None instead of complex things like lambda because if someone needs to add layers of wrappers above such interfaces, it is easier to use `None` as the default: otherwise same piece of code (like `lambda x: x`) would have to be repeated on the upper layers. And if it is to be changed for some reason, it'd have to be changed in all wrapper layers.

Regarding `0` vs `None` for `max_candidates`: I think `None` is semantically more understandable for the absence of limiting the number of candidates than `0`.

OK?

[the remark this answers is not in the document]
AlexeyKruglov: In addition, I reviewed the other default values in ctcdecode-numpy interfaces, and removed the default for `log_probs` in C++ in `CtcDecoderState::append()` and `CtcDecoderStateNumpy::append_numpy()`. Plus removed the defaults for `alpha` and `beta` in `SeqCtcLmDecoder`, and made them required in case LM is provided (when there's no LM, no `alpha` and `beta` are needed).

AlexeyKruglov:
Bump  (@vladimir-dudnik @eizamaliev)

vladimir-dudnik:
Jenkins please retry a build

vladimir-dudnik:
Jenkins please retry a build

vladimir-dudnik (approved):

vladimir-dudnik:
@AlexeyKruglov there are merge conflicts

demos/speech_recognition_demo/python/speech_recognition_demo.py
```diff
@@ -55,56 +65,74 @@ def get_profile(profile_name):
...
+                    time.sleep(to_wait)
+
+    transcription = stt.recognize_audio(None, sampling_rate, finish=True)
+    if args.online:
+        if transcription is not None and len(transcription) > 0:
+            print('\r' + transcription[0]['text'][-args.online_window:])
+
+    if not args.online:
+        # Don't show processing time in online mode because it's being slowed down by time.sleep()
+        print("Processing time (incl. loading audio, MFCC, RNN and beam search): {} s".format(timeit.default_timer() - start_time))
```
eizamaliev, line 131: Looks like `if ... else ...` case

demos/speech_recognition_demo/python/ctcdecode-numpy/ctcdecode_numpy/__init__.py
```diff
@@ -59,18 +128,18 @@ def decode(self, probs, seq_lens=None):
...
         if self._scorer is not None:
-            ctc_decode.reset_params(self._scorer, alpha, beta)
+            self._scorer.reset_params(alpha, beta)
 
-    def __del__(self):
-        if self._scorer is not None:
-            ctc_decode.delete_scorer(self._scorer)
+
+# For compatibility with ctcdecode_numpy version 0.2.0
+CTCBeamDecoder = BatchedCtcLmDecoder
```
Wovchena, line 145: We don't have a requirement to be backward compatible. You broke compatibility any way: at least with `labels`->`alphabet` renaming (I like the new name more :) ) I would completely remove `BatchedCtcLmDecoder` because it becomes dead since we won't ever run it.

I have perception that some things from underlying layers of implementation should also be removed following that principle, though I haven't figured out what happens here yet.
CC @vladimir-dudnik

demos/speech_recognition_demo/python/speech_recognition_demo.py
```diff
@@ -55,56 +65,74 @@ def get_profile(profile_name):
...
+            #   transcription3 = stt.recognize_audio(whole_audio3_block2, sampling_rate, finish=True)
+            # If you need intermediate features, you can call pipeline stage by stage like this:
+            #    audio_features = stt.extract_mfcc(whole_audio, sampling_rate)
+            #    character_probs = stt.extract_per_frame_probs(audio_features)
+            #    transcription = stt.decode_probs(character_probs)
+            #
+            partial_transcr = stt.recognize_audio(audio_block, sampling_rate, finish=False)
+            if args.online:
+                if partial_transcr is not None and len(partial_transcr) > 0:
+                    print('\r' + partial_transcr[0]['text'][-args.online_window:], end='')
```
Wovchena, line 119:
```suggestion
                    print('\r' + partial_transcr[0].text[-args.online_window:], end='')
```

demos/speech_recognition_demo/python/speech_recognition_demo.py
```diff
@@ -55,56 +65,74 @@ def get_profile(profile_name):
...
+                to_wait = play_start_time + audio_pos/sampling_rate - timeit.default_timer()
+                if to_wait > 0:
+                    time.sleep(to_wait)
+
+    transcription = stt.recognize_audio(None, sampling_rate, finish=True)
+    if args.online:
+        if transcription is not None and len(transcription) > 0:
+            print('\r' + transcription[0]['text'][-args.online_window:])
```
Wovchena, line 127:
```suggestion
            print('\r' + transcription[0].text[-args.online_window:])
```

demos/speech_recognition_demo/python/utils/audio_features.py
```diff
@@ -1,10 +1,74 @@
...
+            raise ValueError("window_size(ms)*sampling_rate(kHz) must be a power of two")
+
+        super().__init__(
+            block_len=self.stride_len, context_len=self.window_len - self.stride_len,
+            left_padding_len=0, right_padding_len=0,
+            padding_shape=(), cut_alignment=False)
+
+    def process_data(self, data, finish=False):
+        #if abs(sampling_rate - self.p['model_sampling_rate']) > self.p['model_sampling_rate'] * 0.1:
+        #    raise ValueError("Input audio file should be {} kHz mono".format(self.p['model_sampling_rate']/1e3))
```
Wovchena, line 31:
```suggestion
```
It looks you moved that check

Wovchena (changes requested):

demos/speech_recognition_demo/python/utils/deep_speech_seq_pipeline.py
```diff
@@ -0,0 +1,98 @@
...
+        """
+        if audio is not None:
+            if abs(sampling_rate - self.p['model_sampling_rate']) > self.p['model_sampling_rate'] * 0.1  or  (audio.shape + (1,))[1] != 1:
+                raise ValueError("Input audio file should be {} kHz mono".format(self.p['model_sampling_rate']/1e3))
+        return self.process_data(audio, finish=finish)
+
+    def extract_mfcc(self, audio, sampling_rate, finish=True):
+        # Audio feature extraction
+        if audio is not None:
+            if abs(sampling_rate - self.p['model_sampling_rate']) > self.p['model_sampling_rate'] * 0.1  or  (audio.shape + (1,))[1] != 1:
```
eizamaliev, line 84: Why do you allow 10% percent deviation in sample rate?

demos/speech_recognition_demo/python/ctcdecode-numpy/ctcdecode_numpy/__init__.py
vladimir-dudnik: agree, no reasons to keep a dead code

demos/speech_recognition_demo/python/speech_recognition_demo.py
```diff
@@ -55,56 +65,74 @@ def get_profile(profile_name):
...
-        args = build_argparser().parse_args()
-        profile = get_profile(args.profile)
+    args = build_argparser().parse_args()
+    profile = get_profile(args.profile)
+    if args.block_size is None:
+        sr = profile['model_sampling_rate']
+        args.block_size = round(sr*10) if not args.online else round(sr*profile['frame_stride_seconds']*16)
 
-        stt = DeepSpeechPipeline(
+    with Timer() as timer:
```
eizamaliev, line 74: Since this is only one usage of `Timer` class, it's the best time to delete it.

eizamaliev (changes requested):

vladimir-dudnik:
Jenkins please retry a build

demos/speech_recognition_demo/python/utils/ctc_decoder_seq_pipeline.py
```diff
@@ -0,0 +1,62 @@
...
+        self.p = deepcopy(profile)
+        self.beam_width = beam_width
+        self.max_candidates = max_candidates
+        self.online = online
+
+        alphabet = self.p['alphabet']
+        if alphabet is None:
+            self.alphabet = get_default_alphabet()
+        elif isinstance(alphabet, str):
+            self.alphabet = load_alphabet(alphabet)  # shall not include <blank> token
```
Wovchena, line 26: What is this case for? Default alphabet is always used. Plus you don't need a list of single chars as `self.alphabet`. It can be `str` (and `std::string`, I hope, for c++). As for the blank char `alphabet + ['']` in \_\_init__.py, it looks you can just leave it as `alphabet`.

Wovchena (changes requested):
Python only so far.

Demos shouldn't be written in a library style. Especially parts which don't follow what's written in `common` folder of `demos` dir. Only parts from `common` may be are going to become a library.

QuartzNet is almost there https://github.com/openvinotoolkit/open_model_zoo/pull/2054. It looks your PR makes it impossible to put QuartzNet into `speech_recognition_demo`. It doesn't seem to be possible to have online mode for QuartzNet because the model isn't rnn which one could stop and resume at any timestamp. Do you have advice how to incorporate QuartzNet into that new mode?

Wovchena:
@AlexeyKruglov, since QuartzNet can't be included into this demo, there are going to be two types of demos: online and offline.  This demo must run in online mode only. QuartzNet is going to be in offline_speech_recognition_demo. Later you can extend offline_speech_recognition_demo with mozilla-deepspeech.

vladimir-dudnik:
@AlexeyKruglov do you plan to continue work on this PR?

AlexeyKruglov:
@vladimir-dudnik  Yes, I'm already working on this PR for the past few days.

Wovchena:
1. There is how_are_you_doing.wav on CI available for testing now. I added a test for QuartzNet https://github.com/openvinotoolkit/open_model_zoo/pull/2054 and found that librosa requires additional library installation which matches with what docs say: https://pysoundfile.readthedocs.io/en/latest/#installation. OMZ avoids such dependencies. I’ve checked that for mozilla-deepspeech demo this is also the case: https://github.com/openvinotoolkit/open_model_zoo/pull/2332. Please rewrite the demo without librosa. You probably can refer to AC implementation.
2. Extend tests with mozilla-deepspeech-0.6.1 and mozilla-deepspeech-0.8.2 in your PR
3. Please address this comment https://github.com/openvinotoolkit/open_model_zoo/pull/2054#discussion_r609910692

demos/speech_recognition_demo/python/speech_recognition_demo.py
AlexeyKruglov: Done.  (However, for me this looks as a kind of syllepsis like "he took his hat and his leave" or "шёл дождь и два студента".)

demos/speech_recognition_demo/python/speech_recognition_demo.py
AlexeyKruglov: Done. This bug was added when I rebased my branch. So I fixed it in the commit that adds this code (when I was rebasing another time), not in the last commit containing the other changes -- to keep the state between commits working.

demos/speech_recognition_demo/python/utils/audio_features.py
AlexeyKruglov: Fixed.

demos/speech_recognition_demo/python/utils/deep_speech_seq_pipeline.py
AlexeyKruglov: 1. Because, for example, 1 Hz difference obviously doesn't affect anything, and I've seen such files in the past, which were 1 Hz off some standard frequency.  2. I think up to 10% change in sampling rate (which is equivalent to 10% = ~1 tone change in voice pitch and pace) is probably within the normal variation of speakers, and should be OK.

demos/speech_recognition_demo/python/speech_recognition_demo.py
AlexeyKruglov: Done. Removed Timer class.

demos/speech_recognition_demo/python/utils/audio_features.py
```diff
@@ -1,10 +1,74 @@
...
 
+from utils.pipelines import BlockedSeqPipelineStage
+
+
+class AudioFeaturesSeqPipelineStage(BlockedSeqPipelineStage):
+    def __init__(self, profile):
+        self.p = deepcopy(profile)
+        sampling_rate = self.p['model_sampling_rate']
+        self.window_len = round(sampling_rate * self.p['frame_window_size_seconds'])
+        self.stride_len = round(sampling_rate * self.p['frame_stride_seconds'])
```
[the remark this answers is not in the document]
AlexeyKruglov: OK, done. Removed self.window_len and self.stride_len, now computing them from the underlying block size.  (There was semantic difference, however, and now block size is hardcoded to one frame.)

demos/speech_recognition_demo/python/utils/pipelines.py
```diff
@@ -0,0 +1,165 @@
...
+        self._block_len = block_len
+        self._context_len = context_len
+        self._left_padding_len = left_padding_len
+        self._right_padding_len = right_padding_len
+        self._padding_shape = padding_shape
+        self._cut_alignment = cut_alignment
+        self._reset_state()
+
+    def _reset_state(self):
+        self._buffer = None  # None for buffer without left padding, [] for buffer with empty left padding
```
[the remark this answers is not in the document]
AlexeyKruglov: Replaced comment with `"None" before adding the first chunk of data, list of chunks otherwise`.

demos/speech_recognition_demo/python/utils/pipelines.py
```diff
@@ -0,0 +1,165 @@
...
+        self._buffer = [buffer]
+        self._buffer_len = buffer.shape[0]
+
+        # === Loop over blocks ===
+        # variables accepted from prev.stage: buffer, finalize
+        processed, buffer_skip_len = self._process_blocks(buffer, finish=finish)
+        if finish:
+            self._reset_state()
+        else:
+            # start_pos contains its value for the last iteration of the loop
```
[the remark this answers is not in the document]
AlexeyKruglov: Removed this comment: the code it referred to was already moved to `_process_blocks()`.

demos/speech_recognition_demo/python/utils/ctc_decoder_seq_pipeline.py
```diff
@@ -0,0 +1,62 @@
...
+        else:
+            self.alphabet = alphabet  # list-like
```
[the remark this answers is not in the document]
AlexeyKruglov: OK, not storing/exposing it anymore. And the fields in some other places.

demos/speech_recognition_demo/python/utils/rnn_seq_pipeline.py
```diff
@@ -0,0 +1,106 @@
...
+        assert self.exec_net is not None, "Need to call activate_model(device) method before process_data(...)"
+        if data is not None:
+            assert len(data.shape) == 2
+        return super().process_data(data, finish=finish)
+
+    def _process_block(self, mfcc_features, finish=False):
+        assert mfcc_features.shape[0] == self._block_len + self._context_len, "Wrong data length: _process_block() accepts a single block of data"
+
+        # Create a view into the array with overlapping strides to simulate convolution with FC
+        mfcc_features = np.lib.stride_tricks.as_strided(  # TODO: replacing with conv1d may improve speed a little
```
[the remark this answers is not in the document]
AlexeyKruglov: Replaced "TODO" with "NB" and extended a little.

demos/speech_recognition_demo/python/utils/rnn_seq_pipeline.py
```diff
@@ -0,0 +1,106 @@
...
+            mfcc_features,
+            (self._block_len, self._context_len + 1, self.p['num_mfcc_dct_coefs']),
+            (mfcc_features.strides[0], mfcc_features.strides[0], mfcc_features.strides[1]),
+            writeable = False,
+        )
+
+        if self._rnn_state is None:
+            state_h = np.zeros((1, 2048))
+            state_c = np.zeros((1, 2048))
```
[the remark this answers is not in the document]
AlexeyKruglov: Done.

demos/speech_recognition_demo/python/utils/rnn_seq_pipeline.py
```diff
@@ -0,0 +1,106 @@
...
+            self.p['in_state_h']: state_h,
+            self.p['in_data']: [mfcc_features],
+        })
+
+        state_c = infer_res[self.p['out_state_c']]
+        state_h = infer_res[self.p['out_state_h']]
+        self._rnn_state = (state_h, state_c)
+
+        probs = infer_res[self.p['out_data']].squeeze(1)  # despite its name, this is actually probabilities after softmax
+        return probs
```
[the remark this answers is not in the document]
AlexeyKruglov: Misplaced comment again. Moved it into profiles.py. (It referred to network output being misleadingly called "logits" despite not being logarithmic.)

demos/speech_recognition_demo/python/utils/rnn_seq_pipeline.py
```diff
@@ -0,0 +1,106 @@
...
+        # Plugin initialization for specified device and load extensions library if specified
+        for extension_device, extension_fname in ie_extensions:
+            if extension_fname is None:
+                continue
+            self.ie.add_extension(extension_path=extension_fname, device_name=extension_device)
+
+        # Read IR
+        self.net = self.ie.read_network(model=model_xml_fname, weights=model_bin_fname)
+
+    def activate_model(self, device):
```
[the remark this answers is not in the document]
AlexeyKruglov: OK, done.

demos/speech_recognition_demo/python/utils/rnn_seq_pipeline.py
```diff
@@ -0,0 +1,106 @@
...
+        self.net = self.exec_net = None
+        self.default_device = device
+
+        self.ie = ie if ie is not None else IECore()
+        self._load_net(model, model_bin_fname=model_bin, device=device, ie_extensions=ie_extensions)
+
+        if device is not None:
+            self.activate_model(device)
+
+    def _load_net(self, model_xml_fname, model_bin_fname=None, ie_extensions=[], device='CPU', device_config=None):
```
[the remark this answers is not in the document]
AlexeyKruglov: 1. Removed the `_load_net` method.
2. I thought having `--cpu_extensions` command line option was a requirement -- many other demos have it for no apparent reason. Removed that option and everything related to extensions.
3. Removed `device_config`.  `device` however is a command line option, kept `device`.
4. Removed `weights` (and `model_bin_fname`) everywhere.

demos/speech_recognition_demo/python/utils/profiles.py
```diff
@@ -0,0 +1,50 @@
...
+        'in_state_h': 'previous_state_h',
+        'out_state_c': 'cudnn_lstm/rnn/multi_rnn_cell/cell_0/cudnn_compatible_lstm_cell/BlockLSTM/TensorIterator.2',
+        'out_state_h': 'cudnn_lstm/rnn/multi_rnn_cell/cell_0/cudnn_compatible_lstm_cell/BlockLSTM/TensorIterator.1',
+        'in_data': 'input_node',
+        'out_data': 'logits',
+        'log_probs': False,
+    },
+}
+PROFILES['mds08x_en'] = PROFILES['mds07x_en']
+PROFILES['mds09x_en'] = PROFILES['mds07x_en']
```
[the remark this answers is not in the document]
AlexeyKruglov: OK. And changed `PROFILES['mds07x_en']` initialization to use `dict.update()` method on `PROFILES['mds06x_en'].copy()`.

demos/speech_recognition_demo/python/utils/profiles.py
```diff
@@ -0,0 +1,50 @@
+#
+# Copyright (C) 2019-2021 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+#
+PROFILES = {
+    'mds06x_en': {
+        'alphabet': None,  # the default alphabet
```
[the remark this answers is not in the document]
AlexeyKruglov: OK, defined alphabet here explicitly, no default alphabet now.

demos/speech_recognition_demo/python/utils/ctc_decoder_seq_pipeline.py
AlexeyKruglov: 1. Removed default value for the alphabet. Now either filename, or a list.
2. Loading alphabet from a file because: a) we can read alphabet file in the same format as Mozilla DeepSpeech did, b) users may train their own model and provide its parameters in profile file.
3. Regarding blank symbol. There're actually two alphabets: one in CTC decoder (with blank), and another in LM (without blank). I think it was easier to provide each component with its correct alphabet, than to think where I should add 1, and where I shouldn't.

demos/speech_recognition_demo/python/ctcdecode-numpy/ctcdecode_numpy/__init__.py
AlexeyKruglov: 1. Removed `CTCBeamDecoder` alias for `BatchedCtcLmDecoder`.
2. `BatchedCtcLmDecoder` is used in Accuracy Checker, so the code is not dead.
3. The underlying C++ implementation is a batching wrapper (based on ThreadPool) over the same C++ class `CtcDecoderState` (which is single-threaded).

demos/speech_recognition_demo/python/utils/pipelines.py
```diff
@@ -0,0 +1,165 @@
...
+        """
+            Returns:
+        numpy.ndarray (any stage) or other type (for the last stage only), processing result
+          OR
+        None (empty data)
+        """
+        pass
+
+
+class SeqPipeline(SeqPipelineStage):
```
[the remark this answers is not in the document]
AlexeyKruglov: 1. Moved description of `profile` arg from this docstring into `profiles.py`. Removed the three stage methods from `DeepSpeechSeqPipeline` and removed `SeqPipeline` class (I assumed it to be used for QuartzNet pipeline). Kept only `__init__()` and `recognize_audio()`.
2. I think `speech_recognition_demo.py` will contain too much details without `DeepSpeechSeqPipeline`. 
3. Reordered parameters in `profile.py` and separated by stage explicitly with the comments.
4. I think passing each of them explicitly to stage `__init__()` will lead to too many arguments. Plus now there's separation between parameters which are fixed for the model (`profile`) vs. the parameters that can be changed (like `beam_width`, `max_candidates`, `device`, etc.).

demos/speech_recognition_demo/python/utils/pipelines.py
```diff
@@ -0,0 +1,165 @@
...
+        for start_pos in range(self._context_len, buffer.shape[0] - self._block_len + 1, self._block_len):
+            block = buffer[start_pos - self._context_len:start_pos + self._block_len]
+            processed.append(self._process_block(block, finish=finish and start_pos + self._block_len >= buffer.shape[0]))
+        assert not self._cut_alignment or processed[-1].shape[0] == self._block_len, "Networks with stride != 1 are not supported"
+        # Here start_pos is its value on the last iteration of the loop
+        buffer_skip_len = start_pos + self._block_len - self._context_len
+        return processed, buffer_skip_len
+
+    def _process_block(self, block, finish=False):
+        raise NotImplementedError("_process_block() should have been implemented in case inherited _process_blocks() is used")
```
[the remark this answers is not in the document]
AlexeyKruglov: OK, done.

demos/speech_recognition_demo/python/utils/ctc_decoder_seq_pipeline.py
```diff
@@ -0,0 +1,62 @@
...
+
+        alphabet_decoder = CtcdecoderAlphabet(self.alphabet)
+        self._decoder = SeqCtcLmDecoder(self.alphabet, self.beam_width, max_candidates=max_candidates,
+            scorer_lm_fname=lm, alpha=self.p['alpha'], beta=self.p['beta'],
+            text_decoder=lambda symbols: alphabet_decoder.decode(symbols))
```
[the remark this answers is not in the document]
AlexeyKruglov: It can, but this way seemed easier to me, because `SeqCtcLmDecoder` already loops over candidates.

demos/speech_recognition_demo/python/ctcdecode-numpy/ctcdecode_numpy/ctc_beam_search_decoder.h
```diff
@@ -9,23 +9,80 @@
...
+
+class CtcDecoderState {
+public:
+  CtcDecoderState();
+  CtcDecoderState(const CtcDecoderState&) = delete;  // because copying PathTrie is not inplemented
+  CtcDecoderState(CtcDecoderState&&) = default;
+  CtcDecoderState& operator=(const CtcDecoderState& src) = delete;
+  CtcDecoderState& operator=(CtcDecoderState&& src) = default;
+
+  void init(
```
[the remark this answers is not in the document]
AlexeyKruglov: This is actually close to mutable vs. unmutable objects. Two reasons: 1. I thought mutable-like objects are more convenient to organize thread pools with changing parameters (like LMs, alphabets, etc) -- they can be just held in a vector without any pointers. This can potentially be useful here. 2. In case a child class doesn't add new parameters, `init()` can be inherited.    ( And another general reason: Passing parameters to C++ constructor is inconvenient in some cases where user may be expected to want to conditionally initialize class inside `if/else` without pointers. )

AlexeyKruglov:
> @AlexeyKruglov, since QuartzNet can't be included into this demo,

@Wovchena  Actually QuartzNet could be included if its neural network stage accumulated all the chunks of data, then after the last chunk processed all the data and provided the result. There was a comment (removed in the last commit) in the docstring of `SeqPipelineStage` class about that.

demos/speech_recognition_demo/python/utils/ctc_decoder_seq_pipeline.py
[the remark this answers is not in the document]
Wovchena: To me
> `SeqCtcLmDecoder` already loops over candidates

is actually the argument supporting the idea that `SeqCtcLmDecoder` should map the symbols on its own.

> this way seemed easier

It gets more complicated: `CtcDecoderSeqPipelineStage` needs to know an additional detail about `SeqCtcLmDecoder` and provide `text_decoder` in a way `SeqCtcLmDecoder` expects.

demos/speech_recognition_demo/python/utils/deep_speech_seq_pipeline.py
```diff
@@ -0,0 +1,50 @@
...
+    def __init__(self, model, lm=None, beam_width=500, max_candidates=None,
+            profile=None, ie=None, device='CPU', online_decoding=False):
+        """
+            Args:
+        model (str), filename of IE IR .xml file of the network
+        lm (str), filename of LM (language model)
+        beam_width (int), the number of prefix candidates to retain during decoding in beam search (default 500)
+        max_candidates (int), limit the number of returned candidates; None = do not limit (default None)
+        profile (dict): a dict with pre/post-processing parameters, see profiles.py
+        ie (IECore or None), IECore object to run NN inference with.  Default is to use ie_core_singleton module.
```
Wovchena, line 25: We are tending IECore to always be provided from `main()`. For example https://github.com/openvinotoolkit/open_model_zoo/blob/9ca5dbeff80464bf5728e8be25daedfe9a9208d7/demos/common/python/pipelines/async_pipeline.py#L23 Please, do the same.

demos/speech_recognition_demo/python/utils/deep_speech_seq_pipeline.py
```diff
@@ -0,0 +1,50 @@
...
+            (default None)
+        device (str), inference device for IE, passed here to 1. set default device, and 2. check supported node types
+            in the model load; None = do not check (default 'CPU')
```
Wovchena, line 28:
```suggestion
        device (str), inference device for IE
```

demos/speech_recognition_demo/python/utils/deep_speech_seq_pipeline.py
```diff
@@ -0,0 +1,50 @@
...
+        online_decoding (bool), set to True to return partial decoded text after every input data piece (default False)
+        """
+        assert profile is not None, "profile argument must be provided"
```
Wovchena, line 31: If `profile` must be provided, why assign default value. This is also about `RnnSeqPipelineStage` and `CtcDecoderSeqPipelineStage`.

demos/speech_recognition_demo/python/utils/deep_speech_seq_pipeline.py
```diff
@@ -0,0 +1,50 @@
...
+    def recognize_audio(self, audio, sampling_rate, finish=True):
+        """
+        Run a segment of audio through ASR pipeline.
+        Use finish=True (default) to run recognition once for a whole utterance.
+        Use finish=False for online recognition, provinging data cut into segments and getting updated recognition
+        result after each segment. Set finish=True for the last segment to get the final result and reset pipeline state.
+        """
+        if audio is not None:
+            if abs(sampling_rate - self.p['model_sampling_rate']) > self.p['model_sampling_rate'] * 0.1  or  (audio.shape + (1,))[1] != 1:
+                raise ValueError("Input audio file should be {} kHz mono".format(self.p['model_sampling_rate']/1e3))
```
Wovchena, line 47: `sampling_rate` is passed here just to let `recognize_audio()` verify it. This is strange approach. Move the check to the reset
```py
        assert sample_width == 2, "Only 16-bit WAV PCM supported"
        assert compression_type == 'NONE', "Only linear PCM WAV files supported"
        assert channel_num == 1, "Only mono WAV PCM supported"
```

demos/speech_recognition_demo/python/utils/rnn_seq_pipeline.py
```diff
@@ -0,0 +1,91 @@
...
+        mfcc_features = np.lib.stride_tricks.as_strided(
+            mfcc_features,
+            (self._block_len, self._context_len + 1, self.p['num_mfcc_dct_coefs']),
+            (mfcc_features.strides[0], mfcc_features.strides[0], mfcc_features.strides[1]),
+            writeable = False,
+        )
+
+        if self._rnn_state is None:
+            state_h = np.zeros(self.exec_net.get_exec_graph_info().input_info[self.p['in_state_h']].input_data.shape)
+            state_c = np.zeros(self.exec_net.get_exec_graph_info().input_info[self.p['in_state_c']].input_data.shape)
```
Wovchena, line 76:
```suggestion
            state_h = np.zeros(self.exec_net.input_info[self.p['in_state_h']].input_data.shape)
            state_c = np.zeros(self.exec_net.input_info[self.p['in_state_c']].input_data.shape)
```

Wovchena (commented):
The similar idea about QuartzNet inclusion was discussed in my team. It was rejected.

https://github.com/openvinotoolkit/open_model_zoo/pull/2108/commits/4843b0680ea8724e9135f0a0b2bd4e232728aa6c - cool idea!

demos/speech_recognition_demo/python/utils/audio_features.py
```diff
@@ -1,9 +1,79 @@
...
+
+# Workaround to import librosa on Linux without installed libsndfile.so
+try:
+    import librosa
+except OSError:
+    import sys
+    import types
+    sys.modules['soundfile'] = types.ModuleType('fake_soundfile')
+    import librosa
+    del sys.modules['soundfile']
```
Wovchena, line 17:
```suggestion
```
https://github.com/openvinotoolkit/open_model_zoo/pull/2054#discussion_r622920988

Wovchena:
@AlexeyKruglov, @vladimir-dudnik asked to remind that the release is nearby :)

demos/speech_recognition_demo/python/utils/deep_speech_seq_pipeline.py
AlexeyKruglov: OK, now `profile` is the first argument in all pipeline stages.

demos/speech_recognition_demo/python/utils/deep_speech_seq_pipeline.py
AlexeyKruglov: `sampling_rate` checking is done inside because supplying an audio with wrong sampling rate may more-or-less work, but with impaired quality. With other parameters wrong, it either works, or it doesn't.
If some user just copies our files and invokes ASR from his code with a wrong sampling rate (say, the usual 44.1 kHz or 48 kHz instead of 16 kHz), he'll just have somewhat impaired quality, and wouldn't realize he's doing it wrong.

demos/speech_recognition_demo/python/utils/deep_speech_seq_pipeline.py
AlexeyKruglov: Fixed. And fixed this in the other docstrings.

demos/speech_recognition_demo/python/utils/deep_speech_seq_pipeline.py
AlexeyKruglov: Done, now `ie` is the second arg (next after `profile`).

demos/speech_recognition_demo/python/ctcdecode-numpy/ctcdecode_numpy/__init__.py
```diff
@@ -1,29 +1,97 @@
...
+        self.max_candidates = max_candidates or 0
+        self.cutoff_prob = cutoff_prob
+        self.cutoff_top_n = cutoff_top_n
+
+        if scorer_lm_fname is not None:
+            assert alpha is not None and beta is not None, "alpha and beta arguments must be provided to use LM"
+            self.lm_scorer = impl.ScorerYoklm(alpha, beta, scorer_lm_fname, alphabet)
+        else:
+            self.lm_scorer = None
+        self.decoder_state = impl.CtcDecoderStateNumpy().init(
```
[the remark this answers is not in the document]
AlexeyKruglov: Oops. Seems like I didn't reinstall it when I tested it.
Fixed this in the commit that introduced this bug (to have working code after every commit).

demos/speech_recognition_demo/python/utils/rnn_seq_pipeline.py
AlexeyKruglov: Done.

demos/speech_recognition_demo/python/utils/deep_speech_seq_pipeline.py
Wovchena: There is nothing coupling audio and sampling_rate passed to `recognize_audio()`. `recognize_audio()` has no way to verify the two objects are connected. Basically `recognize_audio()` requires an audio and just a number around 16 kHz. One still may pass wrong audio and after noticing that there is that check put unrelated constant. At least that's how I usually do stuff when something doesn't work :)

demos/speech_recognition_demo/python/utils/audio_features.py
AlexeyKruglov: I checked that deleting `sys.modules['soundfile']` affects importing `soundfile` in the other modules (by adding `raise OSError()` into `try:` and adding `import soundfile; print(dir(soundfile))` into `speech_recognition_demo.py`). With deletion the correct `soundfile` is imported, and without deletion the fake `soundfile` is imported.

`soundfile` may be used for loading audio in the user's code or inside the packages he imports. When importing it people sometimes check for errors to give a comprehensible error message. If we didn't clean up, we'd break this. An example of such checking inside `fairseq`: https://github.com/pytorch/fairseq/blob/425c36eafff535fe7337f8bdd5ace22ebacc78cb/fairseq/data/audio/audio_utils.py#L57 . And even without checking, instead of `ImportError` in proper place we'll have some other incomprehensible error in another place.

`librosa` is actually left in a broken state this way. I think it's more-or-less OK, because we're leaving only one package in a partially broken state, as opposed to leaving completely broken `soundfile`.  If needed, `librosa` can be cleaned up by deleting three cached modules: `librosa`, `librosa.core` and `librosa.core.audio` -- I thought that it may look too complicated for its goal and may be brittle to changes inside `librosa`.

demos/speech_recognition_demo/python/utils/ctc_decoder_seq_pipeline.py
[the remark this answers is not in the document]
AlexeyKruglov: Oh, I misunderstood the question in the last answer.

`text_decoder` is passed because users may want no postprocessing or some other kind of post-processing to the strings: for example, 1) UTF-8 decoding for Chinese or 2) some may need raw symbol indices as numbers, which also comes in two versions: 2.1) as views into one big numpy array, or 2.2) as separate copies of these subarrays.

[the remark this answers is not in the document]
Wovchena: `CtcDecoderSeqPipelineStage` doesn't have an argument similar to `text_decoder`. Thus the argument 
> users may want no postprocessing or some other kind of post-processing to the strings

conflicts with current implementation. (I'm not encouraging you to introduce such arg for `CtcDecoderSeqPipelineStage` :) )

1) Are you sure C++ backend is fine with UTF-8?
2) It seems you can completely split `SeqCtcLmDecoder` and `text_decoder`: calling them sequentially: `text_decoder(self._decoder.decode())`
It conflicts with
> this way seemed easier to me, because `SeqCtcLmDecoder` already loops over candidates.

But such approach should split scopes in a cleaner way.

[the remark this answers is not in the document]
Wovchena: Again I'm not encouraging you to follow 2. :) Just saying there should be put much more thoughts to actually make the code reusable.

demos/speech_recognition_demo/python/utils/deep_speech_seq_pipeline.py
AlexeyKruglov: I think it is obvious from the arg name that it's the sampling rate, and not just some unrelated number. If someone looks at the code that checks its value, he'd notice the error message "Input audio file should be {} kHz mono" suggesting that it's the sampling rate of the audio. And if he doesn't look inside, he's expected to provide values to `audio` and `sampling_rate` according to their names and their usage in the main file.

Wovchena: Ok. Pleas update speech_recognition_offline_demo here as well to follow the same approach.

demos/speech_recognition_demo/python/utils/audio_features.py
Wovchena: > With deletion the correct `soundfile` is imported, and without deletion the fake `soundfile` is imported.

That's what we want. If demo import can't find libsndfile it is reasonble to assume that other similar imports also won't find it. You never know if such imports are going to be used for file reading or a case similar to ours.

> soundfile may be used for loading audio in the user's code or inside the packages he imports.

The solution is a hack anyway. For example I could say that one could copypaste the demo code with that import and than try reading file somhere in `audio_features.py`. The problem is it was never defined how they are suppoed to reuse the code because OMZ demos implementation is not expected to be used blindly. It's not up to us to change the whole OMZ position in this PR.

I'm not much into this discussion, I would prefer to leave it up to you and @IRDonch :)

Wovchena:
To remind
> This demo must run in online mode only.

> Please address this comment #2054 (comment)

Wovchena:
Jenkins ok to test

demos/speech_recognition_demo/python/utils/audio_features.py
Wovchena: Just realized, if Roman ignores this, it is not defined what to do next. You can go with your solution (and update speech_recognition_offline_demo the same way) unless there are complains.

vladimir-dudnik:
11:36:09  ./demos/speech_recognition_demo/python/utils/rnn_seq_pipeline.py:8:1: F401 'os.path' imported but unused

demos/speech_recognition_demo/python/utils/audio_features.py
IRDonch: I don't think we need to worry about what users will do in their own code. This hack is clearly not something that should be used in production code, whether with the deletion or without. We're only doing it so that we can spare the users the trouble of installing `libsndfile` on their computer just to run the demo.

Also FWIW, the proper solution to this would be to open an issue on the librosa tracker and ask the maintainers nicely to make the soundfile dependency optional (i.e. put it in an extra).

demos/speech_recognition_demo/python/utils/ctc_decoder_seq_pipeline.py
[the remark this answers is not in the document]
AlexeyKruglov: OK, removed `text_decoder` callback. And in addition removed `CtcdecoderAlphabet` class -- it was redundant.

demos/speech_recognition_demo/python/utils/deep_speech_seq_pipeline.py
AlexeyKruglov: OK.

demos/speech_recognition_demo/python/utils/audio_features.py
AlexeyKruglov: OK then, removed the clean-up.

AlexeyKruglov:
@Wovchena 
> This demo must run in online mode only.

What do you mean by "online mode" here? The word "online" doesn't have one strict definition and may mean different things I think, and the opposite of your "offline demo" is a streaming demo. This demo always runs in streaming mode. But the demo can show two use-cases: 

  1) transcribing file in streaming mode (with large block size and without partially decoded text)
  2) voice typing use-case (short block size, with artificial delays in data feeding to simulate real-time recording and with partially decoded text).

Both of them are streaming (which means we don't need to store network inputs and outputs and audio data in memory for the whole file).

If by "online" you mean not just "non-offline", but particularly the 2nd mode (the one activated with `--online` argument), why it shouldn't show the first use-case (which I considered the main one -- as we should perform exactly the same as Mozilla on the second one since it's real-time, and may compete on the first one)?

Wovchena:
I think this is reasonable. Can you just add `streaming` to the name then.

CC @vladimir-dudnik, @IRDonch, to check you are fine with that

vladimir-dudnik:
> I think this is reasonable. Can you just add `streaming` to the name then.
> 
> CC @vladimir-dudnik, @IRDonch, to check you are fine with that

@Wovchena I'm ok with name `streaming mode`

Wovchena:
@vladimir-dudnik, wasn't that your and @IRDonch's call to rename to `offline` and `online`?

It was going to be renamed anyway. My question is if it is fine to replace `online` with `streaming` and keep current functionality.

vladimir-dudnik:
> @vladimir-dudnik, wasn't that your and @IRDonch's call to rename to `offline` and `online`?
> 
> It was going to be renamed anyway. My question is if it is fine to replace `online` with `streaming` and keep current functionality.

ok, then. My vote would be for `streaming` word vs `online` in the name

AlexeyKruglov:
1. Rebased everything to the latest `develop` (no conflicts, no changes) to avoid possible merge conflicts
2. Renamed speech_recognition_demo to speech_recognition_streaming_demo everywhere
3. Renamed `utils` directory in the demo to `asr_utils` to make it a little more descriptive

vladimir-dudnik:
@AlexeyKruglov to be honest, your #3 point looks an overkill. There is only one utils folder in demo, why you need this clarification about asr_utils?

demos/speech_recognition_streaming_demo/python/README.md
```diff
@@ -12,12 +12,17 @@ The application accepts
 * n-gram language model file in kenlm quantized binary format, and
 * an audio file in PCM WAV 16 kHz mono format.
 
-After computing audio features, running a neural network to get per-frame character probabilities, and CTC decoding, the demo prints the decoded text together with the timings of the processing stages.
+The application has two modes:
+
+ * *Offline mode* (default). Audio data is fed in 10 second chunks into a pipeline of: computing audio features, running a neural network to get per-frame character probabilities, and CTC decoding. After processing the whole file, the demo prints the decoded text and the time spent.
```
vladimir-dudnik, line 17: for `streaming` demo having default `offline` mode looks strange

demos/speech_recognition_streaming_demo/python/README.md
```diff
@@ -1,8 +1,8 @@
-# Speech Recognition Python\* Demo
+# Speech Recognition Streaming Python\* Demo
```
vladimir-dudnik, line 1: @AlexeyKruglov I tend to think that we have to change this demo name to mozilla_deepspeech_demo. That will distinguish this demo from already presented at OpenVINO `offline_speech_recognition_demo` and `live_speech_recognition_demo` dedicated for Kaldi models and highlight the fact that demo work with Mozilla deepspeech models only.

@Wovchena having both `offline_speech_recognition_demo` in OpenVINO and `speech_recognition_offline_demo` indicate we have an issue.

@dorloff I think demo which work with Kaldi models only should have `kaldi` modifier in their name

vladimir-dudnik (changes requested):

demos/speech_recognition_streaming_demo/python/README.md
Wovchena: Makes sense. For demos this breaks task type based splitting, which isn't that bad to me.

Wovchena:
> to be honest, your #3 point looks an overkill. There is only one utils folder in demo, why you need this clarification about asr_utils?

This may start to conflict with `utils` from `common`, for now we have `utils` only for C++, but this may change. If `offline_speech_recognition_demo` supports `deepspeech`, `asr_utils` will be likely to move into `common` which will make the conflict even more pronounced.
I think it doesn't harm to be explicit after all.

Wovchena:
Jenkins please retry a build

vladimir-dudnik:
@AlexeyKruglov there are merge conflicts

demos/speech_recognition_streaming_demo/python/README.md
AlexeyKruglov: Renamed to `speech_recognition_deepspeech_demo`.

demos/speech_recognition_streaming_demo/python/README.md
AlexeyKruglov: To avoid confusion I renamed "offline" and "online" modes in demo to "normal" and "simulated real-time" modes.

Ex-"offline" mode is still streaming. "Streaming" means the all pipeline stages are executed for each new chunk of data, as opposed to the non-streaming pipeline processing (not employed in this demo), where first stage is executed for all data, then second stage for all data, etc. The difference between "(ex-)offline streaming" and "non-streaming" implementations is: 1) no need to accumulate input data in memory for the long files -- the processed input data can be dismissed, 2) smaller latency between providing the last chunk of data and getting the final result for the long files (e.g. few minutes audio length).

And the demo was renamed to "Speech Recognition DeepSpeech Demo".

AlexeyKruglov:
1. Rebased to the latest develop and removed the last 3 commits.
2. Re-renamed to `speech_recognition_deepspeech_demo`. Renamed `utils` directory to `asr_utils` in the demo.
3. Renamed "offline" and "online" modes in the demo to "normal" and "simulated real-time" modes to avoid confusion.

vladimir-dudnik (approved):

Wovchena (approved):

eizamaliev (approved):

vladimir-dudnik:
Jenkins please retry a build

PR #2117, opened 2021-02-18
Fix computation of audio features in AC for mozilla-deepspeech-* that was broken by 28e5a72

AlexeyKruglov:
Release 2021.3 had WER=6.13% for mozilla-deep-speech-0.8.2 (beam=500), 29 Jan 2021 develop commit deteriorated WER to 6.89%, this PR corrects this: now WER=6.13% again (for beam=500 with accuracy-check-fast.yml config). In addition there're code style corrections + removed an unused config option.


eaidova:
Jenkins ok to test

eaidova (approved):

PR #2120, opened 2021-02-19
very last fix for text-spotting-0004

Ilya-Krylov:
one more fix to match AC metric to original one

### DIFF
--- a/tools/accuracy_checker/accuracy_checker/evaluators/custom_evaluators/text_spotting_evaluator.py
+++ b/tools/accuracy_checker/accuracy_checker/evaluators/custom_evaluators/text_spotting_evaluator.py
@@ -25,7 +25,6 @@ from ...adapters import create_adapter
 from ...config import ConfigError
 from ...launcher import create_launcher
 from ...utils import contains_all, extract_image_representations, get_path
-from ...postprocessor import PostprocessingExecutor
 from ...progress_reporters import ProgressReporter
 from ...logging import print_info
 
--- a/tools/accuracy_checker/accuracy_checker/evaluators/custom_evaluators/text_spotting_evaluator.py
+++ b/tools/accuracy_checker/accuracy_checker/evaluators/custom_evaluators/text_spotting_evaluator.py
@@ -25,6 +25,7 @@ from ...adapters import create_adapter
 from ...config import ConfigError
 from ...launcher import create_launcher
 from ...utils import contains_all, extract_image_representations, get_path
+from ...postprocessor import PostprocessingExecutor
 from ...progress_reporters import ProgressReporter
 from ...logging import print_info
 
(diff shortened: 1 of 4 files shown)


Ilya-Krylov:
I am sorry that there are a log PRs from my side, but this is one more fix to match AC metric to original one
@vladimir-dudnik @eaidova

eaidova (approved):

tools/accuracy_checker/accuracy_checker/metrics/text_detection.py
```diff
@@ -443,7 +443,8 @@ def update(self, annotation, prediction):
         gt_polygons = list(map(polygon_from_points, annotation.points))
         gt_texts = list(annotation.description)
 
-        prediction_polygons = list(map(polygon_from_points, prediction.points))
+        prediction_points_int = np.array(prediction.points, dtype=np.int32)
```
eaidova, line 446: I think these changes may affect a lot of other models (text-detection-*, ctpn, horizontal-text-detection). It may be risky to change it before release, we need also check these models too and look how your changes affect them and probably update docs (otherwise, we can get unexpected accuracy regression)

eaidova: @Ilya-Krylov  can we made casting to int optional or move to postprocessing  (if I right remember, we have cast_to_int and it should works with TextDetectionPrediction?)

Ilya-Krylov: thanks for the hint, fixed

tools/accuracy_checker/accuracy_checker/evaluators/custom_evaluators/text_spotting_evaluator.py
```diff
@@ -35,12 +36,12 @@ def softmax(x):
...
 
 class TextSpottingEvaluator(BaseEvaluator):
-    def __init__(self, dataset_config, launcher, model):
+    def __init__(self, dataset_config, launcher, model, postprocessor):
         self.dataset_config = dataset_config
         self.preprocessing_executor = None
         self.preprocessor = None
         self.dataset = None
-        self.postprocessor = None
+        self.postprocessor = postprocessor
```
eaidova, line 44: please revert changes in constructor and config parsing,
it will be initialized in https://github.com/openvinotoolkit/open_model_zoo/blob/024a9baba4a8150a5b08e7a5620e20a40cf5c323/tools/accuracy_checker/accuracy_checker/evaluators/custom_evaluators/text_spotting_evaluator.py#L234

This method called in process_dataset if we have not initialize dataset yet, so postprocessing should be initialized automatically without explicit setting in constructor
it made for opportunity to change dataset and its attributes (metrics, preprocessing, postprocessing) dynamically during quantization

Ilya-Krylov: fixed

vladimir-dudnik (approved):

Anna-Mironova (approved):

eaidova:
Jenkins please retry a build

PR #2126, opened 2021-02-19
Micro changes to simplify deblurring_demo py

akorobeinikov:
Add demo to list of demos and copy preprocessing and postprocessing from c++ image_processing demo.

### DIFF
--- a/demos/common/python/models/deblurring.py
+++ b/demos/common/python/models/deblurring.py
@@ -21,13 +21,12 @@ class Deblurring(Model):
     def __init__(self, ie, model_path, input_image_shape):
         super().__init__(ie, model_path)
         self.block_size = 32
-        self.input_shape = input_image_shape
-        self.calculate_new_shape()
+        self.reshape(input_image_shape)
         self.input_blob_name = self.prepare_inputs()
         self.output_blob_name = self.prepare_outputs()
 
-    def calculate_new_shape(self):
-        h, w, _ = self.input_shape
+    def reshape(self, base_shape):
+        h, w, _ = base_shape
         new_height = math.ceil(h / self.block_size) * self.block_size
         new_width = math.ceil(w / self.block_size) * self.block_size
 
@@ -50,11 +49,6 @@ class Deblurring(Model):
             self.n, self.c, self.h, self.w = input_size
         else:
             raise RuntimeError("3-channel 4-dimensional model's input is expected")
-
-        self.pad_params = {'mode': 'constant',
-                           'constant_values': 0,
-                           'pad_width': ((0, self.h - self.input_shape[0]), (0, self.w - self.input_shape[1]), (0, 0))
-                           }
 
         return input_blob_name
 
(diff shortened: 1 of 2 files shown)


demos/common/python/models/deblurring.py
```diff
@@ -20,15 +20,16 @@
...
         self.input_blob_name = self.prepare_inputs()
         self.output_blob_name = self.prepare_outputs()
 
-    def calculate_new_shape(self, shape):
-        h, w, _ = shape
-        block_size = 32
-        new_height = math.ceil(h / block_size) * block_size
-        new_width = math.ceil(w / block_size) * block_size
+    def calculate_new_shape(self):
+        h, w, _ = self.input_shape
```
Wovchena, line 30: 1. `Deblurring` shouldn't store `input_image_shape`'s third value because the last value isn't used.
2. `postprocessing()` is meant to revert what was done in `preprocessing()` in terms of resize. `postprocessing()` should resize image back to original size, which you've stored for yourself in `preprocessing()` in `meta`. That means you don't need `self.input_shape` field at all: `meta` should be used instead. BTW, I just noticed, it would be great if you cleaned `meta`: `resized_image.shape` isn't used as well as `image.shape[2]`.

demos/common/python/models/deblurring.py
```diff
@@ -74,15 +75,10 @@ def prepare_outputs(self):
...
-                          'constant_values': 0,
-                          'pad_width': ((0, self.h - image.shape[0]), (0, self.w - image.shape[1]), (0, 0))
-                          }
-            resized_image = np.pad(image, **pad_params)
-        else:
-            resized_image = cv2.resize(image, (self.w, self.h))
+        if not (self.h - self.block_size < image.shape[0] <= self.h and self.w - self.block_size < image.shape[1] <= self.w):
+            self.logger.warn("Chosen model size doesn't match image size. The image is resized")
+
+        resized_image = cv2.resize(image, (self.w, self.h))
```
Wovchena, line 81: The image is resized even if the shape differs a bit. Should the warning be triggered even if a shape is in `block_size` range but doesn't match exactly?

demos/common/python/models/deblurring.py
```diff
@@ -20,15 +20,16 @@
...
+        self.calculate_new_shape()
         self.input_blob_name = self.prepare_inputs()
         self.output_blob_name = self.prepare_outputs()
 
-    def calculate_new_shape(self, shape):
-        h, w, _ = shape
-        block_size = 32
-        new_height = math.ceil(h / block_size) * block_size
-        new_width = math.ceil(w / block_size) * block_size
+    def calculate_new_shape(self):
```
Wovchena, line 29: While you are here, could you rename this method: it not only computes a shape, it actually reshapes and it's what it's called for. The name `reshape` would go for me.

Wovchena (changes requested):

demos/common/python/models/deblurring.py
akorobeinikov: There are 2 variants:
1) Currently. If image shape in range `(input_layer - bs; input_layer + bs)` we do micro resize, but it does not affect accuracy (tested by AC). So I am not liable to print warn. For other images (out of range), I can't guarantee the accuracy. So for them, print the warn.
2) Do check `if image.shape != input_layer.shape` for match exactly and warn for all images that require resize.

I am liable to choose first variant.

demos/common/python/models/deblurring.py
```diff
@@ -74,20 +74,14 @@ def prepare_outputs(self):
...
+
+        resized_image = cv2.resize(image, (self.w, self.h))
         resized_image = resized_image.transpose((2, 0, 1))
         resized_image = np.expand_dims(resized_image, 0)
 
-        meta = {'original_shape': image.shape,
-                'resized_shape': resized_image.shape}
+        meta = {'original_shape': image.shape}
         dict_inputs = {self.input_blob_name: resized_image}
         return dict_inputs, meta
```
Wovchena, line 86:
```suggestion
        dict_inputs = {self.input_blob_name: resized_image}
        return dict_inputs, image.shape[1::-1]
```
So you should rename `meta` in `postprocess()` into something like `dsize` or `widthHeight` to make it easier to understand what it is about. And you can pass `dsize` directly into `cv2.resize()`.

Wovchena (changes requested):

demos/common/python/models/deblurring.py
Wovchena: > tested by AC

It doesn't mean that images that would produce different results don't exist (or can't be created :) ).

I'm not familiar with the net, so it's up to you.

If you keep your approach, the warning message should be modified so it wouldn't be misleading because it's very tempting to assume that if the warning isn't reported, the image isn't resized.

akorobeinikov: Ok, I went back to the previous method. The last question is what about VideoWriter. It works with constant shape and if we use folder with images with different shapes, VideoWriter will work only with images which shape is equal to the first image shape.

Wovchena: Should the prediction be cropped if the image is padded?

VideoWriter works with arbitrary images if filename is % based pattern. I don't think we should do anything about it.

akorobeinikov: Add crop

Wovchena: You can't rely on `Deblurring` state to correspond to current postprocessed image because `Deblurring` could preprocess multiple images before inference fishes, so `self.padded` can't be trusted. `posptorocess()` already has image size so it should deduce on its own if the image was padded.

akorobeinikov: Ok, thank you for noticing.

Wovchena (approved):

akorobeinikov:
Jenkins please retry a build

IRDonch:
Jenkins please retry a build

PR #2137, opened 2021-02-25
AC: Added age_recognition adapter

Anna-Mironova:


tools/accuracy_checker/accuracy_checker/adapters/attributes_recognition.py
```diff
@@ -133,8 +133,8 @@ class AgeGenderAdapter(Adapter):
     def parameters(cls):
         parameters = super().parameters()
         parameters.update({
-            'age_out': StringField(description="Output layer name for age recognition."),
-            'gender_out': StringField(description="Output layer name for gender recognition.")
+            'age_out': StringField(description="Output layer name for age recognition.", optional=True),
```
eaidova, line 136: why these parameters become optional?

tools/accuracy_checker/accuracy_checker/adapters/attributes_recognition.py
```diff
@@ -179,6 +179,26 @@ def process(self, raw, identifiers=None, frame_meta=None):
         return result
 
 
+class AgeRecognitionAdapter(AgeGenderAdapter):
```
eaidova, line 182: not sure that  inheritance logic is correct (should be reversed as the rule).

tools/accuracy_checker/accuracy_checker/adapters/attributes_recognition.py
Anna-Mironova: returned the previous state

tools/accuracy_checker/accuracy_checker/adapters/attributes_recognition.py
Anna-Mironova: Updated

eaidova (approved):

PR #2139, opened 2021-02-25
tools/downloader: print signal name when a child processes is terminated by one

IRDonch:
If a child process is terminated by a signal, there is usually no indication that that happened; it just looks like it exited normally. This is confusing for users looking at the conversion log.

Examine the process exit status and, if it was terminated by a signal, print the signal name and number.

### DIFF
--- a/tools/downloader/common.py
+++ b/tools/downloader/common.py
@@ -22,6 +22,7 @@ import queue
 import re
 import shlex
 import shutil
+import signal
 import subprocess
 import sys
 import threading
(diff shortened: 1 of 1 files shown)


tools/downloader/common.py
```diff
@@ -104,13 +105,31 @@ def check_interrupted(self):
...
 class DirectOutputContext(JobContext):
     def print(self, value, *, end='\n', file=sys.stdout, flush=False):
         print(value, end=end, file=file, flush=flush)
 
     def subprocess(self, args, **kwargs):
-        return subprocess.run(args, **kwargs).returncode == 0
+        return_code = subprocess.run(args, **kwargs).returncode
+
+        if return_code < 0:
+            print(self._signal_message(-return_code), file=sys.stderr)
```
eizamaliev, line 130: Why did you use `-return_code` here?

Fedor-Z: Have you tried it on Windows ? There might be differences in return code notations.

IRDonch: @eizamaliev From https://docs.python.org/3/library/subprocess.html:

> A negative value -N indicates that the child was terminated by signal N (POSIX only).

@Fedor-Z On Windows, signals don't exist, and the return code is always non-negative.

eizamaliev (approved):

tools/downloader/common.py
```diff
@@ -22,6 +22,7 @@
 import re
 import shlex
 import shutil
+import signal
```
deshawn165, line 25: Y'all are disrespectful

tools/downloader/common.py
```diff
@@ -104,13 +105,31 @@ def check_interrupted(self):
     def interrupt(self):
         self._interrupted = True
 
+    @staticmethod
```
deshawn165, line 108: Disrespectful

deshawn165 (commented):
@Subprocess`

PR #2149, opened 2021-02-26
Add PYTHONPATH modification logic for running python demos on Windows

nignatovsky:
Fixed incorrect PYTHONPATH modification with non-Windows delimiters on Windows.

### DIFF
--- a/demos/tests/run_tests.py
+++ b/demos/tests/run_tests.py
@@ -159,8 +159,7 @@ def main():
 
         demo_environment = {**os.environ,
             'PYTHONIOENCODING': 'utf-8',
-            **PYTHONPATH),
-        }
+            **PYTHONPATH}
 
         for demo in demos_to_test:
             print('Testing {}...'.format(demo.subdirectory))
--- a/demos/tests/run_tests.py
+++ b/demos/tests/run_tests.py
@@ -150,12 +150,9 @@ def main():
 
         num_failures = 0
         
-        pythonpath_args = (os.environ['PYTHONPATH'], args.demo_build_dir)
-        
-        if os.name == "nt":
-            PYTHONPATH = {'PYTHONPATH': "{};{};".format(*pythonpath_args)}
-        else:
-            PYTHONPATH = {'PYTHONPATH': "{}:{}/lib".format(*pythonpath_args)}
+        demo_build_subdir = "/lib" if os.pathsep == ":" else ""
+        pythonpath_args = (os.environ['PYTHONPATH'], os.pathsep, args.demo_build_dir, demo_build_subdir)
+        PYTHONPATH = {'PYTHONPATH': "{}{}{}{}".format(*pythonpath_args)}
 
         demo_environment = {**os.environ,
             'PYTHONIOENCODING': 'utf-8',
--- a/demos/tests/run_tests.py
+++ b/demos/tests/run_tests.py
@@ -151,10 +151,11 @@ def main():
 
         num_failures = 0
         
-        demo_build_subdir = "" if platform.system() == "Windows" else "/lib"
+        python_module_subdir = "" if platform.system() == "Windows" else "/lib"
         demo_environment = {**os.environ,
             'PYTHONIOENCODING': 'utf-8',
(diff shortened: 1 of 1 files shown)


demos/tests/run_tests.py
```diff
@@ -149,11 +149,17 @@ def main():
         dl_dir = prepare_models(auto_tools_dir, args.downloader_cache_dir, args.mo, global_temp_dir, demos_to_test)
 
         num_failures = 0
+        
+        pythonpath_args = (os.environ['PYTHONPATH'], args.demo_build_dir)
+        
+        if os.name == "nt":
+            PYTHONPATH = {'PYTHONPATH': "{};{};".format(*pythonpath_args)}
+        else:
+            PYTHONPATH = {'PYTHONPATH': "{}:{}/lib".format(*pythonpath_args)}
```
alalek, line 158: > `if os.name == "nt":`

Consider using `os.pathsep` for that instead.

nignatovsky: Done.

demos/tests/run_tests.py
```diff
@@ -149,11 +149,14 @@ def main():
         dl_dir = prepare_models(auto_tools_dir, args.downloader_cache_dir, args.mo, global_temp_dir, demos_to_test)
 
         num_failures = 0
+        
+        demo_build_subdir = "/lib" if os.pathsep == ":" else ""
+        pythonpath_args = (os.environ['PYTHONPATH'], os.pathsep, args.demo_build_dir, demo_build_subdir)
+        PYTHONPATH = {'PYTHONPATH': "{}{}{}{}".format(*pythonpath_args)}
```
IRDonch, line 155: This variable isn't really necessary; you could just embed the value inside the `demo_environment` initializer.

Also, using an f-string would be more readable than `str.format`.

demos/tests/run_tests.py
```diff
@@ -149,11 +149,14 @@ def main():
         dl_dir = prepare_models(auto_tools_dir, args.downloader_cache_dir, args.mo, global_temp_dir, demos_to_test)
 
         num_failures = 0
+        
+        demo_build_subdir = "/lib" if os.pathsep == ":" else ""
```
IRDonch, line 153: The condition here seems rather indirect; we do not choose the path based on the path separator, we choose it based on the OS. So you should check the OS (you can use `platform.system()` to do that).

IRDonch (changes requested):

demos/tests/run_tests.py
nignatovsky: Done.

demos/tests/run_tests.py
nignatovsky: Done.

demos/tests/run_tests.py
```diff
@@ -149,11 +150,11 @@ def main():
...
 
         num_failures = 0
-
+        
+        demo_build_subdir = "" if platform.system() == "Windows" else "/lib"
         demo_environment = {**os.environ,
             'PYTHONIOENCODING': 'utf-8',
-            'PYTHONPATH': "{}:{}/lib".format(os.environ['PYTHONPATH'], args.demo_build_dir),
-        }
+            'PYTHONPATH': f"{os.environ['PYTHONPATH']}{os.pathsep}{args.demo_build_dir}{demo_build_subdir}"}
```
IRDonch, line 157:
```suggestion
            'PYTHONPATH': f"{os.environ['PYTHONPATH']}{os.pathsep}{args.demo_build_dir}{demo_build_subdir}",
        }
```

demos/tests/run_tests.py
```diff
@@ -149,11 +150,11 @@ def main():
         dl_dir = prepare_models(auto_tools_dir, args.downloader_cache_dir, args.mo, global_temp_dir, demos_to_test)
 
         num_failures = 0
-
+        
+        demo_build_subdir = "" if platform.system() == "Windows" else "/lib"
```
IRDonch, line 154:
```suggestion
        python_module_subdir = "" if platform.system() == "Windows" else "/lib"
```

IRDonch (changes requested):

IRDonch (approved):

IRDonch:
Jenkins: test this please

PR #1897, opened 2020-12-07
Unify Python segmentation demo

anzhella-pankratova:

PR #2150, opened 2021-02-26
segmentation_demo/python: remove the ability to set a custom color palette

IRDonch:
This feature is barely documented, not tested at all, and involves a gratuitous use of `eval`. It's also not present in the C++ version of the demo. I think it's unnecessary.

### DIFF
--- a/demos/segmentation_demo/python/README.md
+++ b/demos/segmentation_demo/python/README.md
@@ -23,8 +23,7 @@ python3 segmentation_demo.py -h
 The command yields the following usage message:
 
 ```
-usage: segmentation_demo.py [-h] -m MODEL -i INPUT
-                            [-d DEVICE] [-c COLORS]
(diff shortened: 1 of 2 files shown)


IRDonch:
CC @akorobeinikov

Wovchena:
There are two reasons to keep this feature:
> [model has 150 segmentation classes, I assume that it may produce ugly result if we will define colors randomly](https://github.com/openvinotoolkit/open_model_zoo/pull/1897#issuecomment-742558466)

@vladimir-dudnik had an idea that someone may hardly want to see segmentation default color set used if such exists. `--colors` allows to provide it.

vladimir-dudnik:
format description of this text file with colors is missed, there is no sample, so not clear how one can create such file

Wovchena (approved):

vladimir-dudnik (changes requested):
can we add description of color palette file to demo rather then delete changes?

PR #2156, opened 2021-03-01
Demos: move nms function

anzhella-pankratova:
Non-maximum Suppression is applied by several models, so is better to create the common place with implementation instead of having one in every model file.
It is moved to `models/utils.py` for python demos and to `utils/nms.hpp` for c++ demos.

### DIFF
--- a/demos/instance_segmentation_demo/python/instance_segmentation_demo/model_utils.py
+++ b/demos/instance_segmentation_demo/python/instance_segmentation_demo/model_utils.py
@@ -17,7 +17,7 @@
 from collections import namedtuple
 import cv2
 import numpy as np
-from utils import nms
+from models.utils import nms
 
 ModelAttributes = namedtuple('ModelAttributes', ['required_outputs', 'postprocessor'])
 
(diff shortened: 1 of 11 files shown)


Wovchena (approved):

demos/instance_segmentation_demo/python/instance_segmentation_demo/model_utils.py
```diff
@@ -14,9 +14,14 @@
  limitations under the License.
 """
 
+import sys
+from pathlib import Path
+sys.path.append(str(Path(__file__).resolve().parents[3] / 'common/python/models'))
+
```
IRDonch, line 20: This shouldn't be needed if the main script of the demo already modifies `sys.path` (which it does).

IRDonch (changes requested):

demos/instance_segmentation_demo/python/instance_segmentation_demo/model_utils.py
anzhella-pankratova: Updated.

IRDonch (approved):

PR #2159, opened 2021-03-01
Option to specify demos to build in build scripts

ivikhrev:
This PR continues work started in #1733.
Added `--target="<list of demos>"` option to build scripts. It allows to specify one or more demos to build.


ivikhrev:
I'm not sure, should we test this feature or not. Because if we should, for example to brute force all the variants, there would be `2^number_of_cmake_targets variants-1`  (cmake gives us 31 targets, if we leave only demos as targets, not python or common modules, there will be 19 targets). It's a huge number and building process would take a lot of time. @vladimir-dudnik, @Wovchena, @eizamaliev  what do you think?

ivikhrev:
I think we can partially test it by specifying each demo and all combinations of two demos, for example.

demos/build_demos.sh
```diff
@@ -82,6 +88,15 @@ fi
 mkdir -p "$build_dir"
 
 (cd "$build_dir" && cmake -DCMAKE_BUILD_TYPE=Release "${extra_cmake_opts[@]}" "$DEMOS_PATH")
-cmake --build "$build_dir" -- "$NUM_THREADS"
+
+if ["&build_targets" == ""]; then
```
Wovchena, line 92:
```suggestion
if [ -z "$build_targets" ]; then
```

Wovchena (approved):

Wovchena:
We don't test these scripts as far as I know. I think this removes the question about different subsets.

demos/README.md
```diff
@@ -214,6 +214,27 @@ For example:
 cmake -DCMAKE_BUILD_TYPE=Release -DENABLE_PYTHON=ON <open_model_zoo>/demos
 ```
 
+### <a name="build_specific_demos"></a>Build Specific Demos
+
+To build specific demos, follow the instructions for building the demo applications above,
+but add `--target <list of demos>` to the `cmake` or `--target="<list of demos>"` to the `build_demos*` command.
```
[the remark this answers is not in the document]
ivikhrev: I've added information about cmake build tool to the instuctions, and also added instructions for Windows, but it's very similar to Linux, so I doubt if we need it.

demos/README.md
```diff
@@ -214,6 +214,27 @@ For example:
...
+Note,  CMake `--build` tool supports multiple targets starting with version 3.15. With lower versions you can specify only one target to the `cmake`
+or use `build_demos*` command.
+For example:
+
+```sh
```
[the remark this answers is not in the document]
ivikhrev: Split for windows and linux.

Wovchena (approved):

demos/README.md
```diff
@@ -169,6 +169,10 @@ cd build
 ```sh
 make
 ```
+Or you can use `cmake --build` tool:
+```sh
+cmake --build .
+```
```
IRDonch, line 175: Do we need to provide both alternatives? I think we can just promote the `cmake --build` approach. It has no downsides (compared to running `make` directly) that I can think of, and it works with non-make generators as well.

demos/README.md
```diff
@@ -202,6 +206,29 @@ build binaries in Debug configuration. Run the appropriate version of the
...
+md build
+```
+2. Go to the created directory:
+```
+cd build
+```
+3. Run CMake to generate project files:
+```
+cmake <open_model_zoo>/demos
```
IRDonch, line 220: I think `-A x64` should be added here, because CMake defaults to 32-bit for Visual Studio generators prior to 2019.

demos/README.md
```diff
@@ -214,6 +241,34 @@ For example:
 cmake -DCMAKE_BUILD_TYPE=Release -DENABLE_PYTHON=ON <open_model_zoo>/demos
 ```
 
+### <a name="build_specific_demos"></a>Build Specific Demos
+
+To build specific demos, follow the instructions for building the demo applications above,
+but add `--target <list of demos>` to the `cmake` or `--target="<list of demos>"` to the `build_demos*` command.
```
IRDonch, line 247:
```suggestion
but add `--target <list of demos>` to the `cmake --build` command or `--target="<list of demos>"` to the `build_demos*` command.
```

I would also suggest replacing `<list of demos>` with `<demo1> <demo2> ...` to make it clear how the list is to be formatted.

demos/README.md
```diff
@@ -214,6 +241,34 @@ For example:
...
+Note,  CMake `--build` tool supports multiple targets starting with version 3.15. With lower versions you can specify only one target to the `cmake`
```
IRDonch, line 248:
```suggestion
Note,  `cmake --build` tool supports multiple targets starting with version 3.15. With lower versions you can specify only one target to the `cmake`
```

Be consistent.

demos/build_demos.sh
```diff
@@ -26,12 +26,18 @@ error() {
...
+build_targets=()
 
 for opt in "$@"; do
     case "$opt" in
     -DENABLE_PYTHON=*)
         extra_cmake_opts+=("$opt")
         ;;
+    --target=*)
+        tmp="${opt%\"}"
+        tmp="${tmp#\"}"
```
IRDonch, line 38: What is this for?

demos/build_demos.sh
```diff
@@ -26,12 +26,18 @@ error() {
...
+        build_targets+=("${tmp//=/ }")
```
IRDonch, line 39: I'm confused by this. AFAICS, this replaces all equals signs in the option with spaces and than adds that option to `build_targets`. But doesn't the option still contain the string `--target` at that point? Why would you add _that_ to `build_targets`?

demos/build_demos.sh
```diff
@@ -82,6 +88,15 @@ fi
 mkdir -p "$build_dir"
 
 (cd "$build_dir" && cmake -DCMAKE_BUILD_TYPE=Release "${extra_cmake_opts[@]}" "$DEMOS_PATH")
-cmake --build "$build_dir" -- "$NUM_THREADS"
+
+if [ -z "$build_targets" ]; then
```
IRDonch, line 92: `$build_targets` substitutes to just the first element of `build_targets`, therefore this doesn't actually check if the array is empty. Use `${#build_targets[@]} -eq 0`.

demos/build_demos.sh
```diff
@@ -82,6 +88,15 @@ fi
...
+    cmake --build "$build_dir" -- "$NUM_THREADS"
+else
+    targets=($build_targets)
+    for t in "${targets[@]:1}"
+    do
```
IRDonch, line 97: Use consistent style; `do` should be on the same line as `for`.

demos/build_demos_msvc.bat
```diff
@@ -34,6 +35,14 @@ if not "%1" == "" (
         shift & shift
         goto argParse
     )
+    rem to build more than one specific demo use quotation marks,
+    rem list the necessary demos separated by space,
+    rem ex. --target="classification_demo segmentation_demo"
+    if "%1" == "--target" (
+        set BUILD_TARGETS=%BUILD_TARGETS% %1 %~2
```
IRDonch, line 42: Again, why are you adding `%1` to `BUILD_TARGETS`?

demos/build_demos.sh
IRDonch: Also, is it really worth it to support combining multiple `--target` options? CMake itself doesn't support it, so why should we?

IRDonch (changes requested):

demos/README.md
ivikhrev: Ok, removed `make` approach.

demos/README.md
ivikhrev: Added

demos/README.md
ivikhrev: Done.

demos/README.md
ivikhrev: Ok.

demos/build_demos.sh
ivikhrev: It removes quotation marks. In `*.bat` file for it we use `%~2`, I didn't find better solution here to do the same.

demos/build_demos.sh
ivikhrev: Ok.

demos/build_demos.sh
ivikhrev: Ok.

demos/build_demos_msvc.bat
ivikhrev: A bit of misunderatanding from my side, fixed it.

demos/build_demos.sh
ivikhrev: Fixed.  
About the neccesity of such option, if it's hard to support or test we can remove it.

demos/README.md
```diff
@@ -214,6 +237,34 @@ For example:
 cmake -DCMAKE_BUILD_TYPE=Release -DENABLE_PYTHON=ON <open_model_zoo>/demos
 ```
 
+### <a name="build_specific_demos"></a>Build Specific Demos
+
+To build specific demos, follow the instructions for building the demo applications above,
+but add `--target <demo1> <demo2> ...` to the `cmake --build` command` or `--target="<demo1> <demo2> ..."` to the `build_demos*` command.
```
IRDonch, line 243:
```suggestion
but add `--target <demo1> <demo2> ...` to the `cmake --build` command or `--target="<demo1> <demo2> ..."` to the `build_demos*` command.
```

demos/build_demos.sh
IRDonch: But why would there be any quotation marks in the first place? If the user runs something like  `build_demos.sh --target="a b"`, then those quotes are shell syntax and will not be present in the value of `opt`.

demos/build_demos_msvc.bat
```diff
@@ -122,8 +131,16 @@ cd "%SOLUTION_DIR64%" && cmake -G "Visual Studio !VS_VERSION!" -A %PLATFORM% %EX
 echo.
 echo ###############^|^| Build Open Model Zoo Demos using MS Visual Studio ^|^|###############
 echo.
-echo cmake --build . --config Release
-cmake --build . --config Release
+echo cmake --build . --config Release %BUILD_TARGETS%
```
IRDonch, line 134: This does not correspond to the actual command(s) being run. You should remove this and add an echo command before each of the `cmake` commands below.

demos/build_demos.sh
IRDonch: It's not about the difficulty of support (I don't think it's that difficult). I just think that if we implement the support for multiple `--target` options now, we might end up not matching CMake if CMake developers decide to implement it with different semantics.

For example, in your current implementation, `--target=a --target=b` is equivalent to `--target 'a b'`. But in a future CMake version, it could be equivalent to `--target b` instead (i.e., they could make the last option override any previous ones). Since we're trying to mimic CMake options here, I think that's an undesirable outcome. Therefore, it would be better to reject multiple `--target` options, like CMake is doing now.

demos/build_demos.sh
```diff
@@ -82,6 +88,13 @@ fi
 mkdir -p "$build_dir"
 
 (cd "$build_dir" && cmake -DCMAKE_BUILD_TYPE=Release "${extra_cmake_opts[@]}" "$DEMOS_PATH")
-cmake --build "$build_dir" -- "$NUM_THREADS"
+
+if [ ${#build_targets[@]} -eq 0 ]; then
+    cmake --build "$build_dir" -- "$NUM_THREADS"
+else
+    for t in $build_targets; do
```
IRDonch, line 95:
```suggestion
    for t in "${build_targets[@]}"; do
```

ivikhrev:
Jenkins please retry a build

demos/build_demos.sh
ivikhrev: I get your point. I will remove it.

IRDonch: I'm quite confused as to what you did here. Now multiple `--target` options are _still_ supported, but multiple targets aren't? That makes no sense.

ivikhrev: Now, I just pass --target to cmake, without calling `cmake --build .  --target A` multiple times in cycle. If cmake version supports multiple targets, they will build, if not - there will be an error from cmake side. Why does it make no sense?

ivikhrev: I just let cmake decides how to interpret --target option, not to mimic it, like you suggest.

IRDonch: (Sorry, it seems that this PR has slipped my mind.)

I have examined everything again, and it seems that I have made a critical mistake - CMake 3.15+ _does_, in fact, support multiple `--target` options (although I don't see this advertised in the documentation). Given this, your solution does indeed make sense.

On a separate note, it's a bit unfortunate that your syntax doesn't correspond to `cmake --build` exactly (since you require the target(s) to be passed via the same argument as the option name, whereas CMake requires them to be passed as separate arguments), but replicating CMake's parsing logic would be a pain and not really worth it.

IRDonch (approved):

IRDonch:
I'll rerun the builds just in case, since it's been a while.

IRDonch:
Jenkins please retry a build

PR #2167, opened 2021-03-03
AC: fix single human pose estimation

eaidova:
current data reading approch does not allow non-unique names as identifiers. Changed identifier for single human pose estimation


tools/accuracy_checker/accuracy_checker/data_readers/__init__.py
```diff
@@ -78,6 +79,7 @@
     'MultiFramesInputIdentifier',
     'ImagePairIdentifier',
     'ListIdentifier',
+    'MultiInstanceIdentifier'
```
eaidova, line 82:
```suggestion
    'MultiInstanceIdentifier',
```

vladimir-dudnik (approved):

Anna-Mironova (approved):

PR #2170, opened 2021-03-04
Demos update CmakeLists.txt

Wovchena:
There is no special reason for this PR


demos/CMakeLists.txt
```diff
@@ -4,17 +4,20 @@
 
 cmake_minimum_required(VERSION 3.10)
 
+# Multi config generators such as Visual Studio ignore CMAKE_BUILD_TYPE. Multi config generators are configured with
+# CMAKE_CONFIGURATION_TYPES, but limiting options in it completely removes such build options
+if(NOT GENERATOR_IS_MULTI_CONFIG AND NOT DEFINED CMAKE_BUILD_TYPE)
+    message(STATUS "CMAKE_BUILD_TYPE not defined, 'Release' will be used")
+    # Setting CMAKE_BUILD_TYPE as CACHE must go before project(). Otherwise project() sets its value and set() doesn't take an effect
+    set(CMAKE_BUILD_TYPE Release CACHE STRING "Choose the type of build, options are: None Debug Release RelWithDebInfo MinSizeRel ...")
```
alalek, line 12: > RelWithDebInfo MinSizeRel

Do all dependencies support these modes properly? Does it checked somewhere?

demos/CMakeLists.txt
```diff
@@ -103,33 +90,17 @@ endfunction()
...
-#               [EXCLUDE_CPPLINT]
-#
-macro(ie_add_sample)
-    set(options EXCLUDE_CPPLINT)
+# add_demo(NAME <target name>
+#     SOURCES <source files>
+#     [HEADERS <header files>]
+#     [INCLUDE_DIRECTORIES <include dir>]
+#     [DEPENDENCIES <dependencies>])
+macro(add_demo)
```
alalek, line 98: omz_add_demo?

vladimir-dudnik:
@Wovchena this seems to be many changes across demo CMakeLists files, could you please add to PR description what it will serve for? Basically why these changes are needed?

demos/CMakeLists.txt
Wovchena: I don't know. This message is what `CMAKE_BUILD_TYPE` usually has, I just put a copy of it to mimic it. Would it be better if I truncated the message to `Choose the type of build`?

demos/CMakeLists.txt
Wovchena: Why? I don't see a reason to stay aligned with style of the old name. Plus this will conflict with `IE_SAMPLE*` naming which I didn't touch here.

Wovchena:
> this seems to be many changes across demo CMakeLists files

I've split PR on multiple commits because of that.

> could you please add to PR description what it will serve for? Basically why these changes are needed?

There is no special reason for this PR, I just updated cmake usage. PR's name reflects exactly that. Regarding ideas staying behind each particular change, see commit messages.

vladimir-dudnik:
@Wovchena PR description is place to put summary info to free up people from looking at each particular commit to deduce what is the chages and whythey are needed. In this case, PR name says what you change (CMakeLists for demos) but it does not say why changes are needed, what is intention.

Wovchena:
I believe `update CmakeLists.txt` is the best summary. As I said, there is no special reason for this PR. What do you want me to put there?

Wovchena:
Jenkins please retry a build

Wovchena:
Jenkins please retry a build

demos/CMakeLists.txt
IRDonch: What was the purpose of this change, anyway? Wouldn't it be sufficient to just add `NOT GENERATOR_IS_MULTI_CONFIG` to the original `if` command?

Wovchena: Setting `CMAKE_BUILD_TYPE` as CACHE puts the correct value into cache. Earlier version shadows original `CMAKE_BUILD_TYPE` cache var.

IRDonch: > Earlier version shadows original CMAKE_BUILD_TYPE cache var.

Yeah, but... is it actually a problem?

demos/CMakeLists.txt
```diff
@@ -103,33 +89,17 @@ endfunction()
...
     set(oneValueArgs NAME)
     set(multiValueArgs SOURCES HEADERS DEPENDENCIES OPENCV_DEPENDENCIES INCLUDE_DIRECTORIES)
```
IRDonch, line 99: Shouldn't `OPENCV_DEPENDENCIES` be removed?

demos/CMakeLists.txt
Wovchena: This resulted in having two wars. While cached value claimed the type is `""`, as well as cache viewers did, in fact Release was generated.

IRDonch (changes requested):

demos/CMakeLists.txt
IRDonch: Yeah, but again... is that actually a problem? We're effectively just "symlinking" the blank configuration to the release configuration.

Wovchena: Blank configuration seems to be a separate valid value, decoupling these two should be cleaner.

IRDonch: Yeah, in the blank configuration none of the `*_DEBUG` or `*_RELEASE` variables are used, resulting in a build with neither optimizations nor debug info. :-)

I'm not convinced that we need to support such a configuration, but I also don't care enough to argue about it, so keep it like that if you want.

IRDonch (approved):

IRDonch:
I'm gonna rerun the builds, in case something was merged in the intervening days that broke them.

IRDonch:
Jenkins please retry a build

PR #2171, opened 2021-03-04
segmentation_demo.py: add description of color palette file

anzhella-pankratova:
Added color palettes for CamVid, Cityscapes and ADE20k datasets; added description how to create a file with custom colors.


vladimir-dudnik (changes requested):
suggest to use this paragraph at line 11:
Upon the start-up the demo application reads command line parameters and loads a network. The demo runs inference and shows results for each image captured from an input. Demo provides default mapping of classes to colors and optionally, allow to specify mapping of classes to colors from simple text file, with using '--colors' argument. Depending on number of inference requests processing simultaneously (-nireq parameter) the pipeline might minimize the time required to process each single image (for nireq 1) or maximize utilization of the device and overall processing performance.

@anzhella-pankratova rephrased a bit to be consistent with your text below

anzhella-pankratova:
> suggest to use this paragraph at line 11:
> Upon the start-up the demo application reads command line parameters and loads a network. The demo runs inference and shows results for each image captured from an input. Demo provides default mapping of classes to colors and optionally, allow to specify mapping of classes to colors from simple text file, with using '--colors' argument. Depending on number of inference requests processing simultaneously (-nireq parameter) the pipeline might minimize the time required to process each single image (for nireq 1) or maximize utilization of the device and overall processing performance.
> 
> @anzhella-pankratova rephrased a bit to be consistent with your text below

Thank you, I modified the documentation.

vladimir-dudnik (approved):

demos/segmentation_demo/python/README.md
```diff
@@ -86,6 +86,14 @@ You can use the following command to do inference on CPU on images captured by a
     python3 segmentation_demo.py -i 0 -m <path_to_model>/semantic-segmentation-adas-0001.xml
 ```
 
+## Color palettes
+
+The color palette is used to visualize predicted classes. By default, the colors from PASCAL VOC dataset are applied. In case then the number of output classes is larger than number of classes provided by PASCAL VOC dataset, the rest classes are randomly colorized.
+Also, one can use predefined colors from other datasets, like CAMVID.
+
+The folder with available colors files is in [palettes](../../../data/palettes).
```
Wovchena, line 94:
```suggestion
Available colors files are in [palettes](../../../data/palettes).
```
The folder with available colors itself is in data.

Wovchena:
Jenkins please retry a build

Wovchena (approved):

demos/segmentation_demo/python/README.md
```diff
@@ -86,6 +86,14 @@ You can use the following command to do inference on CPU on images captured by a
     python3 segmentation_demo.py -i 0 -m <path_to_model>/semantic-segmentation-adas-0001.xml
 ```
 
+## Color palettes
+
+The color palette is used to visualize predicted classes. By default, the colors from PASCAL VOC dataset are applied. In case then the number of output classes is larger than number of classes provided by PASCAL VOC dataset, the rest classes are randomly colorized.
+Also, one can use predefined colors from other datasets, like CAMVID.
+
+Available colors files are in [palettes](../../../data/palettes).
```
vladimir-dudnik, line 94:
```suggestion
Available colors files are in '<omz_dir>/data/palettes`.
```
when openvino onlne documentation generated from markdown files it will not point to relative path

vladimir-dudnik (changes requested):

demos/segmentation_demo/python/segmentation_demo.py
```diff
@@ -71,10 +71,14 @@ def __init__(self, colors_path=None):
 
     def get_palette_from_file(self, colors_path):
         with open(colors_path, 'r') as file:
-            return [eval(line.strip()) for line in file.readlines()]
+            colors = []
+            for line in file.readlines():
+                values = line[line.index('(')+1:line.index(')')].split(',')
+                colors.append([int(v.strip()) for v in values])
```
eaidova, line 77: what if colors are invalid?
Should we handle and provide error message for this case?

vladimir-dudnik (approved):
please look at @eaidova question

demos/segmentation_demo/python/segmentation_demo.py
vladimir-dudnik: @eaidova do you mean validate that 'values' should be a number in 0...255 range?

eaidova: yes and not only:
* string can contains any other symbols which can not be casted to int
* list of values for one color can be not equal number of channels or even be empty

anzhella-pankratova: I am not sure it is necessary. The README provides the fully description of correct file with colors. Also the ValueError occurs then some string has wrong format. There are many ways how invalid colors might look, even the replacement of round brackets by `[ ]` will lead to the error, so the handling of all this cases might make the code more complex.

eaidova: ok

PR #2188, opened 2021-03-10
common-sign-language-0001: fix the task type

IRDonch:
The task it performs is clearly more akin to action recognition than object attribute estimation. In fact, its interface is almost exactly like that of `i3d-rgb-tf`.


IRDonch:
CC @evgeny-izutov

eaidova (approved):

PR #2190, opened 2021-03-10
Try to import MO as Python package

dkurt:
Example MO install:
```
pip install git+https://github.com/openvinotoolkit/openvino.git#subdirectory=model-optimizer
```

### DIFF
--- a/tools/downloader/README.md
+++ b/tools/downloader/README.md
@@ -261,7 +261,7 @@ script. You can override this heuristic with the `--mo` option:
 ./converter.py --all --mo my/openvino/path/model_optimizer/mo.py
 ```
 
-Or add Model Optimizer to `PYTHONPATH`:
+Or add Model Optimizer to `PYTHONPATH`. The path should point to a location of `mo` package.
 ```sh
 export PYTHONPATH=my/openvino/path/model_optimizer/:$PYTHONPATH
 ```
--- a/tools/downloader/README.md
+++ b/tools/downloader/README.md
@@ -255,11 +255,15 @@ If the specified precision is not supported for a model, that model will be skip
 
 The script will attempt to locate Model Optimizer using the environment
 variables set by the OpenVINO&trade; toolkit's `setupvars.sh`/`setupvars.bat`
-script. Or trying to use PyPI distribution.
-You can override this heuristic with the `--mo` option:
+script. You can override this heuristic with the `--mo` option:
 
 ```sh
 ./converter.py --all --mo my/openvino/path/model_optimizer/mo.py
+```
+
+Or add Model Optimizer to `PYTHONPATH`:
+```sh
+export PYTHONPATH=my/openvino/path/model_optimizer/:$PYTHONPATH
 ```
 
 You can add extra Model Optimizer arguments to the ones specified in the model
--- a/tools/downloader/README.md
+++ b/tools/downloader/README.md
@@ -255,7 +255,8 @@ If the specified precision is not supported for a model, that model will be skip
 
 The script will attempt to locate Model Optimizer using the environment
 variables set by the OpenVINO&trade; toolkit's `setupvars.sh`/`setupvars.bat`
-script. You can override this heuristic with the `--mo` option:
+script. Or trying to use PyPI distribution.
+You can override this heuristic with the `--mo` option:
 
 ```sh
 ./converter.py --all --mo my/openvino/path/model_optimizer/mo.py
--- a/tools/downloader/converter.py
+++ b/tools/downloader/converter.py
@@ -115,7 +115,7 @@ def main():
                                           'import mo,sys; sys.stdout.write(mo.__file__)'],
                                          check=True, stdout=subprocess.PIPE, encoding='utf-8').stdout
                 mo_path = (Path(mo_path).parent / '__main__.py').resolve(strict=True)
-            except Exception:
+            except (subprocess.CalledProcessError, OSError, FileNotFoundError):
                 sys.exit('Unable to locate Model Optimizer. '
                     + 'Use --mo or run setupvars.sh/setupvars.bat from the OpenVINO toolkit.')
 
--- a/tools/downloader/converter.py
+++ b/tools/downloader/converter.py
@@ -109,11 +109,11 @@ def main():
             mo_path = Path(os.environ['INTEL_OPENVINO_DIR']) / 'deployment_tools/model_optimizer/mo.py'
         except KeyError:
             try:
-                # For OpenVINO from PyPI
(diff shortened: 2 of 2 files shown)


tools/downloader/converter.py
```diff
@@ -106,7 +106,11 @@ def main():
     if mo_path is None:
         try:
             mo_path = Path(os.environ['INTEL_OPENVINO_DIR']) / 'deployment_tools/model_optimizer/mo.py'
-        except KeyError:
+            # For OpenVINO from pip
```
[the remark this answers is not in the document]
dkurt: Done, thanks!

mryzhov (approved):

dkurt:
@IRDonch, may we ask to review? This is done for MO installed by pip (as a part of `openvino-dev`, in example)

tools/downloader/converter.py
```diff
@@ -107,8 +107,13 @@ def main():
         try:
             mo_path = Path(os.environ['INTEL_OPENVINO_DIR']) / 'deployment_tools/model_optimizer/mo.py'
         except KeyError:
-            sys.exit('Unable to locate Model Optimizer. '
-                + 'Use --mo or run setupvars.sh/setupvars.bat from the OpenVINO toolkit.')
+            # For OpenVINO from PyPI
+            try:
+                import mo
```
IRDonch, line 112: This might not work, because MO will not necessarily be run with the same Python that Model Converter is run with. Perhaps you could do something like `subprocess.run([args.python, '-c', 'import mo; print(mo.__file__)'])` instead and capture the output.

IRDonch (changes requested):
You need to update the README with information about the updated MO search logic.

tools/downloader/converter.py
dkurt: Good point. Fixed

tools/downloader/README.md
```diff
@@ -255,7 +255,8 @@ If the specified precision is not supported for a model, that model will be skip
 
 The script will attempt to locate Model Optimizer using the environment
 variables set by the OpenVINO&trade; toolkit's `setupvars.sh`/`setupvars.bat`
-script. You can override this heuristic with the `--mo` option:
+script. Or trying to use PyPI distribution.
```
IRDonch, line 258: > Or trying to use PyPI distribution.

That is not a complete sentence. It should be combined the preceding one.

But besides that, it also needs reworking. First, it doesn't actually fit the previous sentence. If you combine them, you get "The script will attempt to locate Model Optimizer trying to use PyPI distribution", where both "attempt" and "trying" communicate the same meaning.

And second, it doesn't explain much. What PyPI distribution are you talking about? How is it used to locate Model Optimizer? I would argue that PyPI itself is actually rather irrelevant here.

IRDonch (changes requested):

tools/downloader/README.md
dkurt: Removed PyPI mention at all. With https://github.com/openvinotoolkit/openvino/pull/4801 both PyPI and and just `export PYTHONPATH=/path/to/openvino/model-optimizer/:$PYTHONPATH` may work

tools/downloader/converter.py
```diff
@@ -107,8 +108,15 @@ def main():
         try:
             mo_path = Path(os.environ['INTEL_OPENVINO_DIR']) / 'deployment_tools/model_optimizer/mo.py'
         except KeyError:
-            sys.exit('Unable to locate Model Optimizer. '
-                + 'Use --mo or run setupvars.sh/setupvars.bat from the OpenVINO toolkit.')
+            try:
+                # For OpenVINO from PyPI
+                mo_path = subprocess.run([args.python, '-c',
+                                          'import mo,sys; sys.stdout.write(mo.__file__)'],
+                                         check=True, stdout=subprocess.PIPE).stdout.decode('utf-8')
```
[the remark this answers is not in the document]
dkurt: Enabled `encoding='utf-8'` flag

tools/downloader/README.md
IRDonch: That's a lot better, but I think you should still explicitly say that it will try to locate MO by searching for the `mo` package in the Python environment.

tools/downloader/converter.py
```diff
@@ -107,8 +108,15 @@ def main():
...
             mo_path = Path(os.environ['INTEL_OPENVINO_DIR']) / 'deployment_tools/model_optimizer/mo.py'
         except KeyError:
-            sys.exit('Unable to locate Model Optimizer. '
-                + 'Use --mo or run setupvars.sh/setupvars.bat from the OpenVINO toolkit.')
+            try:
+                # For OpenVINO from PYTHONPATH
+                mo_path = subprocess.run([args.python, '-c',
+                                          'import mo,sys; sys.stdout.write(mo.__file__)'],
+                                         check=True, stdout=subprocess.PIPE, encoding='utf-8').stdout
+                mo_path = (Path(mo_path).parent / '__main__.py').resolve(strict=True)
```
IRDonch, line 116: That's not going to work correctly, because some MO arguments used for converting models refer to files inside the MO directory, and in order for them to be located correctly, `mo_path` must point to the `mo.py` script.

tools/downloader/converter.py
```diff
@@ -107,8 +108,15 @@ def main():
...
+            except Exception:
```
IRDonch, line 117:
```suggestion
            except (subprocess.CalledProcessError, OSError):
```

Any other exception likely means that the code is broken, not that MO is not found.

tools/downloader/converter.py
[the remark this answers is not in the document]
IRDonch: That's good, but that doesn't do the same thing. You still need to make Python output UTF-8, and you can only do that with `PYTHONIOENCODING`.

tools/downloader/converter.py
dkurt: Basic conversion worked for me. Do you know which arguments may be affected?

IRDonch: Any model that uses `$mo_dir`.

SDxKeeper: @dkurt There are another issue with mo installation via proposed command. not all extensions are installed. For example `extensions/front/tf/ssd_v2_support.json`

SDxKeeper: Issue with extensions  installation has been addressed in https://github.com/openvinotoolkit/openvino/pull/5302 
@dkurt can you please finalize changes required to make converter work after installation of openvino-dev python package with model optimizer inside?

IRDonch:
Obsoleted by #2395; closing.

PR #2192, opened 2021-03-10
regnetx-3.2gf: don't download anything at conversion time

IRDonch:
The `pycls.models.model_zoo.regnetx` function downloads the config file, which is bad, because:

1. all downloading is supposed to be done by the downloader, and
2. there's no integrity checking on the downloaded file (and in fact, the file can change, because the link points to the master branch of the repository), which endangers the security and reproducibility of the conversion process.

Add the config file to the list of the model's files and change the model creation code to avoid the unwanted download.

### DIFF
--- a/models/public/regnetx-3.2gf/model.py
+++ b/models/public/regnetx-3.2gf/model.py
@@ -15,7 +15,8 @@
 import pycls.core.checkpoint
 import pycls.models.model_zoo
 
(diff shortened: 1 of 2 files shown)


IRDonch:
CC @akorobeinikov

eaidova (approved):

PR #2205, opened 2021-03-16
AC: fix output dtype for crop with scale

eaidova:


eaidova:
Jenkins please retry a build

Anna-Mironova (approved):

PR #2214, opened 2021-03-18
Resizing in object_detection_demo.py

GalyaZalesskaya:
Added a resizing in object_detection_demo for large input video/images to fit (1280, 720) window

### DIFF
--- a/demos/object_detection_demo/python/object_detection_demo.py
+++ b/demos/object_detection_demo/python/object_detection_demo.py
@@ -204,6 +204,9 @@ def draw_detections(frame, detections, palette, labels, threshold):
             if isinstance(detection, models.DetectionWithLandmarks):
                 for landmark in detection.landmarks:
                     cv2.circle(frame, (int(landmark[0]), int(landmark[1])), 2, (0, 255, 255), 2)
+    y_scale, x_scale = [i / j for i, j in zip(size, (720, 1280))]
+    scale = max(x_scale, y_scale, 1)
+    frame = cv2.resize(frame, (int(size[1]/scale), int(size[0]/scale)))
     return frame
 
 


demos/object_detection_demo/python/object_detection_demo.py
```diff
@@ -204,6 +204,9 @@ def draw_detections(frame, detections, palette, labels, threshold):
             if isinstance(detection, models.DetectionWithLandmarks):
                 for landmark in detection.landmarks:
                     cv2.circle(frame, (int(landmark[0]), int(landmark[1])), 2, (0, 255, 255), 2)
+    y_scale, x_scale = [i / j for i, j in zip(size, (720, 1280))]
+    scale = max(x_scale, y_scale, 1)
```
Wovchena, line 208:
```suggestion
    scale = max(size[0] / 720, size[1] / 1280)
```
Your solution is an overkill :) `1` is out here because it needs to be handled separately.

demos/object_detection_demo/python/object_detection_demo.py
```diff
@@ -204,6 +204,9 @@ def draw_detections(frame, detections, palette, labels, threshold):
...
+    frame = cv2.resize(frame, (int(size[1]/scale), int(size[0]/scale)))
```
Wovchena, line 209: 1. `cv2.resize()` reallocates an array even if `scale` is 1: `frame.__array_interface__['data'][0] == resized.__array_interface__['data'][0]` gives `False`. You should handle case when no resize is needed separately.
2. `resize()` is fine computing `dst` size on its own given `fx` and `fy`. There is no need for you to compute `(int(size[1]/scale), int(size[0]/scale))`. Plus I guess it shoul have been `round()`, not `int()`, though I'm not sure :)
3. Don't give a name to `resize()`'s result. Just return what it gives.

Wovchena (changes requested):
The solution breaks consistency between `draw_detections()` and `print_raw_results()`.
You need to update other demos which follow `object_detection_demo` implementation including C++. We kinda try to unify them :). Please check if C++ `cv::resize()` also always reallocates `cv::Mat`.
Model resolution is usually smaller than `720x1280`, so you can resize the image in the very beginning and that should save some computation when it is resized again to fit model resolution.
Given all that you shouldn’t update `_demo.py` or `main.cpp` files, though some demos like `human_pose_estimation_demo` seem to be an exception because they reshape a net and need a smarter approach. Consider modifying `open_model_zoo/demos/common/python/images_capture.py` and it’s C++ equivalent.

eizamaliev (commented):
@GalyaZalesskaya what is the purpose of such resize?

GalyaZalesskaya:
@eizamaliev In my several experiments the input images were bigger than screen resolution that led to problems in perception and difficulties in debugging. So I considered putting this "simple fix" with resizing. I don't insist on this changing especially if it'll lead to some inconsistency. But I think it is useful. Please tell me if I wrong

GalyaZalesskaya:
@Wovchena Thank you for your detailed feedback and explanations that got me a little confounded. I'll investigate your suggestions

eizamaliev:
@GalyaZalesskaya it could be the problem, but in that case output frame should be resized according to screen resolution.

GalyaZalesskaya:
@eizamaliev Do you mean that the output image should be resized to the screen resolution that may differ from (1280, 720)?

Wovchena:
@GalyaZalesskaya, sorry, I have misled you in my first comment.
It's fine to ignore
> The solution breaks consistency between `draw_detections()` and `print_raw_results()`.

point.

> resize the image in the very beginning and that should save some computation when it is resized again to fit model resolution

seems to be super tricky because of model resolutions, thus ignore it as well.

All in all this is fine to resize the image in the very end, but there are two notes:
1. There is a concern that if bboxes and texts are first drawn and then an image is resized using big or small scale, the result may look ugly. But may be this will be fine :) It would be perfect if you took that into a consideration and checked if this is an issue.
2. Please add an argument for demos to specify target size. If it's not specified, the demos shouldn't resize images as before.

CC @vladimir-dudnik, @eizamaliev, @fzhar, @ivikhrev, @anzhella-pankratova to comment if I mixed something :)

> @GalyaZalesskaya it could be the problem, but in that case output frame should be resized according to screen resolution.
@eizamaliev Do you mean that the output image should be resized to the screen resolution that may differ from (1280, 720)?

Since there will be a cmd arg now, this becomes inaplicable.

GalyaZalesskaya:
@Wovchena Thank you for your comments again and especially for a certain suggestion about cmd parameter. Do you suggest making this change in Python and C++ object_detection demo or in each demo for consistency?

Wovchena:
Each we unified so far. For now these are human_pose_estimation_demo, object_detection_demo, segmentation_demo for both languages.

There is also classification_demo/cpp which has that resize, though it treats it a bit differently. You can take an idea for key naming here, or update classification_demo/cpp's key as well if you can come up with better naming.

As for other demos, it is up to you, feel free to update them as well if you want to.

GalyaZalesskaya:
Hi, everyone. I did changes for the python demos, but had issues with C++ demos, its modification, and building. I still can't fix it, so I decided to close this PR and keep it for a better time. Thank you all for your feedback.

PR #2217, opened 2021-03-19
AC: use defusedxml to read XML files instead of ET/lxml

IRDonch:
This solves several problems:

* it removes the use of the deprecated `cElementTree` module;
* it prevents developers from accidentally using lxml-specific features (AC doesn't require lxml, so that's unreliable);
* it prevents DoS on malicious input (`ElementTree` is vulnerable to entity expansion attacks).

### DIFF
--- a/tools/accuracy_checker/accuracy_checker/utils.py
+++ b/tools/accuracy_checker/accuracy_checker/utils.py
@@ -31,13 +31,9 @@ from warnings import warn
 from collections.abc import MutableSet, Sequence
 from io import BytesIO
 
+import defusedxml.ElementTree as et
 import numpy as np
 import yaml
-
-try:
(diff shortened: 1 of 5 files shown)


tools/accuracy_checker/accuracy_checker/utils.py
```diff
@@ -31,14 +31,10 @@
 from collections.abc import MutableSet, Sequence
 from io import BytesIO
 
+import defusedxml.ElementTree as et
```
eaidova, line 34: please wrap to unsupprted package for avaoid situation, when this package is not installed (by default we install only core dependency on yocto) or move to core-requirements (as I know VOC format converter uses et and it should continue work on yocto)

eaidova (approved):

PR #2223, opened 2021-03-22
Add resizer to ocv_common

ivikhrev:
Some models in object_detection_demo requires letterbox resize for input to produce correct results on image (ex. retinaface), this PR provides it. Also it would be nice to have function which scale bounding box coordinates back to origin image, it implemented in the same class.
- [ ] Update centernet model class

- [ ] Update ssd model class

- [ ] Update yolo model class


ivikhrev:
Closed due to work done by @fzhar in PR #2415

PR #2225, opened 2021-03-23
AC: ignore non image inputs in normalization

eaidova:
by default normalization applied to all model inputs, however not all of them may require normalization. Prevent usage normalization parameters for non image inputs in case, when user explicitly set it

### DIFF
--- a/tools/accuracy_checker/accuracy_checker/preprocessor/README.md
+++ b/tools/accuracy_checker/accuracy_checker/preprocessor/README.md
@@ -95,7 +95,7 @@ Accuracy Checker supports following set of preprocessors:
 * `point_alignment` - aligning keypoints stored in annotation metadata.
   * `draw_points` - allows visualize points.
   * `normalize` - allows to use normalization for keypoints.
-  * `dst_width` and `dst_height` are destination width and height for keypoints resizing respectively. You can also use `size` instead in case when destination sizes are equal.*
+  * `dst_width` and `dst_height` are destination width and height for keypoints resizing respectively. You can also use `size` instead in case when destination sizes are equal.
 * `corner_crop` - Corner crop of the image.
   * `dst_width` and `dst_height` are destination width and height
   * `corner_type` is type of the corner crop. Options are:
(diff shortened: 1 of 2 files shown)


tools/accuracy_checker/accuracy_checker/preprocessor/README.md
```diff
@@ -94,7 +95,7 @@ Accuracy Checker supports following set of preprocessors:
 * `point_alignment` - aligning keypoints stored in annotation metadata.
   * `draw_points` - allows visualize points.
   * `normalize` - allows to use normalization for keypoints.
-  * `dst_width` and `dst_height` are destination width and height for keypoints resizing respectively. You can also use `size` instead in case when destination sizes are equal.
+  * `dst_width` and `dst_height` are destination width and height for keypoints resizing respectively. You can also use `size` instead in case when destination sizes are equal.*
```
eaidova, line 98:
```suggestion
  * `dst_width` and `dst_height` are destination width and height for keypoints resizing respectively. You can also use `size` instead in case when destination sizes are equal.
```

tools/accuracy_checker/accuracy_checker/preprocessor/normalization.py
```diff
@@ -49,18 +49,27 @@ def parameters(cls):
...
         self.std = self.get_value_from_config('std')
+        self.images_only = self.get_value_from_config('images_only')
         if not self.mean and not self.std:
             raise ConfigError('mean or std value should be provided')
 
     def process(self, image, annotation_meta=None):
         def process_data(data, mean, std):
+            if (
+                self.images_only and len(data.shape) not in [2, 3] or
+                (len(data.shape) == 3 and data.shape[-1] not in [1, 3, 4])
```
eaidova, line 70:
```suggestion
                    self.images_only and len(data.shape) not in [2, 3] or
                    (len(data.shape) == 3 and data.shape[-1] not in [1, 3, 4])
```

Anna-Mironova (approved):

PR #2226, opened 2021-03-23
Bump all dependency versions

IRDonch:


eaidova:
@Anna-Mironova  FIY, torch will be updated on 1.8 as part of this request

eaidova (approved):

PR #2227, opened 2021-03-23
demos: use defusedxml to parse XML

IRDonch:
defusedxml prevents DoS due to malicious XML entities.

I don't think we need to add it to the requirements, because we only parse XML in supplementary scripts that likely won't be used by most users of the demos.


IRDonch:
CC @DmitriySidnev @evgeny-izutov

Wovchena (approved):

PR #2244, opened 2021-03-25
demos: replace some dicts with namespaces

IRDonch:
When the keys of a dictionary are identifier-like strings that are known ahead of time, I find `types.SimpleNamespace` to be better suited for the job than `dict`. The attribute access syntax feels more natural than indexing syntax, and the `repr` of namespace instances is more readable than that of `dict`s. Namespaces can also be replaced with more specific types (e.g. dataclasses) to improve type safety in the future, without changing the code that accesses the attributes.

Note that I import the type as `namespace`, because that's the name it uses in the `repr`.

### DIFF
--- a/demos/formula_recognition_demo/python/formula_recognition_demo.py
+++ b/demos/formula_recognition_demo/python/formula_recognition_demo.py
@@ -159,8 +159,8 @@ def non_interactive_demo(model, args):
     renderer = create_renderer()
     show_window = not args.no_show
     for rec in tqdm(model.images_list):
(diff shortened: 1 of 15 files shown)


IRDonch:
Jenkins please retry a build

Wovchena (approved):
Looking good. But should ignorance of C408 be removed now?

IRDonch:
> Looking good. But should ignorance of C408 be removed now?

I do plan to remove it, but IIRC a few more fixes are needed first.

PR #2245, opened 2021-03-25
AC: workaround fp16 output precision issue

eaidova:
there is problem with getting fp16 output from imported blobs (during translation from C++ to Python, it become I16), properly handle this out data type


eaidova:
Jenkins please retry a build

vladimir-dudnik (approved):

PR #2249, opened 2021-03-26
Enable flake8-comprehensions warning C408

IRDonch:
Fix the last remaining violations.

I would've preferred to turn the profiles in `speech_recognition_demo` into instances of `types.SimpleNamespace` (instead of changing the initialization syntax), but they have to be compatible with what `yaml.safe_load` returns, which is a dict.

Granted, it would be possible to convert the result of `yaml.safe_load` into a namespace, but then there would need to be error handling for the case that the keys are not strings; plus if the profile ever needs to contain nested dicts, those would need to be converted too. All in all, it's possible, but it looks like enough of a mess that I don't want to get into it. :-)

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -39,6 +39,3 @@ ignore=
     # ignored because putting a statement after a colon is harmless and
     # sometimes results in more compact code
     E701,E704,
-    # unnecessary <dict/list/tuple> call
-    # TODO: consider enabling this later
-    C408,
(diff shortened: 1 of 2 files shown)


eaidova (approved):

Wovchena (approved):

Wovchena:
Jenkins please retry a build

IRDonch:
Jenkins please retry a build

PR #2253, opened 2021-03-29
cocosnet: remove the PyTorch 1.7 workaround

IRDonch:
We now require PyTorch 1.8+, so it's no longer relevant.

This has no effect on the resulting IRs.


IRDonch:
CC @akorobeinikov

eaidova (approved):

PR #2255, opened 2021-03-30
Encode mask as uint8 to prevent scaling in resize

ngaloppo:
Currently the EncodeSegMask postprocessor encodes masks as int8
datatype. If one adds a ResizeSegmentationMask postprocessor to the
pipeline, unintended scaling to [0, 255] is caused.

This can be prevented by encoding as uint8 in the EncodeSegMask
postprocessor, since the scaling is not performed for such datatype
elements in the ResizeSegmentationMask postprocessor.

### DIFF
--- a/tools/accuracy_checker/accuracy_checker/postprocessor/encode_segmentation_mask.py
+++ b/tools/accuracy_checker/accuracy_checker/postprocessor/encode_segmentation_mask.py
@@ -52,7 +52,7 @@ class EncodeSegMask(PostprocessorWithSpecificTargets):
                 encoded_mask[np.where(
                     np.all(mask == color, axis=-1) if num_channels >= 3 else mask == color
                 )[:2]] = label
-            annotation_.mask = encoded_mask.astype(np.int8)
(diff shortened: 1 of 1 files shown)


eaidova:
@ngaloppo  thanks, but int8 was made intentionally, because in some cases class encoding have -1 as ignored class, so we made more advanced fix soon

eaidova:
@ngaloppo could you please explain or give specific example where you get such behaviour?
As i can see in resize function there is additional case, which is also avoid bytescaling
https://github.com/openvinotoolkit/open_model_zoo/blob/develop/tools/accuracy_checker/accuracy_checker/postprocessor/resize_segmentation_mask.py#L127-L128

We check that dtype is not float and data in [0, 255] range, in this case we cast data to uint8 and provide it to resize, so in your described case if everything is OK and encoded mask is really can be decoded to uint8, it should not be bytescaled

ngaloppo:
@eaidova You are right, the current implementation would not have this issue - I was testing with 2021.2. Looks like the behavior has been fixed since - assuming that the fix is now in 2021.3?

PR #2256, opened 2021-03-30
AC: support tacotron2

eaidova:
TO DO:

- [x]  runtime for onnx
- [x] runtime for openvino
- [x] check quantization
- [x] update docs


eaidova:
@a-a-egorovich I added shape alignment in l2_relative error. Possibly we can switch forward tacotron on this metric too?

a-a-egorovich:
> @a-a-egorovich I added shape alignment in l2_relative error. Possibly we can switch forward tacotron on this metric too?

a-a-egorovich:
Yes, it might be good

Anna-Mironova (approved):

PR #2264, opened 2021-04-01
Typo fix

isanghao:


demos/interactive_face_detection_demo/cpp/detectors.cpp
```diff
@@ -154,7 +154,7 @@ CNNNetwork FaceDetection::read(const InferenceEngine::Core& ie)  {
             throw std::logic_error("Face Detection network output layer should have 7 as a last dimension");
         }
         if (outputDims.size() != 4) {
-            throw std::logic_error("Face Detection network output dimensions not compatible shoulld be 4, but was " +
+            throw std::logic_error("Face Detection network output dimension should be 4, but was " +
```
IRDonch, line 157: It's actually the _number_ of dimensions that should be 4, not the dimension (a tensor doesn't just have a single dimension).

IRDonch (changes requested):

demos/interactive_face_detection_demo/cpp/detectors.cpp
isanghao: @IRDonch Hmm I think there is slight confusion of terminology. Dimension of the output is 4, while the shape(length of each axis) is different. If we draw something on the paper, we are saying that it is a "2-dimensional image", not "the image's number of dimension is 2".
From wikipedia: "_the dimension of a mathematical space (or object) is informally defined as the minimum number of coordinates needed to specify any point within it_." We need 4 numbers to specify a coordinate in this output. That means, the dimension of output is 4.

IRDonch: I think the terminology around dimensions is a bit loose; in fact, the same Wikipedia article you're quoting from later mentions "The four dimensions (4D) of spacetime", in which "dimension" refers to the individual coordinate axes and not the number of them.

But more to the point, in Inference Engine "dimension" clearly refers to the components of the tensor's shape, as evidenced by the naming and description of `TensorDesc::getDims`. Therefore, the number of elements in the shape is the number of dimensions.

If the wording "number of dimensions" bothers you, you can rewrite the message to something more natural, e.g. "Face detection network output should have 4 dimensions, but has X".

isanghao: OK. That sounds reasonable. Let me update the PR.

IRDonch (approved):

PR #2270, opened 2021-04-02
object_detection_demo.py: add import of ONNX models

anzhella-pankratova:
Add support of models in ONNX format for object_detection_demo.py:
- ctdet_coco_dlav0_384
- ctdet_coco_dlav0_512
- faceboxes-pytorch
- ssd-resnet34-1200-onnx
- ultra-lightweight-face-detection-rfb-320
- ultra-lightweight-face-detection-slim-320

The info about mean_values, scale_values and reverse_input_channels is located in `model.yml` for each model.

### DIFF
--- a/demos/common/python/models/centernet.py
+++ b/demos/common/python/models/centernet.py
@@ -58,8 +58,8 @@ class CenterNet(Model):
         resized_image = resized_image.transpose((2, 0, 1))  # Change data layout from HWC to CHW
         resized_image = resized_image.reshape((self.n, self.c, self.h, self.w))
 
-        if self.is_onnx_format and self.mean_values is not None:
-            resized_image = self.scaleshift(resized_image)
+        if self.is_onnx_format:
+            resized_image = (resized_image - self.mean_values[:, None, None]) / self.scale_values[:, None, None]
 
         dict_inputs = {self.image_blob_name: resized_image}
         return dict_inputs, meta
--- a/demos/common/python/models/centernet.py
+++ b/demos/common/python/models/centernet.py
@@ -51,7 +51,7 @@ class CenterNet(Model):
         scale = max(height, width)
         trans_input = self.get_affine_transform(center, scale, 0, [self.w, self.h])
         resized_image = cv2.warpAffine(image, trans_input, (self.w, self.h), flags=cv2.INTER_LINEAR)
-        resized_image = self.input_transform.apply(resized_image)
+        resized_image = self.input_transform(resized_image)
         resized_image = resized_image.transpose((2, 0, 1))  # Change data layout from HWC to CHW
         resized_image = resized_image.reshape((self.n, self.c, self.h, self.w))
 
--- a/demos/common/python/models/centernet.py
+++ b/demos/common/python/models/centernet.py
@@ -23,9 +23,8 @@ from .utils import Detection, load_labels
 
 
 class CenterNet(Model):
-    def __init__(self, ie, model_path, reverse_input_channels, mean_values, scale_values,
-                 labels=None, threshold=0.3):
-        super().__init__(ie, model_path, reverse_input_channels, mean_values, scale_values)
+    def __init__(self, ie, model_path, input_transform, labels=None, threshold=0.3):
+        super().__init__(ie, model_path, input_transform)
 
         assert len(self.net.input_info) == 1, "Expected 1 input blob"
         assert len(self.net.outputs) == 3, "Expected 3 output blobs"
@@ -52,14 +51,12 @@ class CenterNet(Model):
         scale = max(height, width)
         trans_input = self.get_affine_transform(center, scale, 0, [self.w, self.h])
         resized_image = cv2.warpAffine(image, trans_input, (self.w, self.h), flags=cv2.INTER_LINEAR)
-        if self.is_onnx_format and self.reverse_input_channels:
+        if self.input_transform.reverse_input_channels:
             resized_image = cv2.cvtColor(resized_image, cv2.COLOR_BGR2RGB)
+        resized_image = (resized_image - self.input_transform.mean_values) / self.input_transform.scale_values
 
         resized_image = resized_image.transpose((2, 0, 1))  # Change data layout from HWC to CHW
         resized_image = resized_image.reshape((self.n, self.c, self.h, self.w))
-
-        if self.is_onnx_format:
-            resized_image = (resized_image - self.mean_values[:, None, None]) / self.scale_values[:, None, None]
 
         dict_inputs = {self.image_blob_name: resized_image}
         return dict_inputs, meta
--- a/demos/common/python/models/centernet.py
+++ b/demos/common/python/models/centernet.py
@@ -59,7 +59,7 @@ class CenterNet(Model):
         resized_image = resized_image.reshape((self.n, self.c, self.h, self.w))
 
         if self.is_onnx_format and self.mean_values is not None:
-            resized_image = self.normalize(resized_image)
+            resized_image = self.scaleshift(resized_image)
 
         dict_inputs = {self.image_blob_name: resized_image}
         return dict_inputs, meta
--- a/demos/common/python/models/centernet.py
+++ b/demos/common/python/models/centernet.py
@@ -51,10 +51,7 @@ class CenterNet(Model):
         scale = max(height, width)
         trans_input = self.get_affine_transform(center, scale, 0, [self.w, self.h])
         resized_image = cv2.warpAffine(image, trans_input, (self.w, self.h), flags=cv2.INTER_LINEAR)
-        if self.input_transform.reverse_input_channels:
-            resized_image = cv2.cvtColor(resized_image, cv2.COLOR_BGR2RGB)
-        resized_image = (resized_image - self.input_transform.mean_values) / self.input_transform.scale_values
-
+        resized_image = self.input_transform.apply(resized_image)
         resized_image = resized_image.transpose((2, 0, 1))  # Change data layout from HWC to CHW
         resized_image = resized_image.reshape((self.n, self.c, self.h, self.w))
 
--- a/demos/common/python/models/centernet.py
+++ b/demos/common/python/models/centernet.py
@@ -23,8 +23,9 @@ from .utils import Detection, load_labels
 
 
 class CenterNet(Model):
-    def __init__(self, ie, model_path, labels=None, threshold=0.3):
-        super().__init__(ie, model_path)
+    def __init__(self, ie, model_path, reverse_input_channels, mean_values, scale_values,
+                 labels=None, threshold=0.3):
+        super().__init__(ie, model_path, reverse_input_channels, mean_values, scale_values)
 
         assert len(self.net.input_info) == 1, "Expected 1 input blob"
         assert len(self.net.outputs) == 3, "Expected 3 output blobs"
@@ -51,7 +52,14 @@ class CenterNet(Model):
         scale = max(height, width)
         trans_input = self.get_affine_transform(center, scale, 0, [self.w, self.h])
         resized_image = cv2.warpAffine(image, trans_input, (self.w, self.h), flags=cv2.INTER_LINEAR)
-        resized_image = np.transpose(resized_image, (2, 0, 1))
+        if self.is_onnx_format and self.reverse_input_channels:
+            resized_image = cv2.cvtColor(resized_image, cv2.COLOR_BGR2RGB)
+
+        resized_image = resized_image.transpose((2, 0, 1))  # Change data layout from HWC to CHW
+        resized_image = resized_image.reshape((self.n, self.c, self.h, self.w))
+
+        if self.is_onnx_format and self.mean_values is not None:
+            resized_image = self.normalize(resized_image)
 
         dict_inputs = {self.image_blob_name: resized_image}
         return dict_inputs, meta
--- a/demos/common/python/models/model.py
+++ b/demos/common/python/models/model.py
@@ -15,14 +15,19 @@
 """
 
 import logging
-
+import numpy as np
 
 class Model:
-    def __init__(self, ie, model_path):
+    def __init__(self, ie, model_path, reverse_input_channels=None, mean_values=None, scale_values=None):
         self.logger = logging.getLogger()
         self.logger.info('Reading network from IR...')
         self.net = ie.read_network(model_path, model_path.with_suffix('.bin'))
         self.set_batch_size(1)
+
+        self.is_onnx_format = model_path.suffix == '.onnx'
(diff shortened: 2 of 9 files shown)


demos/common/python/models/model.py
```diff
@@ -37,3 +42,9 @@ def set_batch_size(self, batch):
             new_shape = [batch] + self.net.input_info[input_layer].input_data.shape[1:]
             shapes.update({input_layer: new_shape})
         self.net.reshape(shapes)
+
+    def normalize(self, inputs):
+        normalized_inputs = np.zeros(inputs.shape, dtype=np.float32)
+        for i in range(inputs.shape[1]):
+            normalized_inputs[:, i, :, :] = (inputs[:, i, :, :] - self.mean_values[i]) / self.scale_values[i]
+        return normalized_inputs
```
Wovchena, line 50:
```suggestion
        return (inputs - self.mean_values[:, None, None]) / self.scale_values[:, None, None]
```
Just force dtype of `self.mean_values` and `self.scale_values` to `np.float32`.
Since this is one line of code, I don't think you need a wrapping function. Or you can put there checks determining if rescaling should be performed.
If you keep the function, may be name it `scaleshift()` since the image itself isn't always normed, norm may depend on the whole dataset.

demos/common/python/models/centernet.py
```diff
@@ -51,7 +52,14 @@ def preprocess(self, inputs):
...
         trans_input = self.get_affine_transform(center, scale, 0, [self.w, self.h])
         resized_image = cv2.warpAffine(image, trans_input, (self.w, self.h), flags=cv2.INTER_LINEAR)
-        resized_image = np.transpose(resized_image, (2, 0, 1))
+        if self.is_onnx_format and self.reverse_input_channels:
+            resized_image = cv2.cvtColor(resized_image, cv2.COLOR_BGR2RGB)
+
+        resized_image = resized_image.transpose((2, 0, 1))  # Change data layout from HWC to CHW
+        resized_image = resized_image.reshape((self.n, self.c, self.h, self.w))
+
+        if self.is_onnx_format and self.mean_values is not None:
```
Wovchena, line 61: If `self.mean_values` is `None`, `self.scale_values` is silently discarded.

demos/common/python/models/model.py
```diff
@@ -15,15 +15,20 @@
 """
 
 import logging
-
+import numpy as np
 
 class Model:
-    def __init__(self, ie, model_path):
+    def __init__(self, ie, model_path, reverse_input_channels=None, mean_values=None, scale_values=None):
```
Wovchena, line 21:
```suggestion
    def __init__(self, ie, model_path, reverse_input_channels, mean_values, scale_values):
```
`Model`s' children don't have default values (and I think they shouldn't).

Wovchena (changes requested):

vladimir-dudnik:
we also need to think how to enable this in test cases,

demos/common/python/models/model.py
```diff
@@ -15,15 +15,20 @@
...
         self.logger = logging.getLogger()
         self.logger.info('Reading network from IR...')
         self.net = ie.read_network(model_path, model_path.with_suffix('.bin'))
```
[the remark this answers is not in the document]
anzhella-pankratova: Updated.

demos/common/python/models/model.py
anzhella-pankratova: Updated.

demos/common/python/models/centernet.py
anzhella-pankratova: I think condition is correct, because `self.scale_values` are used only if `self.mean_values` are set.

demos/common/python/models/model.py
anzhella-pankratova: The default values are necessary here, otherwise the models constructor with unified API (like segmentation) is failed in `super().__init__(ie, model_path)` part.

demos/common/python/models/model.py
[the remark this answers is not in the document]
Wovchena: I suggest calling `self.net = ie.read_network(model_path)` unconditionally. We can rely on `read_network()` deducing second arg for IR.

demos/common/python/models/model.py
Wovchena: Checking only `scale_values` is None in `scaleshift()` is strange. `mean_values` and `scale_values` are very similar so they should go together, but here you split them by checking `mean_values` in `preprocess()` and `scale_values` in `scaleshift()`. Move `self.mean_values is not None` into `scaleshift()`.

demos/common/python/models/centernet.py
Wovchena: Yes, that's what happens, but it can be unexpected. The only relation of `scale_values` and `mean_values` you mention is `Division is applied after mean values subtraction.` It is ambiguous what happens if `mean_values` is None. One may assume `scale_values` is still applied.

Wovchena: Just realized it can be tempting for you now to only update a help message :) But raising exception or applying `scale_values` is better. Exceptions can't be ignored compared to help messages by a reader. And applying only `scale_values` is what MO does, I assume. I think we should mimic MO.

demos/common/python/models/model.py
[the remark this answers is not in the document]
anzhella-pankratova: Done.

demos/common/python/models/model.py
anzhella-pankratova: I simplified the code, so all cases are considered without additional conditions. Now it seems to be consistent with documentation and MO.

Wovchena: There is no reason to support `None` now. It should be (1.0, 1.0, 1.0) and (0.0, 0.0, 0.0) as default key values. This is more explicit.

MO probably has `None` which doesn't add rescaling at all. So actually the solution with unconditional rescaling is inconsistent with MO. But I think it doesn't matter much for us.

anzhella-pankratova: Changed default values.

Wovchena (approved):

demos/object_detection_demo/python/object_detection_demo.py
```diff
@@ -139,8 +151,9 @@ def __len__(self):
 
 
 def get_model(ie, args):
+    common_args = (ie, args.model, args.reverse_input_channels, args.mean_values, args.scale_values)
```
IRDonch, line 154: IMO, the latter three values should be combined into a class (e.g. `ModelInputTransforms`) and passed around as an object of that class. It would be clearer than adding 3 positional arguments to every model constructor.

demos/object_detection_demo/python/object_detection_demo.py
```diff
@@ -139,8 +151,9 @@ def __len__(self):
...
     if args.architecture_type == 'ssd':
-        return models.SSD(ie, args.model, labels=args.labels, keep_aspect_ratio_resize=args.keep_aspect_ratio)
+        return models.SSD(*common_args, labels=args.labels, keep_aspect_ratio_resize=args.keep_aspect_ratio)
     elif args.architecture_type == 'ctpn':
         return models.CTPN(ie, args.model, input_size=args.input_size, threshold=args.prob_threshold)
```
IRDonch, line 158: If a model doesn't support input transforms, then the demo should produce an error if the user tries to use it with transforms.

IRDonch (changes requested):

demos/common/python/models/centernet.py
```diff
@@ -51,7 +52,14 @@ def preprocess(self, inputs):
         scale = max(height, width)
         trans_input = self.get_affine_transform(center, scale, 0, [self.w, self.h])
         resized_image = cv2.warpAffine(image, trans_input, (self.w, self.h), flags=cv2.INTER_LINEAR)
-        resized_image = np.transpose(resized_image, (2, 0, 1))
+        if self.is_onnx_format and self.reverse_input_channels:
```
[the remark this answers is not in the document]
Wovchena: It is usually encouraged to fuse channel reversal/scaling into IR. Allowing that options in demo will have the opposite effect.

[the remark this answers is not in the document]
IRDonch: Maybe, but so what? It costs us nothing to support this; in fact, it takes extra effort on our part to _prohibit_ this. So why bother?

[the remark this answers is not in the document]
anzhella-pankratova: Removed the checking.

demos/common/python/models/centernet.py
```diff
@@ -51,7 +52,14 @@ def preprocess(self, inputs):
...
+            resized_image = cv2.cvtColor(resized_image, cv2.COLOR_BGR2RGB)
+
+        resized_image = resized_image.transpose((2, 0, 1))  # Change data layout from HWC to CHW
+        resized_image = resized_image.reshape((self.n, self.c, self.h, self.w))
+
+        if self.is_onnx_format:
+            resized_image = (resized_image - self.mean_values[:, None, None]) / self.scale_values[:, None, None]
```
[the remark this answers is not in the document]
anzhella-pankratova: Done.

demos/object_detection_demo/python/object_detection_demo.py
anzhella-pankratova: Done.

demos/object_detection_demo/python/object_detection_demo.py
anzhella-pankratova: Done.

demos/common/python/models/ultra_lightweight_face_detection.py
```diff
@@ -58,6 +59,10 @@ def preprocess(self, inputs):
         resized_image = resize_image(image, (self.w, self.h))
         meta = {'original_shape': image.shape,
                 'resized_shape': resized_image.shape}
+        if self.input_transform.reverse_input_channels:
+            resized_image = cv2.cvtColor(resized_image, cv2.COLOR_BGR2RGB)
+        resized_image = (resized_image - self.input_transform.mean_values) / self.input_transform.scale_values
```
IRDonch, line 64: This code is repeated everywhere; it would make sense to move it to the `InputTransform` class itself (e.g. `resized_image = self.input_transform.apply(resized_image)`).

demos/object_detection_demo/python/object_detection_demo.py
```diff
@@ -139,8 +151,17 @@ def __len__(self):
 
 
 def get_model(ie, args):
+    if args.architecture_type in ('ctpn', 'yolo', 'yolov4', 'retinaface') and \
+       (
+           args.reverse_input_channels or
+           not np.array_equal(args.mean_values, [0., 0., 0.]) or
+           not np.array_equal(args.scale_values, [1., 1., 1.])
+       ):
```
IRDonch, line 159: This also seems like it should be moved to `InputTransform` (e.g. `input_transform.is_trivial`).

demos/object_detection_demo/python/object_detection_demo.py
```diff
@@ -91,6 +91,18 @@ def build_argparser():
...
+    onnx_args.add_argument('--reverse_input_channels', default=False, action='store_true',
+                         help='Optional. Switch the input channels order from '
+                              'BGR to RGB for onnx model.')
+    onnx_args.add_argument('--mean_values', default=(0.0, 0.0, 0.0), type=float, nargs=3,
+                         help='Optional. Normalize input by subtracting the mean values '
+                              'per channel for onnx model. Example: 255 255 255')
+    onnx_args.add_argument('--scale_values', default=(1.0, 1.0, 1.0), type=float, nargs=3,
+                         help='Optional. Divide input by scale values per channel for onnx model. '
+                              'Division is applied after mean values '
+                              'subtraction. Example: 255 255 255')
```
IRDonch, line 104: I think the references to ONNX can now be removed, since the options will work for any type of model.

demos/common/python/models/utils.py
```diff
@@ -42,6 +42,22 @@ def __init__(self, xmin, ymin, xmax, ymax, score, id, landmarks_x, landmarks_y):
...
+        self.mean_values = np.array(mean_values, dtype=np.float32)
+        self.scale_values = np.array(scale_values, dtype=np.float32)
+
+    def is_trivial(self):
+        return np.array_equal(self.mean_values, [0., 0., 0.]) and \
+               np.array_equal(self.scale_values, [1., 1., 1.]) and \
+               not self.reverse_input_channels
+
+    def apply(self, inputs):
+        if self.reverse_input_channels:
```
IRDonch, line 57: If the transform is trivial, you should immediately return `inputs` to avoid useless operations. You should also cache the result of the triviality check so that you don't compare the vectors every time. I suggest you just remove the is_trivial function and instead initialize `is_trivial` in the constructor.

anzhella-pankratova: Done

IRDonch (approved):

demos/common/python/models/utils.py
```diff
@@ -42,6 +42,22 @@ def __init__(self, xmin, ymin, xmax, ymax, score, id, landmarks_x, landmarks_y):
...
 
 
+class InputTransform:
+    def __init__(self, reverse_input_channels, mean_values, scale_values):
+        self.reverse_input_channels = reverse_input_channels
+        self.mean_values = np.array(mean_values, dtype=np.float32)
+        self.scale_values = np.array(scale_values, dtype=np.float32)
+        self.is_trivial = np.array_equal(self.mean_values, [0., 0., 0.]) and \
+                          np.array_equal(self.scale_values, [1., 1., 1.]) and \
+                          not self.reverse_input_channels
```
eizamaliev, line 52: Comparing floating point values are always tricky. I propose to use default value `None` for mean and scales here and in arguments. And this line would transform to
```python
self.use_transform = self.mean_values or self.scale_values or self.reverse_input_channels
```

demos/common/python/models/utils.py
```diff
@@ -42,6 +42,22 @@ def __init__(self, xmin, ymin, xmax, ymax, score, id, landmarks_x, landmarks_y):
...
+
+    def apply(self, inputs):
```
eizamaliev, line 54: I prefer `__call__(self, inputs):` here. It allows using `lambda x: x` in model ctor or anywhere else to set none transforms. And it would be called in model::preprocess as:
```python
data = self.input_transfrom(data)
```

eizamaliev (changes requested):

demos/common/python/models/utils.py
anzhella-pankratova: Done

demos/common/python/models/utils.py
anzhella-pankratova: Done

eizamaliev (approved):

PR #2248, opened 2021-03-26
Aligned format of input/output description in documentation for models

Anna-Mironova:
- Added input/ output description alignment
- Added typo and punctuation fixes

PR #2290, opened 2021-04-09
Updated model description in model.yml files

Anna-Mironova:
Updated public model description in model.yml files according to changes in https://github.com/openvinotoolkit/open_model_zoo/pull/2248


eaidova (approved):

IRDonch:
The Intel model descriptions should not be changed, as they will just get reset the next time the Intel models are republished.

eizamaliev (approved):

IRDonch (approved):

PR #2292, opened 2021-04-09
Rename all model documentation files to README.md

IRDonch:
Currently, all model documentation files are named either `<model-name>.md` or `description/<model-name>.md`. The former is mainly used for public models and the latter for Intel models, although there are exceptions. This inconsistency seems to stem from historic reasons, and doesn't need to exist.

I could just solve the inconsistency by moving the Markdown files out of the `description` directory and be done with it, but I also rename all of them to `README.md` for a number of reasons:

* The documentation file is currently the only file in the model subdirectory whose name varies depending on the name of the model. It's inconsistent with the other files, it's semantically redundant, and it creates extra work when someone is renaming or cloning a model.

* The documentation files for demos are also named `README.md`, so it's consistent with that.

* GitHub automatically shows the `README.md` file when the user opens a directory, which makes it easier to look at model documentation.

* `README.md` is short and doesn't grow with the name of the model, which means we can add models with longer names without running into the path length limit.

Note that the `description` subdirectory remains for models that had images in it. I don't think we should move the images directly to the model directory, because it would clutter the directory with files that the user doesn't need to directly interact with.

It's also worth noting that, while it's tempting to also rename all the illustrations according to a common naming scheme (e.g. `example.png`), it cannot be done, because Doxygen doesn't work correctly if there are multiple images with the same name (because it moves all images to the root of the output directory).


IRDonch:
CC @Anna-Mironova @jkamelin @ntyukaev

Note that this should require no changes on the OpenVINO documentation side, because the documentation page IDs remain unchanged.

eaidova:
@akorobeinikov @morkovka1337  it may affect your requests with addition models, FYI

eaidova (approved):

vladimir-dudnik:
cc @andrew-zaytsev

vladimir-dudnik (approved):
is it worth to consider rename subfolder description to data, as there are no actually any description in it?

IRDonch:
> is it worth to consider rename subfolder description to data, as there are no actually any description in it?

I'm not opposed to it (although I would prefer to name it `assets`), but I don't think it's important enough to add to this PR. Maybe later.

eizamaliev (approved):
Next step - rename `description` subdirectory to `source` or `images` or else.

IRDonch:
Jenkins please retry a build

PR #2324, opened 2021-04-19
demos: fix exit on Esc for Python, move framesProcessed++; to align with Python

Wovchena:


IRDonch:
Looks fine to me, but how does this fix exit on Esc? I don't see anything that could affect that.

Wovchena:
`next_frame_id_to_show += 1` keeps `next_frame_id_to_show` valid now independently on a way the loop is broke. So the next loop, after `pipeline.await_all()`, always takes valid value of `next_frame_id_to_show`.

IRDonch (approved):

PR #2336, opened 2021-04-21
Add telemetry to model downloader

jkamelin:


vladimir-dudnik:
cc @lazarevevgeny (FYI)

tools/downloader/converter.py
```diff
@@ -102,11 +110,29 @@ def main():
...
+    except NameError:
+        pass
+
     mo_path = args.mo
     if mo_path is None:
         try:
             mo_path = Path(os.environ['INTEL_OPENVINO_DIR']) / 'deployment_tools/model_optimizer/mo.py'
         except KeyError:
+            try:
+                telemetry.end_session()
```
eaidova, line 132: how end session connected with error that MO is not found?

eaidova: or at least I think we need also send error like with unknown precision

tools/downloader/converter.py
```diff
@@ -197,8 +252,25 @@ def convert(reporter, model):
     if failed_models:
         reporter.print('FAILED:')
         for failed_model_name in failed_models:
+            try:
+                telemetry.send_event('model_converter', 'failed', failed_model_name)
+            except NameError:
+                pass
             reporter.print(failed_model_name)
+        try:
+            telemetry.send_event('model_converter', 'result', 'fail')
```
eaidova, line 261: if you already send event per model, why do you need additional event with just status?
BTW, twhat the difference between send_error and send_evenl?

tools/downloader/converter.py
```diff
@@ -139,6 +177,10 @@ def convert(reporter, model):
         (output_dir / model.subdirectory).mkdir(parents=True, exist_ok=True)
 
         if not run_pre_convert(reporter, model, output_dir, args):
+            try:
+                telemetry.send_error('Pre-conversion script failed.')
```
eaidova, line 181: should it be extended with model name (just for clarification which script failed)?

tools/downloader/converter.py
```diff
@@ -152,6 +194,10 @@ def convert(reporter, model):
 
         if model.conversion_to_onnx_args:
             if not convert_to_onnx(reporter, model, output_dir, args, template_variables):
+                try:
+                    telemetry.send_error('Conversion to ONNX failed.')
```
eaidova, line 198: the same question about model like with preconvert

tools/downloader/converter.py
jkamelin: My intention of sending the status was to collect statistics about failures and successes. Per-model statistics shows what models fails more.
The difference between `send_event` is that it sends events witch can be classified by category and actions, `send_error` basically sends error message, thought it isn't implemented yet https://github.com/openvinotoolkit/telemetry/blob/master/src/main.py#L84, probably we shouldn't use it.

tools/downloader/converter.py
```diff
@@ -118,13 +145,25 @@ def main():
...
 
     models = common.load_models_from_args(parser, args)
 
     output_dir = args.download_dir if args.output_dir is None else args.output_dir
 
     def convert(reporter, model):
+        try:
+            telemetry = tm.Telemetry()
+            telemetry.send_event('md', 'converter_model_name', model.name)
+            telemetry.send_event('md', 'converter_framework', model.framework)
```
eaidova, line 164: @AnastasiaKazantaeva if we send single event with dict {'model': model_name, 'framework': model.framework}, is it guarantee the same result as these 2 separated events? (I mean can we agregate statistics per model and per framework in the same time or only for union)?

tools/downloader/downloader.py
```diff
@@ -234,6 +242,15 @@ def try_retrieve(reporter, destination, model_file, cache, num_attempts, start_d
     return success
 
 def download_model(reporter, args, cache, session_factory, requested_precisions, model):
+    try:
+        telemetry = tm.Telemetry()
```
eaidova, line 246: why in some places do  you provide parameters to constructor and in other not?

jkamelin: because by the time we got here, the telemetry instance has already been initialized in the `main` https://github.com/openvinotoolkit/telemetry/blob/master/src/main.py#L22

eaidova: ok

tools/downloader/converter.py
AnastasiaKazantaeva: sending 2 events we can separate information: most popular models, most popular framework. If we want to know most popular models and its framework, need to send it in dict in one event. It depends on info you want to get finally

IRDonch: Why do we need to send the framework at all? It's uniquely derivable from the model name.

AnastasiaKazantaeva: @IRDonch if it is included in model name, then there is no need to send framework name

jkamelin: @IRDonch we don't have framework in every model name

eaidova: but we actually know which framework each model belong

jkamelin: So, we don't need to aggregate statistics over frameworks? Models statistics is enough?

AnastasiaKazantaeva: @eaidova @jkamelin but do you need to quickly build info about framework? top1 - framework1, top2 - ..., top3 ...
building this stats will be complicated knowing only model name. But this stats will help to prioritize direct support about frameworks. I'd love to have both

eaidova: @AnastasiaKazantaeva my initial question was just difference between dict and string parameters sending result. I read readme in telemetry repo and this topic looks uncovered for me (now, after your answer,  I understand difference, thanks). I have nothing against frameworks :)

jkamelin: I'd like to have framework statistics. @IRDonch how do you suggest to derive frameworks from models?

AnastasiaKazantaeva: @eaidova great

tools/downloader/src/open_model_zoo/model_tools/converter.py
```diff
@@ -104,11 +109,30 @@ def main():
...
+        telemetry.start_session('md')
+
+        if args.all:
+            telemetry.send_event('md', 'converter_selection_mode', 'all')
+        if args.list:
+            telemetry.send_event('md', 'converter_selection_mode', 'list')
+        if args.name:
+            telemetry.send_event('md', 'converter_selection_mode', 'name')
+    except NameError:
+        pass
```
IRDonch, line 123: All these exception handlers are noisy and could mask `NameError`s that are unrelated to telemetry. I would suggest to instead add something like this to `_common`:

```python
try:
    from openvino_telemetry import Telemetry
except ImportError:
    class Telemetry:
        def __init__(self, tool_name, tid): pass
        ...
```

Then you just import `Telemetry` from `_common` and use it without any error handling.

tools/downloader/src/open_model_zoo/model_tools/converter.py
```diff
@@ -104,11 +109,30 @@ def main():
 
     args = parser.parse_args()
 
+    try:
+        telemetry = tm.Telemetry('Model Converter', tid='UA-194864834-1')
```
IRDonch, line 113: The tid is repeated everywhere; it should be moved to `common`.

tools/downloader/src/open_model_zoo/model_tools/converter.py
```diff
@@ -104,11 +109,30 @@ def main():
...
+    except NameError:
+        pass
+
     mo_path = args.mo
     if mo_path is None:
         try:
             mo_path = Path(os.environ['INTEL_OPENVINO_DIR']) / 'deployment_tools/model_optimizer/mo.py'
         except KeyError:
+            try:
+                telemetry.send_event('md', 'converter_error', str({'error': 'Unable to locate Model Optimizer.', 'model': 'none'}))
```
IRDonch, line 131: I assume that the third argument here is intended for statistical analysis, in which case I have several questions/suggestions about it:

1. Why do you use a stringified Python literal as the value? Wouldn't a more commonly used data interchange format (like JSON) work better?
2. There doesn't seem to be much reason to include a human-readable error message here; a shorter error code (e.g. `unable-to-locate-mo`) would work better.
3. What's the purpose of `'model': 'none'` here? First, it is redundant, and second, "none" here is a string which would be a valid model name, so it's misleading too.

tools/downloader/src/open_model_zoo/model_tools/converter.py
```diff
@@ -104,11 +109,30 @@ def main():
...
+                telemetry.end_session('md')
+                telemetry.force_shutdown(1.0)
```
IRDonch, line 133: I see this fragment everywhere; perhaps it would be better to wrap all code covered by the telemetry in a `try` block and move this code to the `finally` block?

(You can factor out the contents of the `try` block into a new function so that you don't have to reindent everything.)

tools/downloader/src/open_model_zoo/model_tools/converter.py
```diff
@@ -120,13 +144,26 @@ def main():
...
+                pass
             sys.exit('Unknown precisions specified: {}.'.format(', '.join(sorted(unknown_precisions))))
 
     models = _common.load_models_from_args(parser, args)
 
     output_dir = args.download_dir if args.output_dir is None else args.output_dir
 
     def convert(reporter, model):
+        try:
+            telemetry = tm.Telemetry()
```
IRDonch, line 161: Not sure what the point of this is - `telemetry` should still be accessible from the outer function scope.

tools/downloader/src/open_model_zoo/model_tools/converter.py
```diff
@@ -120,13 +144,26 @@ def main():
...
+            telemetry.send_event('md', 'converter_model_name', model.name)
+            telemetry.send_event('md', 'converter_framework', model.framework)
```
IRDonch, line 163: Not sure this is the right time to report this - ideally we should report all models that the user requested at the _beginning_ of the session, so that if the converter is terminated in the middle, we still know what the user _wanted_ to convert.

tools/downloader/src/open_model_zoo/model_tools/converter.py
```diff
@@ -199,8 +253,25 @@ def convert(reporter, model):
     if failed_models:
         reporter.print('FAILED:')
         for failed_model_name in failed_models:
+            try:
+                telemetry.send_event('md', 'converter_failed_models', failed_model_name)
```
IRDonch, line 257: This event is also too late - if a model fails, then we should report that immediately, not at the end.

IRDonch (changes requested):

tools/downloader/src/open_model_zoo/model_tools/converter.py
jkamelin: Done

tools/downloader/src/open_model_zoo/model_tools/converter.py
jkamelin: 1. According to current telemetry realization `event_label` should be a string https://github.com/openvinotoolkit/telemetry/blob/master/src/main.py#L49
2. Changed error codes
3. According to https://github.com/openvinotoolkit/telemetry#readme we need to send all keys if we use dict as a label

tools/downloader/src/open_model_zoo/model_tools/converter.py
jkamelin: Moved to `common` as a `start_telemetry` parameter, although this tid is temporary, when everything is done and tested, this will be removed, since telemetry has tid of common base [backend_ga.py#L18](https://github.com/openvinotoolkit/telemetry/blob/master/src/backend/backend_ga.py#L18)

tools/downloader/src/open_model_zoo/model_tools/converter.py
jkamelin: Done

tools/downloader/src/open_model_zoo/model_tools/converter.py
jkamelin: Removed

tools/downloader/src/open_model_zoo/model_tools/converter.py
jkamelin: Moved to the beginning

tools/downloader/src/open_model_zoo/model_tools/converter.py
jkamelin: Done

tools/downloader/src/open_model_zoo/model_tools/converter.py
```diff
@@ -104,103 +104,131 @@ def main():
...
+
+        extra_mo_args = args.extra_mo_args or []
+
+        if args.precisions is None:
+            requested_precisions = _common.KNOWN_PRECISIONS
+        else:
+            requested_precisions = set(args.precisions.split(','))
+            unknown_precisions = requested_precisions - _common.KNOWN_PRECISIONS
+            if unknown_precisions:
+                telemetry.send_event('md', 'converter_error', str({'error': 'unknown-precisions', 'model': 'none'}))
```
IRDonch, line 139: Is this error worth reporting? It's basically just an invalid argument value - I'm not sure we're gonna get any insights by collecting these.

tools/downloader/src/open_model_zoo/model_tools/converter.py
```diff
@@ -104,103 +104,131 @@ def main():
...
+            if not model_precisions:
+                reporter.print_section_heading('Skipping {} (all conversions skipped)', model.name)
+                reporter.print()
+                return True
+
+            (output_dir / model.subdirectory).mkdir(parents=True, exist_ok=True)
+
+            if not run_pre_convert(reporter, model, output_dir, args):
+                telemetry.send_event('md', 'converter_failed_models', model.name)
+                telemetry.send_event('md', 'converter_error', str({'error': 'pre_convert-script-failed', 'model': model.name}))
```
IRDonch, line 160:
```suggestion
                telemetry.send_event('md', 'converter_error', str({'error': 'pre-convert-script-failed', 'model': model.name}))
```

tools/downloader/src/open_model_zoo/model_tools/converter.py
```diff
@@ -104,103 +104,131 @@ def main():
...
+                    '(DRY RUN) ' if args.dry_run else '', model.name, model_precision)
 
-            reporter.print()
+                reporter.print('Conversion command: {}', _common.command_string(mo_cmd))
 
-        return True
+                if not args.dry_run:
+                    reporter.print(flush=True)
+
+                    telemetry.send_event('md', 'converter_precision', model_precision)
```
IRDonch, line 200: Not sure this makes sense as a separate event. Maybe it would instead be better to add the precision to the `mo-failed` error event so that we know which invocation of MO failed?

tools/downloader/src/open_model_zoo/model_tools/converter.py
```diff
@@ -104,103 +104,131 @@ def main():
...
+                lambda context, model: convert(_common.Reporter(context), model),
+                models)
+
+        failed_models = [model.name for model, successful in zip(models, results) if not successful]
+
+        if failed_models:
+            reporter.print('FAILED:')
+            for failed_model_name in failed_models:
+                reporter.print(failed_model_name)
+            telemetry.send_event('md', 'converter_result', 'fail')
```
IRDonch, line 226: The exit status reporting is common between tools, so I would suggest to unify it. If you implement my suggestion about `telemetry_session` above, then it'll be pretty easy to do by adding some exception handling to the `try` block:

```python
except SystemExit as e:
    telemetry.send_event('md', f'{tool}_result', 'failure' if e.code else 'success')
    raise
except BaseException:
    telemetry.send_event('md', f'{tool}_result', 'exception')
    raise
else:
    telemetry.send_event('md', f'{tool}_result', 'success')
```

(I also added the reporting of exceptions, since I think it could be useful to log them, while distinguishing them from normal exits.)

`tool` would have to be passed in as a parameter. That said, I don't know how the `app_name` parameter is used by the telemetry library - maybe it would be sufficient to distinguish between tools?

IRDonch (changes requested):

tools/downloader/src/open_model_zoo/model_tools/converter.py
IRDonch: > According to current telemetry realization event_label should be a string

Okay, but this doesn't answer the second half of my question.

I guess this would be a question for @AnastasiaKazantaeva - the README claims that the label could be a dictionary yet the library only accepts a string. How are we supposed to reconcile this? Maybe the library could be modified to accept a dictionary, so that we don't have to manually format the data?

> According to https://github.com/openvinotoolkit/telemetry#readme we need to send all keys if we use dict as a label

Fair enough, but I think we should at least use `None` rather than `'none'` so as not to create ambiguity.

@AnastasiaKazantaeva What's the rationale for always sending all keys? And doesn't this policy mean that no keys can be added to an event in the next release (since then the analytics sent by the previous release will lack that key)?

tools/downloader/converter.py
IRDonch: > how do you suggest to derive frameworks from model names?

I don't know much about Google Analytics, but it seems that it allows you to [import custom data](https://support.google.com/analytics/topic/6064627). We could import the model-name-to-framework mapping for each release and use it in conjunction with the events to build reports involving frameworks. The mapping itself is trivial to obtain:

```sh
./tools/downloader/info_dumper.py --all | jq '.[] | [.name, .framework]'
```

The approach that you currently have implemented doesn't scale. You can get info about which frameworks are requested the most, but what if you wanted to know which frameworks cause the most failures? What if you want to group the models by task type instead? Are you going to add all model information to all events that are specific to a model?

tools/downloader/src/open_model_zoo/model_tools/converter.py
```diff
@@ -104,103 +104,131 @@ def main():
...
-            if not convert_to_onnx(reporter, model, output_dir, args, template_variables):
+    try:
+        telemetry = _common.start_telemetry('Model Converter')
+
+        if args.all:
+            telemetry.send_event('md', 'converter_selection_mode', 'all')
+        if args.list:
+            telemetry.send_event('md', 'converter_selection_mode', 'list')
+        if args.name:
+            telemetry.send_event('md', 'converter_selection_mode', 'name')
```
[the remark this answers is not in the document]
jkamelin: Done

tools/downloader/src/open_model_zoo/model_tools/converter.py
jkamelin: Done

tools/downloader/src/open_model_zoo/model_tools/converter.py
jkamelin: Done

tools/downloader/src/open_model_zoo/model_tools/converter.py
jkamelin: Ok, removed this event. Do we need then `unable-to-locate-mo/pot` error reporting?

IRDonch: I'm a bit torn on those, but I think they could be useful - they indicate how often the users try to run our tools without setting up the OpenVINO environment.

tools/downloader/src/open_model_zoo/model_tools/_common.py
```diff
@@ -675,3 +690,21 @@ def quote_arg_windows(arg):
 
 def command_string(args):
     return ' '.join(map(quote_arg, args))
+
+@contextlib.contextmanager
+def telemetry_session(tool):
+    telemetry = Telemetry('Model Downloader', tid='UA-194864834-1')
```
IRDonch, line 696: Hasn't the first argument to `Telemetry` previously been different for each tool?

IRDonch:
One thing that's unclear to me from this is how are we going to know which version of the downloader sent these events? Does the telemetry library add the version of the OpenVINO toolkit to the events?

tools/downloader/src/open_model_zoo/model_tools/converter.py
jkamelin: moved after argument parsing

tools/downloader/src/open_model_zoo/model_tools/quantizer.py
```diff
@@ -135,62 +136,74 @@ def main():
...
+                pot_path = Path(os.environ['INTEL_OPENVINO_DIR']) / 'deployment_tools/tools/post_training_optimization_toolkit/main.py'
+            except KeyError:
+                telemetry.send_event('md', 'quantizer_error', str({'error': 'unable-to-locate-pot', 'model': None}))
+                sys.exit('Unable to locate Post-Training Optimization Toolkit. '
+                    + 'Use --pot or run setupvars.sh/setupvars.bat from the OpenVINO toolkit.')
+
+        # We can't mark it as required, because it's not required when --print_all is specified.
+        # So we have to check it manually.
+        if not args.dataset_dir:
+            telemetry.send_event('md', 'quantizer_error', str({'error': '--dataset_dir-not-specified.', 'model': None}))
```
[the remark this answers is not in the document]
jkamelin: Agree, removed

tools/downloader/src/open_model_zoo/model_tools/_common.py
jkamelin: Yes, fixed

tools/downloader/src/open_model_zoo/model_tools/converter.py
AnastasiaKazantaeva: @IRDonch you may send None instead of 'none'. Rational here is to simplify automation. Library accept only strings, because GA accept strings. But as we want to know more that only one single parameter, we added dict. This dict helps to link parameters. If we send it in diff events - we cannot align them.

jkamelin:
> One thing that's unclear to me from this is how are we going to know which version of the downloader sent these events? Does the telemetry library add the version of the OpenVINO toolkit to the events?

It doesn't, are we going to use OpenVINO version as downloader version or we will have our own versioning?

tools/downloader/src/open_model_zoo/model_tools/converter.py
IRDonch: On second thought, I don't think the telemetry library can even be loaded unless the OpenVINO environment is set up, so maybe those events are pointless too.

IRDonch:
> It doesn't, tool version need to be passed as Telemetry argument. Are we going to use OpenVINO version as downloader version or we will have our own versioning?

I think using the OpenVINO toolkit version would make the most sense. That said, it seems to me that this would be best handled centrally. @AnastasiaKazantaeva Is there no plan to attach the OpenVINO toolkit version to the events in the telemetry library?

tools/downloader/src/open_model_zoo/model_tools/converter.py
IRDonch: > Rational here is to simplify automation.

How does this simplify it?

> Library accept only strings, because GA accept strings. But as we want to know more that only one single parameter, we added dict.

But if you recommend dicts, wouldn't it make sense for the library itself to support dict inputs, so that all components submit data in a standardized format?

tools/downloader/src/open_model_zoo/model_tools/_common.py
```diff
@@ -110,3 +128,44 @@ def get_package_path(python_executable, package_name):
     # For a package, the file is __init__.py, so to get the package path,
     # take its parent directory.
     return file_path.parent
+
+def get_version():
+    get_ie_version = subprocess.run(
+        [str(sys.executable), '-c',
+            'from openvino.inference_engine import ie_api;'
+                'print(ie_api.get_version())',
+        ],
```
IRDonch, line 137: We shouldn't be trying to get the IE version. OMZ is not IE, and IE might not even be in `PYTHONPATH`; furthermore, loading IE would add extra overhead. We should instead use our own `version.txt` file. Make sure to only try to access the file if we are not installed inside a Python environment (IOW, if `_OMZ_ROOT` is set).

tools/downloader/src/open_model_zoo/model_tools/_common.py
```diff
@@ -110,3 +128,44 @@ def get_package_path(python_executable, package_name):
...
+
+    if get_ie_version.returncode == 0:
+        version = get_ie_version.stdout.rstrip('\n')
+        return version
+    else:
+        try:
+            commit_hash = subprocess.check_output(['git', 'rev-parse', 'HEAD']).strip().decode()
+            return f'custom_{commit_hash}'
+        except Exception:
+            return 'unknown version'
```
IRDonch, line 152: AFAIK, the current plan is to only enable telemetry when packaged inside the OpenVINO toolkit, which means that this section is unneeded.

(That said, in case something like this is re-added in the future: make sure to point `git` to the correct repository. By default, it will use the working directory, which could be anything.)

tools/downloader/src/open_model_zoo/model_tools/_common.py
jkamelin: Done

tools/downloader/src/open_model_zoo/model_tools/_common.py
jkamelin: Done

tools/downloader/src/open_model_zoo/model_tools/_common.py
```diff
@@ -110,3 +129,30 @@ def get_package_path(python_executable, package_name):
     # For a package, the file is __init__.py, so to get the package path,
     # take its parent directory.
     return file_path.parent
+
+def get_version():
+    if VERSION_FILE.is_file():
+        with VERSION_FILE.open('r') as version_file:
+            version = version_file.readline().rstrip('\n')
+        return version
```
IRDonch, line 137:
```suggestion
        return f'commit {version}'
```

This will make it easier to switch to a different (non-commit-based) versioning system in the future (if we want to), because we'll be able to identify the versioning system by the first word.

jkamelin: Done

tools/downloader/src/open_model_zoo/model_tools/_common.py
```diff
@@ -110,3 +129,30 @@ def get_package_path(python_executable, package_name):
...
+    else:
+        return 'unknown version'
```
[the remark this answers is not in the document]
jkamelin: Done

tools/downloader/src/open_model_zoo/model_tools/_common.py
```diff
@@ -22,12 +23,14 @@
 PACKAGE_DIR = Path(__file__).resolve().parent
 MODEL_ROOT = PACKAGE_DIR / 'models'
 DATASET_DEFINITIONS = PACKAGE_DIR / 'data/dataset_definitions.yml'
+VERSION_FILE = Path()
```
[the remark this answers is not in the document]
jkamelin: Done

IRDonch (approved):
Looks good for the most part. My only remaining gripe is about the questionable use of `str` to smuggle `dict`s in string values, which a) will make data analysis in languages that are not Python more difficult, and b) should really be handled by the telemetry library, not in individual tools.

But my guess is that this won't be resolved until my move to a different project, so I'm not going to hold up the review process.

vladimir-dudnik (approved):
LGTM, also Draft status can be removed I think

vladimir-dudnik:
@AnastasiaKazantaeva can we merge this?

eaidova (approved):

AnastasiaKazantaeva:
@vladimir-dudnik @jkamelin to merge it, please try to run it on some validation cycle (at least more than 1 user) to check that data is correct, inform me in what dates you've tested it and I'll check it on GA side

PR #2337, opened 2021-04-21
tools/downloader: put the model tools into a Python package

IRDonch:
A package is needed so that we can make them installable via pip later. I don't think you can install individual modules (and even if you could, the names `downloader`, etc., are way too generic and would pollute the global namespace).

I also used this refactoring to rename `common` to `_common`, to signify that it's an internal module.

`openmodelzoo` is a namespace package, so that if we end up creating other installable OMZ functionality (like a model API or something), we can put it in a different subpackage of `openmodelzoo`. `openmodelzoo.modeltools`, however, is not a namespace package, since there should be no reason to extend it.

Note to reviewers: I recommend examining individual commits; otherwise the diff will be incomprehensible.


IRDonch:
CC @mryzhov @slyubimt

tools/downloader/src/openmodelzoo/modeltools/converter.py
```diff
@@ -0,0 +1,206 @@
+# Copyright (c) 2019 Intel Corporation
```
eaidova, line 1:
```suggestion
# Copyright (c) 2021 Intel Corporation
```

tools/downloader/src/openmodelzoo/modeltools/downloader.py
```diff
@@ -0,0 +1,384 @@
+"""
+ Copyright (c) 2018 Intel Corporation
```
eaidova, line 2:
```suggestion
 Copyright (c) 2021 Intel Corporation
```

eaidova:
@IRDonch I'm correctly understand that current PR only for wrapping model tools to package?
The work on making this module installable will be provided in separated request?
I agree that there is no need to install each tool as separated module.
I have the only one note, for me, your namespace name without spaces looks slightly unreadable. possibly `open_model_zoo.model_tools` or `omz.model_tools` will be better

IRDonch:
> @IRDonch I'm correctly understand that current PR only for wrapping model tools to package?

That is correct.

> I have the only one note, for me, your namespace name without spaces looks slightly unreadable. possibly open_model_zoo.model_tools or omz.model_tools will be better

Fair point. I'll change it.

tools/downloader/src/openmodelzoo/modeltools/converter.py
IRDonch: It would be incorrect to remove 2019, but I updated the years to include the current year.

eaidova (approved):

slyubimt:
@IRDonch python3 setup.py bdist_wheel under ./tools/downloader is not working w/o OMZ_ROOT env. var properly set. Is it expected behavior?

IRDonch:
> @IRDonch python3 setup.py bdist_wheel under ./tools/downloader is not working w/o OMZ_ROOT env. var properly set. Is it expected behavior?

I'll investigate it. It's unrelated to this PR, though, since it doesn't include the setuptools configuration.

PR #2338, opened 2021-04-21
Support text-recognition-resnet-fc for text_detection_demo

akorobeinikov:
Add new `text-recognition-resnet-fc` model for text detection demo.

### DIFF
--- a/demos/text_detection_demo/cpp/README.md
+++ b/demos/text_detection_demo/cpp/README.md
@@ -34,7 +34,7 @@ Options:
     -limit "<num>"             Optional. Number of frames to store in output. If 0 is set, all frames are stored.
     -m_td "<path>"               Required. Path to the Text Detection model (.xml) file.
     -m_tr "<path>"               Required. Path to the Text Recognition model (.xml) file.
-    -dt "<type>"               Required. Type of the decoder, either 'simple' for SimpleDecoder or 'ctc' for CTC greedy and CTC beam search decoders.
+    -dt "<type>"               Optional. Type of the decoder, either 'simple' for SimpleDecoder or 'ctc' for CTC greedy and CTC beam search decoders.
     -m_tr_ss "<value>"           Optional. Symbol set for the Text Recognition model.
     -tr_pt_first                   Optional. Specifies if pad token is the first symbol in the alphabet. Default is false
     -tr_o_blb_nm                   Optional. Name of the output blob of the model which would be used as model output. If not stated, first blob of the model would be used.
@@ -66,12 +66,11 @@ For example, use the following command line command to run the application:
 ./text_detection_demo -m_td <path_to_model>/text-detection-0004.xml \
                       -m_tr <path_to_model>/text-recognition-0013.xml \
                       -i <path_to_image>/sample.jpg \
-                      -dt "ctc" \
                       -tr_pt_first \
                       -tr_o_blb_nm "logits"
 ```
 
-For `text-recognition-resnet-fc` you should use `simple` decoder for `-dt` option. For other models use `ctc` decoder.
+For `text-recognition-resnet-fc` you should use `simple` decoder for `-dt` option. For other models use `ctc` decoder (default decoder).
 
 ## Demo Output
 
--- a/demos/text_detection_demo/cpp/README.md
+++ b/demos/text_detection_demo/cpp/README.md
@@ -7,6 +7,7 @@ The demo shows an example of using neural networks to detect and recognize print
 * `horizontal-text-detection-0001`, which is a detection network that works much faster than models above, but it is applicable to finding more or less horizontal text only.
 * `text-recognition-0012`, which is a recognition network for recognizing text.
 * `text-recognition-0013`, which is a recognition network for recognizing text. You should add option `-tr_pt_first` and specify output layer name via `-tr_o_blb_nm` option for this model (see model [description](../../../models/intel/text-recognition-0013/README.md) for details).
+* `text-recognition-resnet-fc`, which is a recognition network for recognizing text. You should add option `-tr_pt_first`.
 * `handwritten-score-recognition-0001`, which is a recognition network for recognizing handwritten score marks like `<digit>` or `<digit>.<digit>`.
 
 For more information about the pre-trained models, refer to the [model documentation](../../../models/intel/index.md).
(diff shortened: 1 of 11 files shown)


models/public/text-recognition-resnet-fc/model.yml
```diff
@@ -291,7 +291,8 @@ conversion_to_onnx_args:
 model_optimizer_args:
   - --input_shape=[1,1,32,100]
   - --input=input
-  - --scale_values=input[255]
+  - --mean_values=input[0.5]
```
eaidova, line 294:
```suggestion
  - --mean_values=input[127.5]
```

eaidova: @morkovka1337 looks like your configuration in internal repo is incorrect

eaidova: @akorobeinikov could you please also reflect origina input/output and normalization parameters in model README.md

demos/text_detection_demo/cpp/models.lst
```diff
@@ -1,4 +1,4 @@
 # This file can be used with the --list option of the model downloader.
 handwritten-score-recognition-????
 text-detection-????
-text-recognition-????
+text-recognition-*
```
eaidova, line 4: I think it may be useful to stay some hints regarding different models usage in model.lst

models/public/text-recognition-resnet-fc/model.yml
morkovka1337: There is a preprocessing in the original model which could be confusing for the first time. The cause is that `Normalize` first divides input tensor on 255 and then normalizes with mean=0.5 and std=0.5.
https://albumentations.ai/docs/api_reference/augmentations/transforms/#albumentations.augmentations.transforms.Normalize
It should be double checked but it seems to me that simplified equation is:
(x / 255 - 0.5) / 0.5 = x / 127.5 - **1**
Thus, preprocessing should be: 
```
--scale_values=input[127.5]
--mean_values=input[1]
```

But again, this should be double checked.

eaidova: MO and AC use another normalization ops  order
(x - mean) / std

If you checked with previous version of AC config and model reproduces accuracy, then (x / 255 - 0.5) / 0.5 = (x - 127.5) / 127.5

morkovka1337: Ok then, your version should be correct. I did not check it, it was just theoretical calculation. =)

demos/text_detection_demo/cpp/models.lst
akorobeinikov: Changed normalization, change ac metrics and add info about normalization to README.md. Add information about `decoder_type` parameter to models.lst.

eaidova (approved):

demos/text_detection_demo/cpp/text_detection_demo.hpp
```diff
@@ -55,6 +56,7 @@ static const char decoder_bandwidth_message[] = "Optional. Bandwidth for CTC bea
 DEFINE_bool(h, false, help_message);
 DEFINE_string(m_td, "", text_detection_model_message);
 DEFINE_string(m_tr, "", text_recognition_model_message);
+DEFINE_string(dt, "", text_recognition_decoder_type_message);
```
morkovka1337, line 59:
```suggestion
DEFINE_string(dt, "ctc", text_recognition_decoder_type_message);
```

If provide default option for decoder, old text recognition models (12 & 13) could be used as is (without setting additional parameters). 
Also, if not provide default choice, documentation for text-recognition-0012 should also be changed.

demos/text_detection_demo/cpp/README.md
```diff
@@ -64,10 +66,13 @@ For example, use the following command line command to run the application:
 ./text_detection_demo -m_td <path_to_model>/text-detection-0004.xml \
                       -m_tr <path_to_model>/text-recognition-0013.xml \
                       -i <path_to_image>/sample.jpg \
+                      -dt "ctc" \
                       -tr_pt_first \
                       -tr_o_blb_nm "logits"
 ```
 
+For `text-recognition-resnet-fc` you should use `simple` decoder for `-dt` option. For other models `ctc` decoder.
```
morkovka1337, line 74:
```suggestion
For `text-recognition-resnet-fc` you should use `simple` decoder for `-dt` option. For other models use `ctc` decoder.
```

demos/text_detection_demo/cpp/README.md
```diff
@@ -34,7 +34,7 @@ Options:
     -limit "<num>"             Optional. Number of frames to store in output. If 0 is set, all frames are stored.
     -m_td "<path>"               Required. Path to the Text Detection model (.xml) file.
     -m_tr "<path>"               Required. Path to the Text Recognition model (.xml) file.
-    -dt "<path>"               Required. Type of the decoder, either 'simple' for SimpleDecoder or 'ctc' for CTC greedy and CTC beam search decoders.
+    -dt "<type>"               Required. Type of the decoder, either 'simple' for SimpleDecoder or 'ctc' for CTC greedy and CTC beam search decoders.
```
eaidova, line 37: please update help message. You define default value, so it is not required anymore

demos/text_detection_demo/cpp/text_detection_demo.hpp
```diff
@@ -17,6 +17,7 @@ DEFINE_OUTPUT_FLAGS
 static const char help_message[] = "Print a usage message.";
 static const char text_detection_model_message[] = "Required. Path to the Text Detection model (.xml) file.";
 static const char text_recognition_model_message[] = "Required. Path to the Text Recognition model (.xml) file.";
+static const char text_recognition_decoder_type_message[] = "Optional. Type of the decoder, either 'simple' for SimpleDecoder or 'ctc' for CTC greedy and CTC beam search decoders.";
```
eaidova, line 20:
```suggestion
static const char text_recognition_decoder_type_message[] = "Optional. Type of the decoder, either 'simple' for SimpleDecoder or 'ctc' for CTC greedy and CTC beam search decoders. Default is 'ctc'";
```

morkovka1337 (approved):

eaidova:
Jenkins please retry a build

PR #2340, opened 2021-04-23
fix segmentation demo

vladimir-dudnik:
remove forcing FP32 output blob type

### DIFF
--- a/demos/common/cpp/models/src/segmentation_model.cpp
+++ b/demos/common/cpp/models/src/segmentation_model.cpp
@@ -114,9 +114,7 @@ std::unique_ptr<ResultBase> SegmentationModel::postprocess(InferenceResult& infR
     }
     else if (blobPtr->getTensorDesc().getPrecision() == Precision::FP32)
     {
-        cv::Mat predictions(outHeight, outWidth, CV_32FC(outChannels), pData);
-
-        float* ptr = predictions.ptr<float>(0, 0);
+        float* ptr = reinterpret_cast<float*>(pData);
         for (int rowId = 0; rowId < outHeight; ++rowId)
         {
             for (int colId = 0; colId < outWidth; ++colId)
--- a/demos/common/cpp/models/src/segmentation_model.cpp
+++ b/demos/common/cpp/models/src/segmentation_model.cpp
@@ -1,5 +1,5 @@
 /*
-// Copyright (C) 2018-2020 Intel Corporation
+// Copyright (C) 2018-2021 Intel Corporation
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


demos/common/cpp/models/src/segmentation_model.cpp
```diff
@@ -66,58 +65,82 @@ void SegmentationModel::prepareInputsOutputs(InferenceEngine::CNNNetwork& cnnNet
...
+    if (outChannels == 1 && blobPtr->getTensorDesc().getPrecision() == Precision::I32)
+    {
+        cv::Mat predictions(outHeight, outWidth, CV_32SC1, pData);
+        predictions.convertTo(result->mask, CV_8UC1);
+    }
+    else if (blobPtr->getTensorDesc().getPrecision() == Precision::FP32)
+    {
+        cv::Mat predictions(outHeight, outWidth, CV_32FC(outChannels), pData);
+
+        float* ptr = predictions.ptr<float>(0, 0);
```
fzhar, line 119:
```suggestion
        float* ptr = reinterpret_cast<float*>(pData);
```

demos/common/cpp/pipelines/src/async_pipeline.cpp
```diff
@@ -69,11 +69,9 @@ int64_t AsyncPipeline::submitData(const InputData& inputData, const std::shared_
...
     auto internalModelData = model->preprocess(inputData, request);
 
-    request->SetCompletionCallback([this,
-        frameID,
-        request,
-        internalModelData,
-        metaData] {
+    request->SetCompletionCallback(
+        [this, frameID, request, internalModelData, metaData]
+        {
```
fzhar, line 74:
```suggestion
        [this, frameID, request, internalModelData, metaData] {
```
as we agreed before, opening bracket should be at the end of the expression line

demos/common/cpp/pipelines/src/async_pipeline.cpp
```diff
@@ -96,14 +101,16 @@ int64_t AsyncPipeline::submitData(const InputData& inputData, const std::shared_
                     }
                 }
             }
+
             condVar.notify_one();
-    });
+        });
```
fzhar, line 106:
```suggestion
    });
```
according to verbally-defined code style, closing bracket should have same indentation as start of opening expression (not opening bracket, as it should be at the end of line).
BTW, look like whole operators block has extra intentation and hould be shifted to the left by one tab

fzhar (approved):

demos/common/cpp/models/src/segmentation_model.cpp
```diff
@@ -49,14 +51,11 @@ void SegmentationModel::prepareInputsOutputs(InferenceEngine::CNNNetwork& cnnNet
...
-    // if the model performs ArgMax, its output type can be I32 but for models that return heatmaps for each
-    // class the output is usually FP32. Reset the precision to avoid handling different types with switch in
-    // postprocessing
-    data.setPrecision(Precision::FP32);
+
     const SizeVector& outSizeVector = data.getTensorDesc().getDims();
     switch (outSizeVector.size()) {
     case 3:
-        outChannels = 0;
+        outChannels = (int)(outSizeVector[0]);
```
eizamaliev, line 58: That is not correct. If model' output is 3D tensor it has shape `B,H,W`, so the `outSizeVector[0]` stands for batch dimension.  It would be more correctly to set it to 1:
```suggestion
        outChannels = 1;
```

eizamaliev (changes requested):

demos/common/cpp/models/src/segmentation_model.cpp
vladimir-dudnik: BTW, at runtime tensor descriptor show CHW as a layout for 3D tensor (in case of deeplabv3 model). Is it Inference Engine issue?

eizamaliev (approved):

demos/common/cpp/models/src/segmentation_model.cpp
eizamaliev: May be
@eaidova could you help us?

eaidova: It is OV limitation. We cannot rely on OV tensor description. t is always report default depends on number of dims. 
There is not any information stored in the model which really store its real layout (majority of frameworks does not contains this info and IR is also do not store it), so it is impossible to somehow correctly get real layout

vladimir-dudnik: confirmed with benchmark_app, in case of batch_size = 2, blob is still reported as CHW layout, but has dimensions (2, 513, 513)

eaidova (approved):

PR #2344, opened 2021-04-24
fix for wave reader mono flag processing

krodyush:
fix average over channels (it was averaged over samples)
add keepdim to be sure that 1 and 2 channels audio files produces output with the same dim [chanel_num, samples_num]

### DIFF
--- a/tools/accuracy_checker/accuracy_checker/data_readers/data_reader.py
+++ b/tools/accuracy_checker/accuracy_checker/data_readers/data_reader.py
@@ -765,7 +765,7 @@ class WavReader(BaseReader):
 
             data = data.reshape(-1, channels).T
             if channels > 1 and self.mono:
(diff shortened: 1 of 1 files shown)


krodyush:
@eaidova can you look at this?

eaidova (approved):

eaidova:
Jenkins ok to test

tools/accuracy_checker/accuracy_checker/data_readers/data_reader.py
```diff
@@ -765,7 +765,7 @@ def read(self, data_id):
 
             data = data.reshape(-1, channels).T
             if channels > 1 and self.mono:
-                data = data.mean(1)
+                data = data.mean(0,keepdims=True)
```
eaidova, line 768:
```suggestion
                data = data.mean(0, keepdims=True)
```

PR #2352, opened 2021-04-27
Refactoring: init paramteres are added to the Results classes

fzhar:
Initializing results object fields using constructors instead of implicit copying after object instantiation.

### DIFF
--- a/demos/common/cpp/models/include/models/results.h
+++ b/demos/common/cpp/models/include/models/results.h
@@ -22,9 +22,11 @@
 
 struct MetaData;
 struct ResultBase {
+    ResultBase(int64_t frameId = -1, const std::shared_ptr<MetaData>& metaData = nullptr) :
+        frameId(frameId), metaData(metaData) {}
     virtual ~ResultBase() {}
 
-    int64_t frameId = -1;
+    int64_t frameId;
 
     std::shared_ptr<MetaData> metaData;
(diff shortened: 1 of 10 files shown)


demos/common/cpp/models/include/models/results.h
```diff
@@ -22,9 +22,11 @@
 
 struct MetaData;
 struct ResultBase {
+    ResultBase(int64_t frameId = -1, const std::shared_ptr<MetaData>& metaData = nullptr) :
+        frameId(frameId), metaData(metaData) {}
     virtual ~ResultBase() {}
 
-    int64_t frameId = -1;
+    int64_t frameId;
```
eizamaliev, line 29: Since you are refactoring this, what do you think about `frameId` -> `resultId` (or even `id`)?

fzhar: I'd prefer to keep it as is - this ID is associated with data frames order (first of all, video data frames), so I'd better keep emphasis that this is sequential frame ID, not just general ID of result.

vladimir-dudnik (approved):
looks ok to me

Wovchena (approved):

eizamaliev (approved):

vladimir-dudnik:
Jenkins please retry a build

PR #2357, opened 2021-04-28
Move checking for links to README.md files from CI to OMZ repo

nignatovsky:
Checks for links to README.md files of models and demos in models/intel/index.md, models/public/index.md and demos/README.md.

### DIFF
--- a/ci/check-documentation.py
+++ b/ci/check-documentation.py
@@ -69,7 +69,7 @@ def main():
                 md_intermediate_parents = list(md_rel_path.parents)[1:-1] # removed root and first parent dirs
 
                 if not any((index_file_path.parent / parent_dir / 'README.md').exists()
-                    for parent_dir in md_intermediate_parents):
+                        for parent_dir in md_intermediate_parents):
                     required_md_links.append(md_file)
 
         index_child_md_links[index_file_path] = sorted(required_md_links)
--- a/ci/check-documentation.py
+++ b/ci/check-documentation.py
@@ -38,9 +38,9 @@ def find_md_files():
 
 def main():
     all_passed = True
-    
-    md_check_cases = (OMZ_ROOT / 'models' / 'intel' / 'index.md', 
-                     OMZ_ROOT / 'models' / 'public' / 'index.md', 
+
+    md_check_cases = (OMZ_ROOT / 'models' / 'intel' / 'index.md',
+                     OMZ_ROOT / 'models' / 'public' / 'index.md',
                      OMZ_ROOT / 'demos' / 'README.md')
 
     all_md_files = tuple(find_md_files())
@@ -54,7 +54,7 @@ def main():
     for check_case_path in md_check_cases:
         valid_md_links = set()
         for md_file in all_md_files:
-            md_file_parents = [parent_dir for parent_dir in md_file.parents][1:]
+            md_file_parents = tuple(md_file.parents)[1:]
             if check_case_path.parent in md_file_parents:
                 md_index = md_file_parents.index(check_case_path.parent)
                 if not any((parent_dir / 'README.md').exists() for parent_dir in md_file_parents[:md_index]):
--- a/ci/check-documentation.py
+++ b/ci/check-documentation.py
@@ -53,26 +53,26 @@ def main():
         print(message, file=sys.stderr)
 
     index_child_md_links = {}
-    for check_case_path in index_file_paths:
-        if not check_case_path.exists():
-            complain(f'{check_case_path}: file not found')
+    for index_file_path in index_file_paths:
+        if not index_file_path.exists():
+            complain(f'{index_file_path}: file not found')
             continue
 
         required_md_links = []
         for md_file in all_md_files:
-            if md_file.name == "README.md":
+            if md_file.name == "README.md" and md_file.parent != index_file_path.parent:
                 try:
-                    md_rel_path = md_file.relative_to(check_case_path.parent)
+                    md_rel_path = md_file.relative_to(index_file_path.parent)
                 except ValueError:
                     continue
 
                 md_intermediate_parents = list(md_rel_path.parents)[1:-1] # removed root and first parent dirs
 
-                if not any((check_case_path.parent / parent_dir / 'README.md').exists()
-                for parent_dir in md_intermediate_parents) and md_file != check_case_path:
+                if not any((index_file_path.parent / parent_dir / 'README.md').exists()
+                    for parent_dir in md_intermediate_parents):
                     required_md_links.append(md_file)
 
-        index_child_md_links[check_case_path] = sorted(required_md_links)
+        index_child_md_links[index_file_path] = sorted(required_md_links)
 
     for md_path in sorted(all_md_files):
         referenced_md_files = set()
@@ -113,6 +113,8 @@ def main():
 
             if md_path in index_child_md_links:
                 referenced_md_files.add(target_path)
+
+        # check for existence of links to README.md files of models and demos
 
         if md_path in index_child_md_links:
             for md_file in index_child_md_links[md_path]:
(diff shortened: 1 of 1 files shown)


ci/check-documentation.py
```diff
@@ -29,22 +30,41 @@
...
     def complain(message):
         nonlocal all_passed
         all_passed = False
         print(message, file=sys.stderr)
 
     for md_path in sorted(find_md_files()):
+        check_md_links = False
+        readme_files = list()
+
+        if os.path.normpath(md_path) in md_check_cases:
```
IRDonch, line 57: This approach is a bit fragile, because if an index is moved/renamed, then the check will be silently skipped. I suggest that you move the index checks into an entirely separate loop over `md_check_cases`.

ci/check-documentation.py
```diff
@@ -29,22 +30,41 @@
...
+            check_md_links = True
+            path_folder = os.path.dirname(md_path)
```
IRDonch, line 59: It's unnecessary to mix `os.path` with `Path` objects - just use `Path` methods.

ci/check-documentation.py
```diff
@@ -29,22 +30,41 @@
...
+
+            md_paths = tuple(Path(path_folder).glob('*/*/README.md')) + \
+                       tuple(Path(path_folder).glob('*/README.md'))
```
IRDonch, line 62: 1. I don't think you actually need to glob here; we already do the globbing in `find_md_files`, so you can just save the result of that into a variable and then filter that variable. (Incidentally, this will also remove the need to filter out the files in `thirdparty`, because `find_md_files` already does that.)

2. Finding README files at only these specific depth levels may work now, but it might silently miss some files if we alter the layout of the demo/model directories. I would suggest to instead change the logic so that it finds all `README.md` files in the subtree whose parent directories don't contain any other `README.md` files (e.g. `demos/speech_recognition_demo/python/ctcdecode-numpy/README.md` should not be included, because `demos/speech_recognition_demo/python/README.md` exists).

ci/check-documentation.py
```diff
@@ -29,22 +30,41 @@
...
+
+            # transforming paths to url format
```
IRDonch, line 64: I don't think that's a good idea. Comparing URLs is generally difficult (i.e. `README.md` is equivalent to `README%2Emd`), and it's not necessary here. Just keep the paths as paths, and convert the URL to path instead. (In fact, you shouldn't even need to do _that_, as the script already does it for link checking.)

ci/check-documentation.py
```diff
@@ -79,6 +99,17 @@ def complain(message):
                     ' does not exist or is not a file')
                 continue
 
+            if check_md_links and url not in readme_files:
+                readme_files.append(url)
```
IRDonch, line 103: The fact that you don't want `readme_files` to contain duplicates suggests that you need a set rather than a list.

ci/check-documentation.py
```diff
@@ -79,6 +99,17 @@ def complain(message):
...
+
+        if check_md_links:
+            for md_file in md_files:
+                # glob uses lowercase filenames on Windows
```
IRDonch, line 107: Does it? That doesn't seem to be the case:

```
>>> g = Path().glob('**/README.md')
>>> next(g)
WindowsPath('README.md')
```

But, even if it does, `WindowsPath` comparison ignores case on Windows, so as long as you stick with `Path` objects, any comparisons should work anyway.

ci/check-documentation.py
```diff
@@ -79,6 +99,17 @@ def complain(message):
...
+                fixed_md_file = md_file.replace("readme.md", "README.md")
+                if fixed_md_file not in readme_files:
+                    complain(f"{fixed_md_file} not in {os.path.basename(md_path)} file")
```
IRDonch, line 110: The error message should follow the common template: "<file>: <problem>".

ci/check-documentation.py
```diff
@@ -79,6 +99,17 @@ def complain(message):
...
+                    continue
```
IRDonch, line 111: This `continue` seems redundant, as it's the end of the loop body anyway.

IRDonch (changes requested):

nignatovsky:
All suggestions were applied.

ci/check-documentation.py
```diff
@@ -39,12 +39,31 @@ def find_md_files():
...
 
+    md_dict = dict()
+    for check_case_path in md_check_cases:
+        valid_md_links = set()
+        for md_file in all_md_files:
+            md_file_parents = tuple(md_file.parents)[1:]
+            if check_case_path.parent in md_file_parents:
+                md_index = md_file_parents.index(check_case_path.parent)
+                if not any((parent_dir / 'README.md').exists() for parent_dir in md_file_parents[:md_index]):
+                    valid_md_links.add(md_file)
```
IRDonch, line 61: Using a `set` here is unnecessary - all elements in `all_md_files` are different, so there's no way you could add duplicate elements.

ci/check-documentation.py
```diff
@@ -79,6 +98,14 @@ def complain(message):
                     ' does not exist or is not a file')
                 continue
 
+            if md_path in md_dict:
+                readme_files.add(target_path)
+
+        if md_path in md_dict:
+            for link in md_dict[md_path]:
+                if link not in readme_files:
+                    complain(f"{link.relative_to(OMZ_ROOT)}: file is not referenced in {md_path.name}")
```
IRDonch, line 107: The error is in the index file, not in the file that isn't referenced. So it's the index file that should go before the colon.

IRDonch (changes requested):

ci/check-documentation.py
IRDonch: You haven't really fixed this - your check will be silently skipped if an index file disappears.

nignatovsky:
All suggestions were applied.

ci/check-documentation.py
```diff
@@ -39,12 +39,44 @@ def find_md_files():
...
+            if md_file.name == "README.md":
+                try:
+                    md_rel_path = md_file.relative_to(check_case_path.parent)
+                except ValueError:
+                    continue
+
+                md_intermediate_parents = list(md_rel_path.parents)[1:-1] # removed root and first parent dirs
+
+                if not any((check_case_path.parent / parent_dir / 'README.md').exists()
+                for parent_dir in md_intermediate_parents) and md_file != check_case_path:
```
IRDonch, line 72:
```suggestion
                        for parent_dir in md_intermediate_parents) and md_file != check_case_path:
```

I would also suggest to move the second half of the condition to the enclosing `if` and rephrase it as `md_file.parent != check_case_path.parent`. That way:

* we don't just exclude the index file itself, but also any (hypothetical) README file in the same directory;
* someone who reads the code can be certain that `md_rel_path.parents` has at least two components, so they don't have to wonder if `[1:-1]` slice works as intended. :-)

ci/check-documentation.py
```diff
@@ -79,6 +111,14 @@ def complain(message):
                     ' does not exist or is not a file')
                 continue
 
+            if md_path in index_child_md_links:
+                referenced_md_files.add(target_path)
+
```
IRDonch, line 116: Please add a comment here to explain the check that the following section performs. (Like the comments that are above and below.)

ci/check-documentation.py
```diff
@@ -39,12 +39,44 @@ def find_md_files():
...
+            if md_file.name == "README.md" and md_file.parent != index_file_path.parent:
+                try:
+                    md_rel_path = md_file.relative_to(index_file_path.parent)
+                except ValueError:
+                    continue
+
+                md_intermediate_parents = list(md_rel_path.parents)[1:-1] # removed root and first parent dirs
+
+                if not any((index_file_path.parent / parent_dir / 'README.md').exists()
+                    for parent_dir in md_intermediate_parents):
```
IRDonch, line 72: This should be indented by one more level, because otherwise it looks like it's part of the body of the `if` statement.

nignatovsky: Done.

IRDonch (approved):

IRDonch:
Jenkins: test this please

PR #2374, opened 2021-05-04
tools/downloader: add setuptools configuration

IRDonch:
This is needed to enable the integration of the model tools into the `openvino-dev` distribution.

### DIFF
--- a/tools/downloader/setup.py
+++ b/tools/downloader/setup.py
@@ -23,14 +23,73 @@ If you're an end user, you most likely don't need to run this. Just execute
 the scripts in this directory (downloader.py, etc.) directly.
 """
 
+import distutils.command.build
+import distutils.command.sdist
+import itertools
+import os
+import shutil
 from pathlib import Path
 
 from setuptools import setup
 
-SETUP_DIR = Path(__file__).parent
+SETUP_DIR = Path(__file__).resolve().parent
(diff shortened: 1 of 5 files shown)


IRDonch:
@slyubimt I think I fixed the issue you were referring to in #2337; could you recheck?

slyubimt (approved):

tools/downloader/setup.py
```diff
@@ -0,0 +1,105 @@
...
+            for path in model_dir.glob('**/*'):
+                if path.is_dir() and not path.is_symlink():
+                    continue
+
+                if path.suffix == '.md' or (model_dir / 'assets') in path.parents:
+                    continue
+
+                path_rel = path.relative_to(OMZ_ROOT)
+
+                if path_rel.suffix in ('.png', '.jpg'):
```
eaidova, line 84: do we really need it in setup.py? I believe, our precommit checks catch such situations, are not they?
Or you plan replace them to install downloader?

IRDonch: FWIW, there isn't an exact equivalent of this check in pre-commit testing.

However, even if there was, the point of this is to be a backup check in case we change the model layout and forget to update the downloader.

eaidova (approved):

PR #2375, opened 2021-05-04
Free lambda callbacks on requests pool destruction

fzhar:


demos/common/cpp/pipelines/src/requests_pool.cpp
```diff
@@ -23,6 +23,13 @@ RequestsPool::RequestsPool(InferenceEngine::ExecutableNetwork& execNetwork, unsi
     }
 }
 
+RequestsPool::~RequestsPool() {
+    // Setting empty callback to free resources allocated for previously assigned lambdas
+    for (auto& pair : requests) {
+        pair.first->SetCompletionCallback([]{});
```
Wovchena, line 29: `AsyncPipeline` is the one assigning callbacks. It makes it its responsibility to free them.

Wovchena: Actually that approach still delays resource freeing (`internalModelData`, `metaData`). We could free them once callback finishes: https://github.com/openvinotoolkit/open_model_zoo/blob/27c311d2f9034d95ffde7180c9472ae635b8e8df/demos/security_barrier_camera_demo/cpp/main.cpp#L494

eizamaliev (approved):

demos/common/cpp/pipelines/src/requests_pool.cpp
fzhar: agreed. Code resetting callback is moved to lambda.

Wovchena (approved):

vladimir-dudnik (approved):

demos/common/cpp/pipelines/src/async_pipeline.cpp
```diff
@@ -71,6 +71,8 @@ int64_t AsyncPipeline::submitData(const InputData& inputData, const std::shared_
 
     request->SetCompletionCallback(
         [this, frameID, request, internalModelData, metaData] {
+            request->SetCompletionCallback([]{});
```
IRDonch, line 74: Won't this cause the currently executing callback to be immediately freed, thus making all subsequent code access invalid memory?

fzhar: Nope, that will just clear function pointer inside the request.  As long as request is still marked as busy in requests pool, it will not start another data submission so all bound data should not change. As for lambda, it will be valid until the end, then resources will be freed.

IRDonch: > As for lambda, it will be valid until the end, then resources will be freed.

Why? An `InferRequest` only has room for one callback. If you reset it, then it has to free the previous one.

fzhar: I've looked into documentation, and You're right - it may lead to function object destruction (together with context). So, I'll move to the end of function.

fzhar:
I've reverted this commit back to original version with clearing up completion callback in InferRequests pool destructor. 
Clearing up callback from insides it's callable function may lead to immediate destorying of lambda and its context and lead to undefined behavior. Clearing up callback from at the end of function is not possible because together with this->requestsPool->setRequestIdle(request); it may lead to race condition and possible cleaning of callback yet hasn't been processed.

demos/common/cpp/pipelines/src/async_pipeline.cpp
Wovchena: Can you share a link to the documentation you are referring to? Just wondering if this is stated somewhere explicitly. Currently provided explanation that a function won't be able to reach a captured context, could be worked around with `std::bind()`. This would keep needed objects on stack until the function finishes.

Anyway IE copies callback right before execution: https://github.com/openvinotoolkit/openvino/blob/4ea09b1f987a2286f3261d224cedc0a28c51cefb/inference-engine/src/plugin_api/cpp_interfaces/impl/ie_infer_async_request_thread_safe_default.hpp#L391. `SetCompletionCallback()` from our callback won't destroy that copy.

fzhar: Well, there's no explicit statement about deletion. However, captured variables are stored inside the closure object inside the lambda/function object (https://en.cppreference.com/w/cpp/language/lambda). So, as long as function object is deleted (and that will happen in case of calling SetCompletionCallback([]{}) - closure object should be destoryed). If you can explain why its lifetime might be extended till the end of callable function (particularly via std::bind) - please give me a hint and a link to documentation.
At last, it's a good catch that IE copies callback to the local variable before calling it - that clears the problem and allows us to call SetCompletionCallback from inside the callable. So I can change the code back. Again.

IRDonch: > Anyway IE copies callback right before execution

I'm a bit wary of relying on this behavior, which appears to be undocumented... That said, it seems that it was implemented specifically to allow usage like this, so I guess it should be OK.

PR #2378, opened 2021-05-06
tools/downloader: dump the information about which models are quantizable

IRDonch:
This has been long overdue - now the users of the info dumper can determine programmatically which models can be quantized.

Instead of a boolean, we report the quantizer output precisions. This way, we can have different precisions per model in the future. Similarly for the added `if` in the quantizer - it's currently useless, as there's no way to skip every precision for a model, but if some model has an unusual set of output precisions in the future, that will become a possibility.

The name `quantization_output_precisions` is a bit bulky, but I thought that if it was called just `quantization_precisions`, it would be unclear whether that was referring to input or output precisions.

### DIFF
--- a/tools/downloader/README.md
+++ b/tools/downloader/README.md
@@ -435,6 +435,10 @@ describing a single model. Each such object has the following keys:
 
   Additional possible values might be added in the future.
 
+* `quantization_output_precisions`: the list of precisions that the model can be quantized to by
+  the model quantizer. Current possible values are `FP16-INT8` and `FP32-INT8`; additional
+  possible values might be added in the future.
+
 * `subdirectory`: the subdirectory of the output tree into which the downloaded or converted files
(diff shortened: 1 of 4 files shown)


tools/downloader/README.md
```diff
@@ -435,6 +435,9 @@ describing a single model. Each such object has the following keys:
 
   Additional possible values might be added in the future.
 
+* `quantization_output_precisions`: the list of precisions that the model can be quantized to by
+  the model quantizer. Possible values are the same as for `precisions`.
```
jkamelin, line 439: This is not quite true. Right now possible values are: FP32-INT8, FP16-INT8

IRDonch: Yeah, fair point. Fixed.

jkamelin (approved):

PR #2382, opened 2021-05-11
Modify machine_translation_demo

anzhella-pankratova:
Add `-d`, `-o` options, modify `-i` option (now demo can accept a .txt file as input).

### DIFF
--- a/demos/machine_translation_demo/python/machine_translation_demo.py
+++ b/demos/machine_translation_demo/python/machine_translation_demo.py
@@ -202,6 +202,14 @@ def build_argparser():
                         help='Optional. Name of the models output node.')
     return parser
 
+def parse_input(input):
+    if input and input[0].endswith('.txt'):
+        try:
+            with open(input[0], 'r', encoding='utf8') as f:
+                input = f.readlines()
+        except OSError:
+            pass
+    return input
 
 def main(args):
     log.basicConfig(format="[ %(levelname)s ] [ %(name)s ] %(message)s", level=log.INFO, stream=sys.stdout)
@@ -215,18 +223,17 @@ def main(args):
         tokenizer_tgt=args.tokenizer_tgt,
         output_name=args.output_name
     )
-    input_data = args.input
-    if input_data and len(input_data) == 1 and input_data[0].endswith('.txt'):
-        input_data = open(input_data[0], 'r', encoding='utf8')
+    input_data = parse_input(args.input)
     if args.output:
-        with open(args.output, 'w') as f:
-            f.truncate()
+        open(args.output, 'w').close()
 
     def sentences():
         if input_data:
             for sentence in input_data:
-                print("> {}".format(sentence))
-                yield sentence
+                sentence = sentence.strip()
+                if sentence:
+                    print("> {}".format(sentence))
+                    yield sentence
         else:
             while True:
                 yield input("> ")
@@ -244,7 +251,7 @@ def main(args):
             logger.info(f"time: {stop - start} s.")
             if args.output:
                 with open(args.output, 'a', encoding='utf8') as f:
-                    f.write(translation + '\n')
+                    print(translation, file=f)
         except Exception:
             log.error("an error occurred", exc_info=True)
 
--- a/demos/machine_translation_demo/python/machine_translation_demo.py
+++ b/demos/machine_translation_demo/python/machine_translation_demo.py
@@ -33,8 +33,8 @@ class Translator:
         tokenizer_src (str): path to src tokenizer.
         tokenizer_tgt (str): path to tgt tokenizer.
     """
-    def __init__(self, model_xml, model_bin, tokenizer_src, tokenizer_tgt, output_name):
-        self.model = TranslationEngine(model_xml, model_bin, output_name)
+    def __init__(self, model_xml, model_bin, device, tokenizer_src, tokenizer_tgt, output_name):
+        self.model = TranslationEngine(model_xml, model_bin, device, output_name)
         self.max_tokens = self.model.get_max_tokens()
         self.tokenizer_src = Tokenizer(tokenizer_src, self.max_tokens)
         self.tokenizer_tgt = Tokenizer(tokenizer_tgt, self.max_tokens)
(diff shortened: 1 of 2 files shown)


demos/machine_translation_demo/python/machine_translation_demo.py
```diff
@@ -230,6 +239,9 @@ def sentences():
             stop = time.perf_counter()
             print(translation)
             logger.info(f"time: {stop - start} s.")
+            if args.output:
+                with open(args.output, 'a') as f:
```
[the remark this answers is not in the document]
anzhella-pankratova: Done

Wovchena (approved):

demos/machine_translation_demo/python/machine_translation_demo.py
```diff
@@ -230,6 +242,9 @@ def sentences():
             stop = time.perf_counter()
             print(translation)
             logger.info(f"time: {stop - start} s.")
+            if args.output:
+                with open(args.output, 'a', encoding='utf8') as f:
+                    f.write(translation + '\n')
```
IRDonch, line 247:
```suggestion
                    print(translation, file=f)
```

(it's more consistent with the other `print`)

demos/machine_translation_demo/python/machine_translation_demo.py
```diff
@@ -204,18 +210,24 @@ def main(args):
...
-
-    if args.input:
-        def sentences():
-            for sentence in args.input:
+    input_data = args.input
+    if input_data and len(input_data) == 1 and input_data[0].endswith('.txt'):
+        input_data = open(input_data[0], 'r', encoding='utf8')
+    if args.output:
+        with open(args.output, 'w') as f:
+            f.truncate()
```
IRDonch, line 223: Opening a file in the `w` mode already truncates it.

demos/machine_translation_demo/python/machine_translation_demo.py
```diff
@@ -204,18 +210,24 @@ def main(args):
...
         tokenizer_tgt=args.tokenizer_tgt,
         output_name=args.output_name
     )
-
-    if args.input:
-        def sentences():
-            for sentence in args.input:
+    input_data = args.input
+    if input_data and len(input_data) == 1 and input_data[0].endswith('.txt'):
+        input_data = open(input_data[0], 'r', encoding='utf8')
```
IRDonch, line 220: This leaks the file handle. Please rewrite the code so that it doesn't happen.

(One simple way would be to just read the whole file at once.)

demos/machine_translation_demo/python/machine_translation_demo.py
```diff
@@ -204,18 +210,24 @@ def main(args):
...
         tokenizer_src=args.tokenizer_src,
         tokenizer_tgt=args.tokenizer_tgt,
         output_name=args.output_name
     )
-
-    if args.input:
-        def sentences():
-            for sentence in args.input:
+    input_data = args.input
+    if input_data and len(input_data) == 1 and input_data[0].endswith('.txt'):
```
IRDonch, line 219: The `len(input_data) == 1` condition seems a bit arbitrary. IMHO, any elements of `input_data` that end with `.txt` should be treated as file names.

demos/machine_translation_demo/python/machine_translation_demo.py
```diff
@@ -204,18 +210,24 @@ def main(args):
...
+        input_data = open(input_data[0], 'r', encoding='utf8')
+    if args.output:
+        with open(args.output, 'w') as f:
+            f.truncate()
+
+    def sentences():
+        if input_data:
+            for sentence in input_data:
```
IRDonch, line 227: If `input_data` is a file, then `sentence` will contain a trailing newline. You should probably strip it.

IRDonch (changes requested):
Oops, forgot to set the status.

anzhella-pankratova:
@IRDonch, suggestions were applied.

demos/machine_translation_demo/python/machine_translation_demo.py
Wovchena: Out of curiosity, won't python close the file when the object is destroyed?

IRDonch: > Out of curiosity, won't python close the file when the object is destroyed?

Yes, but there is no guarantee for when the object is destroyed.

Wovchena: On exit is also fine :)

demos/machine_translation_demo/python/machine_translation_demo.py
IRDonch: You didn't fully fix this - you removed the condition, but you haven't added the support for treating the other elements of `input_data` (besides `input_data[0]`) as text file paths.

demos/machine_translation_demo/python/machine_translation_demo.py
IRDonch: In this particular case it might be fine, but I think it's a good practice to ensure that any resources you allocate are freed when they are no longer needed. An idealistic reason is that the code might be reused in a different context where leaking the resource is not harmless (e.g. a long-running process). A more pragmatic reason is that Python will give you a `ResourceWarning` if you don't do it (visible in [development mode](https://docs.python.org/3/library/devmode.html)).

Wovchena: I see

demos/machine_translation_demo/python/machine_translation_demo.py
anzhella-pankratova: It's not clear for me, which other elements needed to be supported. There are only two cases - when only path to .txt file is submitted and with the location in `input_data[0]`, or when input consists of several sentences. Iteration is the same for all options of '-input'.

IRDonch: There could be multiple txt files, or txt files intermingled with sentences.

anzhella-pankratova: Updated.

anzhella-pankratova:
Jenkins please retry a build

IRDonch (approved):

PR #2385, opened 2021-05-11
Additional labels for inpainiting demo

fzhar:


Wovchena:
Pressing Tab after demo start doesn't show that it takes any effect, but drawing becomes silently disabled. In earlier version these states at least were visually distinguishable. @vladimir-dudnik may be adding “Original” label here wasn’t a good idea :)

I noticed another thing. Do inpainting->press Tab (to switch back to Original). Pressing C after that behaves like Tab, which is strange: C clears current mask, but it is never said that it also switches between Original and Result.

In addition to @vladimir-dudnik note about failed build, there is also merge conflict. Resolve it.

vladimir-dudnik:
21:33:57  ./demos/image_inpainting_demo/python/image_inpainting_demo.py:110:88: E231 missing whitespace after ','

demos/image_inpainting_demo/python/image_inpainting_demo.py
```diff
@@ -106,7 +106,9 @@ def inpaint_auto(img, args):
     #--- Inpaint and show results
     output_image = inpainting_processor.process(masked_image, mask)
     concat_imgs = np.hstack((masked_image, output_image))
-    cv2.putText(concat_imgs, 'summary: {:.1f} FPS'.format(float(1 / inpainting_processor.infer_time)), (5, 15), cv2.FONT_HERSHEY_COMPLEX, 0.5, (0, 0, 200))
+    cv2.putText(concat_imgs, 'original', (5, 15), cv2.FONT_HERSHEY_COMPLEX, 0.5, (0, 0, 100))
+    cv2.putText(concat_imgs, 'result', (concat_imgs.shape[1]-5-cv2.getTextSize('result',cv2.FONT_HERSHEY_COMPLEX, 0.5, 1)[0][0], 15), cv2.FONT_HERSHEY_COMPLEX, 0.5, (0, 0, 100))
```
eizamaliev, line 110: Could you simplify this?

fzhar: we need to align this label to the right, so the only way I see is to calculate text width. I can put it separate variable, but I'm afraid that @Wovchena would ban this change - extra variables are not welcome.

fzhar, line 110:
```suggestion
    cv2.putText(concat_imgs, 'result', (concat_imgs.shape[1]-5-cv2.getTextSize('result', cv2.FONT_HERSHEY_COMPLEX, 0.5, 1)[0][0], 15), cv2.FONT_HERSHEY_COMPLEX, 0.5, (0, 0, 100))
```

fzhar:
> Pressing Tab after demo start doesn't show that it takes any effect, but drawing becomes silently disabled. In earlier version these states at least were visually distinguishable. @vladimir-dudnik may be adding “Original” label here wasn’t a good idea :)
> 
> I noticed another thing. Do inpainting->press Tab (to switch back to Original). Pressing C after that behaves like Tab, which is strange: C clears current mask, but it is never said that it also switches between Original and Result.
> 
> In addition to @vladimir-dudnik note about failed build, there is also merge conflict. Resolve it.

Origin issue is known. But I don't know how to make this behavior better. I may suggest to show other "Original smth" label, but I can't figure out to a good text.
As for C that returns to editing mode - in the time of first release of this demo we've agreed that any editing action (pressing C or something like that" will cancel "Origin" mode. If this needs to be changed - let's discuss better behavior.

fzhar:
code is rebase, precommit issues are fixed

Wovchena:
I was writing mostly to @vladimir-dudnik, asking to reconsider his position on the first topic :)

> As for C that returns to editing mode - in the time of first release of this demo we've agreed that any editing action (pressing C or something like that" will cancel "Origin" mode. If this needs to be changed - let's discuss better behavior.

Could you put that into readme under Interactive Mode section. I think it will fit nicely.

eizamaliev (approved):

fzhar:
> I was writing mostly to @vladimir-dudnik, asking to reconsider his position on the first topic :)
> 
> > As for C that returns to editing mode - in the time of first release of this demo we've agreed that any editing action (pressing C or something like that" will cancel "Origin" mode. If this needs to be changed - let's discuss better behavior.
> 
> Could you put that into readme under Interactive Mode section. I think it will fit nicely.

Done

demos/image_inpainting_demo/python/README.md
```diff
@@ -91,9 +91,11 @@ Also, these hot keys are available:
 * **Backspace or C** to clear current mask
 * **Space or Enter** to inpaint
 * **R** to reset all changes
-* **Tab** to show original image
+* **Tab** to show/hide original image
 * **Esc or Q** to quit
 
+If Backpace,C or R keys are presed while demo is showing original image, demo will hide original image and return back to editing mode.
```
Wovchena, line 97:
```suggestion
If Backpace, C or R keys are presed while demo is showing original image, demo will hide original image and return back to editing mode.
```

fzhar, line 97:
```suggestion
If Backpace, C or R keys are pressed while demo is showing original image, demo will hide original image and return back to editing mode.
```

Wovchena (approved):

PR #2389, opened 2021-05-13
tools/downloader: split the _common module into several modules

IRDonch:
This module has previously been the dumping ground for all functionality that the tools share, but it's grown too big. There are several fairly self-contained sections in there, so split them off into separate modules.


jkamelin (approved):

IRDonch:
Jenkins please retry a build

PR #2395, opened 2021-05-14
Model Converter: try to run Model Optimizer as a package

IRDonch:
If our tools are installed as part of the `openvino-dev` distribution, then it doesn't make sense to ask the user to run `setupvars`, seeing as:

1. It doesn't exist.
2. We know that MO is installed as part of the same distribution.

Therefore, before trying to find MO the old way, try to locate its package and if that succeeds, run it as a package.

I moved the documentation for the `-p` option up, because I'm referencing its value in the documentation for the `--mo` option.

`get_package_path` is in `_common`, because we're likely going to implement the same thing in the quantizer (for POT).

### DIFF
--- a/tools/downloader/README.md
+++ b/tools/downloader/README.md
@@ -260,13 +260,31 @@ option:
 
 If the specified precision is not supported for a model, that model will be skipped.
 
-The script will attempt to locate Model Optimizer using the environment
-variables set by the OpenVINO&trade; toolkit's `setupvars.sh`/`setupvars.bat`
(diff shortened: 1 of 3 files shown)


IRDonch:
CC @SDxKeeper @slyubimt

SDxKeeper (approved):
Looks great! Simplifies life of Open Model Zoo users through usage of openvino-dev! 
Hopefully in the future we would be able to use something smaller like openvino runtime and openvino-mo, openvino-py packages in a native python way :)

jkamelin (approved):

PR #2416, opened 2021-05-20
Drop old text recognition models

morkovka1337:
Drop support for old text recognition models since new models (14 & 15) have higher accuracy and comparable performance


vladimir-dudnik:
@morkovka1337 why do you close this?

PR #2421, opened 2021-05-21
Added checks for <omz_dir> links validity in documentation files.

nignatovsky:
Also added inline code formatting for <omz_dir> links in demos README.md files.

### DIFF
--- a/ci/check-documentation.py
+++ b/ci/check-documentation.py
@@ -114,8 +114,10 @@ def main():
             if md_path in index_child_md_links:
                 referenced_md_files.add(target_path)
 
-        for link in sorted([ref.link for ref in doc_page.omz_references() if '<omz_dir>/' in ref.link]):
-            file_path = OMZ_ROOT / link.replace('<omz_dir>/', '')
+        # check <omz_dir> link validity
+
+        for link in sorted([link for link in doc_page.omz_references() if link.startswith('<omz_dir>')]):
+            file_path = Path(link.replace('<omz_dir>', str(OMZ_ROOT)))
 
             try:
                 file_relative_path = file_path.relative_to(OMZ_ROOT)
--- a/ci/check-documentation.py
+++ b/ci/check-documentation.py
@@ -128,7 +128,7 @@ def main():
             if str(file_relative_path) == md_path_rel: # self-link
                 continue
 
-            if not (file_relative_path.is_file() or file_relative_path.is_dir()):
+            if not (file_path.is_file() or file_path.is_dir()):
                 complain(f'{md_path_rel}: OMZ reference "{file_relative_path}" target'
                     ' does not exist')
 
--- a/ci/check-documentation.py
+++ b/ci/check-documentation.py
@@ -120,15 +120,15 @@ def main():
 
         for code_span in doc_page.code_spans():
             if code_span.startswith(omz_reference_prefix):
-                file_relative_path = code_span[len(omz_reference_prefix):]
-                file_path = OMZ_ROOT / file_relative_path
+                target_path_rel = code_span[len(omz_reference_prefix):]
+                target_path = OMZ_ROOT / target_path_rel
 
-                if ".." in Path(file_relative_path).parts:
+                if ".." in Path(target_path_rel).parts:
                     complain(f'{md_path_rel}: OMZ reference "{code_span}" may'
                         ' point outside the OMZ directory')
                     continue
 
-                if not file_path.exists():
+                if not target_path.exists():
                     complain(f'{md_path_rel}: OMZ reference "{code_span}" target'
                         ' does not exist')
 
--- a/ci/check-documentation.py
+++ b/ci/check-documentation.py
@@ -114,25 +114,22 @@ def main():
             if md_path in index_child_md_links:
                 referenced_md_files.add(target_path)
 
-        # check <omz_dir> link validity
+        # check <omz_dir> reference validity
 
-        for link in sorted([link for link in doc_page.omz_references() if link.startswith('<omz_dir>')]):
-            file_path = Path(link.replace('<omz_dir>', str(OMZ_ROOT)))
+        for link in doc_page.omz_references():
+            file_relative_path = link.removeprefix('<omz_dir>/')
+            file_path = OMZ_ROOT / file_relative_path
 
-            try:
-                file_relative_path = file_path.relative_to(OMZ_ROOT)
-            except ValueError:
-                complain(f'{md_path_rel}: invalid OMZ reference {file_path!r}')
-                continue
+            if ".." in file_relative_path:
+                complain(f'{md_path_rel}: OMZ reference "{link}" may'
+                    ' point outside the OMZ directory')
 
-            if str(file_relative_path) == md_path_rel: # self-link
-                continue
-
-            if not (file_path.is_file() or file_path.is_dir()):
-                complain(f'{md_path_rel}: OMZ reference "{file_relative_path}" target'
+            if not (file_path.exists()):
+                complain(f'{md_path_rel}: OMZ reference "{link}" target'
                     ' does not exist')
 
         # check for existence of links to README.md files of models and demos
+
         if md_path in index_child_md_links:
             for md_file in index_child_md_links[md_path]:
                 if md_file not in referenced_md_files:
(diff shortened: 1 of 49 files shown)


ci/check-documentation.py
```diff
@@ -114,8 +114,25 @@ def complain(message):
             if md_path in index_child_md_links:
                 referenced_md_files.add(target_path)
 
-        # check for existence of links to README.md files of models and demos
+        # check <omz_dir> link validity
+
+        for link in sorted([link for link in doc_page.omz_references() if link.startswith('<omz_dir>')]):
```
IRDonch, line 119: The expected prefix should be `<omz_dir>/`, rather than `<omz_dir>`. If `<omz_dir>` is used by itself, then it's pointless to check it (it's obviously correct); and if it's followed by something other than a slash, then it's not a path and we shouldn't try to check it as a path.

ci/check-documentation.py
```diff
@@ -114,8 +114,25 @@ def complain(message):
...
+            file_path = Path(link.replace('<omz_dir>', str(OMZ_ROOT)))
```
IRDonch, line 120: This will replace all occurrences of `<omz_dir>` in the string, but only the initial occurrence should be replaced.

ci/check-documentation.py
```diff
@@ -114,8 +114,25 @@ def complain(message):
...
+
+            try:
+                file_relative_path = file_path.relative_to(OMZ_ROOT)
+            except ValueError:
```
IRDonch, line 124: How can this exception ever occur? `file_path` always begins with `OMZ_ROOT` by construction, so `relative_to` should never fail.

Moreover, it seems counterproductive to add `OMZ_ROOT` to the path and then to try to remove it again. It would be more straightforward to initialize `file_relative_path` first (by stripping the `<omz_dir>/` prefix from the reference) and _then_ build `file_path`.

ci/check-documentation.py
IRDonch: Also, I don't think the links should be sorted; it would be more useful for the developers to see the list of broken links in order of appearance.

ci/check-documentation.py
```diff
@@ -114,8 +114,25 @@ def complain(message):
...
+                complain(f'{md_path_rel}: invalid OMZ reference {file_path!r}')
+                continue
+
+            if str(file_relative_path) == md_path_rel: # self-link
+                continue
```
IRDonch, line 129: What's the purpose of this check? It's true that a self-link should always be valid and thus doesn't _need_ to be checked, but the cost for checking it anyway is negligible, so there's no need to special-case it. (Also, I'm not sure why we would even have any `<omz_dir>` references that point to the document they're in.)

ci/check-documentation.py
```diff
@@ -114,8 +114,25 @@ def complain(message):
...
 
+            if not (file_path.is_file() or file_path.is_dir()):
```
IRDonch, line 131:
```suggestion
            if not file_path.exists():
```

ci/check-documentation.py
```diff
@@ -114,8 +114,25 @@ def complain(message):
...
+
+        for link in sorted([link for link in doc_page.omz_references() if link.startswith('<omz_dir>')]):
+            file_path = Path(link.replace('<omz_dir>', str(OMZ_ROOT)))
+
+            try:
+                file_relative_path = file_path.relative_to(OMZ_ROOT)
+            except ValueError:
+                complain(f'{md_path_rel}: invalid OMZ reference {file_path!r}')
+                continue
+
```
IRDonch, line 127: It would be useful to check whether the reference points outside the OMZ directory, like we do for URLs. (_Unlike_ URLs, the `<omz_dir>` references are always rooted at the OMZ root, so it would be sufficient to check whether the path contains any `..` components.)

ci/check-documentation.py
```diff
@@ -114,8 +114,25 @@ def complain(message):
             if md_path in index_child_md_links:
                 referenced_md_files.add(target_path)
 
-        # check for existence of links to README.md files of models and demos
+        # check <omz_dir> link validity
```
IRDonch, line 117:
```suggestion
        # check <omz_dir> reference validity
```

Your function is called `omz_references`, so you should call them references elsewhere, as well.

ci/check-documentation.py
```diff
@@ -114,8 +114,25 @@ def complain(message):
...
+                file_relative_path = file_path.relative_to(OMZ_ROOT)
+            except ValueError:
+                complain(f'{md_path_rel}: invalid OMZ reference {file_path!r}')
+                continue
+
+            if str(file_relative_path) == md_path_rel: # self-link
+                continue
 
+            if not (file_path.is_file() or file_path.is_dir()):
+                complain(f'{md_path_rel}: OMZ reference "{file_relative_path}" target'
```
IRDonch, line 132:
```suggestion
                complain(f'{md_path_rel}: OMZ reference "{link}" target'
```

`file_relative_path` is not the reference - `link` is.

IRDonch (changes requested):

nignatovsky:
All suggestions were applied.

ci/check-documentation.py
```diff
@@ -114,6 +116,21 @@ def complain(message):
             if md_path in index_child_md_links:
                 referenced_md_files.add(target_path)
 
+        # check <omz_dir> reference validity
+
+        for link in doc_page.code_spans():
```
IRDonch, line 121:
```suggestion
        for code_span in doc_page.code_spans():
```

This variable can contain values that are not links.

ci/check-documentation.py
IRDonch: You didn't remove the parentheses.

ci/check-documentation.py
```diff
@@ -114,6 +116,21 @@ def complain(message):
...
+            if link.startswith(omz_prefix):
+                file_relative_path = link[len(omz_prefix):]
+                file_path = OMZ_ROOT / file_relative_path
+
+                if ".." in file_relative_path:
```
IRDonch, line 126: This doesn't check for `..` components - it checks whether the string contains the two characters `..` anywhere. You should turn `file_relative_path` into a `Path` and use `file_relative_path.parts` here.

ci/check-documentation.py
```diff
@@ -74,6 +74,8 @@ def complain(message):
 
         index_child_md_links[index_file_path] = sorted(required_md_links)
 
+    omz_prefix = '<omz_dir>/'
```
IRDonch, line 77:
```suggestion
    omz_reference_prefix = '<omz_dir>/'
```

ci/check-documentation.py
```diff
@@ -114,6 +116,21 @@ def complain(message):
             if md_path in index_child_md_links:
                 referenced_md_files.add(target_path)
 
+        # check <omz_dir> reference validity
+
+        for link in doc_page.code_spans():
+            if link.startswith(omz_prefix):
+                file_relative_path = link[len(omz_prefix):]
+                file_path = OMZ_ROOT / file_relative_path
```
IRDonch, line 124: I think it would make sense to rename these to `target_path_rel` and `target_path`, which would be consistent with the preceding loop, as well as more accurate (since these path do not necessarily point to a file).

nignatovsky:
Done, all suggestions were applied.

ci/check-documentation.py
```diff
@@ -114,6 +116,22 @@ def complain(message):
...
                 referenced_md_files.add(target_path)
 
+        # check <omz_dir> reference validity
+
+        for code_span in doc_page.code_spans():
+            if code_span.startswith(omz_reference_prefix):
+                target_path_rel = code_span[len(omz_reference_prefix):]
+                target_path = OMZ_ROOT / target_path_rel
+
+                if ".." in Path(target_path_rel).parts:
```
IRDonch, line 126: You shouldn't convert `target_path_rel` to `Path` _here_. This variable represents a path, so it should contain a `Path` object already. You should move the `Path` call to the variable initialization.

ci/check-documentation.py
```diff
@@ -114,6 +116,22 @@ def complain(message):
...
+                    complain(f'{md_path_rel}: OMZ reference "{code_span}" may'
+                        ' point outside the OMZ directory')
```
IRDonch, line 128: I think a more direct message would be more helpful here, e.g. `OMZ reference "{code_span}" contains a ".." component`.

IRDonch (approved):

IRDonch:
Jenkins: test this please

PR #2427, opened 2021-05-24
AC: improve mnist dataset support

eaidova:
* handle non-exsist directory for conver images
* handle csv files without header
* support original binary data format


tools/accuracy_checker/accuracy_checker/annotation_converters/mnist.py
```diff
@@ -112,12 +215,12 @@ def convert(self, check_content=False, progress_callback=None, progress_interval
...
-        for x in range(28):
-            for y in range(28):
-                pixel = int(features[column_template.format(x+1, y+1)])
-                image[x, y] = pixel
+        image = np.array(features[1:], dtype=np.uint8).reshape((28, 28))
+        # column_template = '{}x{}'
+        # for x in range(28):
+        #     for y in range(28):
+        #         pixel = int(features[column_template.format(x+1, y+1)])
+        #         image[x, y] = pixel
```
eaidova, line 223:
```suggestion
```

Anna-Mironova (approved):

tools/accuracy_checker/accuracy_checker/utils.py
```diff
@@ -304,9 +304,11 @@ def read_yaml(file: Union[str, Path], *args, **kwargs):
         return yaml.safe_load(content, *args, **kwargs)
 
 
-def read_csv(file: Union[str, Path], *args, **kwargs):
+def read_csv(file: Union[str, Path], is_dict=True, *args, **kwargs):
```
eaidova, line 307:
```suggestion
def read_csv(file: Union[str, Path], *args, is_dict=True, **kwargs):
```

PR #2430, opened 2021-05-24
Model Quantizer: try to run POT as a package

IRDonch:
Basically, the same thing as in 1df8e872, but for POT instead of MO. The code is simpler here, because we don't need to know the directory POT is installed in.


IRDonch:
CC @slyubimt

jkamelin (approved):

slyubimt (approved):

PR #2439, opened 2021-05-26
tools/downloader: access composite models by composite model name

jkamelin:


jkamelin:
cc @IRDonch

tools/downloader/src/open_model_zoo/model_tools/_configuration.py
```diff
@@ -402,8 +424,12 @@ def load_models_or_die(args):
 # requires the --print_all, --all, --name and --list arguments to be in `args`
 def load_models_from_args(parser, args):
     if args.print_all:
+        model_names = set()
         for model in load_models_or_die(args):
-            print(model.name)
+            name_to_print = model.name if model.composite_model_name is None else model.composite_model_name
```
IRDonch, line 429: Composite model names should not be printed here. This breaks the invariant that for every name printed, there is a corresponding entry in the info dumper output.

(I think potentially it could be a good idea for info dumper to have an option to print information about composite models, but this should be done in a backwards-compatible manner.)

tools/downloader/src/open_model_zoo/model_tools/_configuration.py
```diff
@@ -359,9 +360,23 @@ def deserialize(cls, model, name, subdirectory):
             task_type = validate_string_enum('"task_type"', model['task_type'],
                 _common.KNOWN_TASK_TYPES)
 
+            composite_model_name = model.get('composite_model_name', None)
+            if composite_model_name is not None:
+                composite_model_name = validate_string('"composite_model_name"', composite_model_name)
```
IRDonch, line 365: Any error message produced here would be incorrect - `model.yml` files do not _have_ `composite_model_name` keys.

And more broadly, it makes no sense to first add `composite_model_name` to `model` and then extract it back out. It's redundant, and it will produce strange results if someone does add a `composite_model_name` key to a `model.yml` file. Instead, you should just pass `composite_model_name` into this function as a parameter. Moreover, any validation on composite model names should happen when you first find the `composite-model.yml` file, not here.

tools/downloader/src/open_model_zoo/model_tools/_configuration.py
```diff
@@ -359,9 +360,23 @@ def deserialize(cls, model, name, subdirectory):
...
+
             return cls(name, subdirectory, files, postprocessing, mo_args, framework,
                 description, license_url, precisions, quantization_output_precisions,
-                task_type, conversion_to_onnx_args)
+                task_type, conversion_to_onnx_args, composite_model_name)
+
+def check_composite_model_dir(model_dir):
+    if list(model_dir.glob('*/*/*')):
+        raise DeserializationError('{} directory should not contain any other folders'.format(model_dir))
```
IRDonch, line 373: This check seems too broad. The only thing that really matters is that there should not be any `model.yml` files in any subdirectories that are not direct children of the composite model directory (as well as in the composite model directory itself).

tools/downloader/src/open_model_zoo/model_tools/_configuration.py
```diff
@@ -359,9 +360,23 @@ def deserialize(cls, model, name, subdirectory):
...
+
+    model_name = model_dir.name
+    model_stages = list(model_dir.glob('*/model.yml'))
+    for model in model_stages:
+        if not model.parent.name.startswith(model_name):
```
IRDonch, line 378:
```suggestion
        if not model.parent.name.startswith(f'{model_name}-'):
```

tools/downloader/src/open_model_zoo/model_tools/_configuration.py
```diff
@@ -359,9 +360,23 @@ def deserialize(cls, model, name, subdirectory):
...
+            raise DeserializationError('Unknown model name {}'.format(model.parent.name))
```
IRDonch, line 379: This error message is strange. How is the model name "unknown"? You do _know_ the name of the model - it's `model.parent.name`.

tools/downloader/src/open_model_zoo/model_tools/_configuration.py
```diff
@@ -370,10 +385,17 @@ def load_models(args):
     for config_path in sorted(_common.MODEL_ROOT.glob('**/model.yml')):
         subdirectory = config_path.parent.relative_to(_common.MODEL_ROOT)
 
+        composite_model_config = config_path.parents[1] / 'composite-model.yml'
+        composite_model_name = None
+        if composite_model_config.is_file():
+            check_composite_model_dir(composite_model_config.parent)
```
IRDonch, line 391: Seems like this will recheck the composite model directory for every component of the composite model, which is not ideal. I would suggest that you first find all `composite-model.yml` files, validate them, remember where they were, and then, for each model, determine whether it belongs to one of the composite models.

tools/downloader/src/open_model_zoo/model_tools/_configuration.py
```diff
@@ -435,8 +461,12 @@ def load_models_from_args(parser, args):
         models = collections.OrderedDict() # deduplicate models while preserving order
 
         for pattern in patterns:
-            matching_models = [model for model in all_models
-                if fnmatch.fnmatchcase(model.name, pattern)]
+            matching_models = []
+            for model in all_models:
+                if fnmatch.fnmatchcase(model.name, pattern):
+                    matching_models.append(model)
+                elif model.composite_model_name and fnmatch.fnmatchcase(model.composite_model_name, pattern):
```
IRDonch, line 468: The documentation for the `--name` and `--list` options should be updated to note that both model names and composite model names are accepted.

tools/downloader/README.md
```diff
@@ -424,6 +424,9 @@ describing a single model. Each such object has the following keys:
 
 * `name`: the identifier of the model, as accepted by the `--name` option.
 
+* `composite_model_name`: the identifier of the composite model name, if the model is a composition
```
IRDonch, line 427: This explanation is mixing the concepts. The _composite model_ is the composition of several models; the _model_ is a component of that composition.

jkamelin: Fixed

tools/downloader/src/open_model_zoo/model_tools/_configuration.py
jkamelin: Done

tools/downloader/src/open_model_zoo/model_tools/_configuration.py
jkamelin: Done

tools/downloader/src/open_model_zoo/model_tools/_configuration.py
jkamelin: Fixed

tools/downloader/src/open_model_zoo/model_tools/_configuration.py
jkamelin: Done

tools/downloader/src/open_model_zoo/model_tools/_configuration.py
jkamelin: Done

tools/downloader/src/open_model_zoo/model_tools/_configuration.py
```diff
@@ -361,15 +362,42 @@ def deserialize(cls, model, name, subdirectory):
 
             return cls(name, subdirectory, files, postprocessing, mo_args, framework,
                 description, license_url, precisions, quantization_output_precisions,
-                task_type, conversion_to_onnx_args)
+                task_type, conversion_to_onnx_args, composite_model_name)
+
+def check_composite_model_dir(model_dir):
+    if list(model_dir.glob('*/*/model.yml')):
```
IRDonch, line 368:
```suggestion
    if list(model_dir.glob('*/*/**/model.yml')):
```

tools/downloader/src/open_model_zoo/model_tools/_configuration.py
```diff
@@ -361,15 +362,42 @@ def deserialize(cls, model, name, subdirectory):
...
+        raise DeserializationError(
+            '{} directory should not contain any model.yml files in any subdirectories '
+            'that are not direct children of the composite model directory'.format(model_dir))
+
+    if list(model_dir.glob('model.yml')):
+        raise DeserializationError('{} directory should not contain any model.yml files'.format(model_dir))
```
IRDonch, line 374:
```suggestion
    if (model_dir / 'model.yml').exists():
        raise DeserializationError('{} directory should not contain a model.yml file'.format(model_dir))
```

(There can only be one `model.yml` file in a directory.)

tools/downloader/src/open_model_zoo/model_tools/_configuration.py
```diff
@@ -361,15 +362,42 @@ def deserialize(cls, model, name, subdirectory):
...
     models = []
     model_names = set()
 
+    composite_models = []
+
+    for composite_model_config in sorted(_common.MODEL_ROOT.glob('**/composite-model.yml')):
+        check_composite_model_dir(composite_model_config.parent)
+        composite_models.append(composite_model_config.parent.name)
+
+        if composite_models[-1] in composite_models[:-1]:
```
IRDonch, line 392: This way you make a copy of (most of) `composite_models` on every iteration. You should check for duplicates _before_ you add the new name.

tools/downloader/src/open_model_zoo/model_tools/_configuration.py
```diff
@@ -361,15 +362,42 @@ def deserialize(cls, model, name, subdirectory):
...
+            raise DeserializationError(
+                'Duplicate composite model name "{}"'.format(composite_models[-1]))
+
     for config_path in sorted(_common.MODEL_ROOT.glob('**/model.yml')):
         subdirectory = config_path.parent.relative_to(_common.MODEL_ROOT)
 
+        composite_model_name = subdirectory.parent.name if subdirectory.parent.name in composite_models else None
```
IRDonch, line 399: This check is not quite correct. For example, consider these files:

```
a/composite-model.yml
b/a/b/model.yml
```

Here, the grandparent of `model.yml` is `a`, but the model `b` does not belong to the `a` composite model.

tools/downloader/src/open_model_zoo/model_tools/_configuration.py
```diff
@@ -361,15 +362,42 @@ def deserialize(cls, model, name, subdirectory):
...
 
 def load_models(args):
     models = []
     model_names = set()
 
+    composite_models = []
+
+    for composite_model_config in sorted(_common.MODEL_ROOT.glob('**/composite-model.yml')):
+        check_composite_model_dir(composite_model_config.parent)
+        composite_models.append(composite_model_config.parent.name)
```
IRDonch, line 390: The composite model name should be validated with the same regex that is used for simple model names.

tools/downloader/src/open_model_zoo/model_tools/_configuration.py
```diff
@@ -361,15 +362,42 @@ def deserialize(cls, model, name, subdirectory):
...
+            'that are not direct children of the composite model directory'.format(model_dir))
+
+    if list(model_dir.glob('model.yml')):
+        raise DeserializationError('{} directory should not contain any model.yml files'.format(model_dir))
+
+    model_name = model_dir.name
+    model_stages = list(model_dir.glob('*/model.yml'))
+    for model in model_stages:
+        if not model.parent.name.startswith(f'{model_name}-'):
+            raise DeserializationError('Names of composite model parts should start with composite model name')
```
IRDonch, line 380: This message is much better, but it doesn't tell the user _where_ the problem is (and there is still no deserialization context, so that doesn't help either).

tools/downloader/src/open_model_zoo/model_tools/_configuration.py
jkamelin: Done

tools/downloader/src/open_model_zoo/model_tools/_configuration.py
jkamelin: Added deserialization context

tools/downloader/src/open_model_zoo/model_tools/_configuration.py
jkamelin: Fixed

tools/downloader/src/open_model_zoo/model_tools/_configuration.py
jkamelin: Fixed

tools/downloader/src/open_model_zoo/model_tools/_configuration.py
jkamelin: Done

tools/downloader/src/open_model_zoo/model_tools/_configuration.py
jkamelin: Done

Wovchena (approved):

tools/downloader/src/open_model_zoo/model_tools/_configuration.py
```diff
@@ -361,14 +362,51 @@ def deserialize(cls, model, name, subdirectory):
...
+                raise DeserializationError(
+                    'Duplicate composite model name "{}"'.format(composite_model_name))
+            composite_models.append(composite_model_name)
+
     for config_path in sorted(_common.MODEL_ROOT.glob('**/model.yml')):
-        subdirectory = config_path.parent.relative_to(_common.MODEL_ROOT)
+        subdirectory = config_path.parent
+
+        is_composite = ((subdirectory.parent / 'composite-model.yml').exists() and
+            subdirectory.parent.name in composite_models)
```
IRDonch, line 406: Seems like the first half of the condition implies the second half.

jkamelin: Thanks, fixed

IRDonch (approved):

eizamaliev (approved):

PR #2446, opened 2021-05-28
Add suppression list for test cases

eizamaliev:


Wovchena:
`unsupported-devices.txt` contains prior knowledge. It doesn't depend on anything like host machine or the way `run_tests.py` is used. `unsupported-devices.txt` shouldn't be a config, it should be hardcoded somewhere in `demos/tests`.

demos/tests/run_tests.py
```diff
@@ -202,6 +219,14 @@ def option_to_args(key, value):
...
                 device_args = demo.device_args(args.devices.split())
                 for test_case_index, test_case in enumerate(demo.test_cases):
+                    def get_models(case, keys):
+                        models = []
+                        for key in keys:
+                            model = case.options.get(key, None)
+                            if model:
+                                models.append(model.name if isinstance(model, ModelArg) else model.model_name)
+                        return models
+                    test_case_models = get_models(test_case, demo.model_keys)
```
fzhar, line 229: Do we really need this written as separate local function ?  Is it called anywhere more than once ? I'd suggest either write it as separate global function, or local function, but at the very beginning of main (definetely not inside the cycle), or just put its code inside the cycle without creating extra function.

demos/tests/run_tests.py
```diff
@@ -219,6 +244,14 @@ def option_to_args(key, value):
                         continue
 
                     for device, dev_arg in device_args.items():
+                        skip = False
+                        for model in test_case_models:
+                            if unsupported_devices and device in unsupported_devices.get(model, []):
+                                print('Test case #{}/{}: Model {} is not supported by device'
```
fzhar, line 250: Maybe "not supported" is not good term for logging output here. In some test scenarios models/devices might be excluded from testing, but generally they are supported by device.

fzhar (changes requested):

demos/tests/run_tests.py
vladimir-dudnik: well, might be "Model {} is suppressed on device" ?

eizamaliev: Done

demos/tests/run_tests.py
eizamaliev: Done

demos/tests/run_tests.py
```diff
@@ -63,6 +63,8 @@ def parse_args():
         help='list of devices to test')
     parser.add_argument('--report-file', type=Path,
         help='path to report file')
+    parser.add_argument('--unsupported-devices', type=Path, required=False,
+        help='path to file with unsupported devices for each model')
```
vladimir-dudnik, line 67:
```suggestion
    parser.add_argument('--suppressed-devices', type=Path, required=False,
        help='path to file with suppressed devices for each model')
```

demos/tests/run_tests.py
```diff
@@ -139,9 +141,33 @@ def prepare_models(auto_tools_dir, downloader_cache_dir, mo_path, global_temp_di
...
+    if not path:
+        return None
+    unsupported_devices = {}
+    with open(path, "r") as f:
+        for line in f:
+            if line.startswith('#'):
+                continue
+            parsed = line.rstrip('\n').split(sep=',')
+            unsupported_devices[parsed[0]] = parsed[1:]
+    return unsupported_devices
```
vladimir-dudnik, line 154:
```suggestion
def parse_suppressed_device_list(path):
    if not path:
        return None
    suppressed_devices = {}
    with open(path, "r") as f:
        for line in f:
            if line.startswith('#'):
                continue
            parsed = line.rstrip('\n').split(sep=',')
            suppressed_devices[parsed[0]] = parsed[1:]
    return suppressed_devices
```

demos/tests/run_tests.py
```diff
@@ -139,9 +141,33 @@ def prepare_models(auto_tools_dir, downloader_cache_dir, mo_path, global_temp_di
...
+        model = case.options.get(key, None)
+        if model:
+            models.append(model.name if isinstance(model, ModelArg) else model.model_name)
+    return models
+
+
 def main():
     args = parse_args()
 
+    unsupported_devices = parse_unsupported_device_list(args.unsupported_devices)
```
vladimir-dudnik, line 169:
```suggestion
    suppressed_devices = parse_suppressed_device_list(args.suppressed_devices)
```

demos/tests/run_tests.py
```diff
@@ -219,6 +246,14 @@ def option_to_args(key, value):
                         continue
 
                     for device, dev_arg in device_args.items():
+                        skip = False
+                        for model in test_case_models:
+                            if unsupported_devices and device in unsupported_devices.get(model, []):
```
vladimir-dudnik, line 251:
```suggestion
                            if suppressed_devices and device in suppressed_devices.get(model, []):
```

vladimir-dudnik (commented):
suggest to not use word `unsupported` as within a time device may get support particular model, while until tests updated may cause confusion

demos/tests/run_tests.py
fzhar: OK with me

fzhar (approved):

vladimir-dudnik:
?

vladimir-dudnik (approved):

PR #2454, opened 2021-06-01
[Docs]: update input/outputs for bert large

eaidova:
Note: test cases in demo testing are not updated as part of this request. It should be done, when model package will be uploaded


vladimir-dudnik:
cc @Wovchena

vladimir-dudnik (approved):

PR #2456, opened 2021-06-01
Added check for existence of absolute links to files in OMZ Repo

nignatovsky:
Also, replaced all absolute links to files in OMZ Repo by relative ones.


vladimir-dudnik:
Jenkins please retry a build

README.md
```diff
@@ -1,5 +1,5 @@
 # [OpenVINO™ Toolkit](https://01.org/openvinotoolkit) - Open Model Zoo repository
-[![Stable release](https://img.shields.io/badge/version-2021.4-green.svg)](https://github.com/openvinotoolkit/open_model_zoo/releases/tag/2021.4)
```
IRDonch, line 2: This link does not refer to a _file_ in the repository, so it should not be changed. Links to files have `/blob/` after the repository name.

There could also be links to directories (although I don't think the documentation contains any at the moment). They have `/tree/` in the URL and should be handled the same way as links to files.

models/intel/driver-action-recognition-adas-0002/README.md
```diff
@@ -2,7 +2,7 @@
 
 ## Use Case and High-Level Description
 
-This is an action recognition composite model for the driver monitoring use case, consisting of encoder and decoder parts. The encoder model uses Video Transformer approach with MobileNetV2 encoder. It is able to recognize actions such as drinking, doing hair or making up, operating the radio, reaching behind, safe driving, talking on the phone, texting. The full list of recognized actions is located [here](https://github.com/openvinotoolkit/open_model_zoo/blob/develop/demos/action_recognition_demo/python/driver_actions.txt).
+This is an action recognition composite model for the driver monitoring use case, consisting of encoder and decoder parts. The encoder model uses Video Transformer approach with MobileNetV2 encoder. It is able to recognize actions such as drinking, doing hair or making up, operating the radio, reaching behind, safe driving, talking on the phone, texting. The full list of recognized actions is located [here](../../../../../blob/develop/demos/action_recognition_demo/python/driver_actions.txt).
```
IRDonch, line 5: Links to non-Markdown files should be replaced by `<omz_dir>` references, not relative links.

ci/check-documentation.py
```diff
@@ -86,6 +88,22 @@ def complain(message):
         # check local link validity
 
         for url in sorted([ref.url for ref in doc_page.external_references()]):
+            if url.startswith(omz_github_url):
+                suggested_reference = url[len(omz_github_url):]
```
IRDonch, line 92: This value is not the suggested reference and you shouldn't call it that.

ci/check-documentation.py
```diff
@@ -86,6 +88,22 @@ def complain(message):
...
+
+                if suggested_reference.startswith(('blob', 'tree')):
```
IRDonch, line 94:
```suggestion
                if suggested_reference.startswith(('blob/', 'tree/')):
```

ci/check-documentation.py
```diff
@@ -86,6 +88,22 @@ def complain(message):
...
+                    md_path_parent_dir_components = '../' * len(md_path_rel.parent.parts)
```
IRDonch, line 95: This algorithm is rather crude, the suggested URL that it builds goes all the way up to the root directory and then down to the target file. That's usually not a good suggestion. For example, if `a/b/c.md` links to `https://github.com/openvinotoolkit/open_model_zoo/blob/a/b/d.md`, then it should be replaced with just `d.md`, not `../../a/b/d.md`.

ci/check-documentation.py
```diff
@@ -86,6 +88,22 @@ def complain(message):
...
+
+                    if url.endswith('.md'):
+                        suggested_reference = blob_tree_components + md_path_parent_dir_components \
+                            + suggested_reference
+                    else:
+                        suggested_reference = omz_reference_prefix + suggested_reference
```
IRDonch, line 101: 1. The (previous) value of `suggested_reference` may contain URL escape sequences, which should be removed with `urllib.request.url2pathname`. You can then use `as_posix` from `Path` to make sure the path uses forward slashes.
2. The suggestion should include the backtick characters around the reference to make it clear to the developer that they should be included.

nignatovsky:
All suggestions were applied.

ci/check-documentation.py
```diff
@@ -86,6 +87,31 @@ def complain(message):
         # check local link validity
 
         for url in sorted([ref.url for ref in doc_page.external_references()]):
+            if url.startswith(omz_github_url):
+                omz_relative_url = url[len(omz_github_url):]
+                omz_relative_path = Path(urllib.request.url2pathname(omz_relative_url))
+
+                if omz_relative_path.as_posix().startswith(('blob/', 'tree/')):
```
IRDonch, line 94:
```suggestion
                if omz_relative_url.startswith(('blob/', 'tree/')):
```

ci/check-documentation.py
```diff
@@ -86,6 +87,31 @@ def complain(message):
...
+                    omz_relative_path = Path(*omz_relative_path.parts[1:]) # removed blob/ or tree/
+                    omz_relative_posix_path = omz_relative_path.as_posix()
+
+                    distance_to_md_parent_dir = 0
```
IRDonch, line 98: This variable is only used in one branch of the following `if`, so it should be moved there.

ci/check-documentation.py
```diff
@@ -86,6 +87,31 @@ def complain(message):
...
+
+                    if url.endswith('.md'):
+                        for parent in md_path_rel.parents:
+                            try:
+                                omz_relative_posix_path = omz_relative_path.relative_to(parent).as_posix()
+                                break
+                            except ValueError:
+                                distance_to_md_parent_dir += 1
+                        suggested_path = '../' * distance_to_md_parent_dir + omz_relative_posix_path
```
IRDonch, line 107: This value should be converted back to a URL path (with `pathname2url`), since it's going to be used as the link target.

nignatovsky:
All suggestions were applied.

IRDonch (approved):

IRDonch:
Jenkins: test this please

tools/accuracy_checker/README.md
```diff
@@ -166,7 +166,7 @@ models:
     - name: dataset_name
 ```
 Optionally you can use global configuration. It can be useful for avoiding duplication if you have several models which should be run on the same dataset.
-Example of global definitions file can be found [here](https://github.com/openvinotoolkit/open_model_zoo/blob/master/tools/accuracy_checker/dataset_definitions.yml). Global definitions will be merged with evaluation config in the runtime by dataset name.
+Example of global definitions file can be found at `<omz_dir>/tools/accuracy_checker/dataset_definitions.yml`. Global definitions will be merged with evaluation config in the runtime by dataset name.
```
eaidova, line 169:
```suggestion
Example of global definitions file can be found at `<omz_dir>/data/dataset_definitions.yml`. Global definitions will be merged with evaluation config in the runtime by dataset name.
```

eaidova (approved):

PR #2458, opened 2021-06-01
demos/text_detection: reuse what’s already there to simplify `Cnn`s construction and reduce load time for EncoderDecoderCNN

Wovchena:


demos/text_detection_demo/cpp/src/cnn.cpp
```diff
@@ -28,37 +28,35 @@ Cnn::Cnn(const std::string &model_path, Core & ie, const std::string & deviceNam
     if (inputInfo.size() != 1) {
         throw std::runtime_error("The network should have only one input");
     }
+    input_name_ = inputInfo.begin()->first;
     InputInfo::Ptr inputInfoFirst = inputInfo.begin()->second;
 
     SizeVector input_dims = inputInfoFirst->getInputData()->getTensorDesc().getDims();
+    if (input_dims.size() != 4) {
+        throw std::runtime_error("The network should have 4-dimensional input");
+    }
```
morkovka1337, line 37: I would not restrict the network to have 4 inputs. The easiest example is the same network but with LSTM decoder instead of GRU. LSTM would have 2 hidden tensors, and the whole network thus would have 5 inputs.

Wovchena: This check is about one particular input dimensionality, not number of inputs. As for the number of inputs, this part corresponds to encoder, but not decoder

vladimir-dudnik (approved):

demos/text_detection_demo/cpp/src/cnn.cpp
morkovka1337: Did not see this is the encoder. Agree.

morkovka1337 (approved):

PR #2489, opened 2021-06-08
update smart_classroom_demo doc

vladimir-dudnik:
fix demo doc


vladimir-dudnik:
cc @Wovchena @mpashchenkov G-API demo doc also should be corrected in the same way

sovrasov (approved):

PR #2504, opened 2021-06-10
retinaface-resnet50-pytorch C++ model wrapper

fzhar:
TODO list:

- [x] Update Docs
- [x] Update Tests

### DIFF
--- a/demos/common/cpp/models/include/models/detection_model_retinaface_pt.h
+++ b/demos/common/cpp/models/include/models/detection_model_retinaface_pt.h
@@ -20,8 +20,7 @@
 #include <utils/nms.hpp>
 #include <string>
 
-class ModelRetinaFacePT
-    : public DetectionModel {
+class ModelRetinaFacePT : public DetectionModel {
 public:
     struct Box {
         float cX;
@@ -47,8 +46,7 @@ public:
     /// @param confidenceThreshold - threshold to eliminate low-confidence detections.
     /// Any detected object with confidence lower than this threshold will be ignored.
     /// @param useAutoResize - if true, image will be resized by IE.
-    /// @param labels - array of labels for every class. If this array is empty or contains less elements
-    /// than actual classes number, default "Label #N" will be shown for missing items.
+    /// @param boxIOUThreshold - threshold for NMS boxes filtering, varies in [0.0, 1.0] range.
     ModelRetinaFacePT(const std::string& modelFileName, float confidenceThreshold, bool useAutoResize, float boxIOUThreshold);
     std::unique_ptr<ResultBase> postprocess(InferenceResult& infResult) override;
 
--- a/demos/common/cpp/models/include/models/detection_model_retinaface_pt.h
+++ b/demos/common/cpp/models/include/models/detection_model_retinaface_pt.h
@@ -18,6 +18,7 @@
 #include <vector>
 #include "detection_model.h"
 #include <utils/nms.hpp>
+#include <string>
 
 class ModelRetinaFacePT
     : public DetectionModel {
@@ -41,8 +42,6 @@ public:
         float getYCenter() const { return top + (getHeight() - 1.0f) / 2.0f; }
     };
 
-    static constexpr int LANDMARKS_NUM = 5;
-
     /// Loads model and performs required initialization
     /// @param model_name name of model to load
     /// @param confidenceThreshold - threshold to eliminate low-confidence detections.
@@ -52,15 +51,11 @@ public:
     /// than actual classes number, default "Label #N" will be shown for missing items.
     class ModelRetinaFacePT(const std::string& model_name, float confidenceThreshold, bool useAutoResize, float boxIOUThreshold);
     std::unique_ptr<ResultBase> postprocess(InferenceResult& infResult) override;
-    std::unique_ptr<ResultBase> ModelRetinaFacePT::process_output(InferenceResult& infResult);
-protected:
 
-    bool shouldDetectMasks;
-    bool shouldDetectLandmarks;
+protected:
+    size_t landmarksNum;
     const float boxIOUThreshold;
-    const float maskThreshold;
     float variance[2] = { 0.1f, 0.2f };
-    float landmarkStd;
 
     enum EOutputType {
         OT_BBOX,
@@ -69,15 +64,15 @@ protected:
         OT_MAX
     };
 
-    std::vector<ModelRetinaFacePT::Box> priorData;
+    std::vector<ModelRetinaFacePT::Box> priors;
 
     std::vector<uint32_t> doThresholding(const InferenceEngine::MemoryBlob::Ptr& rawData, const float confidenceThreshold);
     std::vector<float> getFilteredScores(const InferenceEngine::MemoryBlob::Ptr& rawData, const std::vector<uint32_t>& indicies);
     std::vector<cv::Point2f> getFilteredLandmarks(const InferenceEngine::MemoryBlob::Ptr& rawData,
-        const std::vector<uint32_t>& indicies, std::vector<ModelRetinaFacePT::Box> priors, int imgWidth, int imgHeight);
+        const std::vector<uint32_t>& indicies, int imgWidth, int imgHeight);
     std::vector<ModelRetinaFacePT::Box> generatePriorData();
     std::vector<ModelRetinaFacePT::Rect> getFilteredProposals(const InferenceEngine::MemoryBlob::Ptr& rawData,
-        std::vector<ModelRetinaFacePT::Box> priors, const std::vector<uint32_t>& indicies, int imgWidth, int imgHeight);
+        const std::vector<uint32_t>& indicies, int imgWidth, int imgHeight);
 
     void prepareInputsOutputs(InferenceEngine::CNNNetwork& cnnNetwork) override;
 };
--- a/demos/common/cpp/models/include/models/detection_model_retinaface_pt.h
+++ b/demos/common/cpp/models/include/models/detection_model_retinaface_pt.h
@@ -64,7 +64,7 @@ protected:
 
     std::vector<ModelRetinaFacePT::Box> priors;
 
-    std::vector<uint32_t> doThresholding(const InferenceEngine::MemoryBlob::Ptr& rawData, const float confidenceThreshold);
+    std::vector<uint32_t> filterByScore(const InferenceEngine::MemoryBlob::Ptr& rawData, const float confidenceThreshold);
     std::vector<float> getFilteredScores(const InferenceEngine::MemoryBlob::Ptr& rawData, const std::vector<uint32_t>& indicies);
     std::vector<cv::Point2f> getFilteredLandmarks(const InferenceEngine::MemoryBlob::Ptr& rawData,
         const std::vector<uint32_t>& indicies, int imgWidth, int imgHeight);
--- a/demos/common/cpp/models/include/models/detection_model_retinaface_pt.h
+++ b/demos/common/cpp/models/include/models/detection_model_retinaface_pt.h
@@ -1,0 +1,83 @@
+/*
+// Copyright (C) 2020-2021 Intel Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+*/
+
+#pragma once
+#include <vector>
+#include "detection_model.h"
+#include <utils/nms.hpp>
+
+class ModelRetinaFacePT
+    : public DetectionModel {
+public:
+    struct Box {
+        float cX;
+        float cY;
+        float width;
+        float height;
+    };
+
+    struct Rect {
+        float left;
+        float top;
+        float right;
+        float bottom;
+
+        float getWidth() const { return (right - left) + 1.0f; }
+        float getHeight() const { return (bottom - top) + 1.0f; }
+        float getXCenter() const { return left + (getWidth() - 1.0f) / 2.0f; }
+        float getYCenter() const { return top + (getHeight() - 1.0f) / 2.0f; }
+    };
+
+    static constexpr int LANDMARKS_NUM = 5;
+
+    /// Loads model and performs required initialization
+    /// @param model_name name of model to load
+    /// @param confidenceThreshold - threshold to eliminate low-confidence detections.
+    /// Any detected object with confidence lower than this threshold will be ignored.
+    /// @param useAutoResize - if true, image will be resized by IE.
+    /// @param labels - array of labels for every class. If this array is empty or contains less elements
+    /// than actual classes number, default "Label #N" will be shown for missing items.
+    class ModelRetinaFacePT(const std::string& model_name, float confidenceThreshold, bool useAutoResize, float boxIOUThreshold);
+    std::unique_ptr<ResultBase> postprocess(InferenceResult& infResult) override;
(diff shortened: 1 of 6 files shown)


eizamaliev (changes requested):

demos/common/cpp/models/src/detection_model_retinaface_pt.cpp
```diff
@@ -0,0 +1,249 @@
...
+        else if (output.first.find("landmark") != std::string::npos) {
+            outputsNames.resize(std::max(outputsNames.size(), (size_t)OT_LANDMARK + 1));
+            outputsNames[OT_LANDMARK] = output.first;
+            shouldDetectLandmarks = true;
+        }
+        else {
+            continue;
+        }
+    }
+
```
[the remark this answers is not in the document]
fzhar: It is checked below. Landmarks are optional

demos/common/cpp/models/include/models/detection_model_retinaface_pt.h
```diff
@@ -0,0 +1,83 @@
...
+    /// Loads model and performs required initialization
+    /// @param model_name name of model to load
+    /// @param confidenceThreshold - threshold to eliminate low-confidence detections.
+    /// Any detected object with confidence lower than this threshold will be ignored.
+    /// @param useAutoResize - if true, image will be resized by IE.
+    /// @param labels - array of labels for every class. If this array is empty or contains less elements
+    /// than actual classes number, default "Label #N" will be shown for missing items.
+    class ModelRetinaFacePT(const std::string& model_name, float confidenceThreshold, bool useAutoResize, float boxIOUThreshold);
+    std::unique_ptr<ResultBase> postprocess(InferenceResult& infResult) override;
+    std::unique_ptr<ResultBase> ModelRetinaFacePT::process_output(InferenceResult& infResult);
```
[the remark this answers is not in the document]
fzhar: removed. it's nott needed anymore

demos/common/cpp/models/include/models/detection_model_retinaface_pt.h
```diff
@@ -0,0 +1,78 @@
...
+*/
+
+#pragma once
+#include <vector>
+#include "detection_model.h"
+#include <utils/nms.hpp>
+#include <string>
+
+class ModelRetinaFacePT
+    : public DetectionModel {
```
eizamaliev, line 24:
```suggestion
class ModelRetinaFacePT: public DetectionModel {
```

demos/common/cpp/models/src/detection_model_retinaface_pt.cpp
```diff
@@ -0,0 +1,250 @@
...
+        outputsNames.resize(2);
+        if (output.first.find("bbox") != std::string::npos) {
+            outputsNames[OT_BBOX] = output.first;
+        }
+        else if (output.first.find("cls") != std::string::npos) {
+            outputsNames[OT_SCORES] = output.first;
+        }
+        else if (output.first.find("landmark") != std::string::npos) {
+            outputsNames.resize(std::max(outputsNames.size(), (size_t)OT_LANDMARK + 1));
```
eizamaliev, line 77: It is better to use explicit value `3` or the `OT_MAX` instead of `(size_t)OT_LANDMARK + 1` expression.

demos/common/cpp/models/include/models/detection_model_retinaface_pt.h
```diff
@@ -0,0 +1,78 @@
...
+    /// than actual classes number, default "Label #N" will be shown for missing items.
+    ModelRetinaFacePT(const std::string& modelFileName, float confidenceThreshold, bool useAutoResize, float boxIOUThreshold);
+    std::unique_ptr<ResultBase> postprocess(InferenceResult& infResult) override;
+
+protected:
+    size_t landmarksNum;
+    const float boxIOUThreshold;
+    float variance[2] = { 0.1f, 0.2f };
+
+    enum EOutputType {
```
eizamaliev, line 60: I think, you should explicitly set the values of enumerations.

demos/common/cpp/models/include/models/detection_model_retinaface_pt.h
```diff
@@ -0,0 +1,78 @@
...
+        float getYCenter() const { return top + (getHeight() - 1.0f) / 2.0f; }
+    };
+
+    /// Loads model and performs required initialization
+    /// @param model_name name of model to load
+    /// @param confidenceThreshold - threshold to eliminate low-confidence detections.
+    /// Any detected object with confidence lower than this threshold will be ignored.
+    /// @param useAutoResize - if true, image will be resized by IE.
+    /// @param labels - array of labels for every class. If this array is empty or contains less elements
+    /// than actual classes number, default "Label #N" will be shown for missing items.
```
eizamaliev, line 51:
```suggestion
    /// @param boxIOUThreshold - threshold for NMS boxes filtering, varies in [0.0, 1.0] range.
```

demos/common/cpp/models/include/models/detection_model_retinaface_pt.h
```diff
@@ -0,0 +1,78 @@
...
+    enum EOutputType {
+        OT_BBOX,
+        OT_SCORES,
+        OT_LANDMARK,
+        OT_MAX
+    };
+
+    std::vector<ModelRetinaFacePT::Box> priors;
+
+    std::vector<uint32_t> doThresholding(const InferenceEngine::MemoryBlob::Ptr& rawData, const float confidenceThreshold);
```
eizamaliev, line 69: May be `filterBoxes` more suitable?

demos/common/cpp/models/src/detection_model_retinaface_pt.cpp
```diff
@@ -0,0 +1,250 @@
...
+        }
+    }
+
+    if (outputsNames[OT_BBOX] == "" || outputsNames[OT_SCORES] == "") {
+        throw std::runtime_error("Bbox or cls layers are not found");
+    }
+
+    if (outputsNames.size() != 2 && outputsNames.size() != 3) {
+        throw std::logic_error("Expected 2 or 3 output blobs");
+    }
```
eizamaliev, line 92: The `outputsNames.size` will be always 2 or 3.

demos/common/cpp/models/src/detection_model_retinaface_pt.cpp
```diff
@@ -0,0 +1,250 @@
...
+
+std::vector<uint32_t> ModelRetinaFacePT::doThresholding(const InferenceEngine::MemoryBlob::Ptr& rawData, const float confidenceThreshold) {
+    std::vector<uint32_t> indicies;
+    auto desc = rawData->getTensorDesc();
+    auto sz = desc.getDims();
+    InferenceEngine::LockedMemory<const void> outputMapped = rawData->rmap();
+    const float *memPtr = outputMapped.as<float*>();
+
+    for (uint32_t x = 0; x < sz[1]; ++x) {
+        auto idx = (x * sz[2] + 1);
```
eizamaliev, line 106: I'd prefer named constant for sz[1] and sz[2] values.
*Hereinafter*

demos/common/cpp/models/src/detection_model_retinaface_pt.cpp
```diff
@@ -0,0 +1,250 @@
...
+
+std::vector<cv::Point2f> ModelRetinaFacePT::getFilteredLandmarks(const InferenceEngine::MemoryBlob::Ptr& rawData, const std::vector<uint32_t>& indicies, int imgWidth, int imgHeight) {
+    InferenceEngine::LockedMemory<const void> outputMapped = rawData->rmap();
+    const float *memPtr = outputMapped.as<float*>();
+    auto desc = rawData->getTensorDesc();
+    auto sz = desc.getDims();
+
+    std::vector<cv::Point2f> landmarks(landmarksNum*indicies.size());
+
+    for (size_t i = 0; i < indicies.size(); i++) {
```
eizamaliev, line 139: Looks like the range based for more suitable. It could save us line of code!

demos/common/cpp/models/src/detection_model_retinaface_pt.cpp
```diff
@@ -0,0 +1,250 @@
...
+    return landmarks;
+}
+
+std::vector<ModelRetinaFacePT::Box> ModelRetinaFacePT::generatePriorData() {
+    float globalMinSizes[][2] = { {16, 32}, {64, 128}, {256, 512} };
+    float steps[] = { 8., 16., 32. };
+    std::vector<ModelRetinaFacePT::Box> anchors;
+    for (size_t stepNum = 0; stepNum < _countof(steps); stepNum++) {
+        int featureW = (int)std::round(netInputWidth / steps[stepNum]);
+        int featureH = (int)std::round(netInputHeight / steps[stepNum]);
```
eizamaliev, line 156:
```suggestion
        const int featureW = (int)std::round(netInputWidth / steps[stepNum]);
        const int featureH = (int)std::round(netInputHeight / steps[stepNum]);
```

eizamaliev (changes requested):

demos/common/cpp/models/src/detection_model_retinaface_pt.cpp
```diff
@@ -0,0 +1,250 @@
...
+// limitations under the License.
+*/
+
+#include <ngraph/ngraph.hpp>
+#include <utils/common.hpp>
+#include <utils/slog.hpp>
+#include "models/detection_model_retinaface_pt.h"
+
+#if !defined(_countof)
+#define _countof(x) (sizeof(x) / sizeof(x[0]))
```
Wovchena, line 23: Reuse `arraySize()` from common.

demos/common/cpp/models/src/detection_model_retinaface_pt.cpp
```diff
@@ -0,0 +1,250 @@
...
+        if (output.first.find("bbox") != std::string::npos) {
+            outputsNames[OT_BBOX] = output.first;
+        }
+        else if (output.first.find("cls") != std::string::npos) {
+            outputsNames[OT_SCORES] = output.first;
+        }
+        else if (output.first.find("landmark") != std::string::npos) {
+            outputsNames.resize(std::max(outputsNames.size(), (size_t)OT_LANDMARK + 1));
+            outputsNames[OT_LANDMARK] = output.first;
+            landmarksNum = output.second->getDims()[2]/2; // Each landmark consist of 2 variables (x and y)
```
Wovchena, line 79:
```suggestion
            landmarksNum = output.second->getDims()[2] / 2; // Each landmark consist of 2 variables (x and y)
```

demos/common/cpp/models/src/detection_model_retinaface_pt.cpp
```diff
@@ -0,0 +1,250 @@
...
+
+
+std::vector<uint32_t> ModelRetinaFacePT::doThresholding(const InferenceEngine::MemoryBlob::Ptr& rawData, const float confidenceThreshold) {
+    std::vector<uint32_t> indicies;
+    auto desc = rawData->getTensorDesc();
+    auto sz = desc.getDims();
+    InferenceEngine::LockedMemory<const void> outputMapped = rawData->rmap();
+    const float *memPtr = outputMapped.as<float*>();
+
+    for (uint32_t x = 0; x < sz[1]; ++x) {
```
Wovchena, line 105: `sz[1]` is `size_t`. Why is `indicies` a vector of `uint32_t`? To me it should be `size_t` as well.

Wovchena (changes requested):
The model itself is in release branch so you should retarget to release.

demos/common/cpp/models/include/models/detection_model_retinaface_pt.h
Wovchena: Same for demos\common\cpp\models\include\models\detection_model_retinaface.h then

demos/common/cpp/models/include/models/detection_model_retinaface_pt.h
fzhar: It does not filter box (that's done below), it calculates indicies of items with score > threshold. I'll rename it to filterByScore

demos/common/cpp/models/src/detection_model_retinaface_pt.cpp
fzhar: done

demos/common/cpp/models/src/detection_model_retinaface_pt.cpp
fzhar: we should resize array to fit OT_LANDMARK to the last item of array. So, we should resize it to OT_LANDMARK+1. It might look little bit clumsy, but internally correct, as numeric value of OT_LANDMARK or OT_MAX may change in future (if we'll adopt this wrapper to another similar model). I'll leave comment about this

demos/common/cpp/models/src/detection_model_retinaface_pt.cpp
fzhar: removed. Assuming that we just skip extra outputs if they are present. But I kept check for outputs names

demos/common/cpp/models/src/detection_model_retinaface_pt.cpp
fzhar: I will make code more complex. Generally, we have CHW layout for outputs, but values there are not H/W dimensions of image or something, but just general rows/columns count. I may put constants like SZ_ROWS and SZ_COLS but they will not give much explanation as they have different meaning in every functions (particularly, COLS may be equal to number of box coordinates, or to number of anchors multipliesd by point coordinates, etc...) . So I'd keep numeric indicies here - at least it will be consistent with model outputs layouts in model documentation.

demos/common/cpp/models/src/detection_model_retinaface_pt.cpp
fzhar: I'll merge this, but frankly... is const modifier THAT important here ?

demos/common/cpp/models/src/detection_model_retinaface_pt.cpp
Wovchena: `indicies` is still a vector of `uint32_t`

fzhar: OK, looks I misunderstood the question. Max index value depends on priors size which is calculated by formula netWidth*netHeight*21/512. That gives us that to overflow uint_32 we need net size of 323598x323598 or greater - and that is definetely not our case. Same time, indicies array may have significant size, so changing uint32_t to size_t would double the RAM amount it consumes. sz[1] is size_t because it is returned by common function suitable for any network. So I suggest to keep indicies as vector of uint32_t to avoid useless overconsumption of RAM.

vladimir-dudnik (approved):

demos/common/cpp/models/src/detection_model_retinaface_pt.cpp
Wovchena: > changing uint32_t to size_t would double the RAM amount it consumes

`size_t` is not grunted to be of size 64 bits.

If you cared that much about memory, you should have used `unit16_t`.

`indicies` is used for indexing and `size_t` is the type for indexing. I'm not sure that your `uint32_t` isn't casted to something similar to hardware equivalent of `size_t` when used, so what you may end up with is extra casts for `uint32_t` and even more, if it is performed by compiler than you will also have a casted copy of a value used: `uint32_t` and `size_t`. That will make things even worse for memory for the moment when the var is used (though I expect this not to happen in most cases).

fzhar: uint32 is CPU natively supported type unless you're going to run this on Z80. 
Just to avoid extra discussion: if you insist on this useless change, I'll do it, just to save time. Please answer here: do it or don't do it.

Wovchena: Do it

fzhar: Done

Wovchena:
You have a compile warning

> The model itself is in release branch so you should retarget to release.

Wovchena:
> The model itself is in release branch so you should retarget to release.

No longer actual

Wovchena (approved):

eizamaliev (approved):

PR #2518, opened 2021-06-15
AC: pin imagecodecs for python3.6 (release)

eaidova:


vladimir-dudnik (approved):

vladimir-dudnik:
@eaidova should we inform someone from other teams to align on python deps?

eaidova:
> @eaidova should we inform someone from other teams to align on python deps?

no, it is only our dependency.
I pin due to issue in accuracy_validation

PR #2522, opened 2021-06-16
ci: bump OpenVINO version

Wovchena:


Wovchena:
There is no 2021.4.563 for Windows

ci/dependencies.yml
```diff
@@ -1,2 +1,2 @@
-openvino_linux: '2021.4.555'
-openvino_windows: '2021.4.555'
+openvino_linux: '2021.4.563'
+openvino_windows: '2021.4.562'
```
vladimir-dudnik, line 2: 562?

vladimir-dudnik: ok, got it, never mind

vladimir-dudnik (approved):

PR #2526, opened 2021-06-17
Support for amazon product data dataset and DIEN model 

mzuevx:


eaidova:
Jenkins ok to test

tools/accuracy_checker/accuracy_checker/adapters/classification.py
```diff
@@ -73,15 +81,20 @@ def process(self, raw, identifiers, frame_meta):
         if self.block:
             if self.argmax_output:
                 single_prediction = ArgMaxClassificationPrediction(identifiers[0], prediction)
+            elif self.fixed_output:
+                single_prediction = ArgMaxClassificationPrediction(identifiers[0],
```
eaidova, line 85: does it mean that batching is not supported for this model type?

tools/accuracy_checker/accuracy_checker/annotation_converters/amazon.py
```diff
@@ -0,0 +1,327 @@
...
+import numpy as np
+
+from ..representation import ClassificationAnnotation
+from ..config import NumberField, StringField, PathField, BoolField
+from ..utils import get_path
+from .format_converter import BaseFormatConverter
+from .format_converter import ConverterReturn
+
+def unicode_to_utf8(d):
+    return dict((key, value) for (key,value) in d.items())
```
eaidova, line 28: hmm... looks like function name is not correspond its behaviour.
Also not sure that we need function wrapper for single line operation

tools/accuracy_checker/accuracy_checker/annotation_converters/amazon.py
```diff
@@ -0,0 +1,327 @@
...
+
+def load_dict(filename):
+    with open(filename, 'rb') as f:
+        return unicode_to_utf8(pickle.load(f))
```
eaidova, line 32: you can provide encoding to pickle reader if need

tools/accuracy_checker/accuracy_checker/annotation_converters/amazon.py
```diff
@@ -0,0 +1,327 @@
...
+
+class DataIterator:
+
```
eaidova, line 35:
```suggestion
```

tools/accuracy_checker/accuracy_checker/metrics/README.md
```diff
@@ -19,6 +19,7 @@ Accuracy Checker supports following set of metrics:
 Supported representations: `ClassificationAnnotation`, `TextClassificationAnnotation`, `ClassificationPrediction`, `ArgMaxClassificationPrediction`.
   * `top_k` - the number of classes with the highest probability, which will be used to decide if prediction is correct.
   * `match` - Batch-oriented binary classification metric. Metric value calculated for each record in batch. Parameter `top_k` ignored in this mode.
+  * `cast_to_int` - in `match` mode predictions rounded to nearest integer before calculating metric value.
```
eaidova, line 22: we have postprocessor for this purposes, possibly better to extend it to work with classification data instead manipulation in metric

mzuevx: Accuracy metric requires integer prediction values. Roc_auc_score metric assumes that prediction values passed "as is" without casting to int. Cast_to_int postprocessor enabled for classification data but we will loose roc_auc_score.

tools/accuracy_checker/accuracy_checker/adapters/classification.py
mzuevx: No, model supports batching. For each batch model outputs pair of values, direct and reverse probability, we need only direct one.

tools/accuracy_checker/accuracy_checker/annotation_converters/README.md
```diff
@@ -569,6 +569,26 @@ The main difference between this converter and `super_resolution` in data organi
   * `labels_file` - path to file with class labels in csv format.
   * `images_dir` - path to images folder (Optional).
   * `label_start` - specifies label index start in label map. You can provide another value, if you want to use this dataset for separate label validation (Optional, default value is 1).
+* `amazon_product_data` - converts book reviews from `Amazon Product Data` (http://snap.stanford.edu/data/amazon/productGraph) dataset preprocessed for DIEN model to `ClassificationAnnotation`.
+  * `test_data` - path to file with user reviews
+  * `uid_voc` - path to user identification vocabulary
+  * `mid_voc' - path to
```
eaidova, line 575:
```suggestion
  * `mid_voc` - path to
```

tools/accuracy_checker/accuracy_checker/annotation_converters/README.md
```diff
@@ -569,6 +569,26 @@ The main difference between this converter and `super_resolution` in data organi
...
+  * `cat_voc` - path to category vocabulary
+  * `item_info` - path to `item-info` file
+  * `reviews_info` - path to `reviews-info` file
+  * `preprocessed_dir` - path to preprocessed data
+  * `separator` - separator between input name and batch identifier in preprocessed data filename. (Optional, default `#`)`
```
eaidova, line 580:
```suggestion
  * `separator` - separator between input name and batch identifier in preprocessed data filename. (Optional, default `#`)
```

tools/accuracy_checker/accuracy_checker/metrics/classification.py
```diff
@@ -345,7 +349,7 @@ def update(self, annotation, prediction):
         return 0
 
     def evaluate(self, annotations, predictions):
-        all_results = np.concatenate([t.squeeze() for t in self.results])
+        all_results = np.concatenate([t.squeeze() if len(t.shape) > 1 or t.shape[0] > 1 else t for t in self.results])
```
eaidova, line 352: 1. ndim returns numer of dimensions
2. there is no guaranties that data is np arays
```suggestion
        all_results = np.concatenate([t.squeeze() if np.ndim(t) > 1 or np.shape(t)[0] > 1 else t for t in self.results])
```
Also possibly better to perform this modification during adding values to results instead

tools/accuracy_checker/accuracy_checker/annotation_converters/README.md
eaidova: path to where?

tools/accuracy_checker/accuracy_checker/metrics/classification.py
mzuevx: Yes, this implementation of roc_auc_score metric was introduced for huge DLRM model in multibatch mode. Since huge DLRM is out of scope, it seems better to remove squeeze() of results at all.

eaidova (approved):

PR #2527, opened 2021-06-17
AC: Added saving of intermediate metrics results to csv file

Anna-Mironova:


tools/accuracy_checker/accuracy_checker/main.py
```diff
@@ -215,6 +215,11 @@ def add_tool_settings_args(parser):
         default=1000,
         required=False
     )
+    tool_settings_args.add_argument(
+        '--csv_intermediate_result',
```
eaidova, line 219: I think no need in separated file, just dump on the same file like final result

tools/accuracy_checker/accuracy_checker/evaluators/custom_evaluators/cocosnet_evaluator.py
```diff
@@ -332,6 +335,15 @@ def configure_intermediate_metrics_results(config):
     def dataset_size(self):
         return self.dataset.size
 
+    def _write_results_to_csv(self, csv_file, ignore_results_formatting, metric_interval, processing_info):
```
eaidova, line 338: I believe this method behavior is the same for all evaluators, please move to base class

tools/accuracy_checker/accuracy_checker/evaluators/module_evaluator.py
```diff
@@ -36,6 +36,7 @@ def from_configs(cls, config, *args, **kwargs):
         return cls(load_module(module, python_path).from_configs(module_config, *args, **kwargs), config)
 
     def process_dataset(self, stored_predictions, progress_reporter, *args, **kwargs):
+        kwargs['processing_info'] = self._internal_module.get_processing_info(self._config)
```
eaidova, line 39: no need to provide it into evaluator, you can get it inside internal module directly

eaidova (approved):

PR #2529, opened 2021-06-17
AC docs: remove wrong parameters description from yolo v3 adapter (develop)

eaidova:
also fixed grammar and typo


vladimir-dudnik:
I think we will take this from release branch, no need to have second PR for the same change

PR #2542, opened 2021-06-24
Support yolo-v4 tf2 for object detection demo

akorobeinikov:


vladimir-dudnik:
are there issues with yolo-v4 support in python object_detection_demo?

akorobeinikov:
> are there issues with yolo-v4 support in python object_detection_demo?

Yes. Output names for `yolo-v4-tf` looks like `StatefulPartitionedCall/model/conv2d_<number>/BiasAdd/Add`, but no `conv2d_<number>/BiasAdd/Add`. The demo requires names to start with `conv`.

vladimir-dudnik:
missed in the release?

akorobeinikov:
> missed in the release?

Yes, missed

eizamaliev (approved):

vladimir-dudnik:
@akorobeinikov @eizamaliev line `if name.startswith('conv'):` was added by @dkurt to distinguish yolo-v4-tiny model. were there any changes in MO regarding yolo-v4-tiny model, which makes this line not needed?

dkurt:
@vladimir-dudnik , seems like that because I've verified the demo when added this check.

demos/common/python/models/yolo.py
```diff
@@ -245,10 +245,9 @@ def _get_output_info(self):
 
         output_info = {}
         for i, (name, layer) in enumerate(outputs):
-            if name.startswith('conv'):
```
dkurt, line 248: Maybe let's replace to `if name.find('conv') != -1` ?

dkurt (approved):

vladimir-dudnik:
closing this, the same change will come from release branch

PR #2551, opened 2021-06-28
object_detection_demo.py: move boxes clipping to postprocess

anzhella-pankratova:
1. `draw_detections` and `print_raw_results` functions don't have to apply thresholding, because all models (except ssd) filter detections by threshold in postprocess. Added thresholding to SSD models instead
2. In two stage pipelining the second model may require clipped boxes, so the clipping should be moved to postprocess (otherwise the clipping is applied twice)

### DIFF
--- a/demos/object_detection_demo/python/object_detection_demo.py
+++ b/demos/object_detection_demo/python/object_detection_demo.py
@@ -164,7 +164,8 @@ def get_model(ie, args):
         raise ValueError("{} model doesn't support input transforms.".format(args.architecture_type))
 
     if args.architecture_type == 'ssd':
-        return models.SSD(*common_args, labels=args.labels, keep_aspect_ratio_resize=args.keep_aspect_ratio)
+        return models.SSD(*common_args, labels=args.labels, keep_aspect_ratio_resize=args.keep_aspect_ratio,
+                          threshold=args.prob_threshold)
     elif args.architecture_type == 'ctpn':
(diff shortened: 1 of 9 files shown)


demos/object_detection_demo/python/object_detection_demo.py
```diff
@@ -187,41 +188,32 @@ def get_model(ie, args):
...
-            cv2.putText(frame, '{} {:.1%}'.format(det_label, detection.score),
-                        (xmin, ymin - 7), cv2.FONT_HERSHEY_COMPLEX, 0.6, color, 1)
-            if isinstance(detection, models.DetectionWithLandmarks):
-                for landmark in detection.landmarks:
-                    landmark = output_transform.scale(landmark)
-                    cv2.circle(frame, (int(landmark[0]), int(landmark[1])), 2, (0, 255, 255), 2)
+        class_id = int(detection.id)
+        color = palette[class_id]
+        det_label = labels[class_id] if labels and len(labels) >= class_id else '#{}'.format(class_id)
+        xmin, ymin, xmax, ymax = detection.get_coords()
```
Wovchena, line 197: What is `get_coords()` for? Why not `detection.xmin, detection.ymin` etc

Wovchena: Never mind

Wovchena (approved):

PR #2553, opened 2021-06-29
AC: support single value in text classification

eaidova:


tools/accuracy_checker/accuracy_checker/adapters/nlp.py
```diff
@@ -271,32 +272,39 @@ def parameters(cls):
             'classification_out': StringField(
                 optional=True,
                 description='Classification output layer name. If not provided, first output will be used.'
-            )
+            ),
+            'single_score': BoolField(optional=True, default=False)
```
Anna-Mironova, line 276:
```suggestion
            'single_score': BoolField(
                optional=True, default=False, description='flag that highlight that model return single value '
                'representing class id or probability belonging to class 1 in binary classification case'
            )
```

Anna-Mironova (approved):

PR #2568, opened 2021-07-05
Python demos: add "Required" to parameters helping messages

anzhella-pankratova:
In addition, made `--input` as required argument in _image_inpainting_demo_, _time_series_forecasting_demo_

### DIFF
--- a/demos/segmentation_demo/python/segmentation_demo.py
+++ b/demos/segmentation_demo/python/segmentation_demo.py
@@ -107,8 +107,8 @@ def build_argparser():
     args.add_argument('-h', '--help', action='help', default=SUPPRESS, help='Show this help message and exit.')
     args.add_argument('-m', '--model', help='Required. Path to an .xml file with a trained model.',
                       required=True, type=Path)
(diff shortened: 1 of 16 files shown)


demos/segmentation_demo/python/segmentation_demo.py
```diff
@@ -107,7 +107,7 @@ def build_argparser():
     args.add_argument('-h', '--help', action='help', default=SUPPRESS, help='Show this help message and exit.')
     args.add_argument('-m', '--model', help='Required. Path to an .xml file with a trained model.',
                       required=True, type=Path)
-    args.add_argument('-at', '--architecture_type', help='Required. Specify the model\'s architecture type.',
+    args.add_argument('-at', '--architecture_type', help='Optional. Specify the model\'s architecture type.',
```
Wovchena, line 110: Could you make it required instead? -at corresponds to -m and it is strange to make assumptions about -at and not about -m.

anzhella-pankratova: Done

Wovchena (approved):

PR #2574, opened 2021-07-05
Calculate MAE and MSE as float to prevent integer overflows

mzuevx:


tools/accuracy_checker/accuracy_checker/metrics/regression.py
```diff
@@ -542,11 +542,15 @@ def calculate_distance(x_coords, y_coords, selected_points):
 
 
 def mae_differ(annotation_val, prediction_val):
+    annotation_val = annotation_val.astype(float) if annotation_val.dtype == np.uint8 else annotation_val
```
eaidova, line 545: I think it should be more generic (if value is integer then cast to float)
np.issubdtype(annotation_value.dtype, np.integer)

tools/accuracy_checker/accuracy_checker/metrics/regression.py
```diff
@@ -542,11 +542,15 @@ def calculate_distance(x_coords, y_coords, selected_points):
...
+    prediction_val = prediction_val.astype(float) if prediction_val.dtype == np.uint8 else prediction_val
     return np.abs(annotation_val - prediction_val)
 
 
 def mse_differ(annotation_val, prediction_val):
-    return (annotation_val - prediction_val)**2
+    annotation_val = annotation_val.astype(float) if annotation_val.dtype == np.uint8 else annotation_val
+    prediction_val = prediction_val.astype(float) if prediction_val.dtype == np.uint8 else prediction_val
+    return np.power(annotation_val - prediction_val, 2)
```
eaidova, line 553: why notation is changed?

eaidova:
Jenkins ok to test

tools/accuracy_checker/accuracy_checker/metrics/regression.py
eaidova: and probably before diff call in base class to prevent the issue for other regression metrics)

mzuevx: Fixed

tools/accuracy_checker/accuracy_checker/metrics/regression.py
mzuevx: Returned back

mzuevx:
@eaidova , please review

tools/accuracy_checker/accuracy_checker/metrics/regression.py
```diff
@@ -110,19 +110,37 @@ def _calculate_diff_regression_rep(self, annotation, prediction):
...
                 if np.ndim(diff) > 1:
                     diff = np.mean(diff)
                 diff_dict[key] = diff
             return diff_dict
         if isinstance(prediction.value, dict):
             if len(prediction.value) != 1:
                 raise ConfigError('annotation for all predictions should be provided')
-            diff = self.value_differ(annotation.value, next(iter(prediction.value.values())))
+            a_val = annotation.value
+            a_val = a_val.astype(float) if not np.isscalar(a_val) and np.issubdtype(a_val.dtype,
```
eaidova, line 128: could you please create helper-function to_float for this case to avoid code duplication?

eaidova (approved):

PR #2600, opened 2021-07-15
Update ModelAPI for YoloV4

eizamaliev:
- Total classes now resolved by heuristic from output blobs
- Fix error for tiny model modification
- The anchors and mask are parametrizable now
- The `py/object_detection_demo` can set anchors and masks for YOLOV4

### DIFF
--- a/demos/object_detection_demo/python/object_detection_demo.py
+++ b/demos/object_detection_demo/python/object_detection_demo.py
@@ -227,6 +227,10 @@ def print_raw_results(detections, labels):
 
 def main():
     args = build_argparser().parse_args()
+    if args.architecture_type != 'yolov4' and args.anchors:
+        log.warning('The "--anchors" options works only for "-at==yolov4". Option will be omitted')
+    if args.architecture_type != 'yolov4' and args.masks:
+        log.warning('The "--masks" options works only for "-at==yolov4". Option will be omitted')
 
     log.info('Initializing Inference Engine...')
     ie = IECore()
--- a/demos/object_detection_demo/python/object_detection_demo.py
+++ b/demos/object_detection_demo/python/object_detection_demo.py
@@ -69,6 +69,12 @@ def build_argparser():
                                    help='Optional. The first image size used for CTPN model reshaping. '
                                         'Default: 600 600. Note that submitted images should have the same resolution, '
                                         'otherwise predictions might be incorrect.')
+    common_model_args.add_argument('--anchors', default=None, type=float, nargs='+',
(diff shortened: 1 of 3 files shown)


demos/object_detection_demo/python/object_detection_demo.py
```diff
@@ -69,6 +69,12 @@ def build_argparser():
                                    help='Optional. The first image size used for CTPN model reshaping. '
                                         'Default: 600 600. Note that submitted images should have the same resolution, '
                                         'otherwise predictions might be incorrect.')
+    common_model_args.add_argument('--anchors', default=None, type=float, nargs='+',
+                                   help='Optional. A space separated list of anchors. '
+                                        'By default used default anchors for model. Only for YOLOV4 architecture type.')
+    common_model_args.add_argument('--masks', default=None, type=int, nargs='+',
+                                   help='Optional. A space separated list of mask for anchors. '
+                                        'By default used default masks for model. Only for YOLOV4 architecture type.')
```
Wovchena, line 77: You should report an error if they are set for anything else.

eizamaliev: Good point. But it better be a warning, not error.

Wovchena: Why? warnings are easy to ignore. Errors that are easy to fix should crash demos to make run it correctly

eizamaliev: If any of these option set for another architecture, it should cause no effect at all. Because demo in that case works correctly, I don't see any reason to abort demo.

Wovchena: That would be strange if there was an option that causes no effect

eizamaliev:
```
Only for YOLOV4 architecture type
```

Wovchena: Exactly. You claim it is only for YOLOV4, but accept it for any.

fzhar: IMHO, error states that demo cannot be executed with such options set (i.e. incorrect model input shape - demo would just crash if we continue). Warnings - if demo wolud continue to work correctly, but with some assumptions - i.e. set default value , ignore something. 
For this particular case I'd leave warning: "Warning: --mask parameter is ignored as model type is different from YOLOv4"

fzhar (approved):

PR #2603, opened 2021-07-15
fix object_detection_demo work on hddl

vladimir-dudnik:
Fix for [issue](https://community.intel.com/t5/forums/forumtopicpage/board-id/distribution-openvino-toolkit/message-id/24012#M24012) reported on user forum (internal id CVS-59597)
Free lambda callbacks on requests pool destruction because callback cleanup does not work from callback itself for HDDL
sync processing of tail (in main.cpp) with segmentation demo

### DIFF
--- a/demos/common/cpp/pipelines/src/async_pipeline.cpp
+++ b/demos/common/cpp/pipelines/src/async_pipeline.cpp
@@ -107,9 +107,9 @@ int64_t AsyncPipeline::submitData(const InputData& inputData, const std::shared_
             condVar.notify_one();
     });
 
-        inputFrameId++;
-        if (inputFrameId < 0)
-            inputFrameId = 0;
+    inputFrameId++;
+    if (inputFrameId < 0)
+        inputFrameId = 0;
 
     request->StartAsync();
 
--- a/demos/common/cpp/pipelines/src/async_pipeline.cpp
+++ b/demos/common/cpp/pipelines/src/async_pipeline.cpp
@@ -74,8 +74,7 @@ int64_t AsyncPipeline::submitData(const InputData& inputData, const std::shared_
     preprocessMetrics.update(startTime);
 
     request->SetCompletionCallback(
-        [this, frameID, request, internalModelData, metaData, startTime]()
-        {
+        [this, frameID, request, internalModelData, metaData, startTime]() {
             {
                 const std::lock_guard<std::mutex> lock(mtx);
                 inferenceMetrics.update(startTime);
--- a/demos/common/cpp/pipelines/src/async_pipeline.cpp
+++ b/demos/common/cpp/pipelines/src/async_pipeline.cpp
@@ -105,7 +105,7 @@ int64_t AsyncPipeline::submitData(const InputData& inputData, const std::shared_
                 }
             }
             condVar.notify_one();
-        });
+    });
 
         inputFrameId++;
         if (inputFrameId < 0)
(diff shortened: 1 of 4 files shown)


demos/common/cpp/pipelines/src/async_pipeline.cpp
```diff
@@ -72,12 +76,11 @@ int64_t AsyncPipeline::submitData(const InputData& inputData, const std::shared_
     preprocessMetrics.update(startTime);
 
     request->SetCompletionCallback(
-        [this, frameID, request, internalModelData, metaData, startTime] {
-            request->SetCompletionCallback([]{});
-
+        [this, frameID, request, internalModelData, metaData, startTime]()
+        {
```
fzhar, line 80:
```suggestion
        [this, frameID, request, internalModelData, metaData, startTime]() {
```

demos/common/cpp/pipelines/src/async_pipeline.cpp
```diff
@@ -87,28 +89,27 @@ int64_t AsyncPipeline::submitData(const InputData& inputData, const std::shared_
...
-                        this->callbackException = std::current_exception();
+                    if (!callbackException) {
+                        callbackException = std::current_exception();
                     }
                 }
             }
-
             condVar.notify_one();
-    });
+        });
```
fzhar, line 108:
```suggestion
    });
```

demos/common/cpp/pipelines/src/async_pipeline.cpp
```diff
@@ -87,28 +89,27 @@ int64_t AsyncPipeline::submitData(const InputData& inputData, const std::shared_
...
 
-    inputFrameId++;
-    if (inputFrameId < 0)
-        inputFrameId = 0;
+        inputFrameId++;
+        if (inputFrameId < 0)
+            inputFrameId = 0;
```
fzhar, line 112:
```suggestion
    inputFrameId++;
    if (inputFrameId < 0)
        inputFrameId = 0;
```

demos/common/cpp/pipelines/src/async_pipeline.cpp
fzhar: it should be aligned to the left position of request->setCompletionCallback.
You may shift whole lambda body to the left, or just move closing bracket alignment.

demos/object_detection_demo/cpp/main.cpp
```diff
@@ -402,32 +401,35 @@ int main(int argc, char *argv[]) {
...
+                metrics.update(result->metaData->asRef<ImageMetaData>().timeStamp,
+                    outFrame, { 10, 22 }, cv::FONT_HERSHEY_COMPLEX, 0.65);
+                if (videoWriter.isOpened() && (FLAGS_limit == 0 || framesProcessed <= FLAGS_limit - 1)) {
+                    videoWriter.write(outFrame);
+                }
+                if (!FLAGS_no_show) {
+                    cv::imshow("Detection Results", outFrame);
+                    //--- Updating output window
+                    cv::waitKey(1);
+                }
```
fzhar, line 432: This will not work in case of slow networks. getResult would return nullptr if result is not ready yet and body of "if" will not be executed, but framesProcessed will be increased - as in case if we got processed data. So, we'll miss results from some submissions as framesProcessed would reach frameNum before all submissions are processed.
Please revert this part as it was before.

fzhar (dismissed):

demos/object_detection_demo/cpp/main.cpp
vladimir-dudnik: that what we have in segmentation_demo

fzhar: yes, you're right - after waitForTotalCOmpletion  this code will work. So we can leave it as is.

PR #2609, opened 2021-07-16
cpp demos: changed getDevices implementation in config factory.

ivikhrev:
Minor imrovements in config factory.


vladimir-dudnik:
this will fill devices only once, Not sure we don't want to read them every time. This function does not look to be on performance critical path, so what we optimize there?

ivikhrev:
> this will fill devices only once, Not sure we don't want to read them every time. This function does not look to be on performance critical path, so what we optimize there?

It was @fzhar's suggestion. The idea was to unite parsing and get operations for devices. What for filling it only once - it's ok, config factory is created only once and devices doesn't change during demo's run.

vladimir-dudnik (approved):

PR #2630, opened 2021-07-21
tools/downloader: dataset files downloading

jkamelin:


jkamelin:
Jenkins please retry a build

tools/downloader/src/open_model_zoo/model_tools/data_downloader.py
```diff
@@ -0,0 +1,43 @@
...
+import argparse
+import shutil
+
+from pathlib import Path
+
+from open_model_zoo.model_tools import _common
+
+def copy_data(output_dir):
+    data_path = output_dir / 'data'
+    if data_path.exists():
```
eaidova, line 24: not sure that we need remove user content directory, it can contains not OMZ specific files

jkamelin: probably yes, thanks

eaidova:
FYI @slyubimt this request adding one more tool in downloader tools, which should be a part openvino-dev package. Probably we also need provide some update for package description

eaidova (approved):

PR #2633, opened 2021-07-23
social_distance: downgrade frameCounter to 32-bit

Wovchena:
Follow up https://github.com/openvinotoolkit/open_model_zoo/pull/2071

### DIFF
--- a/demos/social_distance_demo/cpp/main.cpp
+++ b/demos/social_distance_demo/cpp/main.cpp
@@ -166,9 +166,9 @@ struct Context {  // stores all global data for tasks
         std::vector<PersonTrackers> personTracker;
         std::vector<int> minW;
         std::vector<int> maxW;
-        std::vector<std::atomic<int64_t>> lastProcessedIds;
+        std::vector<std::atomic<int32_t>> lastProcessedIds;
 
-        //manual definition of constructor is needed only for creating vector<std::atomic<int64_t>>
+        //manual definition of constructor is needed only for creating vector<std::atomic<int32_t>>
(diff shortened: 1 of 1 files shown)


vladimir-dudnik (approved):

ivikhrev:
Is it needed to replace all `uint64_t` types with `uint32_t` or only those are used as std::atomic? And in this demo we also have 
`std::vector<std::atomic<int64_t>> lastProcessedIds;`. Is this needed to be replaced also?

Wovchena:
The problem is about making 64 bit var atomic on 32 bit system.
Replaced

ivikhrev:
> Is it needed to replace all uint64_t types with uint32_t or only those are used as std::atomic? And in this demo we also have
> std::vector<std::atomic<int64_t>> lastProcessedIds;. Is this needed to be replaced also? Because if we are going to make lastProcessedIds as int32_t, we will have to make all ids as int32_t.

updated

ivikhrev:
> Because if we are going to make lastProcessedIds as `int32_t`, we will have to make all ids as `int32_t`.

Others ids are int64_t, so there would be comparison between `int32_t` and `int64_t`.
` return lastProcessedFrameId == sharedVideoFrame->frameId;`

Wovchena:
Well, it is able to cast them. And https://github.com/openvinotoolkit/open_model_zoo/blob/12e8e4f3bd194e5307d48469721a1982efcd61ee/demos/common/cpp/pipelines/include/pipelines/async_pipeline.h#L85 uses int64_t.

I think the question is wider: what type to use to count frames. int64_t probably isn't friendly to 32 bit system even when it's not atomic. Though I won't answer it in this PR :)

ivikhrev (approved):

PR #2634, opened 2021-07-23
security barrier camera demo fix

vladimir-dudnik:
adopt to model license-plate-recognition-0007 I32 output blob type

### DIFF
--- a/models/public/license-plate-recognition-barrier-0007/README.md
+++ b/models/public/license-plate-recognition-barrier-0007/README.md
@@ -134,7 +134,7 @@ is an integer number encoding a character according to this dictionary:
 
 ### Converted Model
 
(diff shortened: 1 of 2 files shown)


models/public/license-plate-recognition-barrier-0007/README.md
```diff
@@ -134,7 +134,7 @@ is an integer number encoding a character according to this dictionary:
 
 ### Converted Model
 
-Encoded vector of floats, name: `d_predictions:0`, shape: `1, 88`. Each float
+Encoded vector of integerss, name: `d_predictions:0`, shape: `1, 88`. Each value
```
vladimir-dudnik, line 137:
```suggestion
Encoded vector of integers, name: `d_predictions:0`, shape: `1, 88`. Each value
```

Wovchena (approved):

PR #2639, opened 2021-07-27
Add Bandit scanning for security issues in Python code

anzhella-pankratova:


.bandit
```diff
@@ -0,0 +1,399 @@
+
```
Wovchena, line 1: You should exclude it in `.gitattributes` so we wouldn't pack the file.

ci/prepare-documentation.py
```diff
@@ -42,7 +42,7 @@
 import sys
 import urllib.parse
 import urllib.request
-import xml.etree.ElementTree as ET
+import xml.etree.ElementTree as ET # nosec - disable B405:import-xml-etree check
```
anzhella-pankratova, line 45: The recommendation here is to replace `xml.etree.ElementTree` with the equivalent `defusedxml` package. However, the  `defusedxml` doesn't contain `Element` and `SubElement` classes or equivalent one, therefore this replacement would require changes in `add_page` implementation. 
For now the implementation with `xml.etree` is left.

.bandit
anzhella-pankratova: Updated

ci/prepare-documentation.py
Wovchena: Ok

Wovchena (approved):
Thank you!

Wovchena:
CC @IRDonch

Wovchena:
Jenkins please retry a build

eaidova (approved):

PR #2642, opened 2021-07-27
cpp demos: logging fix

ivikhrev:
Replaced remaing `\n` with` slog::endl` and made print statstream only if it isn't emtpy.


vladimir-dudnik (approved):

vladimir-dudnik:
16:54:56    c:\jenkins\workspace\demos_windows10ce29c7ab\src\demos\common\cpp\utils\include\utils\slog.hpp(66): error C2679: binary '<<': no operator found which takes a right-hand operand of type 'const T' (or there is no acceptable conversion)

vladimir-dudnik (approved):

PR #2670, opened 2021-08-11
Code refactoring in python gesture recognition demo

AlexeyLebedev1:
Action recognizer infers only in sync mode, person tracker infers only in async mode, so removed methods are useless.

### DIFF
--- a/demos/gesture_recognition_demo/python/gesture_recognition_demo/person_detector.py
+++ b/demos/gesture_recognition_demo/python/gesture_recognition_demo/person_detector.py
@@ -47,8 +47,10 @@ class PersonDetector(IEModel):
 
         return in_frame, initial_h, initial_w, scale_h, scale_w
 
-    def _process_output(self, result, initial_h, initial_w, scale_h, scale_w, ):
+    def _process_output(self, result):
         """Converts network output to the internal format"""
+
+        scale_h, scale_w = self.last_scales
 
         if result.shape[-1] == 5:  # format: [xmin, ymin, xmax, ymax, conf]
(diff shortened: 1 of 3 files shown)


demos/gesture_recognition_demo/python/gesture_recognition_demo/person_detector.py
```diff
@@ -78,19 +80,8 @@ def wait_request(self, req_id):
         if result is None:
             return None
 
-        initial_h, initial_w = self.last_sizes
```
Wovchena, line 81: I think the rest mentions of `self.last_sizes` should also be removed.

demos/gesture_recognition_demo/python/gesture_recognition_demo/person_detector.py
```diff
@@ -78,19 +80,8 @@ def wait_request(self, req_id):
...
-
-    def __call__(self, frame):
-        """Runs model on the specified input"""
-
-        in_frame, initial_h, initial_w, scale_h, scale_w = self._prepare_frame(frame)
-        result = self.infer(in_frame)
-        out = self._process_output(result, initial_h, initial_w, scale_h, scale_w)
+        out = self._process_output(result)
 
         return out
```
Wovchena, line 87: You have `scale_h, scale_w = self.last_scales` left. While you are here, this whole part should be replaced with just `return self._process_output(result)`

Wovchena (changes requested):

Wovchena (approved):
Thanks

PR #2698, opened 2021-08-18
AC: handle undefined values for detection report

eaidova:


tools/accuracy_checker/accuracy_checker/metrics/detection.py
```diff
@@ -232,9 +232,9 @@ def prepare_summary(label_stats):
                 continue
             ap.append(stat['result'])
             recall.append(stat['recall'])
-            recall_v.append(recall[-1][-1])
+            recall_v.append(recall[-1][-1] if np.size(recall[-1]) else np.NAN),
```
eaidova, line 235:
```suggestion
            recall_v.append(recall[-1][-1] if np.size(recall[-1]) else np.NAN)
```

tools/accuracy_checker/accuracy_checker/metrics/detection.py
```diff
@@ -232,9 +232,9 @@ def prepare_summary(label_stats):
...
             precision.append(stat['precision'])
-            precision_v.append(precision[-1][-1])
+            precision_v.append(precision[-1][-1] if np.size(precision[-1]) else np.NAN),
```
eaidova, line 237:
```suggestion
            precision_v.append(precision[-1][-1] if np.size(precision[-1]) else np.NAN)
```

vladimir-dudnik (approved):

PR #2720, opened 2021-08-31
AC: Added support for YOLOX model

Anna-Mironova:


tools/accuracy_checker/accuracy_checker/adapters/yolo.py
```diff
@@ -653,3 +653,80 @@ def process(self, raw, identifiers, frame_meta):
...
+    def configure(self):
+        self.threshold = self.get_value_from_config('threshold')
+        self.output_name = self.get_value_from_config('output_name')
+
+    @staticmethod
+    def boxes_postprocess(boxes, img_size):
+        grids = []
+        expanded_strides = []
+
+        strides = [8, 16, 32]
```
eaidova, line 681: there is a lot of common with yolor, is it possible to use common adapter or reuse some its parts?

Anna-Mironova: Done

tools/accuracy_checker/accuracy_checker/adapters/yolo.py
```diff
@@ -635,15 +633,25 @@ def xywh2xyxy(x):
...
         result = []
         raw_outputs = self._extract_predictions(raw, frame_meta)
+        self.select_output_blob(raw_outputs)
+        self.output_name = self.output_name or self.output_blob
 
         for identifier, output, meta in zip(identifiers, raw_outputs[self.output_name], frame_meta):
-            valid_predictions = output[output[..., 4] > self.threshold]
+            mask = output[..., 4] > self.threshold
+            boxes = self.boxes_postprocess(output[:, :4], (meta['preferable_width'],
+                                                           meta['preferable_height']))[mask, :]
```
eaidova, line 649: are you sure that preferable_height and width are always in meta, they are depends from preprocessing stage. Maybe we can use more universal values?

tools/accuracy_checker/accuracy_checker/adapters/yolo.py
```diff
@@ -653,3 +661,30 @@ def process(self, raw, identifiers, frame_meta):
...
+        strides = [8, 16, 32]
+        hsizes = [img_size[0] // stride for stride in strides]
+        wsizes = [img_size[1] // stride for stride in strides]
+
+        for hsize, wsize, stride in zip(hsizes, wsizes, strides):
+            xv, yv = np.meshgrid(np.arange(wsize), np.arange(hsize))
+            grid = np.stack((xv, yv), 2).reshape(1, -1, 2)
+            grids.append(grid)
+            shape = grid.shape[:2]
+            expanded_strides.append(np.full((*shape, 1), stride))
```
eaidova, line 683: as I can see, these calculation does not depend from model output. Maybe we can use some cache and recalculate grid only if image size is changed from previous step?

tools/accuracy_checker/accuracy_checker/adapters/README.md
```diff
@@ -79,6 +79,10 @@ AccuracyChecker supports following set of adapters:
   * `output_name` - name of output layer.
   * `threshold` - minimal objectness score value for valid detections (Optional, default 0.001).
   * `num` - num parameter from DarkNet configuration file (Optional, default 5).
+* `yolox` - converting output of YOLOX model to `DetectionPrediction` representation.
+  * `output_name` - name of output layer.
```
eaidova, line 83: is it always required or optional?

tools/accuracy_checker/accuracy_checker/adapters/yolo.py
Anna-Mironova: Done

tools/accuracy_checker/accuracy_checker/adapters/README.md
Anna-Mironova: It is optional parameter

tools/accuracy_checker/accuracy_checker/adapters/yolo.py
Anna-Mironova: Done

eaidova (approved):

PR #2724, opened 2021-09-02
modelAPI cpp refactoring - sharing mat data

fzhar:
Previously model wrapper had a limitation that in case of autoResize input mat had to be kept until request is processed. That was workarounded by putting this mat to pipeline's internal data, but it still made troubles while using model wrappers alone.
Now this limitiation is eliminated.


demos/common/cpp/utils/include/utils/shared_blob_allocator.h
```diff
@@ -0,0 +1,32 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+*/
+
+#pragma once
+#include "ie_allocator.hpp"
+#include "opencv2/core.hpp"
+
+class SharedBlobAllocator : public InferenceEngine::IAllocator {
```
eizamaliev, line 21: Could it be a template class?

demos/common/cpp/models/include/models/internal_model_data.h
```diff
@@ -38,23 +38,12 @@ struct InternalImageModelData : public InternalModelData
...
-
-struct InternalScaleMatData : public InternalModelData
-{
-    InternalScaleMatData(float scaleX, float scaleY, cv::Mat&& mat) :
-        x(scaleX), y(scaleY), mat(std::move(mat)) {}
-
-    float x;
-    float y;
-    cv::Mat mat;
+    float scaleX;
```
vladimir-dudnik, line 47: names causing code like
`    const auto& scale = infResult.internalModelData->asRef<InternalScaleData>();
    float scaleX = stride / upsampleRatio * scale.scaleX;
    float scaleY = stride / upsampleRatio * scale.scaleY;`
could not it be something like
`const auto& scale_factors = infResult.internalModelData->asRef<InternalScaleData>();` ?

vladimir-dudnik (approved):

demos/common/cpp/models/include/models/internal_model_data.h
fzhar: Frankly we probably need to refactor this code and store actual sizes instead of scales. But this is feature for separate PR

Wovchena (changes requested):

demos/common/cpp/utils/src/shared_blob_allocator.cpp
```diff
@@ -0,0 +1,42 @@
...
+        return img.data;
+    }
+    return nullptr;
+}
+
+void SharedBlobAllocator::unlock(void * handle) noexcept {
+}
+
+void * SharedBlobAllocator::alloc(size_t size) noexcept {
+    return img.data;
```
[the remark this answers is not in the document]
fzhar: not clear. What's the divivirgence exactly ?

demos/common/cpp/utils/include/utils/shared_blob_allocator.h
fzhar: there will be differences is alloc/lock functions if we use random type for input data. So I don't see any way how template could help us here.

demos/common/cpp/utils/src/shared_blob_allocator.cpp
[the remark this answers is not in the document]
Wovchena: It seems you are supposed to verify the requested size matches img size https://github.com/openvinotoolkit/openvino/blob/22f29f092ce9e83f931508586ebcb82ad9937a6d/inference-engine/src/inference_engine/include/ie/details/ie_pre_allocator.hpp#L48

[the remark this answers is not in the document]
fzhar: check is added

eizamaliev (approved):

vladimir-dudnik:
14:03:07  /jenkins/workspace/omz/precommit/demos_ubuntu18/src/demos/common/cpp/utils/include/utils/ocv_common.hpp:110:35: error: operands to ?: have different types

vladimir-dudnik (approved):

PR #2732, opened 2021-09-07
Support for Criteo binary dataset

mzuevx:


tools/accuracy_checker/accuracy_checker/annotation_converters/criteo_kaggle_dac.py
```diff
@@ -95,36 +143,34 @@ def convert(self, check_content=False, **kwargs):
...
+                    self.dense_features: np.log1p(x_int),
                     self.lso_features: np.dot(np.expand_dims(np.linspace(0, self.batch - 1, num=self.batch), -1),
-                                              np.ones((1, cat_feat))).T
+                                              np.ones((1, self.cat_feat))).T
                 }
 
                 for name in self.sparse_features.keys():
-                    sample[name] = x_cat[i:i+self.batch, self.sparse_features[name]].T
+                    x_cat_batch = x_cat[:, self.sparse_features[name]]
+                    x_cat_batch = x_cat_batch % self.max_ind_range if self.max_ind_range > 0 else x_cat_batch
```
eaidova, line 172: not clear logic for me, better to use None as identifier that max_ind_range is not set instead of 0.
0 gives user opprotunity to set this value manually even it does not required

tools/accuracy_checker/accuracy_checker/annotation_converters/criteo_kaggle_dac.py
```diff
@@ -32,10 +33,14 @@ def parameters(cls):
         parameters = super().parameters()
         parameters.update({
             'testing_file': PathField(description="Path to testing file."),
+            "binary": BoolField(optional=True, default=False,
+                                    description="Allows input file in binary mode instead of .npz mode"),
             "batch": NumberField(optional=True, default=128, description="Model batch"),
+            "max_ind_range": NumberField(optional=True, default=0,
```
eaidova, line 39: please provide constrains, min_value=1 and value_type=int as you use in mod

tools/accuracy_checker/accuracy_checker/annotation_converters/criteo_kaggle_dac.py
```diff
@@ -83,6 +90,47 @@ def parse_sparse_features(self):
...
+            self.cat_feat = spa_fea
+
+            self.binfile = open(self.src, 'rb') if self.save_preprocessed_features else None
+        else:
+            data = np.load(self.src)
+            self._x_int = data['X_int']
+            self._x_cat = data['X_cat']
+            self._y = data['y']
+            self.count, self.cat_feat = self._x_cat.shape
+            self.count = math.ceil(self.count / self.batch)
```
eaidova, line 114: you can avoid import math if it is need only in this place, the same operation exists in numpy

tools/accuracy_checker/accuracy_checker/annotation_converters/criteo_kaggle_dac.py
```diff
@@ -83,6 +90,47 @@ def parse_sparse_features(self):
...
+            bytes_per_feature = 4
+            tar_fea = 1  # single target
+            den_fea = 13  # 13 dense  features
+            spa_fea = 26  # 26 sparse features
+            tad_fea = tar_fea + den_fea
+            tot_fea = tad_fea + spa_fea
+            self._bytes_per_entry = bytes_per_feature * tot_fea * self.batch
+            self._fea_shape = (self.batch, tot_fea)
+
+            self.count = math.ceil(os.path.getsize(self.src) / self._bytes_per_entry)
```
eaidova, line 104: we avoid to use os.path in AC, please use pathlib:
https://docs.python.org/3/library/pathlib.html#pathlib.Path.stat

mzuevx: Fixed

tools/accuracy_checker/accuracy_checker/annotation_converters/criteo_kaggle_dac.py
mzuevx: Fixed

tools/accuracy_checker/accuracy_checker/annotation_converters/criteo_kaggle_dac.py
mzuevx: Fixed

eaidova (approved):

PR #2741, opened 2021-09-09
Update auto-labeler tool

eizamaliev:


.github/labeler.yml
```diff
@@ -3,3 +3,14 @@
...
+demo:
+  - any: ['demos/**/*', '!demos/**/*.md']
+documentation:
+  - '*.md'
+  - '**/*.md'
+model:
+  - 'models/**/*.yml'
+accuracy checker:
+  - 'tools/accuracy_checker/**/*'
+downloader:
```
vladimir-dudnik, line 15: we are going to rename folder downloader to model_tools because it is not just about downloader. So, label should be changed later.
https://github.com/openvinotoolkit/open_model_zoo/pull/2723

vladimir-dudnik (approved):
cc @Wovchena @nignatovsky @eaidova

.github/labeler.yml
eizamaliev: New path is added.

.github/labeler.yml
```diff
@@ -3,3 +3,16 @@
 dependencies:
   - '**/*requirements*.*'
   - '**/setup.py'
+demo:
+  - any: ['demos/**/*', '!demos/**/*.md']
+documentation:
+  - '*.md'
+  - '**/*.md'
+model:
+  - 'models/**/*.yml'
```
eaidova, line 12: not sure that it is correct.
now, label model will be added only if you modify model.yml or accuracy-check.yml, but the second may have more relation to AC, and what about model documentation and extra files vocabs?

eaidova: e.g. for intel trained models we can not have model.yml at all when it will be added manually before package uploaded

eaidova (approved):

vladimir-dudnik:
Jenkins please retry a build

PR #2747, opened 2021-09-09
clipping is added to detection models

fzhar:
Rectangles clipping to avoid rectangles going out-of-bounds of image

### DIFF
--- a/demos/common/cpp/utils/include/utils/common.hpp
+++ b/demos/common/cpp/utils/include/utils/common.hpp
@@ -39,6 +39,11 @@ constexpr std::size_t arraySize(const T (&)[N]) noexcept {
     return N;
 }
 
+template <typename T>
+T clamp(T value, T low, T high) {
+    return value<low ? low : (value>high ? high : value);
+}
+
 // Redefine operator<< for LogStream to print IE version information.
 inline slog::LogStream& operator<<(slog::LogStream& os, const InferenceEngine::Version& version) {
     os << "OpenVINO Inference Engine" << slog::endl;
--- a/demos/common/cpp/utils/include/utils/common.hpp
+++ b/demos/common/cpp/utils/include/utils/common.hpp
@@ -41,7 +41,7 @@ constexpr std::size_t arraySize(const T (&)[N]) noexcept {
 
 template <typename T>
 T clamp(T value, T low, T high) {
-    return value<low ? low : (value>high ? high : value);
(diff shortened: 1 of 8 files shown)


vladimir-dudnik (approved):

demos/common/cpp/utils/include/utils/common.hpp
```diff
@@ -39,6 +39,11 @@ constexpr std::size_t arraySize(const T (&)[N]) noexcept {
     return N;
 }
 
+template <typename T>
+T clamp(T value, T low, T high) {
+    return value<low ? low : (value>high ? high : value);
```
ivikhrev, line 44:
```suggestion
    return value < low ? low : (value > high ? high : value);
```
Why is function called `clamp`? I think `clip` is more suitable here, similar function in python called the same.
And I would rename `low` and `high` to `minVal` and `maxVal` (or just `min` and `max`) respectively.

eizamaliev (approved):

demos/common/cpp/utils/include/utils/common.hpp
fzhar: There's std::clamp in C++17, as soon as it will be fully supported by our code, we'll switch to it. Signature is taken from that standard.

Wovchena (approved):

Wovchena:
Jenkins please retry a build

ivikhrev:
@fzhar since clipping right now happens in wrappers, we can remove this in `renderDetectionData` function in [object_detection_demo ](https://github.com/openvinotoolkit/open_model_zoo/blob/f90d68ceaccd3dfbd9cb76072973a9572d4a9df5/demos/object_detection_demo/cpp/main.cpp#L258):
```
        if (FLAGS_r) {
            slog::debug << " "
                << std::left << std::setw(9) << obj.label << " | "
                << std::setw(10) << obj.confidence << " | "
                << std::setw(4) << std::max(int(obj.x), 0) << " | "
                << std::setw(4) << std::max(int(obj.y), 0) << " | "
                << std::setw(4) << std::min(int(obj.x + obj.width), outputImg.cols) << " | "
                << std::setw(4) << std::min(int(obj.y + obj.height), outputImg.rows)
                << slog::endl;
        }
```

fzhar:
@ivikhrev , clipping in demo is removed

ivikhrev (approved):

PR #2768, opened 2021-09-20
AC: update namespace

eaidova:
This PR is to align OMZ tools namespace in OpenVINO Python environment. The Accuracy Checker tool will now be located under openvino.tools namespace and accessible through command line as accuracy_check commands, just like it is in openvino-dev PyPi package starting from OpenVINO 2021.4 release

TO DO:

- [x] do not forget bump version before merge


eaidova:
fix for docs https://github.com/openvinotoolkit/open_model_zoo/pull/2772

vladimir-dudnik (approved):

PR #2776, opened 2021-09-22
mobilenet-v3 tf models changed to tf2

akorobeinikov:


models/public/mobilenet-v3-large-1.0-224-tf/accuracy-check.yml
```diff
@@ -1,39 +1,11 @@
 models:
-  - name: mobilenet-v3-large-1.0-224-tf
-
-    launchers:
-      - framework: tf
```
eaidova, line 5: is it possible to add configuration for saved_model (obtained during preconvert script) and tf2 launcher of AC?

models/public/mobilenet-v3-large-1.0-224-tf/accuracy-check.yml
```diff
@@ -1,39 +1,11 @@
...
     launchers:
       - framework: dlsdk
         adapter: classification
 
     datasets:
-      - name: imagenet_1001_classes
+      - name: imagenet_1000_classes
         preprocessing:
           - type: resize
             size: 256
```
eaidova, line 11: please update ref numbers in configs too

akorobeinikov: updated

models/public/mobilenet-v3-large-1.0-224-tf/accuracy-check.yml
akorobeinikov: Added

models/public/mobilenet-v3-large-1.0-224-tf/accuracy-check.yml
eaidova: we do not plan to compare reference now using AC, configs will be executed using --ignore_reference
but you can send patch (we need update docs with note that ref should be normalized for metrics where it can be used as percentage and use scaling for both value and scale before comparing or compare then apply scale).

P.S. please make these changes as separated PR

models/public/mobilenet-v3-small-1.0-224-tf/model.yml
```diff
@@ -19,21 +19,16 @@ description: >-
...
-  - --input=input
-  - --mean_values=input[127.5,127.5,127.5]
-  - --scale_values=input[127.5]
-  - --output=MobilenetV3/Predictions/Softmax
-  - --input_model=$dl_dir/v3-small_224_1.0_float/v3-small_224_1.0_float.pb
+  - --input=input_1
+  - --output=StatefulPartitionedCall/MobilenetV3small/Predictions/Softmax
+  - --saved_model_dir=$dl_dir/mobilenet_v3_small_224_1.0_float.savedmodel
 framework: tf
+quantizable: yes
```
eaidova, line 33: as I know both small and large models required custom configuration for quantization, it is not enough just add yes.
please ask @l-bat about correct pot configuration or remove this status for now (we add it automatically near to release time)

l-bat: Yes, we need to add ignored scopes for small and large models

akorobeinikov: I added `quantization.yml` for small and large models. @l-bat , could you please review these files

eaidova (approved):

l-bat (approved):

PR #2782, opened 2021-09-24
Python classification demo

ivikhrev:
Added python wrapper for classification models and python demo for classification. 
C++ demo was renamed to **Classification Benchmark Demo**.


vladimir-dudnik:
Jenkins please retry a build

ivikhrev:
Jenkins please retry a build

demos/classification_demo/python/classification_demo.py
```diff
@@ -0,0 +1,308 @@
...
+    ie = IECore()
+
+    plugin_config = get_user_config(args.device, args.num_streams, args.num_threads)
+
+    log.info('Reading model {}'.format(args.model))
+    model = models.Classification(ie, args.model, ntop=args.ntop, labels=args.labels, logger=log)
+    log_blobs_info(model)
+
+    gt_indices = None
+    if args.ground_truth and isinstance(cap, DirReader):
```
anzhella-pankratova, line 172:
```suggestion
    if args.ground_truth and cap.get_type() == 'DIR':
```
I think the demo should reports the warning message when a user submits a video/image/camera_id as demo input and ground truth indices via `--ground_truth`.  Also, the documentation should say that the key is applied only with the folder input

demos/common/python/openvino/model_zoo/model_api/models/classification.py
```diff
@@ -0,0 +1,76 @@
...
+        if in_size[2] != in_size[3]:
+            raise Exception('Model input has incorrect image shape. Must be NxN square.\
+                             Got {} x {}.'.format(in_size[2], in_size[3]))
+
+    def _get_outputs(self):
+        out_blob_name = next(iter(self.net.outputs))
+        out_blob = self.net.outputs[out_blob_name]
+        out_size = out_blob.shape
+        if len(out_size) != 2 and len(out_size) != 4:
+            raise Exception('Demo supports topologies only with 2-dimensional or 4-dimensional output')
```
anzhella-pankratova, line 57:
```suggestion
            raise RuntimeError('The Classification model wrapper supports topologies only with 2D or 4D output')
```
Please, specify all Exceptions (it is quite generic) in wrapper and demo by RuntimeError/ValueError or something else.

demos/common/python/openvino/model_zoo/model_api/models/classification.py
```diff
@@ -0,0 +1,76 @@
...
+        if len(out_size) == 4 and out_size[2] != 1 and out_size[3] != 1:
+            raise Exception('Demo supports topologies only with 4-dimensional output which has last two dimensions of size 1')
```
anzhella-pankratova, line 59:
```suggestion
        if len(out_size) == 4 and (out_size[2] != 1 or out_size[3] != 1):
            raise RuntimeError('The Classification model wrapper topologies only with 4-dimensional output which has last two dimensions of size 1')
```

demos/common/python/openvino/model_zoo/model_api/models/classification.py
```diff
@@ -0,0 +1,76 @@
...
+
+        if (out_size[1] == len(self.labels) + 1):
+            self.labels.insert(0, 'other')
+            if self.log:
+                self.log.warning("\tInserted 'other' label as first.")
```
anzhella-pankratova, line 64:
```suggestion
            self.logger.warning("\tInserted 'other' label as first.")
```
Also, you don't need to pass logger to Classification wrapper' constructor, it is initialized in `BaseModel`

demos/tests/cases.py
```diff
@@ -672,6 +672,25 @@ def single_option_cases(key, *args):
         }),
     )),
 
+    PythonDemo(name='classification_demo',
+            device_keys=['-d'],
+            test_cases=combine_cases(
+        TestCase(options={
+            '--no_show': None,
+            '-i': DataDirectoryOrigFileNamesArg('classification'),
+            '--labels': str(OMZ_DIR / 'data/dataset_classes/imagenet_2012.txt')}),
```
anzhella-pankratova, line 681: The `classification_benchmark_demo` is tested with `-gt` key, maybe in this demo it is also worth to test?

anzhella-pankratova (changes requested):
It would be great, if you add an example picture of the demo output (with green-colored correct classification among all ntop classifications) to the documentation file

demos/common/python/openvino/model_zoo/model_api/models/classification.py
```diff
@@ -0,0 +1,76 @@
...
+        if out_size[1] != len(self.labels):
+            raise Exception("Model's number of classes and parsed \
+                 labels must match {} and {}".format(out_size[1], len(self.labels)))
```
anzhella-pankratova, line 67:
```suggestion
            raise RuntimeError("Model's number of classes does not match the number of "
                               "loaded labels ({} != {})".format(out_size[1], len(self.labels)))
```

demos/common/python/openvino/model_zoo/model_api/models/classification.py
ivikhrev: Why `or` is used? It supports models with NxCx1x1, not NxCxHx1 or NxCx1xW,

anzhella-pankratova: If the model requires `out_size[2]` and `out_size[3]` to equal 1, the condition `out_size[2] != 1 and out_size[3] != 1` is True only only if we have `NxCxHxW`. For `NxCxHx1`, `NxCx1xW` cases the exception is not raised, but they are also not suitable, as you said

demos/common/python/openvino/model_zoo/model_api/models/classification.py
ivikhrev: The phrase is taken from https://github.com/openvinotoolkit/open_model_zoo/blob/b7c83530a470734e3e4bde8fe0d80dcc2d9e9b2c/demos/common/cpp/models/src/classification_model.cpp#L111
 I am not sure that we should change it.

anzhella-pankratova: You might leave the phrase the same, but with the addition of the missed brackets `({} and {})`

demos/classification_demo/python/classification_demo.py
```diff
@@ -0,0 +1,308 @@
...
+    initial_labels_pos =  frame.shape[0] - label_height * (2 * len(classifications) + 1)
+
+    if (initial_labels_pos < 0):
+        initial_labels_pos = label_height
+        log.warning('Too much labels to display on this frame, some will be omitted')
+    offset_y = initial_labels_pos
+
+    for cl in classifications:
+        class_id = cl[0]
+        conf = cl[1]
```
anzhella-pankratova, line 111:
```suggestion
    for class_id, conf in classifications:
```

demos/classification_benchmark_demo/cpp/README.md
```diff
@@ -147,7 +147,7 @@ and `<omz_dir>/data/dataset_classes/imagenet_2012.txt` labels file with all othe
 Running the application with the `-h` option yields the following usage message:
 
 ```
-classification_demo [OPTION]
+classification_benchmark_demo [OPTION]
```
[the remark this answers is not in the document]
ivikhrev: updated in 3d324f2

demos/classification_demo/python/README.md
```diff
@@ -0,0 +1,227 @@
...
+  -r, --raw_output_message
+                        Optional. Output inference results raw values showing
+```
+
+Running the application with the empty list of options yields an error message.
+
+For example, use the following command-line command to run the application:
+
+```sh
+./classification_demo -m <path_to_classification_model> \
```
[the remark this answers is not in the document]
ivikhrev: updated in 2be1924

demos/classification_demo/python/README.md
```diff
@@ -0,0 +1,227 @@
...
+If you want to see classification results, you must use "-gt" and "-labels" flags to specify two .txt files containing lists of classes and labels.
+
+"The ground truth" file is used for matching image file names with correct object classes.
+
+It has the following format:
+
+```
+./ILSVRC2012_val_00000001.JPEG 65
+./ILSVRC2012_val_00000002.JPEG 970
+./ILSVRC2012_val_00000003.JPEG 230
```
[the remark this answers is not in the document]
ivikhrev: updated in 2be1924

demos/classification_demo/python/classification_demo.py
ivikhrev: Done in 4ac5f7e.

demos/common/python/openvino/model_zoo/model_api/models/classification.py
ivikhrev: Done in 8e7e8f8.

demos/common/python/openvino/model_zoo/model_api/models/classification.py
ivikhrev: Yeah, good catch, thank you. I fixed it in 8e7e8f8.

demos/common/python/openvino/model_zoo/model_api/models/classification.py
ivikhrev: Removed in 8e7e8f8.

demos/common/python/openvino/model_zoo/model_api/models/classification.py
```diff
@@ -0,0 +1,76 @@
...
+                labels.append(s[(begin_idx + 1):end_idx])
+        return labels
+
+    def _check_inputs(self):
+        in_size = self.inputs[self.image_blob_name].input_data.shape
+        if len(in_size) == 4 and in_size[1] != 3:
+            raise Exception("3-channel 4-dimensional model's input is expected")
+        if in_size[2] != in_size[3]:
+            raise Exception('Model input has incorrect image shape. Must be NxN square.\
+                             Got {} x {}.'.format(in_size[2], in_size[3]))
```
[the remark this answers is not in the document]
ivikhrev: Done in 8e7e8f8.

demos/common/python/openvino/model_zoo/model_api/models/classification.py
ivikhrev: Added brackets in 8e7e8f8.

demos/common/python/openvino/model_zoo/model_api/models/classification.py
ivikhrev: and in 0c79d8f

demos/classification_demo/python/classification_demo.py
ivikhrev: Done in 0c79d8f

demos/common/python/openvino/model_zoo/model_api/models/classification.py
anzhella-pankratova: The long strings, related to helping and error messages, should be split into the two strings. Please, split it in all cases, where you use `\` symbol in messages, because in python it creates a lot of whitespaces when outputted.

demos/common/python/openvino/model_zoo/model_api/models/classification.py
anzhella-pankratova: There is should be `self.logger` instead of `self.log`

demos/common/python/openvino/model_zoo/model_api/models/classification.py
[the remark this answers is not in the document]
anzhella-pankratova: Why not just `if self.h != self.w`? 
The `self.inputs[self.image_blob_name].input_data.shape` is already unpacked in `ImageModel` constructor

demos/classification_demo/python/classification_demo.py
anzhella-pankratova: Apply the suggestion also to `print_raw_results` function, please

demos/common/python/openvino/model_zoo/model_api/models/classification.py
ivikhrev: Done in  6708930

demos/common/python/openvino/model_zoo/model_api/models/classification.py
ivikhrev: fixed in  6708930

demos/common/python/openvino/model_zoo/model_api/models/classification.py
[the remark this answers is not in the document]
ivikhrev: Done in  6708930

demos/classification_demo/python/classification_demo.py
ivikhrev: thx! fixed in 6708930

anzhella-pankratova (approved):

Wovchena (changes requested):

demos/tests/cases.py
ivikhrev: -gt key and associated code was removed.

demos/classification_demo/python/classification_demo.py
```diff
@@ -0,0 +1,258 @@
...
+    log.debug(' Class ID | {:^{width}s}| Confidence '.format('Label', width=label_max_len))
+    for class_id, conf in classifications:
+        log.debug('{:^9} | {:^{width}s}| {:^10f} '.format(class_id, labels[class_id], conf, width=label_max_len))
+
+
+def main():
+    args = build_argparser().parse_args()
+
+    cap = open_images_capture(args.input, args.loop)
+
```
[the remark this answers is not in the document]
ivikhrev: Shouldn't we introduce such key for folder / sequence of images input? Without it images just fly by and you won't catch anything.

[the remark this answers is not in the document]
Wovchena: `delay` != 1 slows down the demo, so the solution becomes controversial. That's why I asked about the source of --pause.

Though I'm fine with that. I only want you to replace it with automatic detection of images input, like `delay = int(cap.get_type() in {'VIDEO', 'CAMERA'})`

demos/classification_demo/python/classification_demo.py
```diff
@@ -0,0 +1,258 @@
...
+                         help='Optional. Enable reading the input in a loop.')
+    io_args.add_argument('-o', '--output', required=False,
+                         help='Optional. Name of the output file(s) to save.')
+    io_args.add_argument('--pause', required=False, default=1, type=int,
+                         help='Optional. Pause in ms between frames to show.')
+    io_args.add_argument('-limit', '--output_limit', required=False, default=1000, type=int,
+                         help='Optional. Number of frames to store in output. '
+                              'If 0 is set, all frames are stored.')
+    io_args.add_argument('--no_show', help="Optional. Don't show output.", action='store_true')
+    io_args.add_argument('--output_resolution', default=None, type=resolution,
```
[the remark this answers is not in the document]
anzhella-pankratova: `set_inputs_preprocessing()` had been added as `Model`'s method before the PR with `ImageModel` implementation was  merged. I will change it.
If we reuse the `blobFromImage()` now, won't we change the implementation of `InputTransform` in 2022.1 patch? There will be also the opportunity to set mean, scale values to Blob in the preprocessing stage.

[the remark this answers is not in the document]
Wovchena: Good point

demos/common/python/openvino/model_zoo/model_api/models/classification.py
```diff
@@ -0,0 +1,69 @@
...
+        if out_size[1] != len(self.labels):
+            raise RuntimeError("Model's number of classes and parsed "
+                               'labels must match ({} != {})'.format(out_size[1], len(self.labels)))
+
+        return out_blob_name
+
+    def postprocess(self, outputs, meta):
+        probs = outputs[self.output_blob_name].squeeze()
+        indices = np.argsort(probs)
+        max_indices = indices[-self.ntop:][::-1] # indices sorted by probs in descended order
```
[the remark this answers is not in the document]
ivikhrev: To add softmax as ngraph op we have to be able to differ models with probs output from models with logits, isn't it happening only after inference?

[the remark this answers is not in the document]
ivikhrev: Of course, if there a way to find out, if model had SoftMax layer, then it would be easy. Is there any way to find out it?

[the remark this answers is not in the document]
Wovchena: You should be able to check net layers how it is done to extract params for yolo through ngraph

demos/classification_demo/python/classification_demo.py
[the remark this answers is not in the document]
ivikhrev: Done in 1594b56

demos/classification_demo/python/classification_demo.py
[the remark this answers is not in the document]
ivikhrev: Options added in 50192a7

demos/common/python/openvino/model_zoo/model_api/models/utils.py
```diff
@@ -139,7 +140,18 @@ def resize_image_letterbox(image, size):
     return resized_image
 
 
+def resize_with_center_square_crop(image, size):
+    if (image.shape[0] > image.shape[1]):
+        offset = (image.shape[0] - image.shape[1]) // 2
+        cropped_frame = image[offset:image.shape[0] - offset, 0:image.shape[1]]
+    else:
+        offset = (image.shape[1] - image.shape[0]) // 2
+        cropped_frame = image[0:image.shape[0], offset:image.shape[1] - offset]
```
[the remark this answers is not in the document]
ivikhrev: Thanks!

demos/common/python/openvino/model_zoo/model_api/models/classification.py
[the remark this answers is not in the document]
ivikhrev: Added softmax and topk in 28ba3d2. Should I do the same for cpp demo in this PR?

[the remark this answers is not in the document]
Wovchena: I suspect @eizamaliev won't approve it quickly and the patch will hang here for some time, so yes, you should.

demos/common/python/openvino/model_zoo/model_api/models/classification.py
```diff
@@ -26,9 +27,30 @@ def __init__(self, ie, model_path, labels, ntop, resize_type='crop'):
...
+        function = ng.function_from_cnn(self.net)
+        if 'Softmax' not in [node.get_type_name() for node in function.get_ops()]:
+            logits = function.get_output_op(0)
+            logits = logits.inputs()[0].get_source_output().get_node()
+            logits.set_friendly_name('logits')
+            probs_node = ng.softmax(logits, 1)
+        else:
+            probs_node = function.get_output_op(0)
+            probs_node = probs_node.inputs()[0].get_source_output().get_node()
+            probs_node.set_friendly_name('probs')
```
Wovchena, line 44: I think we shouldn't set names: we don't use them

demos/common/python/openvino/model_zoo/model_api/models/classification.py
```diff
@@ -26,9 +27,30 @@ def __init__(self, ie, model_path, labels, ntop, resize_type='crop'):
...
+
+        topk_node = ng.topk(probs_node, ntop, 1, "max", "value")
```
Wovchena, line 46: I suggest keeping topk and added softmax independent. With that IE will have a chance to compute it in parallel, though I'm not sure it exploits that.

Wovchena (changes requested):

demos/common/python/openvino/model_zoo/model_api/models/classification.py
Wovchena: topk reduces number of elements in tensor. Never mind then

anzhella-pankratova:
Please, align the Running section in README with ArgumentParser args (remove --pause, add input transform parameters to .md file). It is also recommended to add at least one `TestCase` for ONNX models

demos/common/python/openvino/model_zoo/model_api/models/classification.py
```diff
@@ -0,0 +1,83 @@
...
+"""
+
+import ngraph as ng
+from openvino.inference_engine import IENetwork
+
+from .image_model import ImageModel
+
+
+class Classification(ImageModel):
+    def __init__(self, ie, model_path, labels, ntop, resize_type='crop'):
```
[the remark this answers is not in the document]
ivikhrev: Done.

demos/common/python/openvino/model_zoo/model_api/models/classification.py
```diff
@@ -0,0 +1,83 @@
...
+        super().__init__(ie, model_path, resize_type=resize_type)
+        self._check_io_number(1, 1)
+        if self.h != self.w:
+            raise RuntimeError('Model input has incorrect image shape. Must be NxN square. '
+                                'Got {} x {}.'.format(self.h, self.w))
```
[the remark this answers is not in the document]
ivikhrev: Removed.

demos/common/python/openvino/model_zoo/model_api/models/classification.py
```diff
@@ -0,0 +1,83 @@
+"""
+ Copyright (c) 2020 Intel Corporation
```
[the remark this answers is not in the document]
ivikhrev: Done.

demos/classification_demo/python/classification_demo.py
```diff
@@ -0,0 +1,273 @@
...
+        if async_pipeline.callback_exceptions:
+            raise async_pipeline.callback_exceptions[0]
+        # Process all completed requests
+        results = async_pipeline.get_result(next_frame_id_to_show)
+        if results:
+            classifications, frame_meta = results
+            frame = frame_meta['frame']
+            start_time = frame_meta['start_time']
+
+            if len(classifications) and args.raw_output_message:
```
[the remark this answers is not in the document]
ivikhrev: Ok, done

demos/common/python/openvino/model_zoo/model_api/models/classification.py
```diff
@@ -0,0 +1,78 @@
...
+
+        desc_order = scores.argsort()[::-1]
+        scores = scores[desc_order]
+        indices = indices[desc_order]
+        if self.apply_softmax:
+            scores = softmax(scores)
+        labels = [self.labels[i] if self.labels else "" for i in indices]
+        return list(zip(indices,
+                        labels,
+                        scores))
```
eizamaliev, line 78:
```suggestion
        return list(zip(indices, labels, scores))

```

eizamaliev (approved):

demos/classification_demo/python/classification_demo.py
```diff
@@ -0,0 +1,273 @@
...
+        if delay or args.no_show:
+            render_metrics.update(rendering_start_time)
+            metrics.update(start_time, frame)
+
+        if video_writer.isOpened() and (args.output_limit <= 0 or next_frame_id_to_show <= args.output_limit-1):
+            video_writer.write(frame)
+
+        if not args.no_show:
+            cv2.imshow('Classification Results', frame)
+            key = cv2.waitKey(delay)
```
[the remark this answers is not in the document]
anzhella-pankratova: bump

demos/classification_demo/python/classification_demo.py
```diff
@@ -0,0 +1,289 @@
...
+                    raise RuntimeError("Can't open video writer")
+            # Submit for inference
+            async_pipeline.submit_data(frame, next_frame_id, {'frame': frame, 'start_time': start_time})
+            next_frame_id += 1
+
+        else:
+            # Wait for empty request
+            async_pipeline.await_any()
+
+    if key not in {ord('q'), ord('Q'), ESC_KEY}:
```
anzhella-pankratova, line 244: What's the purpose of this check? I launched the demo with a video and then exit by escape, the demo always hangs out here

demos/classification_demo/python/classification_demo.py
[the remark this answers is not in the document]
ivikhrev: In this case tail of processed frames (if it was there) would fly by even if Esc wasn't pressed.
Changed it another way, I added condition and skip the tail only if exit button was pressed:
https://github.com/openvinotoolkit/open_model_zoo/blob/10a973b7e191326fffebb39861d516806c336e62/demos/classification_demo/python/classification_demo.py#L244

demos/classification_demo/python/classification_demo.py
ivikhrev: I'll recheck it

demos/classification_demo/python/classification_demo.py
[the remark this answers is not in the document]
anzhella-pankratova: Okay

demos/classification_demo/python/classification_demo.py
anzhella-pankratova: Probably the moving of `await_all()` before the condition resolves the issue

anzhella-pankratova (approved):

Wovchena (approved):
Just for ur future work, oversimplified version is as follows. Layer is a function. And inputs (openvino 2.0 calls them `Output`s) is what this function takes as argument. And returns outpts. In `Classification` class you deal with outputs, not with layers.

ivikhrev:
> Just for ur future work, oversimplified version is as follows. Layer is a function. And inputs (openvino 2.0 calls them `Output`s) is what this function takes as argument. And returns outpts. In `Classification` class you deal with outputs, not with layers.

@Wovchena  Ok, I'll keep this in mind.
But in this case,  `Classification`  class is wrapper for ModelAPI, and I called some variables there as layers just to stay consistent with model API naming:
https://github.com/openvinotoolkit/open_model_zoo/blob/c4023466b639330afed61864d0681ae8d47225d0/demos/common/python/openvino/model_zoo/model_api/models/model.py#L129

Does this make sense?

Wovchena:
It does. I've approved it after all

vladimir-dudnik (approved):

PR #2785, opened 2021-09-27
Raw comparison for reference

akorobeinikov:


tools/accuracy_checker/tests/test_presenter.py
```diff
@@ -657,37 +657,6 @@ def test_vector_presenter_with_vector_data_with_vector_scale(self, mocker):
         ]
         mock_write_scalar_res.assert_has_calls(calls)
 
-    def test_vector_presenter_with_vector_data_with_dict_ref(self, mocker):
```
eaidova, line 660: why do you remove test for testing reference represented as dictionary in the config?

akorobeinikov: I return this test, but I removed scaling for vector metric at all. Now scale using only for present metric. So when we calculate `mean` of vector metric we don't multiply values by scale.

eaidova (approved):

tools/accuracy_checker/openvino/tools/accuracy_checker/metrics/README.md
```diff
@@ -16,6 +16,14 @@ reference:
   std: 0.89
 ```
 
+If you want provide reference for metric expressed as a percentage, you should normalized reference value by `scale = 100`. For example, if you use accuracy metric for classification models, you should specify reference as:
```
eaidova, line 19:
```suggestion
*Note*: all reference values should be provided in raw format without scaling (metric specific, you can see raw metrics results using --ignore_results_formatting True parameter) e.g.
If you want provide reference for metric expressed as a percentage, you should normalized reference value by `scale = 100`. For example, if you use accuracy metric for classification models, you should specify reference as:
```

PR #2790, opened 2021-09-29
Model Quantizer: update pot import

jkamelin:


tools/model_tools/src/openvino/model_zoo/omz_quantizer.py
```diff
@@ -162,9 +162,9 @@ def main():
 
         pot_path = args.pot
         if pot_path is None:
-            if _common.get_package_path(args.python, 'pot'):
```
eaidova, line 165: should we preserve both old import way and new with try expect (atleast until openvino-dev has old import on pypi)?

eaidova (approved):

PR #2807, opened 2021-10-06
AC: remove all mentions about FPGA

eaidova:


tools/accuracy_checker/tests/test_dlsdk_launcher.py
```diff
@@ -1167,15 +979,15 @@ def setup(self):
 
     def test_hetero_correct(self):
         self.config.validate(update_dict(self.launcher, device='HETERO:CPU'))
-        self.config.validate(update_dict(self.launcher, device='HETERO:CPU,FPGA'))
+        self.config.validate(update_dict(self.launcher, device='HETERO:CPU,CPU'))
```
vladimir-dudnik, line 982:
```suggestion
        self.config.validate(update_dict(self.launcher, device='HETERO:CPU,GPU'))
```

do you mean CPU,GPU ?

vladimir-dudnik (commented):
question

tools/accuracy_checker/tests/test_dlsdk_launcher.py
eaidova: no, I wrote correctly. by default GPU can be not available in the system, this test do not run inference, so we do not need take care on device availability

vladimir-dudnik (approved):

PR #2809, opened 2021-10-06
Introduce sequence maximum length checking mode

mzuevx:
Introduce sequence maximum length checking mode, with padding to maximum length


tools/accuracy_checker/openvino/tools/accuracy_checker/annotation_converters/amazon.py
```diff
@@ -229,14 +229,31 @@ def configure(self):
...
+                    new_seqs_mid.append(inp[3])
+                    new_seqs_cat.append(inp[4])
+                    new_lengths_x.append(l_x)
+            lengths_x = new_lengths_x
+            seqs_mid = new_seqs_mid
+            seqs_cat = new_seqs_cat
 
         n_samples = len(seqs_mid)
         maxlen_x = np.max(lengths_x)
+        maxlen_x = max(maxlen, maxlen_x)
```
eaidova, line 256: this is dengerous, because maxlen can be None if it is not provided

PR #2810, opened 2021-10-06
extend test cases for VPU

vladimir-dudnik:
extend test cases for VPU to test optimized models


vladimir-dudnik:
cc @mlyashko @Maxim-Doronin

PR #2813, opened 2021-10-07
[Tests] Add testing of demos performance

anzhella-pankratova:
The PR adds the possibility to run the different scopes of test cases. The performance scope was created - after the demo execution the `PerformanceParser` collects the performance metrics, such as latency and FPS, and writes it to the csv file in the following format:  

"Device, ModelName, Latency, FPS, Decoding, Preprocessing, Inference, Postprocessing, Rendering, Requests, Streams, Threads"

The example for object_detection_demo.py:

 example

To update test cases of some demo for performance scope, one can inherit the all base test cases of the demo, left test cases only with desired models (via `only_models`, `exclude_models` methods), update some option (for example, update `-i` parameter,  because we need to test demos on the long data sequences) and add options '**nireq**', '**nstreams**', '**nthreads**' to test cases.


vladimir-dudnik:
Jenkins please retry a build

vladimir-dudnik:
@Wovchena are you ok with this?

Wovchena (approved):
Ok

PR #2854, opened 2021-10-31
Add scipy to requirements of modelAPI

akorobeinikov:
Scipy is needed for HPE model

### DIFF
--- a/demos/common/python/requirements.txt
+++ b/demos/common/python/requirements.txt
@@ -1,3 +1,3 @@
 numpy==1.19.5
 opencv-python==4.5.*
-scipy==1.7.1
+scipy~=1.5.4
--- a/demos/common/python/requirements.txt
+++ b/demos/common/python/requirements.txt
@@ -1,2 +1,3 @@
 numpy==1.19.5
 opencv-python==4.5.*
+scipy==1.7.1


demos/common/python/requirements.txt
```diff
@@ -1,2 +1,3 @@
 numpy==1.19.5
 opencv-python==4.5.*
+scipy==1.7.1
```
eaidova, line 3: @akorobeinikov  could you please explain why it is required for model api?

eaidova: ok, I found https://github.com/openvinotoolkit/open_model_zoo/blob/develop/demos/common/python/openvino/model_zoo/model_api/models/hpe_associative_embedding.py#L18

eaidova: fyi @eizamaliev @anzhella-pankratova

eaidova: there is common dependency resolving policy for all OV components, please use the same scipy version, which used in 
https://github.com/openvinotoolkit/open_model_zoo/blob/develop/tools/accuracy_checker/requirements.in#L25
and https://github.com/openvinotoolkit/openvino/blob/master/tools/pot/setup.py#L44

eaidova (approved):

PR #2789, opened 2021-09-28
Fixed pylint(2.10.2) errors in AC files

Anna-Mironova:
This PR includes fixes for linter errors in accuracy checker files since upgrade to pylint version 2.10.2

PR #2894, opened 2021-11-17
AC: update profiling

eaidova:


Anna-Mironova (approved):

Anna-Mironova:
It seems we need to add changes from https://github.com/openvinotoolkit/open_model_zoo/pull/2789

eaidova:
> It seems we need to add changes from #2789 to release branch

I started, but looks like it contains a lot of significant changes, will be done as separated request

PR #2910, opened 2021-11-20
correct output format for classification models

vladimir-dudnik:
check and correct description for classification model's output format (probability in [0, 1] range and logits format)


vladimir-dudnik:
cc @jkamelin @Anna-Mironova @Wovchena

eaidova (approved):

PR #1161, opened 2020-05-29
Contribute the handwritten-simplified-chinese recognition-0001 model

cocokale:
We want to contribute the handwritten-simplified-chinese recognition-0001 model. 

Including:
- Its markdown format documentation,
- Its configuration file of accuracy checker and also the corresponding update in dataset_definitions.yml
- Integration of the model's demo into the previous handwritten-japanese-recognition demo (by renaming and updating the model.lst and readme)
- Renaming the existing annotation converter to a more general one.
- Some other updates in several READMEs, e.g. the one in demos/, etc.

Note: the dataset's license has not been finalized yet. Currently we are not merging it.

Thanks!

PR #2917, opened 2021-11-23
Contribute the handwritten-english-recognition-0001 model

jnzw:
Hi, 

We'd like to add a model for handwritten English text recognition. This pull request builds upon previous works for the Japanese (#869) and Chinese (#1161) text recognition models.


demos/handwritten_text_recognition_demo/python/handwritten_text_recognition_demo.py
```diff
@@ -83,10 +83,9 @@ def main():
     net = ie.read_network(args.model, os.path.splitext(args.model)[0] + ".bin")
 
     assert len(net.input_info) == 1, "Demo supports only single input topologies"
-    assert len(net.outputs) == 1, "Demo supports only single output topologies"
 
     input_blob = next(iter(net.input_info))
-    out_blob = next(iter(net.outputs))
+    out_blob = 'output' # name of output blob of model
```
Wovchena, line 88: I guess these are debug changes. Revert them.

jnzw: It wasn't a debug change, actually. The model uses 5 layers of bi-LSTM so after converting to IR, it has 10 extra LSTM hidden cell outputs as well, and I'm not sure how to discard/suppress those extra outputs. The only two methods I can think of is either calling `next` 10 times on `iter(net.outputs)` or hardcoding the actual output name, as shown in line 88. If there's a better way to handle this, could you please advice? Thanks.
image

jnzw: I've updated `output_blob` to be a command line argument.

Wovchena: Can you update 'models/intel/handwritten-english-recognition-0001/README.md' to mention more than 1 output

data/dataset_classes/gnhk.txt
```diff
@@ -0,0 +1,94 @@
+
```
eaidova, line 1: not sure that it is dataset classes, more looks like model vocabulary file

demos/handwritten_text_recognition_demo/python/data/test_img_id_gt.txt
```diff
@@ -0,0 +1,2189 @@
+eng_AF_004_line0.jpg,DFI Rocks
```
eaidova, line 1: we do not store annotations as part of demos

models/intel/handwritten-english-recognition-0001/accuracy-check.yml
```diff
@@ -0,0 +1,33 @@
...
+            size: 96
+          - type: padding
+            use_numpy: True
+            numpy_pad_mode: edge
+            dst_height: 96
+            dst_width: 2000
+            pad_type: right_bottom
+        #  - type: normalization
+        #    mean: 127.5
+        #    std: 127.5
```
eaidova, line 29: please remove these lines

models/intel/handwritten-english-recognition-0001/README.md
```diff
@@ -0,0 +1,52 @@
...
+## Specification
+
+| Metric                    | Value     |
+| ------------------------- | --------- |
+| GFlops                    | 1.3182    |
+| MParams                   | 0.1413    |
+| Accuracy on GNHK test subset (excluding images wider than 2000px after resized to height 96px with aspect ratio) | 81.5%     |
+| Source framework          | PyTorch\* |
+
+> **Note:** to achieve the accuracy, images from the GNHK test set should be preprocessed into single-line text images, using the coordinates from the accompanying JSON annotation files in the GNHK dataset, and then binarized using adaptive thresholding.
```
eaidova, line 21: is it possible to link some scripts for preprocessing?

models/intel/handwritten-english-recognition-0001/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Accuracy Values
+
+This model adopts [label error rate](https://dl.acm.org/doi/abs/10.1145/1143844.1143891) as the metric for accuracy.
```
eaidova, line 25: probably it should be a comment to achived accuracy metric, no need in separated paragraph

models/intel/handwritten-english-recognition-0001/accuracy-check.yml
```diff
@@ -0,0 +1,33 @@
...
+
+        metrics:
+          - type: label_level_recognition_accuracy
+
```
eaidova, line 33: please provide metric reference in accuracy config

data/dataset_classes/gnhk.txt
jnzw: Hi, this file contains the list of characters (the predicted classes) supported by the model. The previous models for Chinese and Japanese also have similar files under `dataset_classes`: [scut_ept.txt](https://github.com/openvinotoolkit/open_model_zoo/blob/master/data/dataset_classes/scut_ept.txt) and [kondate_nakayosi.txt](https://github.com/openvinotoolkit/open_model_zoo/blob/master/data/dataset_classes/kondate_nakayosi.txt), respectively.

demos/handwritten_text_recognition_demo/python/handwritten_text_recognition_demo.py
jnzw: The README has been updated.

data/dataset_classes/gnhk.txt
eaidova: looks like it was wrong decision, because these files have more relation to model, then dataset by itself (somebody can train model with another symblos order or include/exclude symblos on the same datasets and it also does not mean that it works only for these dataset, you can use them for decoding on different data).
But looks like transition should be done later, for all affected models

demos/handwritten_text_recognition_demo/python/data/test_img_id_gt.txt
jnzw: Removed.

models/intel/handwritten-english-recognition-0001/accuracy-check.yml
jnzw: Removed.

models/intel/handwritten-english-recognition-0001/accuracy-check.yml
jnzw: Reference metric added.

models/intel/handwritten-english-recognition-0001/README.md
jnzw: Updated.

models/intel/handwritten-english-recognition-0001/README.md
jnzw: I've added the script `preprocess_gnhk.py` under `demos/`.

eaidova: this script does not have relation to demos, please move it to directory with your model, like here 
https://github.com/openvinotoolkit/open_model_zoo/blob/master/models/public/mozilla-deepspeech-0.8.2/pbmm_to_pb.py

demos/README.md
```diff
@@ -26,7 +26,7 @@ The Open Model Zoo includes the following demos:
 - [Gaze Estimation C++ G-API Demo](./gaze_estimation_demo/cpp_gapi/README.md) - Face detection followed by gaze estimation, head pose estimation and facial landmarks regression. G-API version.
 - [Gesture Recognition Python\* Demo](./gesture_recognition_demo/python/README.md) - Demo application for Gesture Recognition algorithm (e.g. American Sign Language gestures), which classifies gesture actions that are being performed on input video.
 - [GPT-2 Text Prediction Python\* Demo](./gpt2_text_prediction_demo/python/README.md) - GPT-2 text prediction demo.
-- [Handwritten Text Recognition Python\* Demo](./handwritten_text_recognition_demo/python/README.md) - The demo demonstrates how to run Handwritten Japanese Recognition models and Handwritten Simplified Chinese Recognition models.
+- [Handwritten Text Recognition Python\* Demo](./handwritten_text_recognition_demo/python/README.md) - The demo demonstrates how to run Handwritten Text Recognition models for Japanese, Simplified Chinese, and English.
```
eaidova, line 29:
```suggestion
- [Handwritten Text Recognition Python\* Demo](./handwritten_text_recognition_demo/python/README.md) - The demo demonstrates how to run Handwritten Text Recognition models for Japanese, Simplified Chinese and English.
```

demos/handwritten_text_recognition_demo/python/handwritten_text_recognition_demo.py
```diff
@@ -48,6 +48,7 @@ def build_argparser():
                       help="Path to the decoding char list file. Default is for Japanese")
     args.add_argument("-dc", "--designated_characters", type=str, default=None, help="Optional. Path to the designated character file")
     args.add_argument("-tk", "--top_k", type=int, default=20, help="Optional. Top k steps in looking up the decoded character, until a designated one is found")
+    args.add_argument("-ob", "--output_blob", type=str, default="output", help="Optional. Name of the output layer of the model. Default is \'output\'")
```
eaidova, line 51: I think it will be better if default behavior to get output from model instead hardcoding name (it preserves previous demo behavior and allow to run custom models in the demo, where single output, but name is different)

demos/handwritten_text_recognition_demo/python/handwritten_text_recognition_demo.py
```diff
@@ -77,16 +78,16 @@ def main():
     log.info('OpenVINO Inference Engine')
     log.info('\tbuild: {}'.format(get_version()))
     ie = IECore()
+    ie.set_config(config={"GPU_ENABLE_LOOP_UNROLLING": "NO", "CACHE_DIR": "./"}, device_name="GPU")
```
eaidova, line 81: why this config become required for running demo? Why you use current directory as cache?

jnzw: Hi, this config is to speed up loading the network on GPUs, without the config, loading the model is very time consuming.

eaidova: but, what if user does not have intel gpu on running machine? I think it will be better to enable it only if GPU selected as running device

Wovchena: That would be hard to handle all cases when gpu may be used with this auto plugin and stuff. `device_name` arg already sets it only for gpu. I would leave it as it is for now.

models/intel/handwritten-english-recognition-0001/README.md
jnzw: Moved.

demos/handwritten_text_recognition_demo/python/handwritten_text_recognition_demo.py
jnzw: I've updated the script to reuse previous code unless this value is set to not `None`.

Wovchena (approved):

eaidova (approved):

PR #2920, opened 2021-11-23
AC: support different image types in voc segmentation

eaidova:


tools/accuracy_checker/openvino/tools/accuracy_checker/annotation_converters/pascal_voc.py
```diff
@@ -14,6 +14,7 @@
 limitations under the License.
 """
 
+from pathlib import Path
```
eaidova, line 17:
```suggestion
```

Anna-Mironova (approved):

PR #2953, opened 2021-12-02
chmod for preprocess

vladimir-dudnik:
+x to preprocess_gnhk.py


vladimir-dudnik:
@nignatovsky trailing space check should be fixed, we got a case where we have only single character in line and character is SPACE. There should be way to allow this.
19:10:20  data/dataset_classes/gnhk.txt:1: trailing whitespace.

Wovchena:
> @nignatovsky trailing space check should be fixed, we got a case where we have only single character in line and character is SPACE. There should be way to allow this. 19:10:20 data/dataset_classes/gnhk.txt:1: trailing whitespace.

Done https://github.com/openvinotoolkit/open_model_zoo/pull/2954

eaidova (approved):

PR #2966, opened 2021-12-06
model_tools: update composite models support

jkamelin:


jkamelin:
Jenkins please retry a build

models/intel/action-recognition-0001/composite-model.yml
```diff
@@ -1 +1,6 @@
+description: >-
+  General-purpose action recognition model for Kinetics-400 dataset based on Video
+  Transformer Network approach. Encoder par
```
eaidova, line 3: looks like description is broken
```suggestion
  Transformer Network approach.
```

tools/model_tools/src/openvino/model_zoo/_configuration.py
```diff
@@ -192,15 +234,38 @@ def load_models(models_root, args):
...
-            if composite_model_name in composite_models:
-                raise validation.DeserializationError(
-                    'Duplicate composite model name "{}"'.format(composite_model_name))
-            composite_models.append(composite_model_name)
+            with composite_model_config.open('rb') as config_file, \
+                validation.deserialization_context('In config "{}"'.format(composite_model_config)):
+
+                composite_model = yaml.safe_load(config_file)
+                model_stages = {}
+                for stage in sorted(composite_model_config.parent.glob('*/model.yml')):
```
eaidova, line 242: is it possible to have composition of model composition in future? :)
if yes, rglob usage will be preferable instead of glob

models/intel/action-recognition-0001/composite-model.yml
jkamelin: Done

tools/model_tools/src/openvino/model_zoo/_configuration.py
jkamelin: Right now it is not allowed https://github.com/openvinotoolkit/open_model_zoo/blob/master/tools/model_tools/src/openvino/model_zoo/_configuration.py#L164. I think composition of composition models would require more effort than using rglob, probably it's better to do when we have one.

eaidova: pk, then keep as is

eaidova:
@nignatovsky  could yo please update CI script parsing logic?

nignatovsky:
> @nignatovsky could yo please update CI script parsing logic?

Done.

vladimir-dudnik:
Jenkins please retry a built

vladimir-dudnik:
Jenkins please retry a build

vladimir-dudnik (approved):

eaidova (approved):

PR #2976, opened 2021-12-09
model_tools: fix subdirectory for composite models

jkamelin:
Make subdirectory for composite models relative to model root, same as for regular models.


vladimir-dudnik (approved):

vladimir-dudnik:
cc @ematroso

just-sparta (approved):

PR #2893, opened 2021-11-16
OV2.0  c++ demos/multi_channel: move to new API, set PERFORMANCE_HINT THROUGHPUT, remove -nireq

Wovchena:
The demo is designed to be used in throughput mode. That allows to
ignore latency scenarios and to simplify usage by guessing -nireq.

There are bugs in openvino affecting
multi_channel_object_detection_demo_yolov3, so don’t expect it to work
now.

PR #2547, opened 2021-06-25
cpp demos logging unification

ivikhrev:
Unified demos logging. The following stages are left:

1. IE version
2. Model loading completion (with name and device)
3. IE parameters (num of streams, threads, requests)
4. Perfomance metrics.

Also done some improvements for text visualization (replaced putText with putHighlitedText in several places).

PR #3003, opened 2021-12-20
OV2.0 c++ noise suppression

vladimir-dudnik:
OV2.0 in noise_suppression_demo

### DIFF
--- a/demos/common/cpp/utils/include/utils/common.hpp
+++ b/demos/common/cpp/utils/include/utils/common.hpp
@@ -252,7 +252,7 @@ inline void logExecNetworkInfo(const InferenceEngine::ExecutableNetwork& execNet
 }
 
 inline
-void log_compiled_model_info(
+void logCompiledModelInfo(
     const ov::runtime::CompiledModel& compiledModel,
     const std::string& modelName,
     const std::string& deviceName,
@@ -281,7 +281,7 @@ void log_compiled_model_info(
 }
 
 inline
-void log_model_info(const std::shared_ptr<ov::Model>& model) {
+void logBasicModelInfo(const std::shared_ptr<ov::Model>& model) {
     slog::info << "model name: " << model->get_friendly_name() << slog::endl;
 
     // Prepare input blobs
--- a/demos/common/cpp/utils/include/utils/common.hpp
+++ b/demos/common/cpp/utils/include/utils/common.hpp
@@ -23,6 +23,7 @@
 #include <iostream>
 
 #include <inference_engine.hpp>
+#include "openvino/openvino.hpp"
 #include "utils/slog.hpp"
 #include "utils/args_helper.hpp"
 
@@ -248,4 +249,70 @@ inline void logExecNetworkInfo(const InferenceEngine::ExecutableNetwork& execNet
             catch (const InferenceEngine::Exception&) {}
         }
     }
+}
+
+inline
+void log_compiled_model_info(
+    const ov::runtime::CompiledModel& compiledModel,
+    const std::string& modelName,
+    const std::string& deviceName,
+    const std::string& modelType = "")
+{
+    slog::info << "The " << modelType << (modelType.empty() ? "" : " ") << "model " << modelName << " is loaded to " << deviceName << slog::endl;
+    std::set<std::string> devices;
+    for (const std::string& device : parseDevices(deviceName)) {
+        devices.insert(device);
+    }
+
+    if (devices.find("AUTO") == devices.end()) { // do not print info for AUTO device
+        for (const auto& device : devices) {
+            try {
+                slog::info << "\tDevice: " << device << slog::endl;
+                std::string nstreams = compiledModel.get_config(device + "_THROUGHPUT_STREAMS").as<std::string>();
+                slog::info << "\t\tNumber of streams: " << nstreams << slog::endl;
+                if (device == "CPU") {
+                    std::string nthreads = compiledModel.get_config("CPU_THREADS_NUM").as<std::string>();
+                    slog::info << "\t\tNumber of threads: " << (nthreads == "0" ? "AUTO" : nthreads) << slog::endl;
+                }
+            }
+            catch (const ov::Exception&) {}
+        }
+    }
+}
+
+inline
+void log_model_info(const std::shared_ptr<ov::Model>& model) {
+    slog::info << "model name: " << model->get_friendly_name() << slog::endl;
+
+    // Prepare input blobs
+
+    // Taking information about all topology inputs
+    ov::OutputVector inputs = model->inputs();
+    ov::OutputVector outputs = model->outputs();
+
+    for (const ov::Output<ov::Node> input : inputs)
+    {
+        slog::info << slog::endl;
+        const std::string name = input.get_any_name();
+        const ov::element::Type type = input.get_element_type();
+        const ov::Shape shape = input.get_shape();
+
+        slog::info << "input name: " << name << slog::endl;
+        slog::info << "input type: " << type << slog::endl;
+        slog::info << "input shape: " << shape << slog::endl;
+    }
+
+    for (const ov::Output<ov::Node> output : outputs)
+    {
+        slog::info << slog::endl;
+        const std::string name = output.get_any_name();
+        const ov::element::Type type = output.get_element_type();
+        const ov::Shape shape = output.get_shape();
+
+        slog::info << "output name: " << name << slog::endl;
+        slog::info << "output type: " << type << slog::endl;
+        slog::info << "output shape: " << shape << slog::endl;
+    }
+
+    return;
 }
--- a/demos/common/cpp/utils/include/utils/common.hpp
+++ b/demos/common/cpp/utils/include/utils/common.hpp
@@ -1,4 +1,4 @@
-// Copyright (C) 2018-2019 Intel Corporation
+// Copyright (C) 2018-2022 Intel Corporation
 // SPDX-License-Identifier: Apache-2.0
 //
 
@@ -295,7 +295,7 @@ void log_model_info(const std::shared_ptr<ov::Model>& model) {
         slog::info << slog::endl;
         const std::string name = input.get_any_name();
         const ov::element::Type type = input.get_element_type();
-        const ov::Shape shape = input.get_shape();
+        const ov::PartialShape shape = input.get_partial_shape();
 
         slog::info << "input name: " << name << slog::endl;
         slog::info << "input type: " << type << slog::endl;
@@ -307,7 +307,7 @@ void log_model_info(const std::shared_ptr<ov::Model>& model) {
         slog::info << slog::endl;
         const std::string name = output.get_any_name();
         const ov::element::Type type = output.get_element_type();
-        const ov::Shape shape = output.get_shape();
+        const ov::PartialShape shape = output.get_partial_shape();
 
         slog::info << "output name: " << name << slog::endl;
         slog::info << "output type: " << type << slog::endl;
--- a/demos/common/cpp/utils/include/utils/common.hpp
+++ b/demos/common/cpp/utils/include/utils/common.hpp
@@ -284,34 +284,28 @@ inline
 void logBasicModelInfo(const std::shared_ptr<ov::Model>& model) {
     slog::info << "model name: " << model->get_friendly_name() << slog::endl;
 
-    // Prepare input blobs
-
-    // Taking information about all topology inputs
+    // Dump information about model inputs/outputs
     ov::OutputVector inputs = model->inputs();
     ov::OutputVector outputs = model->outputs();
 
+    slog::info << "inputs: " << slog::endl;
     for (const ov::Output<ov::Node> input : inputs)
     {
-        slog::info << slog::endl;
         const std::string name = input.get_any_name();
         const ov::element::Type type = input.get_element_type();
         const ov::PartialShape shape = input.get_partial_shape();
 
-        slog::info << "input name: " << name << slog::endl;
-        slog::info << "input type: " << type << slog::endl;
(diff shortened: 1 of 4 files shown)


vladimir-dudnik:
Jenkins please retry a build

vladimir-dudnik:
Jenkins please retry a build

vladimir-dudnik:
Jenkins please retry a build

demos/common/cpp/utils/include/utils/common.hpp
```diff
@@ -249,3 +250,69 @@ inline void logExecNetworkInfo(const InferenceEngine::ExecutableNetwork& execNet
         }
     }
 }
+
+inline
+void log_compiled_model_info(
```
Wovchena, line 255: I think my way is better :)
https://github.com/openvinotoolkit/open_model_zoo/pull/2893/files#diff-746af3a0af0e254bedb0fa42f1604fc533a856185093fb9db870a2cf951e395bR236

demos/common/cpp/utils/include/utils/common.hpp
```diff
@@ -249,3 +250,69 @@ inline void logExecNetworkInfo(const InferenceEngine::ExecutableNetwork& execNet
...
+                    slog::info << "\t\tNumber of threads: " << (nthreads == "0" ? "AUTO" : nthreads) << slog::endl;
+                }
+            }
+            catch (const ov::Exception&) {}
+        }
+    }
+}
+
+inline
+void log_model_info(const std::shared_ptr<ov::Model>& model) {
```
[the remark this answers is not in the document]
vladimir-dudnik: yes, I find it useful to dump basic model info just after we read it. for compiled models we usually dump only device configuration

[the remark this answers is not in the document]
Wovchena: @fzhar, @ivikhrev, @anzhella-pankratova, I have perception, that the decision was not to have it. Is this true?

demos/common/cpp/utils/include/utils/common.hpp
vladimir-dudnik: well, reviewed your implementation. I do not think we need involve templates to keep old func name `logExecNetworkInfo` with new API. Better to make clear difference from old API.

demos/noise_suppression_demo/cpp/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-# Copyright (C) 2018-2019 Intel Corporation
+# Copyright (C) 2021 Intel Corporation
```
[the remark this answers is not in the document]
vladimir-dudnik: changed

demos/noise_suppression_demo/cpp/main.cpp
```diff
@@ -116,131 +118,155 @@ void read_wav(const std::string& file_name, std::vector<int16_t>& wave, RiffWave
...
-        std::vector<std::pair<std::string, std::string>> state_names;
+        ov::OutputVector inputs = model->inputs();
+        ov::OutputVector outputs = model->outputs();
+
+        // get state names pairs (inp,out) and compute overall states size
         size_t state_size = 0;
-        for(auto& inp: inputs) {
-            std::string inp_state_name = inp.first;
-            if (inp_state_name.find("inp_state") == std::string::npos)
+        std::regex state_regex{ "^(inp|out)(_state_)(\\d\\d\\d)" };
```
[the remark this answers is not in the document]
vladimir-dudnik: reverted

demos/common/cpp/utils/include/utils/common.hpp
ivikhrev: I suggest to adhere camel case naming for functions, as we always do -` logCompiledModelInfo`.

demos/noise_suppression_demo/cpp/main.cpp
```diff
@@ -116,131 +117,152 @@ void read_wav(const std::string& file_name, std::vector<int16_t>& wave, RiffWave
...
                     // first iteration. set input state to zero tensor.
-                    const InferenceEngine::TensorDesc &state_desc = inputs[inp_state_name]->getInputData()->getTensorDesc();
-                    auto blob_ptr = InferenceEngine::make_shared_blob<float>(state_desc);
-                    blob_ptr->allocate();
-                    for (auto &val: *blob_ptr) {
-                        val = 0;
-                    }
-                    infer_request.SetBlob(inp_state_name,blob_ptr);
+                    ov::Shape state_shape = model->input(inp_state_name).get_shape();
+                    ov::runtime::Tensor blob_ptr(ov::element::f32, state_shape);
```
ivikhrev, line 238: Are we going to use snake case now?  Because previously we stuck to camel case.

vladimir-dudnik: OpenVINO 2.0 switched to snake case, so there will be mix in any case, although switch in OpenVINO does not dictate immediate changes for demos, but in general, it would be better to be consistent.

vladimir-dudnik: BTW, in OV2.0 there is no such matter as blobs, so rename it to tensor

demos/common/cpp/utils/include/utils/common.hpp
vladimir-dudnik: ok, renamed functions to logBasicModelInfo and logCompiledModelInfo, as change in style would be to massive for OMZ and not well justified.

demos/common/cpp/utils/include/utils/common.hpp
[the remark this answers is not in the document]
anzhella-pankratova: Actually, I don't remember the reason, why it was not introduced along with Python implementation, but this is an useful option. 
 @vladimir-dudnik could you please align logging messages with [Python version](https://github.com/openvinotoolkit/open_model_zoo/blob/master/demos/common/python/openvino/model_zoo/model_api/models/model.py#L198)? The info per layer should be printed on the one line, not splitted into 3 lines.

demos/noise_suppression_demo/cpp/main.cpp
```diff
@@ -116,131 +117,152 @@ void read_wav(const std::string& file_name, std::vector<int16_t>& wave, RiffWave
...
+        slog::info << ov::get_openvino_version() << slog::endl;
+
         // Loading Inference Engine
-        InferenceEngine::Core ie;
+        ov::runtime::Core core;
 
-        InferenceEngine::CNNNetwork network = ie.ReadNetwork(FLAGS_m);
-        InferenceEngine::InputsDataMap inputs = network.getInputsInfo();
+        std::shared_ptr<ov::Model> model = core.read_model(FLAGS_m);
+        slog::info << "model file: " << FLAGS_m << slog::endl;
```
anzhella-pankratova, line 148:
```suggestion
        slog::info << "Reading model " << FLAGS_m << slog::endl;
```

demos/noise_suppression_demo/cpp/main.cpp
```diff
@@ -116,131 +117,152 @@ void read_wav(const std::string& file_name, std::vector<int16_t>& wave, RiffWave
...
+            if (inp_state_name.find("inp_state_") == std::string::npos)
                 continue;
+
             std::string out_state_name(inp_state_name);
             out_state_name.replace(0, 3, "out");
+
+            // find corresponding output state
+            auto scmp = [&](ov::Output<ov::Node> output) { return output.get_any_name() == out_state_name; };
+            if (std::end(outputs) == std::find_if(outputs.begin(), outputs.end(), scmp))
+                throw std::logic_error("model output state name not corresponf input state name");
```
anzhella-pankratova, line 168:
```suggestion
                throw std::logic_error("model output state name does not correspond input state name");
```

demos/common/cpp/utils/include/utils/common.hpp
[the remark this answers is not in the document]
ivikhrev: Initially we agreed only to this logging stages (https://github.com/openvinotoolkit/open_model_zoo/pull/2547):

> 1. IE version
> 2. Model loading completion (with name and device)
> 3. IE parameters (num of streams, threads, requests)
> 4. Perfomance metrics.

Argument against the stage with logging information about input/output layers - that this would be too excessive for demos. In python this stage was introduced after we done logging unification, and I don't remember discussions about it.

demos/noise_suppression_demo/cpp/main.cpp
```diff
@@ -116,131 +117,152 @@ void read_wav(const std::string& file_name, std::vector<int16_t>& wave, RiffWave
...
-                    auto blob_ptr = InferenceEngine::make_shared_blob<float>(state_desc);
-                    blob_ptr->allocate();
-                    for (auto &val: *blob_ptr) {
-                        val = 0;
-                    }
-                    infer_request.SetBlob(inp_state_name,blob_ptr);
+                    ov::Shape state_shape = model->input(inp_state_name).get_shape();
+                    ov::runtime::Tensor state_tensor(ov::element::f32, state_shape);
+                    memset(state_tensor.data<float>(), 0, state_tensor.get_byte_size());
+                    infer_request.set_tensor(inp_state_name, state_tensor);
```
anzhella-pankratova, line 240: Setting tensor should be out of the condition

demos/noise_suppression_demo/cpp/main.cpp
```diff
@@ -116,131 +117,152 @@ void read_wav(const std::string& file_name, std::vector<int16_t>& wave, RiffWave
...
-        InferenceEngine::InferRequest infer_request = executable_network.CreateInferRequest();
+
+        if (state_size == 0)
+            throw std::logic_error("no expected model state inputs found");
+
+        std::cout << "State_param_num = " << state_size << " (" << state_size * 4e-6f << "Mb)" << std::endl;
+
+        // sort names
+        auto scmp = [](const std::pair<std::string, std::string>& a, const std::pair<std::string,std::string>& b) { return a.first < b.first; };
+        std::sort(state_names.begin(), state_names.end(), scmp);
```
anzhella-pankratova, line 187: Does the model will work correctly without sorting the names? I suppose so. After the inference the `inp_state_*` is filled by the specific `out_state_*` and the order is does not matter, because the sequential number of `inp_state_*` is compared with its shape (the states have different shapes)

vladimir-dudnik: did not get what you are saying

demos/common/cpp/utils/include/utils/common.hpp
```diff
@@ -249,3 +250,69 @@ inline void logExecNetworkInfo(const InferenceEngine::ExecutableNetwork& execNet
...
+            catch (const ov::Exception&) {}
+        }
+    }
+}
+
+inline
+void logBasicModelInfo(const std::shared_ptr<ov::Model>& model) {
+    slog::info << "model name: " << model->get_friendly_name() << slog::endl;
+
+    // Prepare input blobs
```
Wovchena, line 287: You don't modify them thus it's not preparation

demos/noise_suppression_demo/cpp/CMakeLists.txt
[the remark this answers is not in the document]
Wovchena: It's still 2021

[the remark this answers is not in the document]
vladimir-dudnik: there were no changes in CMakeLists.txt in 2022. Anyway, if we are going to follow openvino approach and change copyright year in each and every file (what we did not in the past), then it is better to do at one step in separate PR to all files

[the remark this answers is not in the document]
vladimir-dudnik: oh, I see your point, this is new file, so year should reflect actual commit time. yeah, will change

demos/noise_suppression_demo/cpp/main.cpp
vladimir-dudnik: @anzhella-pankratova what you are trying to achieve with this change? The reading is continuous process, while we dump info for the model which already read (we have no physical capability to dump info during the reading is not it?).

demos/noise_suppression_demo/cpp/CMakeLists.txt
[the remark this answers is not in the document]
vladimir-dudnik: changed to 2022

demos/noise_suppression_demo/cpp/main.cpp
vladimir-dudnik: got it, changed

demos/noise_suppression_demo/cpp/main.cpp
vladimir-dudnik: yes, moved out of condition

demos/noise_suppression_demo/cpp/main.cpp
vladimir-dudnik: that's correct, input state should correspond to output state, but the order in which output states from previous inference copied to input state for the next  inference does not matter while all states copied, so sorting removed

demos/common/cpp/utils/include/utils/common.hpp
vladimir-dudnik: yes, corrected

demos/noise_suppression_demo/cpp/main.cpp
anzhella-pankratova: The `infer_request.set_tensor(inp_state_name, state_tensor)` in else block is not removed

vladimir-dudnik: yeah, good catch. fixed

anzhella-pankratova (approved):

PR #3008, opened 2021-12-21
Merge releases/2021/4 into master

Wovchena:


Wovchena:
Strangely, when I compare master and releases/2021/4 github shows there are conflicts and 4 commits difference, but here it shows only one commit.

eaidova:
> Strangely, when I compare master and releases/2021/4 github shows there are conflicts and 4 commits difference, but here it shows only one commit.

I see the same, but probably, it is my squashed commits porting bugfixes for DL Workbench from master to 2021.4

eaidova (approved):

PR #3011, opened 2021-12-22
AC: Fixed set_input_and_output in case exec_network is not set

Anna-Mironova:


tools/accuracy_checker/openvino/tools/accuracy_checker/evaluators/custom_evaluators/base_models.py
```diff
@@ -195,13 +195,18 @@ def automatic_model_search(self, network_info):
...
 
     def set_input_and_output(self):
-        has_info = hasattr(self.exec_network, 'input_info')
-        input_info = self.exec_network.input_info if has_info else self.exec_network.inputs
+        if self.exec_network is not None:
+            has_info = hasattr(self.exec_network, 'input_info')
+            input_info = self.exec_network.input_info if has_info else self.exec_network.inputs
+        else:
+            has_info = hasattr(self.network, 'input_info')
+            input_info = self.network.input_info if has_info else self.network.inputs
```
eaidova, line 203:
```suggestion
        network self.exec_network if self.exec_network is not None else network
        has_info = hasattr(network, 'input_info')
        input_info = network.input_info if has_info else network.inputs
        output_blob = next(iter(network.outputs))
```

Anna-Mironova: Done

eaidova (approved):

PR #3013, opened 2021-12-22
OV2.0 python action_recognition_demo

ivikhrev:
migrate python action_recognition demo to OV2.0 API

### DIFF
--- a/demos/action_recognition_demo/python/action_recognition_demo.py
+++ b/demos/action_recognition_demo/python/action_recognition_demo.py
@@ -57,9 +57,6 @@ def build_argparser():
     decoder_args.add_argument('--seq', dest='decoder_seq_size',
                               help='Optional. Length of sequence that decoder takes as input.',
                               default=16, type=int)
-    args.add_argument('-l', '--cpu_extension',
-                      help='Optional. For CPU custom layers, if any. Absolute path to a shared library with the '
-                           'kernels implementation.', type=str, default=None)
     args.add_argument('-d', '--device',
                       help='Optional. Specify a target device to infer on. CPU, GPU, HDDL or MYRIAD is '
                            'acceptable. The demo will look for a suitable plugin for the device specified. '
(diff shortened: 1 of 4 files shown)


ivikhrev:
Jenkins please retry a build

demos/action_recognition_demo/python/action_recognition_demo/models.py
```diff
@@ -73,41 +75,48 @@ def infer(self, model_input, frame=None):
...
+
         self.num_requests = num_requests
+        self.infer_requests = [self.compiled_model.create_infer_request() for _ in range(self.num_requests)]
         log.info('The {} model {} is loaded to {}'.format(model_type, model_path, target_device))
 
     def infer(self, frame):
         input_data = {self.input_name: frame}
-        infer_result = self.exec_net.infer(input_data)
+        infer_result = self.compiled_model.infer(input_data)
         return infer_result[self.output_name]
```
anzhella-pankratova, line 110:
```suggestion
        return next(iter(self.infer_requests[0].infer(input_data).values()))
```
and `self.output_name` should be removed in this case

ivikhrev: You suggest to emulate sync infer via async api with 1 ireq, it can be done with `async_infer` func already, so there is no need to change this.
But maybe functions for sync inference should be removed at all, because it isn't used in this demo.

anzhella-pankratova: Yeah, I thought about it. I suppose it was implemented in order to call sync model inference via `class ActionRecognitionSequential`, although the demo does not provide an option to run it in a synchronous mode.
So, I agree with you, the `infer` method with the corresponding class might be removed at all.

anzhella-pankratova (approved):

vladimir-dudnik:
Jenkins please retry a build

demos/action_recognition_demo/python/action_recognition_demo.py
```diff
@@ -88,34 +88,34 @@ def main():
...
+    core = Core()
 
     if 'MYRIAD' in args.device:
         myriad_config = {'VPU_HW_STAGES_OPTIMIZATION': 'YES'}
-        ie.set_config(myriad_config, 'MYRIAD')
+        core.set_config(myriad_config, 'MYRIAD')
 
     if args.cpu_extension and 'CPU' in args.device:
-        ie.add_extension(args.cpu_extension, 'CPU')
+        core.add_extension(args.cpu_extension, 'CPU')
```
vladimir-dudnik, line 98: I think we need to remove extensions, as we do not have models with extended layers

ivikhrev: Removed.

PR #3030, opened 2021-12-29
models/public: move to https://storage.openvinotoolkit.org/repositories/open_model_zoo/public/2022.1

Wovchena:


models/public/alexnet/model.yml
```diff
@@ -32,7 +32,7 @@ files:
   - name: alexnet.caffemodel
     size: 243862414
     checksum: f102235bc9d8f422a04d5254a21fd1862986eae91c7337714fd5d78dcf37f835d05ed7943e80056ae19ca9e9c054c181
-    source: http://dl.caffe.berkeleyvision.org/bvlc_alexnet.caffemodel
```
alalek, line 35: It makes sense to keep original URL somewhere, e.g. in `description` field.

vladimir-dudnik: actually, I agree. @Wovchena @jkamelin can we replace source keyword with new, original_source keyword. So, when we upload public models package to storage.openvinotoolkit.org, CI will add us keywords source for actual link where we will download model by default, and original_source will be kept in model.yml files for reference. And potentially, we even may add option to downloader to get models from original links ?

Wovchena: The original URL is kept in `source.txt` similar to how it was done for 2021.4 public models

vladimir-dudnik: yeah, but having this as a keyword in model.yml allows to use it for downloading as an option. and this way we keep all info in single place

vladimir-dudnik (approved):

vladimir-dudnik:
hash mismatch for mtcnn models?

models/public/Sphereface/model.yml
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) 2019 Intel Corporation
+# Copyright (c) 2021 Intel Corporation
```
eaidova, line 1: possibly, it should be 2022 :)

eaidova:
@Wovchena @vladimir-dudnik what about updating CONTRIBUTING.md file with info, how public models should be contributed according to this changes now (it may confuse some external contributors...)?

Wovchena:
@jkamelin, can you help me with the error? The clue I have is that concatenation of .prototxt with itself gives the actual hash.

Wovchena:
@jkamelin also running with `use_downloader_cache` results in printing the error twice: http://213.221.44.203/job/omz/job/precommit/job/downloader_ubuntu18/1491/console

jkamelin:
> @jkamelin also running with `use_downloader_cache` results in printing the error twice: http://213.221.44.203/job/omz/job/precommit/job/downloader_ubuntu18/1491/console

This happens because mtcnn and mtcnn-o, mtcnn-p, mtcnn-r are in [tested_models.lst](http://213.221.44.203/job/omz/job/precommit/job/downloader_ubuntu18/1493/artifact/tested_models.lst/*view*/), so when downloading the second time, all files are written to the end of previously downloaded ones. I'll create a fix on downloader side.

Wovchena:
Jenkins please retry a build

models/public/alexnet/model.yml
Wovchena: Done

models/public/Sphereface/model.yml
Wovchena: Updated

eaidova (approved):

PR #3037, opened 2021-12-30
ov 2.0: python text_spotting_demo

anzhella-pankratova:
Migrate python text_spotting_demo to OV2.0 API.

### DIFF
--- a/demos/text_spotting_demo/python/text_spotting_demo.py
+++ b/demos/text_spotting_demo/python/text_spotting_demo.py
@@ -182,7 +182,7 @@ def main():
         n, c, h, w = mask_rcnn_model.input(input_tensor_name).shape
         if n != 1:
             raise RuntimeError('Only batch 1 is supported by the demo application')
-    except:
+    except RuntimeError:
         raise RuntimeError('Demo supports only topologies with the following input tensor name: {}'.format(input_tensor_name))
 
     required_output_names = {'boxes', 'labels', 'masks', 'text_features.0'}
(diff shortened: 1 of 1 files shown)


vladimir-dudnik:
Jenkins please retry a build

vladimir-dudnik (approved):

demos/text_spotting_demo/python/text_spotting_demo.py
```diff
@@ -347,7 +353,6 @@ def main():
     metrics.log_total()
     for rep in presenter.reportMeans():
         log.info(rep)
-    cv2.destroyAllWindows()
```
vladimir-dudnik, line 350: why is this removed?

anzhella-pankratova: When the demo exits, all cv2 windows are destroyed by the demo anyway, so this line is not necessary at the end of the script

vladimir-dudnik: well, when application exits, all allocated memory will be deallocated anyway, but good practice is to deallocate resources you've allocated.

anzhella-pankratova: By the way, it's the recommendation from OpenCV doc to not call it in simple cases (https://docs.opencv.org/3.4/d7/dfc/group__highgui.html#ga5afdf8410934fd099df85c75b2e0888b)

> but good practice is to deallocate resources you've allocated

I agree with it. But it's more applicable for sophisticated scenarios with a potential risk that some resource will be not deallocated automatically. 
In Python demos we are certain, what OpenCV objects, like window, capture and video_writter, are deallocated properly. Otherwise, we need to call `release()` for these objects (capture, video_writter).  

Since we don't apply this functionality in py demos (either destroy windows or release OpenCV objects at the end of script), I also removed the line in text_spotting_demo and instance_segmentation_demo to be consistent with other computer vision py demos, where it was deleted a long time ago

eaidova (approved):

PR #3041, opened 2022-01-10
Add openvino-telemetry to requirements

jkamelin:
update for openvino telemetry


Wovchena:
Was `openvino_telemetry` being installed on CI before? You may need to suppress actual data sending.

vladimir-dudnik (approved):

eaidova:
> Was `openvino_telemetry` being installed on CI before? You may need to suppress actual data sending.

I believe, we need to check (my suspicion that it is not applied now...)

nignatovsky (approved):

PR #3046, opened 2022-01-10
AC: remove usage MO

eaidova:
usage mo or model converter before running AC is the preferable way to get model. Remove legacy code, which allows to run MO indirectly via ac config


eaidova:
Jenkins please retry a build

vladimir-dudnik (approved):

PR #3056, opened 2022-01-12
ov 2.0: python image_inpainting_demo

anzhella-pankratova:
Migrate python image_inpainting_demo to OV2.0 API

### DIFF
--- a/demos/image_inpainting_demo/python/inpainting.py
+++ b/demos/image_inpainting_demo/python/inpainting.py
@@ -50,12 +50,14 @@ class ImageInpainting:
         return next(iter(output.values()))
 
 
-    def process(self, src_image, mask):
-        image = np.transpose(src_image, (2, 0, 1))
-        mask = np.transpose(mask, (2, 0, 1))
+    def process(self, image, mask):
+        if self.nchw_layout:
+            image = np.transpose(image, (2, 0, 1))
+            mask = np.transpose(mask, (2, 0, 1))
         image = np.expand_dims(image, axis=0)
         mask = np.expand_dims(mask, axis=0)
         output = self.infer(image, mask)
 
-        output = np.transpose(output, (0, 2, 3, 1)).astype(np.uint8)
-        return output[0]
+        if self.nchw_layout:
+            output = np.transpose(output, (0, 2, 3, 1))
+        return output.astype(np.uint8)[0]
--- a/demos/image_inpainting_demo/python/inpainting.py
+++ b/demos/image_inpainting_demo/python/inpainting.py
@@ -1,5 +1,5 @@
 """
- Copyright (c) 2019-2020 Intel Corporation
+ Copyright (c) 2019-2022 Intel Corporation
  Licensed under the Apache License, Version 2.0 (the "License");
  you may not use this file except in compliance with the License.
  You may obtain a copy of the License at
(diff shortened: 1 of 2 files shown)


vladimir-dudnik:
Jenkins please retry a build

vladimir-dudnik:
@anzhella-pankratova 01:19:48  RuntimeError: Attempt to get a name for a Tensor without names
either we did something wrong in new API or there are issues with API itself

cc @AlexeyLebedev1 @AnastasiaKazantaeva

anzhella-pankratova:
> @anzhella-pankratova 01:19:48 RuntimeError: Attempt to get a name for a Tensor without names either we did something wrong in new API or there are issues with API itself
> 
> cc @AlexeyLebedev1 @AnastasiaKazantaeva

I successfully have launched the demo locally with IR version 10, however with IR version 11 this error is raised. So the problem is not caused by new API, the Model Optimizer converts the model with tensors without names

vladimir-dudnik:
@andrei-kochin is it known issue?

andrei-kochin:
> @andrei-kochin is it known issue?

Yes it is known issue with missing tensor names

demos/image_inpainting_demo/python/inpainting.py
```diff
@@ -15,32 +15,36 @@
...
-        assert channels == 3, "Expected 3-channel input"
+        _, channels, input_height, input_width = model.input(self.image_input_layer).shape
+        if channels != 3:
+            raise RuntimeError("The model expects 3 channels for {} input layer".format(self.image_input_layer))
 
-        _, channels, mask_height, mask_width = model.input_info[self._input_layer_names[1]].input_data.shape
-        assert channels == 1, "Expected 1-channel input"
+        _, channels, mask_height, mask_width = model.input(self.mask_input_layer).shape
+        if channels != 1:
+            raise RuntimeError("The model expects 3 channels for {} input layer".format(self.mask_input_layer))
```
eaidova, line 37: mask is single channel input
```suggestion
            raise RuntimeError("The model expects 1 channels for {} input layer".format(self.mask_input_layer))
```

anzhella-pankratova: Thanks, it was a typo

eaidova:
> > @andrei-kochin is it known issue?
> 
> Yes it is known issue with missing tensor names

eaidova:
@

> > > @andrei-kochin is it known issue?
> > 
> > 
> > Yes it is known issue with missing tensor names

@andrei-kochin  @vladimir-dudnik  should we wait for fix or we can workaround the issue (using indexes instead of tensor names for inference)?

anzhella-pankratova:
Jenkins please retry a build

anzhella-pankratova:
Jenkins please retry a build

eaidova:
`RuntimeError: Can't SetBlob with name: Placeholder, because model input and blob are incompatible`

@anzhella-pankratova
looks like something wrong with data shape (old api automatically fit data to input shape, if it has the same number of elements, but different shape representation (e.g. input (1, 1, 10, 10) and data (10, 10) - ok), new api expects that you provide data with the same shape

anzhella-pankratova:
> `RuntimeError: Can't SetBlob with name: Placeholder, because model input and blob are incompatible`
> 
> @anzhella-pankratova looks like something wrong with data shape (old api automatically fit data to input shape, if it has the same number of elements, but different shape representation (e.g. input (1, 1, 10, 10) and data (10, 10) - ok), new api expects that you provide data with the same shape

I've updated the image preprocessing step with NHWC support, now it's fine.

demos/image_inpainting_demo/python/inpainting.py
```diff
@@ -15,40 +15,49 @@
...
+            raise RuntimeError("Mask size is expected to be equal to image size")
         self.input_height = input_height
         self.input_width = input_width
 
 
     def infer(self, image, mask):
-        output = self._exec_model.infer(inputs={self._input_layer_names[0]: image, self._input_layer_names[1]: mask})
-        return output[self._output_layer_name]
+        output = self.infer_request.infer(inputs={self.image_input_layer: image, self.mask_input_layer: mask})
+        return next(iter(output.values()))
```
eaidova, line 50: do you plan unify logic with output[self.output_tensor] later for all demos for models with single output?

anzhella-pankratova: Yeah, I will change it in all demos in a separate PR

eaidova (approved):

PR #3063, opened 2022-01-13
Fixed some demo mistakes

mpashchenkov:
Fixes for problems with Klocwork.

### DIFF
--- a/demos/gaze_estimation_demo/cpp/src/landmarks_estimator.cpp
+++ b/demos/gaze_estimation_demo/cpp/src/landmarks_estimator.cpp
@@ -98,7 +98,7 @@ std::vector<cv::Point2i> LandmarksEstimator::heatMapPostprocess(cv::Rect faceBou
 
 std::vector<cv::Mat> LandmarksEstimator::split(std::vector<float>& data, const std::vector<unsigned long>& shape) {
     std::vector<cv::Mat> flattenData(shape[1]);
-    float* output = &data.front();
+    float* output = data.data();
     for (size_t i = 0; i < flattenData.size(); i++) {
         flattenData[i] = cv::Mat(shape[2], shape[3], CV_32FC1, output + i * shape[2] * shape[3]);
     }
--- a/demos/gaze_estimation_demo/cpp/src/landmarks_estimator.cpp
+++ b/demos/gaze_estimation_demo/cpp/src/landmarks_estimator.cpp
@@ -98,7 +98,10 @@ std::vector<cv::Point2i> LandmarksEstimator::heatMapPostprocess(cv::Rect faceBou
 
 std::vector<cv::Mat> LandmarksEstimator::split(std::vector<float>& data, const std::vector<unsigned long>& shape) {
     std::vector<cv::Mat> flattenData(shape[1]);
-    float* output = data.data();
+    float* output = new float[data.size()];
+    for (size_t i = 0; i < data.size(); i++) {
+        output[i] = data[i];
+    }
     for (size_t i = 0; i < flattenData.size(); i++) {
         flattenData[i] = cv::Mat(shape[2], shape[3], CV_32FC1, output + i * shape[2] * shape[3]);
     }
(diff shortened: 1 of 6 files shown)


demos/gaze_estimation_demo/cpp/src/landmarks_estimator.cpp
```diff
@@ -99,10 +98,7 @@ std::vector<cv::Point2i> LandmarksEstimator::heatMapPostprocess(cv::Rect faceBou
 
 std::vector<cv::Mat> LandmarksEstimator::split(std::vector<float>& data, const std::vector<unsigned long>& shape) {
     std::vector<cv::Mat> flattenData(shape[1]);
-    float* output = new float[data.size()];
-    for (size_t i = 0; i < data.size(); i++) {
-        output[i] = data[i];
-    }
+    float* output = &data.front();
```
Wovchena, line 101:
```suggestion
    float* output = data.data;
```
But I'm not sure `data` vector outlives `flattenData`. Is this guaranteed?

demos/gesture_recognition_demo/cpp_gapi/src/custom_kernels.cpp
```diff
@@ -112,6 +112,11 @@ struct Params {
         prepared_mat = params.prepared_mat;
         last_id.fetch_add(params.last_id);
     }
+    Params& operator=(const Params& params) {
```
Wovchena, line 115: I think it's better to rely on struct initialization. You can remove constructor and this operator. And in `setup()` just do
```cpp
state = std::make_shared<BatchState::Params>({cv::Mat(), cv::Mat(
                                   std::vector<int>{1,
                                                    int(net_size[0]),
                                                    int(net_size[1]),
                                                    int(net_size[2]),
                                                    int(net_size[3])}
), 0});
```

Wovchena (changes requested):

demos/gaze_estimation_demo/cpp/src/landmarks_estimator.cpp
Wovchena: Even better. `flattenData[i] = cv::Mat(shape[2], shape[3], CV_32FC1, output + i * shape[2] * shape[3]);` can be replaced with `flattenData[i] = cv::Mat(shape[2], shape[3], CV_32FC1, data.data + i * shape[2] * shape[3]);`

mpashchenkov: Fixed.
But function returns copy of vector `flattenData`, isn't it? Is this full copy of `std::vector<cv::Mat>`? Now landmarks work correctly.

demos/gesture_recognition_demo/cpp_gapi/src/custom_kernels.cpp
mpashchenkov: Unfortunately atomic<size_t> doesn't have copy constructor (deleted function). I am thinking about this now.

demos/gaze_estimation_demo/cpp/src/landmarks_estimator.cpp
Wovchena: It copies mats, not underlying data:
> [Matrix constructors that take data and step parameters do not allocate matrix data.](https://docs.opencv.org/4.x/d3/d63/classcv_1_1Mat.html#a51615ebf17a64c968df0bf49b4de6a3a)

mpashchenkov: Ok. Memory isn't shared between threads and lives as long as the result (`std::vector<cv::Point2i>`) is built with this memory. I don't know the meaning of this `new`.

Wovchena: I'm not sure I follow. To me the hypothetical problem is that mats stored in `flattenData` borrow memory from `data` vector. This can be a real problem if `flattenData` outlives `data`.

mpashchenkov: Revert.

Wovchena (approved):

PR #3065, opened 2022-01-13
ov 2.0: python speech_recognition_deepspeech_demo

anzhella-pankratova:
Migrate python speech_recognition_deepspeech_demo to OV2.0 API.
**The demo does not work, because the model's input/output nodes names are changed by original tensorflow names.** So, the profile file will contain different names for old and new API. Waiting for updates from MO side solving the issue of models nodes naming

### DIFF
--- a/demos/speech_recognition_deepspeech_demo/python/asr_utils/deep_speech_seq_pipeline.py
+++ b/demos/speech_recognition_deepspeech_demo/python/asr_utils/deep_speech_seq_pipeline.py
@@ -1,5 +1,5 @@
 #
-# Copyright (C) 2019-2021 Intel Corporation
+# Copyright (C) 2019-2022 Intel Corporation
 # SPDX-License-Identifier: Apache-2.0
 #
 # This file is based in part on deepspeech_openvino_0.5.py by Feng Yen-Chang at
@@ -13,12 +13,12 @@ from asr_utils.ctc_decoder_seq_pipeline import CtcDecoderSeqPipelineStage
 
 
 class DeepSpeechSeqPipeline:
-    def __init__(self, profile, ie, model, lm=None, beam_width=500, max_candidates=None,
+    def __init__(self, profile, core, model, lm=None, beam_width=500, max_candidates=None,
             device='CPU', online_decoding=False):
         """
             Args:
         profile (dict), a dict with pre/post-processing parameters, see profiles.py
-        ie (IECore or None), IECore object for model loading/compilation/inference
+        core (Core or None), Core object for model loading/compilation/inference
         model (str), filename of IE IR .xml file of the network
         lm (str), filename of LM (language model)
         beam_width (int), the number of prefix candidates to retain during decoding in beam search (default 500)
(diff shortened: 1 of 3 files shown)


eaidova:
@anzhella-pankratova 

> The demo does not work, because the model's input/output nodes names are changed by original tensorflow names. (CVS-75841)

We do not need to preserve demo support for both old and new API in the same time, so I think it is not big problem to change model output names, especially if they aligned with original framework

anzhella-pankratova:
@eaidova We would do it, if both models mozilla-deepspeech-0.6.1 and mozilla-deepspeech-0.8.2 were contain only tensorflow names for nodes. But mozilla-deepspeech-0.6.1 has both old and new names, allowing launch the demo with old and new API correspondently. In this case I believe to modify the names for the second model is a better option and we can wait for update. 

Alternately, I can replace the names in the demo in order to run tests. If the tests are passed, we will merge the PR, and I will provide the other PR to support both API when the conversion for mozilla-deepspeech-0.8.2 is fixed. @vladimir-dudnik what do you think?

vladimir-dudnik:
@anzhella-pankratova when you say "mozilla-deepspeech-0.6.1 has both old and new names, allowing launch the demo with old and new API" you mean IR v10 and IR v11?

anzhella-pankratova:
> @anzhella-pankratova when you say "mozilla-deepspeech-0.6.1 has both old and new names, allowing launch the demo with old and new API" you mean IR v10 and IR v11?

Yes. For example, for IR v10 we have "logits" name, for IR v11 "logits:0". For mozilla-deepspeech-0.6.1 the both these names can be accessed, but for mozilla-deepspeech-0.8.2 the input name contains only "logits:0" name

vladimir-dudnik:
> > @anzhella-pankratova when you say "mozilla-deepspeech-0.6.1 has both old and new names, allowing launch the demo with old and new API" you mean IR v10 and IR v11?
> 
> @vladimir-dudnik Yes. For example, for IR v10 we have "logits" name, for IR v11 "logits:0". For mozilla-deepspeech-0.6.1 the both these names can be accessed, but for mozilla-deepspeech-0.8.2 the input name contains only "logits:0" name

@andrei-kochin it seems like MO issue ? We do not expect changes in names between IR v10 and IR v11 ?.

ilyachur:
> > > @anzhella-pankratova when you say "mozilla-deepspeech-0.6.1 has both old and new names, allowing launch the demo with old and new API" you mean IR v10 and IR v11?
> > 
> > 
> > @vladimir-dudnik Yes. For example, for IR v10 we have "logits" name, for IR v11 "logits:0". For mozilla-deepspeech-0.6.1 the both these names can be accessed, but for mozilla-deepspeech-0.8.2 the input name contains only "logits:0" name
> 
> @andrei-kochin @ilya-lavrenov @ilyachur it seems like MO or runtime issue ? We do not expect changes in names between IR v10 and IR v11 ?.

Why don't you expect changes in names after migration to v11?
In IR v11 we are using tensor names, in IR v10 operation names. It means that we are using different names and they can be different.

vladimir-dudnik:
@ilyachur basically you say, 2022.1 applications will have complexities if they have to support 2021.4.2 IRs.

vladimir-dudnik:
@anzhella-pankratova we will generate intel models with IR v11 soon (hope on this week), you do not need to support IR v10 names in 2022.1 application (if one need this there is always 2021.4.2 release available)

anzhella-pankratova:
> @anzhella-pankratova we will generate intel models with IR v11 soon (hope on this week), you do not need to support IR v10 names in 2022.1 application (if one need this there is always 2021.4.2 release available)

@vladimir-dudnik Okay, then I will update the names.

eaidova:
looks like tests, passed, but need to add sys.exit for avoid issue with memory deallocation on python api side

eaidova (approved):

PR #3072, opened 2022-01-14
model_tools: read dataset classes

as-suvorov:


tools/model_tools/src/openvino/model_zoo/omz_data_downloader.py
```diff
@@ -30,6 +30,12 @@ def copy_data(output_dir):
         str(data_path),
     )
 
+
+def read_dataset_classes(file_name):
+    file_path = _common.PACKAGE_DIR / 'data' / 'dataset_classes' / file_name
```
eaidova, line 35: do we need to check that such file exists?

tools/model_tools/src/openvino/model_zoo/omz_data_downloader.py
```diff
@@ -30,6 +30,12 @@ def copy_data(output_dir):
...
+    return file_path.read_text()
```
eaidova, line 36: as I can see, for some datasets, classes represented in json format, probably we need to handle file type during reading?

Which expected result from reading? just get text and parse it outside of tool or retrun list with class names?

tools/model_tools/src/openvino/model_zoo/omz_data_downloader.py
as-suvorov: check added

tools/model_tools/src/openvino/model_zoo/omz_data_downloader.py
as-suvorov: parsing added

eaidova (approved):

PR #3086, opened 2022-01-19
OV2.0 c++ mri reconstruction demo

ivikhrev:
Migrate c++ mri reconstruction demo to new API.

### DIFF
--- a/demos/mri_reconstruction_demo/cpp/main.cpp
+++ b/demos/mri_reconstruction_demo/cpp/main.cpp
@@ -1,7 +1,6 @@
 // Copyright (C) 2021-2022 Intel Corporation
 // SPDX-License-Identifier: Apache-2.0
 //
-#include <inference_engine.hpp>
 #include <openvino/openvino.hpp>
 #include <opencv2/opencv.hpp>
 #include <utils/common.hpp>
@@ -39,12 +38,13 @@ int main(int argc, char** argv) {
         return 0;
     }
 
-    slog::info << *InferenceEngine::GetInferenceEngineVersion() << slog::endl;
+    slog::info << ov::get_openvino_version() << slog::endl;
     ov::runtime::Core core;
 
-    std::shared_ptr<ov::Model> model = core.read_model(FLAGS_m);
     slog::info << "Reading model " << FLAGS_m << slog::endl;
-    logLayersInfo(model);
+    std::shared_ptr<ov::Model> model = core.read_model(FLAGS_m);
+    logBasicModelInfo(model);
+
     std::string outputTensorName = "";
     for (const auto& output : model->outputs()) {
         if (output.get_any_name().find("89") != std::string::npos) {
@@ -54,6 +54,7 @@ int main(int argc, char** argv) {
     if (outputTensorName == "") {
         throw std::logic_error("Not found suitable output!");
     }
+
     ov::preprocess::PrePostProcessor ppp(model);
     ppp.input().model().set_layout("NHWC");
     model = ppp.build();
--- a/demos/mri_reconstruction_demo/cpp/main.cpp
+++ b/demos/mri_reconstruction_demo/cpp/main.cpp
@@ -52,7 +52,7 @@ int main(int argc, char** argv) {
             outputTensorName = output.get_any_name();
         }
     }
-    if (outputTensorName == "") {
+    if (outputTensorName.empty()) {
         throw std::logic_error("Not found suitable output!");
     }
 
--- a/demos/mri_reconstruction_demo/cpp/main.cpp
+++ b/demos/mri_reconstruction_demo/cpp/main.cpp
@@ -46,8 +46,9 @@ int main(int argc, char** argv) {
     logBasicModelInfo(model);
 
     std::string outputTensorName = "";
+    ov::Layout outputLayout("NHWC");
     for (const auto& output : model->outputs()) {
-        if (output.get_any_name().find("89") != std::string::npos) {
+        if (output.get_shape()[ov::layout::channels_idx(outputLayout)] == 1) {
             outputTensorName = output.get_any_name();
         }
     }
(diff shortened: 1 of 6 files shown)


demos/mri_reconstruction_demo/cpp/main.cpp
```diff
@@ -1,7 +1,8 @@
-// Copyright (C) 2018-2021 Intel Corporation
+// Copyright (C) 2021-2022 Intel Corporation
 // SPDX-License-Identifier: Apache-2.0
 //
 #include <inference_engine.hpp>
```
vladimir-dudnik, line 4: why you still need in old API header?

vladimir-dudnik:
please use common functions from #3003 (merged)

demos/mri_reconstruction_demo/cpp/main.cpp
ivikhrev: removed

demos/mri_reconstruction_demo/cpp/main.cpp
```diff
@@ -38,16 +38,31 @@ int main(int argc, char** argv) {
...
-    logExecNetworkInfo(execNet, FLAGS_m, FLAGS_d);
+    std::string outputTensorName = "";
+    for (const auto& output : model->outputs()) {
+        if (output.get_any_name().find("89") != std::string::npos) {
+            outputTensorName = output.get_any_name();
+        }
+    }
+    if (outputTensorName == "") {
+        throw std::logic_error("Not found suitable output!");
+    }
```
anzhella-pankratova, line 56: The demo expects the model with 1 output tensor, I think in that case we might not check the output name. The check is preferable for models with 2 or more outputs to extract data from specific tensor. So you can simplify it like this:
```suggestion
   OPENVINO_ASSERT(model->outputs().size() == 1, "The demo supports models with 1 output only");
   std::string outputTensorName = model->outputs()[0].get_any_name();
```

ivikhrev: I see that model has 2 outputs:
```
[ INFO ] Reading model ../../intel64/Debug/hybrid-cs-model-mri.xml
[ INFO ] model name: hybrid-cs-model-mri
[ INFO ] inputs:
[ INFO ] Func/StatefulPartitionedCall/input/_0:0, f32, {1,256,256,2}
[ INFO ] outputs:
[ INFO ] Func/StatefulPartitionedCall/output/_90:0, f32, {1,256,256,2}
[ INFO ] Func/StatefulPartitionedCall/output/_89:0, f32, {1,256,256,1}
```
And we need the one with `89` in name.

ivikhrev: Maybe second output emerged after recent changed that were done in MO.

anzhella-pankratova: Yeah, the number of outputs is probably changed with IR version 11. Then I will add the same check to the Python demo. Thank you for the clarification!

demos/mri_reconstruction_demo/cpp/main.cpp
```diff
@@ -38,16 +38,31 @@ int main(int argc, char** argv) {
...
-    net.getInputsInfo().begin()->second->setLayout(InferenceEngine::Layout::NHWC);
+    slog::info << "Reading model " << FLAGS_m << slog::endl;
+    std::shared_ptr<ov::Model> model = core.read_model(FLAGS_m);
+    logBasicModelInfo(model);
 
-    InferenceEngine::ExecutableNetwork execNet = ie.LoadNetwork(net, FLAGS_d);
-    logExecNetworkInfo(execNet, FLAGS_m, FLAGS_d);
+    std::string outputTensorName = "";
+    for (const auto& output : model->outputs()) {
+        if (output.get_any_name().find("89") != std::string::npos) {
```
anzhella-pankratova, line 50: I have checked the output names with Python API, the desired output tensor has 3 possible names: `{'StatefulPartitionedCall/model/conv2d_43/BiasAdd:0', 'Identity:0', 'Func/StatefulPartitionedCall/output/_89:0', 'StatefulPartitionedCall/Identity:0'}`  

 The `get_any_name()` might return any of them. So please, use `get_names()` method, it's safe and deterministic. Probably you should iterate through all of them to check the condition

ivikhrev: Maybe more stable approach is to find desired output by it's shape. In this case we need output with only 1 channel, so it would be better to find and get output with 1 channel.

vladimir-dudnik: yes, also possible. would not say it is more stable, but should be good enough. This particular demo support only one specific model, so most stable approach is to check all input/output params (name, type, shape), num of inputs/outputs but it is not critically important, demo any way will not work with wrong model.

vladimir-dudnik:
Jenkins please retry a build

demos/mri_reconstruction_demo/cpp/main.cpp
anzhella-pankratova: I support the approach with the verification by the last dimension in shape, let's apply it

demos/mri_reconstruction_demo/cpp/main.cpp
```diff
@@ -38,16 +38,32 @@ int main(int argc, char** argv) {
...
-    InferenceEngine::ExecutableNetwork execNet = ie.LoadNetwork(net, FLAGS_d);
-    logExecNetworkInfo(execNet, FLAGS_m, FLAGS_d);
+    std::string outputTensorName = "";
+    ov::Layout outputLayout("NHWC");
+    for (const auto& output : model->outputs()) {
+        if (output.get_shape()[ov::layout::channels_idx(outputLayout)] == 1) {
+            outputTensorName = output.get_any_name();
+        }
+    }
+    if (outputTensorName == "") {
```
anzhella-pankratova, line 55:
```suggestion
    if (outputTensorName.empty()) {
```

anzhella-pankratova (approved):

vladimir-dudnik (approved):

PR #3100, opened 2022-01-24
deps: move paddle version

eaidova:
paddle 2.1.3 has incompatible dependency with tf on windows, in 2.2.0 this problem is resolved

### DIFF
--- a/ci/requirements-conversion.txt
+++ b/ci/requirements-conversion.txt
@@ -87,7 +87,7 @@ onnx==1.10.1
     #   -r tools/model_tools/requirements-pytorch.in
 opt-einsum==3.3.0
     # via tensorflow
-paddlepaddle==2.1.3
+paddlepaddle==2.2.0
     # via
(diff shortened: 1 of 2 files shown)


vladimir-dudnik (approved):

vladimir-dudnik:
@nignatovsky can general name for opencv package be reason of failed build in precommit (you probably start uploading a new version of package)?
15:21:24  + tar -xf /mnt/dist/opencv/opencv_linux.tar.gz

build issue which we did not have before:
15:28:33  /jenkins/workspace/omz/precommit/demos_ubuntu18/openvino/opencv/lib/libopencv_dnn.so.4.5.5: undefined reference to `ov::Node::evaluate_upper(std::vector<ov::Tensor, std::allocator<ov::Tensor> >&) const'

nignatovsky:
Yes; I will add OpenCV versions to `ci/dependencies.yml` file to avoid such situations.

PR #3103, opened 2022-01-24
OV2.0 c++ interactive_face_detection_demo

Wovchena:


demos/interactive_face_detection_demo/cpp/detectors.cpp
```diff
@@ -222,7 +154,7 @@ void FaceDetection::fetchResults() {
         }
     }
 
-    for (int i = 0; i < maxProposalCount && objectSize == 7; i++) {
+    for (int i = 0; i < ndetections; i++) {
         float image_id = detections[i * objectSize + 0];
```
ivikhrev, line 158:
```suggestion
        float image_id = detections[i * objectSize];
```

demos/interactive_face_detection_demo/cpp/detectors.cpp
```diff
@@ -187,10 +119,10 @@ void FaceDetection::fetchResults() {
                 continue;
             }
 
-            r.location.x = static_cast<int>(detections[i * objectSize + 0] / network_input_width * width);
-            r.location.y = static_cast<int>(detections[i * objectSize + 1] / network_input_height * height);
-            r.location.width = static_cast<int>(detections[i * objectSize + 2] / network_input_width * width - r.location.x);
-            r.location.height = static_cast<int>(detections[i * objectSize + 3] / network_input_height * height - r.location.y);
+            r.location.x = static_cast<int>(detections[i * objectSize + 0] / model_input_width * width);
```
ivikhrev, line 122:
```suggestion
            r.location.x = static_cast<int>(detections[i * objectSize] / model_input_width * width);
```

demos/interactive_face_detection_demo/cpp/detectors.cpp
```diff
@@ -16,169 +15,102 @@
...
-void BaseDetection::submitRequest() {
-    if (!enabled() || request == nullptr) return;
-    if (isAsync) {
-        request->StartAsync();
-    } else {
-        request->Infer();
-    }
-}
+namespace {
+constexpr size_t ndetections = 200;
```
ivikhrev, line 24: It is replacement for maxProposalCount, isn't it? Then it was determined based on model shape, why now it's hardcoded?

demos/interactive_face_detection_demo/cpp/detectors.cpp
```diff
@@ -268,226 +200,147 @@ void FaceDetection::fetchResults() {
...
 }
 
 void AntispoofingClassifier::submitRequest() {
     if (!enquedFaces)
         return;
-    if (isBatchDynamic) {
-        request->SetBatch(enquedFaces);
-    }
-    BaseDetection::submitRequest();
+    request.set_input_tensor(ov::Tensor{request.get_input_tensor(), {0, 0, 0, 0}, {enquedFaces, inShape[1], inShape[2], inShape[3]}});
```
ivikhrev, line 214: It's kind of tricky. You've filled input tensor in `AntispoofingClassifier::enqueue` by selecting roi's. And now you select roi with whole batch and set it as input. But if it was already filled earlier, why do we need to set it?

ivikhrev (changes requested):

demos/interactive_face_detection_demo/cpp/detectors.cpp
```diff
@@ -268,226 +200,147 @@ void FaceDetection::fetchResults() {
...
     enquedFaces++;
 }
 
-float AntispoofingClassifier::operator[] (int idx) const {
-    InferenceEngine::Blob::Ptr  ProbBlob = request->GetBlob(prob_output);
-    InferenceEngine::LockedMemory<const void> ProbBlobMapped =
-        InferenceEngine::as<InferenceEngine::MemoryBlob>(ProbBlob)->rmap();
-    // use prediction for real face only
-    float r = ProbBlobMapped.as<float*>()[2 * idx] * 100;
+float AntispoofingClassifier::operator[](int idx) {
```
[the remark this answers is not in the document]
Wovchena: Why? To me I should have done the same removal for `detectors.h` as well

demos/interactive_face_detection_demo/cpp/main.cpp
```diff
@@ -71,63 +63,25 @@ int main(int argc, char *argv[]) {
...
-        Load(headPoseDetector).into(ie, FLAGS_d_hp, FLAGS_dyn_hp);
-        Load(emotionsDetector).into(ie, FLAGS_d_em, FLAGS_dyn_em);
-        Load(facialLandmarksDetector).into(ie, FLAGS_d_lm, FLAGS_dyn_lm);
-        Load(antispoofingClassifier).into(ie, FLAGS_d_am, FLAGS_dyn_am);
+        Load(faceDetector).into(core, FLAGS_d);
+        Load(ageGenderDetector).into(core, FLAGS_d);
+        Load(headPoseDetector).into(core, FLAGS_d);
+        Load(emotionsDetector).into(core, FLAGS_d);
+        Load(facialLandmarksDetector).into(core, FLAGS_d);
+        Load(antispoofingClassifier).into(core, FLAGS_d);
```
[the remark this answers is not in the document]
Wovchena: MULTI uses greedy approach when choosing which device to use for current request. But mostly yes, MULTI should be sufficient because people are unlikely to mess with choosing which particular device to assign to each of 6 nets.

`-d MULTI:CPU,CPU` works, but even just `-d GPU` fails with `Only dynamic batch is supported!`. That's broken GPU plugin.

demos/interactive_face_detection_demo/cpp/detectors.cpp
Wovchena: None of the plugins currently support https://github.com/openvinotoolkit/openvino/blob/069b3839cec79d26eb4bf8ee90a3b7437f96c106/src/inference/include/openvino/runtime/infer_request.hpp#L173
To work this around I create a batch of size equal to max possible number of detections. So non detection models need to know this limit, which is `ndetections`. And a detection model verifies that `ndetections = 200` is correct guess. After all detections are filed in a batch for non detection models, I shrink the tensor to be of smaller size to avoid inferencing all `ndetections` elements.

demos/interactive_face_detection_demo/cpp/detectors.cpp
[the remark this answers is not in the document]
ivikhrev: yeah, that's would be better

demos/interactive_face_detection_demo/cpp/detectors.cpp
ivikhrev: Understand, thanks for explanation

demos/interactive_face_detection_demo/cpp/detectors.cpp
```diff
@@ -703,58 +500,38 @@ std::vector<float> FacialLandmarksDetection::operator[] (int idx) const {
...
-            config[InferenceEngine::PluginConfigParams::KEY_DYN_BATCH_ENABLED] = InferenceEngine::PluginConfigParams::YES;
-        }
-
-        detector.net = ie.LoadNetwork(detector.read(ie), deviceName, config);
-        logExecNetworkInfo(detector.net, detector.pathToModel, deviceName, detector.topoName);
-        slog::info << "\tBatch size is set to " << detector.maxBatch << slog::endl;
+void Load::into(ov::Core& core, const std::string & deviceName) const {
+    if (!detector.pathToModel.empty()) {
+        ov::runtime::CompiledModel cml = core.compile_model(detector.read(core), deviceName);
+        logCompiledModelInfo(cml, detector.pathToModel, deviceName, detector.pathToModel);
```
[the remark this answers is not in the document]
Wovchena: Removed

vladimir-dudnik (approved):

demos/common/cpp/utils/include/utils/ocv_common.hpp
```diff
@@ -183,6 +182,17 @@ static UNUSED ov::Tensor wrapMat2Tensor(const cv::Mat& mat) {
     return ov::Tensor(ov::element::u8, ov::Shape{ 1, height, width, channels }, mat.data);
 }
 
+static inline void resize2tensor(const cv::Mat& mat, const ov::Tensor& tensor) {
+    static const ov::Layout layout{"NHWC"};
+    const ov::Shape& shape = tensor.get_shape();
+    cv::Size size{int(shape[ov::layout::height_idx(layout)]), int(shape[ov::layout::width_idx(layout)])};
```
vladimir-dudnik, line 188:
```suggestion
    cv::Size size{int(shape[ov::layout::width_idx(layout)]), int(shape[ov::layout::height_idx(layout)])};
```

demos/common/cpp/utils/include/utils/ocv_common.hpp
```diff
@@ -183,6 +182,17 @@ static UNUSED ov::Tensor wrapMat2Tensor(const cv::Mat& mat) {
     return ov::Tensor(ov::element::u8, ov::Shape{ 1, height, width, channels }, mat.data);
 }
 
+static inline void resize2tensor(const cv::Mat& mat, const ov::Tensor& tensor) {
+    static const ov::Layout layout{"NHWC"};
+    const ov::Shape& shape = tensor.get_shape();
+    cv::Size size{int(shape[ov::layout::width_idx(layout)]), int(shape[ov::layout::height_idx(layout)])};
+    assert(tensor.get_element_type() == ov::element::u8);
```
vladimir-dudnik, line 189: why you use assert for validating parameters ? is not it better to have just regular check for params and explicitly throw exception?

Wovchena: Normally these check's aren't needed: `PrePostProcessor` should do its job. So they are removed in release build but kept in debug

vladimir-dudnik: ok

Wovchena:
Jenkins please retry a build

ivikhrev (approved):

PR #3109, opened 2022-01-25
GAPI. Text Detection Demo

OrestChura:
This is G-API implementation of OMZ Text Detection Demo

 - `getFramesCount()` functionality as a method of `PerformanceMetrics` is also proposed to be added


demos/common/cpp/utils/src/performance_metrics.cpp
```diff
@@ -104,6 +104,11 @@ void PerformanceMetrics::logTotal() const {
     slog::info << "\tFPS: " << metrics.fps << slog::endl;
 }
 
+uint32_t PerformanceMetrics::getFrameCount() const {
```
mpashchenkov, line 107: What is this for?

demos/text_detection_demo/cpp/src/cnn.cpp
```diff
@@ -203,7 +203,7 @@ InferenceEngine::BlobMap EncoderDecoderCNN::Infer(const cv::Mat &frame) {
             break;
         input_data_decoder[0] = float(argmax);
 
-        infer_request_decoder_.SetBlob(in_dec_hidden_name_, infer_request_decoder_.GetBlob(out_enc_hidden_name_));
+        infer_request_decoder_.SetBlob(in_dec_hidden_name_, infer_request_decoder_.GetBlob(out_dec_hidden_name_));
```
mpashchenkov, line 206: Does it work now? It is OMZ variant of demo.

demos/text_detection_demo/cpp_gapi/main.cpp
```diff
@@ -0,0 +1,509 @@
...
+        if (trRequired) {
+            outVector += cv::gout(outTexts);
+        }
+        outVector += cv::gout(outPts);
+        /** ---------------- The execution part ---------------- **/
+        pipeline.setSource(cv::gin(cv::gapi::wip::make_src<custom::CommonCapSrc>(cap)));
+        pipeline.start();
+        std::chrono::steady_clock::time_point beginFrame = std::chrono::steady_clock::now();
+        while (pipeline.pull(std::move(outVector))) {
+            cv::Mat demoImage = image.clone();
```
mpashchenkov, line 389: Why clone?

demos/text_detection_demo/cpp_gapi/main.cpp
```diff
@@ -0,0 +1,509 @@
...
+            // Displaying system parameters and FPS
+            putHighlightedText(demoImage, "Found: " + std::to_string(numRecognized),
+                               cv::Point(10, 80), cv::FONT_HERSHEY_COMPLEX, 0.65,
+                               cv::Scalar(0, 0, 255), 2);
+
+            presenter.drawGraphs(demoImage);
+            metrics.update(beginFrame, demoImage, { 10, 22 }, cv::FONT_HERSHEY_COMPLEX, 0.65);
+
+            if (videoWriter.isOpened() &&
+                (outputFramesLimit == 0 || metrics.getFrameCount() <= outputFramesLimit)) {
```
mpashchenkov, line 463: It is implemented in the `ImagesCapture` class.

demos/text_detection_demo/cpp_gapi/src/shared_functions.cpp
```diff
@@ -0,0 +1,42 @@
...
+            secondLeftIdx = i;
+        }
+    }
+    if (secondLeft.y < mostLeft.y) {
+        mostLeft = secondLeft;
+        mostLeftIdx = secondLeftIdx;
+    }
+    return mostLeftIdx;
+}
+std::string custom::getWeightsPath(const std::string &modelPath) {
```
mpashchenkov, line 34: Remove this.
Use 
https://github.com/OrestChura/open_model_zoo/blob/cc7cf629ad129e556ffd5a5c21167f14223493ae/demos/gaze_estimation_demo/cpp_gapi/main.cpp#L198
instead.

demos/text_detection_demo/cpp/src/cnn.cpp
OrestChura: right. but it was a mistake, not recognizable because for the particular network `out_enc_hidden_name_ == out_dec_hidden_name_`

demos/text_detection_demo/cpp_gapi/main.cpp
OrestChura: `demoImage` is used for the user output, so I'm cloning not to corrupt the original frame, just in case

demos/common/cpp/utils/src/performance_metrics.cpp
OrestChura: this is to avoid counting the frames by hands while they are already counted in `PerformanceMetrics`

demos/text_detection_demo/cpp_gapi/main.cpp
OrestChura: You are about metrics, right? Yes, I didn't know, will change to use ImagesCapture metrics instead

vladimir-dudnik:
Jenkins please retry a build

demos/text_detection_demo/cpp_gapi/src/custom_kernels.cpp
```diff
@@ -0,0 +1,407 @@
...
+                }
+            }
+        }
+    }
+    return newData;
+}
+
+std::vector<std::size_t> dimsToShape(const cv::MatSize& sz) {
+    const int nDims = sz.dims();
+    std::vector<std::size_t> result(nDims, 0);
```
mpashchenkov, line 60: M B `std::vector<int> vec(sz.p, sz.p + sz.dims());` instead?

demos/text_detection_demo/cpp_gapi/src/custom_kernels.cpp
```diff
@@ -0,0 +1,407 @@
...
+                    const size_t                 numClasses,
+                    const size_t                 endToken,
+                          std::vector<cv::Mat>&  res) {
+        constexpr int maxDecodedSymbols = 20;
+        GAPI_DbgAssert(hiddens_.size() == features.size());
+        const size_t numDetectedLabels = features.size();
+
+        std::vector<cv::Mat> states(numDetectedLabels, cv::Mat({1}, CV_32FC1));
+        std::for_each(states.begin(), states.end(),
+                      [](cv::Mat& m){ m.dims = 1; *(m.begin<float>()) = 0; });
```
mpashchenkov, line 363: Could you explain this?

demos/text_detection_demo/cpp_gapi/src/custom_kernels.cpp
```diff
@@ -0,0 +1,407 @@
...
+        std::vector<cv::Mat> hiddens(hiddens_);
+
+        res.reserve(numDetectedLabels);
+        for (auto i = 0U; i < numDetectedLabels; i++) {
+            res.emplace_back(cv::Mat({ maxDecodedSymbols, 1, static_cast<int>(numClasses) },
```
mpashchenkov, line 368:
```suggestion
            res.emplace_back({ maxDecodedSymbols, 1, static_cast<int>(numClasses) },
```

demos/text_detection_demo/cpp_gapi/src/custom_kernels.cpp
```diff
@@ -0,0 +1,407 @@
...
+        std::vector<std::vector<cv::Point>> contours;
+        cv::findContours(bboxMask, contours, cv::RETR_CCOMP, cv::CHAIN_APPROX_SIMPLE);
+        if (contours.empty())
+            continue;
+        cv::RotatedRect r = cv::minAreaRect(contours[0]);
+        if (std::min(r.size.width, r.size.height) < minHeight)
+            continue;
+        if (r.size.area() < minArea)
+            continue;
+        bboxes.emplace_back(r);
```
mpashchenkov, line 97: push_back

demos/text_detection_demo/cpp_gapi/src/custom_kernels.cpp
```diff
@@ -0,0 +1,407 @@
...
+                    points = centralCropPoints(r);
+                    cv::resize(image(r), crop, outSize);
+                } else {
+                    points = imgPts;
+                    cv::resize(image, crop, outSize);
+                }
+            }
+            crop.reshape(1, blobShape).convertTo(blob, CV_32F);
+            out.emplace_back(blob.clone());
+            pts.emplace_back(points);
```
mpashchenkov, line 345: push_backs

demos/text_detection_demo/cpp_gapi/src/custom_kernels.cpp
```diff
@@ -0,0 +1,407 @@
...
+                    auto r = centralCropRect(image.cols, image.rows);
+                    points = centralCropPoints(r);
+                    cv::resize(image(r), crop, outSize);
+                } else {
+                    points = imgPts;
+                    cv::resize(image, crop, outSize);
+                }
+            }
+            crop.reshape(1, blobShape).convertTo(blob, CV_32F);
+            out.emplace_back(blob.clone());
```
mpashchenkov, line 344: M B create `out = vector<cv::Mat>(rrs.size(), cv::Mat(blobShape, CV_32FC1))`
and use `out[i]` instead blob:
`crop.reshape(1, blobShape).convertTo(out[i], CV_32F);`

demos/text_detection_demo/cpp_gapi/src/nets_configuration.cpp
```diff
@@ -0,0 +1,211 @@
...
+    }
+    trAlphabet = trPadSymbolFirst
+        ? std::string(decoderStartIndex + 1, kPadSymbol) + trSymbolsSet
+        : trSymbolsSet + kPadSymbol;
+}
+
+void custom::NetsConfig::configureTR(const std::string& trDevice) {
+    auto trNet = cv::gapi::ie::Params<nets::TextRecognition> {
+        trModelPath, fileNameNoExt(trModelPath) + ".bin", trDevice
+    }.cfgOutputLayers({trOutputName});
```
mpashchenkov, line 148: Please add [info about nets](https://github.com/openvinotoolkit/open_model_zoo/blob/24346485099c98f368afb2cb001eaf34925e0018/demos/gaze_estimation_demo/cpp_gapi/main.cpp#L166)

demos/common/cpp_gapi/utils_gapi/src/stream_source.cpp
```diff
@@ -16,14 +16,14 @@ void CommonCapSrc::preparation() {
     if (!tmp.data) {
         GAPI_Assert(false && "Couldn't grab the first frame");
     }
-    first = tmp.clone();
+    first = tmp;
```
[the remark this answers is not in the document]
OrestChura: No. cv::Mat counts references, and dies only when there's no references

demos/common/cpp_gapi/utils_gapi/src/stream_source.cpp
```diff
@@ -16,14 +16,14 @@ void CommonCapSrc::preparation() {
...
 }
 
 bool CommonCapSrc::pull(cv::gapi::wip::Data &data) {
     if (!first_pulled) {
         GAPI_Assert(!first.empty());
         first_pulled = true;
-        data = first;
+        data = first.clone();
```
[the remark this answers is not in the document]
OrestChura: Maybe..
How it is connected with this particular snippet?

The reason of this change is: I've faced a problem that, if you not `clone()` the `first`, then `data` will basically posses the pointed to `first`'s data. And when `data` is cleared due to g-api internal logic, `first` is cleared too.

So, we need to `clone()` it here

demos/common/cpp_gapi/utils_gapi/src/stream_source.cpp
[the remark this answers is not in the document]
OrestChura: I mean, it works that way :)

demos/tests/cases.py
```diff
@@ -660,6 +660,50 @@ def single_option_cases(key, *args):
...
+                ]),
+            *combine_cases(
+                TestCase(options={'-dt': 'simple'}),
+                [
+                    TestCase(options={'-m_tr': ModelArg('text-recognition-0015-encoder'),
+                                      '-tr_pt_first': None,
+                                      '-tr_o_blb_nm': 'logits',
+                                      '-m_tr_ss': '?0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ'},
+                             extra_models=[ModelArg('text-recognition-0015-decoder')]),
+                    # TestCase(options={'-m_tr': ModelArg('text-recognition-0016-encoder'), #TODO
```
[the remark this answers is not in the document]
OrestChura: The net is supported, however it cannot be retrieved from OMZ yet, as I understand
`cpp` case also has it commented

demos/text_detection_demo/cpp_gapi/src/custom_kernels.cpp
OrestChura: This is mostly setting all the elements of `states` to a default starting value - which is a 1-dimentional Mat with the single value equal 0.

It is impossible to create 1-dimentional cv::Mat by constructor, so I'm setting each `.dims` to 1

demos/text_detection_demo/cpp_gapi/src/custom_kernels.cpp
OrestChura: No, I need to create `CV_32FC1` Mat

And even without it, vector don't want to create a Mat out of a brace-enclosed initializer list :(

OrestChura: Oh, I got what you are saying: `emplace` semantics
Unfortunately, it also can't figure out what is the initializer list. I'll have to add `std::vector<int>`, but I'll fix `emplace` so it will construct Mat in place

demos/text_detection_demo/cpp_gapi/src/custom_kernels.cpp
OrestChura: no, unfortunately, then, without cloning, every `out[i]` would has the shared memory - I mean, all the Mats would be the same
and if handle this, it wouldn't be any shorter

demos/text_detection_demo/cpp_gapi/text_detection_demo_gapi.hpp
```diff
@@ -0,0 +1,169 @@
...
+DEFINE_int32(max_rect_num, -1, trMaxRectNumMsg);
+DEFINE_string(d_td, "CPU", tdDeviceMsg);
+DEFINE_string(d_tr, "CPU", trDeviceMsg);
+DEFINE_bool(no_show, false, noShowMsg);
+DEFINE_bool(r, false, outRawMsg);
+DEFINE_string(u, "", utilMonitorsMsg);
+DEFINE_uint32(b, 0, decoderBandwidthMsg);
+DEFINE_uint32(start_index, 0, decoderStartIndexMsg);
+DEFINE_string(pad, "#", padSymbolMsg);
+DEFINE_bool(gapi_regular_mode, false, gapiRegularMsg);
```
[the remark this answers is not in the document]
OrestChura: added, thanks!

demos/text_detection_demo/cpp_gapi/src/custom_kernels.cpp
```diff
@@ -0,0 +1,407 @@
...
+
+#include "custom_kernels.hpp"
+#include "shared_functions.hpp"
+#include "custom_nets.hpp"
+
+#include <opencv2/imgproc.hpp>
+
+void softmax(std::vector<float>& rdata) {
+    const size_t lastDim = 2;
+    for (size_t i = 0 ; i < rdata.size(); i+=lastDim) {
```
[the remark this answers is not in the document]
OrestChura: OK

demos/text_detection_demo/cpp_gapi/src/custom_kernels.cpp
```diff
@@ -0,0 +1,407 @@
...
+        float m = std::max(rdata[i], rdata[i+1]);
```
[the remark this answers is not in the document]
OrestChura: OK

demos/text_detection_demo/cpp_gapi/src/custom_kernels.cpp
```diff
@@ -0,0 +1,407 @@
...
+    if (shape.size() != axes.size()) {
+        throw std::runtime_error("Shape and axes must have the same dimension.");
+    }
+
+    for (size_t a : axes) {
+        if (a >= shape.size()) {
+            throw std::runtime_error("Axis must be less than dimension of shape.");
+        }
+    }
+    size_t totalSize = shape[0]*shape[1]*shape[2]*shape[3];
```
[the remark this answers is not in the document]
OrestChura: OK

demos/text_detection_demo/cpp_gapi/src/custom_kernels.cpp
OrestChura: Yes, works well thanks

demos/text_detection_demo/cpp_gapi/src/custom_kernels.cpp
OrestChura: Done

demos/text_detection_demo/cpp_gapi/src/custom_kernels.cpp
OrestChura: OK, figured out push/emplace differences, thanks!

demos/text_detection_demo/cpp_gapi/src/custom_kernels.cpp
OrestChura: OK

demos/text_detection_demo/cpp_gapi/src/nets_configuration.cpp
OrestChura: OK, added

demos/common/cpp/utils/src/performance_metrics.cpp
Wovchena: But you still count them by hands in this function

demos/common/cpp_gapi/utils_gapi/src/stream_source.cpp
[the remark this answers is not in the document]
Wovchena: Well, since you mess with `CommonCapSrc`
Without that `clone()` the whole `preparation()` body should be rewritten as
```cpp
    GAPI_Assert(first.empty());
    first = cap->read();
    GAPI_Assert(!first.data && "Couldn't grab the first frame");
```
and
```cpp
bool CommonCapSrc::pull(cv::gapi::wip::Data &data) {
    if (first.data) {
        data = std::move(first);
        first = cv::Mat{};
        return true;
    }
    cv::Mat frame = cap->read();
    data = frame;
    return !frame.data;
}
```

demos/tests/cases.py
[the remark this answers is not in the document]
Wovchena: It's there now. Please uncomment

demos/text_detection_demo/cpp_gapi/main.cpp
```diff
@@ -0,0 +1,386 @@
...
+        const auto decoderBandwidth  = FLAGS_b;
+        const auto decoderStartIndex = FLAGS_start_index;
+        const size_t tdNewInputWidth  = FLAGS_w_td;
+        const size_t tdNewInputHeight = FLAGS_h_td;
+        const size_t tdMaxRectsNum    = FLAGS_max_rect_num;
+        const bool centralCrop        = FLAGS_cc;
+        const float segmConfThreshold = static_cast<float>(FLAGS_cls_pixel_thr);
+        const float linkConfThreshold = static_cast<float>(FLAGS_link_pixel_thr);
+        const double trMinConfidence  = FLAGS_thr;
+        const bool trPadSymbolFirst   = FLAGS_tr_pt_first;
```
Wovchena, line 105: You can just read `FLAGS_*` themselves. No need to copy

demos/text_detection_demo/cpp_gapi/main.cpp
OrestChura: removed

demos/common/cpp/utils/src/performance_metrics.cpp
OrestChura: @Wovchena Not really. I meant, this function can be used in **demo** instead of counting frames there by creating local variable `int frameCount`

see main.cpp:339
```
if (videoWriter.isOpened() && outputFramesLimit == 0 || **metrics.getFrameCount()** <= outputFramesLimit)) {
    videoWriter.write(image);
}
```

OrestChura: From my point of view, it's a small but really convenient feature, and `PerfMetrics` lacks it - I think it can be used in almost all the OMZ demos where frames are being counted, as I said, "by hands" instead of just retrieving already existing info from `Metrics`

But maybe I should've suggested it in a separate PR, with suggesting it in demos where I see it can be helpful

demos/common/cpp_gapi/utils_gapi/src/stream_source.cpp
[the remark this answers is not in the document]
OrestChura: that's a good point, and actually these changes almost don't depend on this `clone()` fix

I think this rewriting deserves a separate PR, this PR is too big already. @Wovchena  Do you agree with a separate PR being opened?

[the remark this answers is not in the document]
Wovchena: Ok

demos/common/cpp_gapi/utils_gapi/src/stream_source.cpp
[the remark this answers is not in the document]
mpashchenkov: We have already cloned in prep. I agree with move(). But not clone here.
Also:
https://github.com/opencv/opencv/blob/3eeec4faae0c8b020d0efd30a51c1e97a0f444a5/modules/gapi/include/opencv2/gapi/streaming/cap.hpp#L72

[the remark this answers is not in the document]
Wovchena: yes, for some reason I though `CommonCapSrc` is a subclass of `GCaptureSource`...

demos/text_detection_demo/cpp_gapi/src/text_recognition.cpp
```diff
@@ -0,0 +1,209 @@
...
+    for (auto i = begin; i != end; i++) {
+        sum += std::exp((*i) - maxVal);
+    }
+    if (std::fabs(sum) < std::numeric_limits<double>::epsilon()) {
+        throw std::logic_error("sum can't be equal to zero");
+    }
+    *prob = 1.0f / static_cast<float>(sum);
+}
+
+std::vector<float> softmax(const std::vector<float>::const_iterator& begin,
```
mpashchenkov, line 55: You already have softmax. Can you use version from custom_kernels.cpp?

demos/common/cpp_gapi/utils_gapi/src/stream_source.cpp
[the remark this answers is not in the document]
Wovchena: I added a flag for `openImagesCapture()`: https://github.com/openvinotoolkit/open_model_zoo/pull/3190

demos/common/cpp/utils/src/performance_metrics.cpp
Wovchena: I see

Wovchena: I have an idea which includes complete hiding counters from `main()`: https://github.com/openvinotoolkit/open_model_zoo/pull/3110

I think you should wait adding `getFrameCount()` func

OrestChura: OK, then I'll remove these changes for now, leaving a counter in main()

demos/text_detection_demo/cpp_gapi/src/nets_configuration.cpp
```diff
@@ -0,0 +1,211 @@
...
+            tdOutputNames[1] = pair.first;
+            break;
+        case tdHorizBoxesLayerChannels:
+            tdOutputNames[0] = pair.first;
+            break;
+        case tdHorizLabelsLayerChannels:
+            tdOutputNames[1] = pair.first;
+            break;
+        default:
+            break;
```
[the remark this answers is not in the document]
OrestChura: No, here shouldn't be a throw, as the cycle iterates through all the outputs and therefore have to filter them for some of the nets
The throw happens after the `switch-case`, below
Maybe the default case should just be removed here

demos/text_detection_demo/cpp_gapi/src/text_recognition.cpp
OrestChura: I thought about that too - but they.. are different, for some reasons. One is "softmaxing" pairs in a vector, the other is doing it with all the vector values

But I think I can unite them, or, specifically speaking, use this function in the function of `custom_kernels`

demos/text_detection_demo/cpp_gapi/src/custom_kernels.cpp
```diff
@@ -0,0 +1,401 @@
...
+
+std::vector<cv::RotatedRect> maskToBoxes(const cv::Mat& mask, const float minArea,
+                                         const float minHeight, const cv::Size& imgSize) {
+    std::vector<cv::RotatedRect> bboxes;
+    double minVal = 0.;
+    double maxVal = 0.;
+    cv::minMaxLoc(mask, &minVal, &maxVal);
+    int maxBboxIdx = static_cast<int>(maxVal);
+    cv::Mat resizedMask;
+    cv::resize(mask, resizedMask, imgSize, 0, 0, cv::INTER_NEAREST);
```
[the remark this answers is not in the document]
OrestChura: Idk but it's done with this interpolation in cpp demo

demos/text_detection_demo/cpp_gapi/src/custom_kernels.cpp
```diff
@@ -0,0 +1,401 @@
...
+    for (size_t i = 0; i < pixelMask.size(); i++) {
+        pixelMask[i] = segmData[i] >= segmConfThreshold;
+        if (pixelMask[i]) {
+            points.emplace_back(i % w, i / w);
+            groupMask[i] = -1;
+        }
+    }
+    std::vector<uchar> linkMask(linkData.size(), 0);
+    for (size_t i = 0; i < linkMask.size(); i++) {
+        linkMask[i] = linkData[i] >= linkConfThreshold;
```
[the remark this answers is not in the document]
OrestChura: I don't think so - there is comparison of two floats producing a bool which then is stored in uchar. Seems good

demos/text_detection_demo/cpp_gapi/src/nets_configuration.cpp
```diff
@@ -0,0 +1,211 @@
...
+    throw std::runtime_error("Name '" + name + "' does not exist in the network");
+}
+template<class NamesArray>
+void checkIONames(const std::vector<std::string>& layers, const NamesArray& names) {
+    for (auto name : names) {
+        if (std::find(layers.begin(), layers.end(), name) == layers.end()) {
+            ThrowNameNotFound(name);
+        }
+    }
+}
```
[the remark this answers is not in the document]
OrestChura: I didn't separate these two functions because they're connected - the first is used only in the second

demos/common/cpp/utils/src/performance_metrics.cpp
OrestChura: Done

demos/tests/cases.py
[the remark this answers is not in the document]
OrestChura: OK, uncommented

demos/text_detection_demo/cpp_gapi/main.cpp
OrestChura: OK, removed unnecessary redefinitions

demos/text_detection_demo/cpp_gapi/main.cpp
```diff
@@ -0,0 +1,386 @@
...
+        }
+
+        custom::NetsConfig config(tdModelPath, trModelPath);
+        if (tdRequired) {
+            config.getTDinfo();
+            config.configureTD(tdDevice, tdNewInputWidth, tdNewInputHeight);
+            slog::info << "The Text Detection model " << tdModelPath << " is loaded to " <<
+                tdDevice << slog::endl;
+        }
+        if (trRequired) {
```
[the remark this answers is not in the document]
OrestChura: Yes, make sense

demos/text_detection_demo/cpp_gapi/src/nets_configuration.cpp
```diff
@@ -0,0 +1,211 @@
...
+        trNetwork = ie.ReadNetwork(trModelPath);
+    }
+}
+
+void ThrowNameNotFound(const std::string& name) {
+    throw std::runtime_error("Name '" + name + "' does not exist in the network");
+}
+template<class NamesArray>
+void checkIONames(const std::vector<std::string>& layers, const NamesArray& names) {
+    for (auto name : names) {
```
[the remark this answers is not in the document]
OrestChura: OK

demos/text_detection_demo/cpp_gapi/src/nets_configuration.cpp
```diff
@@ -0,0 +1,211 @@
...
+    if ("" != tdModelPath_) {
+        tdNetwork = ie.ReadNetwork(tdModelPath_);
+    }
+    if ("" != trModelPath) {
+        trNetwork = ie.ReadNetwork(trModelPath);
+    }
+}
+
+void ThrowNameNotFound(const std::string& name) {
+    throw std::runtime_error("Name '" + name + "' does not exist in the network");
```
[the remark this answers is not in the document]
OrestChura: you're right. in cpp demo there were multiple throws, but here it doesn't make sense

demos/text_detection_demo/cpp_gapi/src/nets_configuration.cpp
```diff
@@ -0,0 +1,211 @@
...
+}
+template<class NamesArray>
```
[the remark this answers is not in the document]
OrestChura: you're right. rewritten a bit

demos/text_detection_demo/cpp_gapi/src/nets_configuration.cpp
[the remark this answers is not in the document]
OrestChura: removed

demos/text_detection_demo/cpp_gapi/src/custom_kernels.cpp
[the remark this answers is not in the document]
mpashchenkov: Is mask 2d image? M B bilinear interpolation improves result. Up to you

demos/text_detection_demo/cpp_gapi/src/custom_kernels.cpp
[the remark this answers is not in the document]
mpashchenkov: Ok. I suggested uchar()

demos/text_detection_demo/cpp_gapi/src/custom_kernels.cpp
mpashchenkov: out --> labels --> infer. What shared memory do you want handle?

demos/text_detection_demo/cpp_gapi/src/custom_kernels.cpp
```diff
@@ -0,0 +1,408 @@
...
+        graph = std::make_shared<cv::GComputation>(cv::GIn(inDec, inHidden, feature),
+                                                   cv::GOut(outHidden, outDec));
+    }
+    static void run(const std::vector<cv::Mat>&  hiddens_,
+                    const std::vector<cv::Mat>&  features,
+                    const cv::gapi::GNetPackage& net,
+                    const size_t                 numClasses,
+                    const size_t                 endToken,
+                          std::vector<cv::Mat>&  res,
+                          cv::GComputation& graph) {
```
mpashchenkov, line 359: Does it make sense to keep `GCompiled` in state?

demos/text_detection_demo/cpp_gapi/src/custom_kernels.cpp
OrestChura: so, the problem of this string
`crop.reshape(1, blobShape).convertTo(out[i], CV_32F);`
is that the `crop` is the one which is being assigned to `out[i]`. Without `clone()`. So `out[i]` and `crop`  are bounded

Then, on the next cycle iteration, crop will be changed and assigned to `out[i+1]` - but, `out[i]` will also be changed.

mpashchenkov: Ok. What about reallocation? Have blobs and crops the same data pointers after `convertTo`?

demos/text_detection_demo/cpp_gapi/src/custom_kernels.cpp
mpashchenkov: Parameter for cv::compile_args for main graph + tag for parameter --> cv::gapi::getCompileArg in setup.
Or, does ArrayDesc come in runtime?

mpashchenkov: Is it for the net?

demos/text_detection_demo/cpp_gapi/src/custom_kernels.cpp
```diff
@@ -0,0 +1,408 @@
...
+        }
+
+        std::vector<cv::Mat> states(numDetectedLabels, cv::Mat({1}, CV_32FC1));
+        std::for_each(states.begin(), states.end(),
+                      [](cv::Mat& m){ m.dims = 1; *(m.begin<float>()) = 0; });
+        std::vector<cv::Mat> hiddens(hiddens_);
+
+        res.reserve(numDetectedLabels);
+        for (auto i = 0U; i < numDetectedLabels; i++) {
+            res.emplace_back(std::vector<int>{ maxDecodedSymbols, 1, static_cast<int>(numClasses) },
```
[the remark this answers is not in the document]
mpashchenkov: or create()  ?

demos/text_detection_demo/cpp_gapi/src/custom_kernels.cpp
```diff
@@ -0,0 +1,408 @@
...
+                            cv::gout(hidden, out),
+                            cv::compile_args(net));
+
+                auto maxElem = std::max_element(out.begin<float>(),
+                                                out.begin<float>() + numClasses);
+                auto argmax = static_cast<size_t>(std::distance(out.begin<float>(), maxElem));
+                for (size_t i = 0; i < numClasses; i++) {
+                    res[nLabel].begin<float>()[nSymbol * numClasses + i] = out.begin<float>()[i];
+                }
+                if (endToken == argmax) {
```
mpashchenkov, line 393: MB put this check after `auto argmax = static_cast...`.
Or `auto maxElem = std...` after check.

demos/text_detection_demo/cpp_gapi/src/text_recognition.cpp
```diff
@@ -0,0 +1,208 @@
...
+            }
+            alphabet += vocabLine;
+        }
+        if (alphabet.empty()) {
+            throw std::logic_error("File is empty: " + filename);
+        }
+        return true;
+    } else {
+        return false;
+    }
```
mpashchenkov, line 36: I suggest:
    return true;
}
return false;

Without else we save one line :)

demos/text_detection_demo/cpp_gapi/src/text_recognition.cpp
```diff
@@ -0,0 +1,208 @@
...
+    *prob = 1.0f / static_cast<float>(sum);
+}
+
+std::vector<float> softmax(const std::vector<float>::const_iterator& begin,
+                           const std::vector<float>::const_iterator& end) {
+    std::vector<float> prob(end - begin, 0.f);
+    std::transform(begin, end, prob.begin(), [](float x) { return std::exp(x); });
+    float sum = std::accumulate(prob.begin(), prob.end(), 0.0f);
+    for (int i = 0; i < static_cast<int>(prob.size()); i++)
```
mpashchenkov, line 60: for (auto& el : prob) {
    el /= sum;
}

demos/text_detection_demo/cpp_gapi/src/text_recognition.cpp
```diff
@@ -0,0 +1,208 @@
...
+    std::string result = "";
+    const int numClasses = alphabet.length();
+    *conf = 1;
+
+    for (auto it = data.begin() + startIdx * numClasses; it != data.end(); it += numClasses) {
+        int argmax;
+        float prob;
+
+        softmaxAndChoose(it, it + numClasses, &argmax, &prob);
+        (*conf) *= prob;
```
mpashchenkov, line 88: MB the reference will be simpler the pointer.

demos/text_detection_demo/cpp_gapi/src/text_recognition.cpp
```diff
@@ -0,0 +1,208 @@
...
+      (*conf) *= prob;
+
+      auto symbol = alphabet[argmax];
+      if (symbol != padSymbol) {
+          if (result.empty() || padPrev || (!result.empty() && symbol != result.back())) {
+            padPrev = false;
+            result += symbol;
+          }
+      } else {
+        padPrev = true;
```
mpashchenkov, line 121: Alignment

demos/text_detection_demo/cpp_gapi/src/text_recognition.cpp
```diff
@@ -0,0 +1,208 @@
...
+}
+
+std::string CTCBeamSearchDecoder(const std::vector<float>& data, const std::string& alphabet,
+                                 char padSymbol, double *conf, int bandwidth) {
+    const int numClasses = alphabet.length();
+
+    std::vector<BeamElement> curr;
+    std::vector<BeamElement> last;
+
+    last.push_back(BeamElement{std::vector<int>(), 1.f, 0.f});
```
mpashchenkov, line 135: MB `emplace_back(std::vector<int>(), 1.f, 0.f)`

demos/text_detection_demo/cpp_gapi/src/custom_kernels.cpp
```diff
@@ -0,0 +1,408 @@
...
+    return getAll(points, w, h, groupMask);
+}
+
+void processData(const float* ptr, const size_t dataTotal, const cv::MatSize& dataShape,
+                 std::vector<float>& out, std::vector<int>& newShape) {
+    std::vector<float> inData(ptr, ptr + dataTotal);
+    inData = transpose4d(inData, dimsToShape(dataShape), {0, 2, 3, 1});
+    softmax(inData);
+    out = sliceAndGetSecondChannel(inData);
+    newShape = { dataShape[0], dataShape[2], dataShape[3], dataShape[1] / 2 };
```
mpashchenkov, line 206: Why do you cut channels and use NHWC layout for out shapes?

mpashchenkov: Also I think N is extra here.
You use dataShape[2], dataShape[3] for segmNewShape and C for linkNewShape.

demos/text_detection_demo/cpp_gapi/main.cpp
```diff
@@ -0,0 +1,357 @@
...
+        // Getting information about frame from ImagesCapture
+        std::shared_ptr<ImagesCapture> cap = openImagesCapture(FLAGS_i, FLAGS_loop);
+        const cv::Mat tmp = cap->read();
+        cap.reset();
+        if (!tmp.data) {
+            throw std::runtime_error("Couldn't grab first frame");
+        }
+        cap = openImagesCapture(FLAGS_i, FLAGS_loop);
+
+        auto kernels  = custom::kernels();
```
mpashchenkov, line 212:
```suggestion
        auto kernels = custom::kernels();
```

demos/text_detection_demo/cpp_gapi/include/shared_functions.hpp
```diff
@@ -0,0 +1,12 @@
+// Copyright (C) 2022 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+//
+
+#pragma once
+
+#include <vector>
+#include <opencv2/core.hpp>
+
+namespace custom {
```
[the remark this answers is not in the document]
mpashchenkov: Does one function deserve two files?

demos/text_detection_demo/cpp_gapi/include/nets_configuration.hpp
```diff
@@ -0,0 +1,66 @@
...
+    void configureTR(const std::string& trDevice);
+    void getTRcompositeInfo(const std::array<std::string,2>& encoderOutputNames_,
+                            const std::array<std::string,3>& decoderInputNames_,
+                            const std::array<std::string,2>& decoderOutputNames_,
+                            const bool trPadSymbolFirst, const char kPadSymbol,
+                            const std::string& trSymbolsSet, const std::string& decoderType);
+    void configureTRcomposite(const std::string& trDevice);
+
+    cv::Size                  tdInputSize{};
+    size_t                    tdInputChannels = 0;
```
mpashchenkov, line 52:
```suggestion
    cv::Size tdInputSize{};
    size_t    tdInputChannels = 0;
```

demos/text_detection_demo/cpp_gapi/src/custom_kernels.cpp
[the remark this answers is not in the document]
OrestChura: vector's constructor can't be used as it binds all Mats-elements (shares the memory).
I came up with some refinements though, hope they're fine

demos/text_detection_demo/cpp_gapi/src/custom_kernels.cpp
OrestChura: No - `maxElem` is used for calculation `argmax` so it cannot be calculated after check.
And I cannot move the check upper - algorithm needs to write `out` down to `res` even if `endToken` has already been reached. Tried, doesn't work other way

demos/text_detection_demo/cpp_gapi/src/custom_kernels.cpp
OrestChura: `Why?` - I can't really give exhaustive answer. It's bound with math in deeper functions such as `transpose4d`.

I think you're right about N, thanks

demos/text_detection_demo/cpp_gapi/include/shared_functions.hpp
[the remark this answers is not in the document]
OrestChura: I don't see why to change namespace - all other stuff like kernels and their functions are in `custom` too

I need to share this function. If I finish combining `softmax`es - there would be 3 funcitons :)

demos/common/cpp_gapi/utils_gapi/src/stream_source.cpp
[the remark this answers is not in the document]
OrestChura: applied required changes from https://github.com/openvinotoolkit/open_model_zoo/pull/3190

demos/text_detection_demo/cpp_gapi/src/custom_kernels.cpp
OrestChura: `convertTo` does not change the source Mat, so they are different with output after operation.

The problem is/was in `vector`'s constructor which binds all elements

I've figured out a way to do it better (I think) - please take a look on this spot again

demos/text_detection_demo/cpp_gapi/main.cpp
```diff
@@ -0,0 +1,386 @@
...
+        if (gapiStreaming) {
+            continueExecution = pipeline.pull(getOutVector());
+        } else {
+            image = cap->read();
+            if (!image.data) {
+                throw std::runtime_error("Can't read an image from the input");
+            }
+            continueExecution = bool(image.data);
+        }
+        while (continueExecution) {
```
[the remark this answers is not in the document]
OrestChura: OK

demos/text_detection_demo/cpp_gapi/main.cpp
```diff
@@ -0,0 +1,386 @@
...
+            outVector += cv::gout(outPts);
+            return outVector;
+        };
+
+        /** ---------------- The execution part ---------------- **/
+        PerformanceMetrics metrics;
+        std::chrono::steady_clock::time_point beginFrame = std::chrono::steady_clock::now();
+        bool continueExecution;
+        if (gapiStreaming) {
+            continueExecution = pipeline.pull(getOutVector());
```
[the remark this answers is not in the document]
OrestChura: OK

demos/text_detection_demo/cpp_gapi/main.cpp
```diff
@@ -0,0 +1,386 @@
...
+        } else {
+            image = cap->read();
```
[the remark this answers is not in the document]
OrestChura: Refined the cycle handling (this and 2 previous comments)

Thanks for pointing this out!

demos/text_detection_demo/cpp_gapi/src/custom_kernels.cpp
OrestChura: Yes, probably..
Anyway - rewritten the kernel, it keeps `GCompiled` now. Added taking sizes from the net and passing it via `compile_args`

demos/text_detection_demo/cpp_gapi/src/text_recognition.cpp
OrestChura: OK, nice one :)

demos/text_detection_demo/cpp_gapi/src/text_recognition.cpp
OrestChura: I guess

I didn't look to these functions deep as they are almost untouched-ly copied from `cpp` demo.

demos/text_detection_demo/cpp_gapi/src/text_recognition.cpp
OrestChura: OK

demos/text_detection_demo/cpp_gapi/src/text_recognition.cpp
OrestChura: OK

demos/text_detection_demo/cpp_gapi/src/text_recognition.cpp
OrestChura: NO, `BeamElement` struct doesn't have a constructor so `emplace` doesn't work - and I don't want to change a lot in these functions, let's leave them as similar with `cpp` as it is reasonable

demos/text_detection_demo/cpp_gapi/src/custom_kernels.cpp
```diff
@@ -0,0 +1,408 @@
...
+                    }
+                }
+                neighbour++;
+            }
+        }
+    }
+    return getAll(points, w, h, groupMask);
+}
+
+void processData(const float* ptr, const size_t dataTotal, const cv::MatSize& dataShape,
```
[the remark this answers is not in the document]
OrestChura: yes, good one, changed

demos/text_detection_demo/cpp_gapi/main.cpp
OrestChura: OK

demos/text_detection_demo/cpp_gapi/include/nets_configuration.hpp
OrestChura: OK

Wovchena:
Jenkins please retry a build

demos/text_detection_demo/cpp_gapi/main.cpp
```diff
@@ -0,0 +1,349 @@
...
+int main(int argc, char* argv[]) {
+    try {
+        // This demo covers certain topologies only
+        // Parsing and validating input arguments
+        if (!ParseAndCheckCommandLine(argc, argv)) {
+            return 0;
+        }
+        const bool gapiStreaming = !FLAGS_gapi_regular_mode;
+        const bool tdEnabled  = !FLAGS_m_td.empty();
+        const bool trEnabled  = !FLAGS_m_tr.empty();
```
Wovchena, line 88:
```suggestion
        const bool tdEnabled = !FLAGS_m_td.empty();
        const bool trEnabled = !FLAGS_m_tr.empty();
```

demos/text_detection_demo/cpp_gapi/main.cpp
```diff
@@ -0,0 +1,349 @@
...
+            cv::GMat inDet, det1, det2;
+            if (1 == config.tdInputChannels) {
+                inDet = cv::gapi::BGR2Gray(in);
+            } else {
+                inDet = in;
+            }
+            std::tie(det1, det2) = cv::gapi::infer<nets::TextDetection>(inDet);
+            // ROI for each text piece
+            rrs = custom::DetectionPostProcess::on(det1, det2, size, config.tdInputSize,
+                                                   segmConfThreshold, linkConfThreshold,
```
Wovchena, line 161:
```suggestion
                                                   static_cast<float>(FLAGS_cls_pixel_thr),
                                                   static_cast<float>(FLAGS_link_pixel_thr),
```
So you can remove `segmConfThreshold` and `linkConfThreshold`

demos/text_detection_demo/cpp_gapi/main.cpp
```diff
@@ -0,0 +1,349 @@
...
+        const bool trEnabled  = !FLAGS_m_tr.empty();
+        const bool trComposite = std::string::npos != FLAGS_m_tr.find("encoder");
+        const float segmConfThreshold = static_cast<float>(FLAGS_cls_pixel_thr);
+        const float linkConfThreshold = static_cast<float>(FLAGS_link_pixel_thr);
+        const std::array<std::string,2> encoderOutputNames { FLAGS_out_enc_hidden_name,
+                                                             FLAGS_features_name        };
+        const std::array<std::string,3> decoderInputNames { FLAGS_in_dec_symbol_name,
+                                                            FLAGS_in_dec_hidden_name,
+                                                            FLAGS_features_name         };
+        const std::array<std::string,2> decoderOutputNames { FLAGS_out_dec_hidden_name,
```
Wovchena, line 97: You can move definition of `encoderOutputNames`, `decoderInputNames`, `decoderOutputNames` to scopes where the vars are actually used to be passed to `getTRcompositeInfo()`

demos/text_detection_demo/cpp_gapi/main.cpp
```diff
@@ -0,0 +1,349 @@
...
+                                                             FLAGS_out_dec_symbol_name  };
+        if (FLAGS_pad.length() != 1) {
+            throw std::invalid_argument("Pad symbol should be 1 character");
```
Wovchena, line 100: Move this check to `ParseAndCheckCommandLine()`

demos/text_detection_demo/cpp_gapi/src/text_recognition.cpp
```diff
@@ -0,0 +1,208 @@
...
+#include <numeric>
+#include <fstream>
+
+#include "text_recognition.hpp"
+
+bool fileExists(const std::string& filename) {
+    std::ifstream f(filename.c_str());
+    return f.good();
+}
+bool tryReadVocabFile(const std::string& filename, std::string& alphabet) {
```
Wovchena, line 16: You can let `tryReadVocabFile()` decide how to interpret `FLAGS_m_tr_ss`. This function should just return alphabet string given `FLAGS_m_tr_ss`. In simplest case it should just return passed `FLAGS_m_tr_ss`.
Please add similar function to cpp version

demos/text_detection_demo/cpp_gapi/main.cpp
```diff
@@ -0,0 +1,349 @@
...
+        }
+
+        custom::NetsConfig config(FLAGS_m_td, FLAGS_m_tr);
+        if (tdEnabled) {
+            config.getTDinfo();
+            config.configureTD(FLAGS_d_td, FLAGS_w_td, FLAGS_h_td);
+            slog::info << "The Text Detection model " << FLAGS_m_td << " is loaded to " <<
+                FLAGS_d_td << slog::endl;
+        }
+        if (trEnabled) {
```
Wovchena, line 119: I think you should put the body of this if and `if (tdEnabled)`'s body to `NetsConfig`'s constructor. Let the object manage itself.

demos/text_detection_demo/cpp_gapi/src/custom_kernels.cpp
```diff
@@ -0,0 +1,422 @@
...
+            processData(det2, segmData, segmNewShape);
+            auto mask = decodeImageByJoin(segmData, segmNewShape, linkData, linkNewShape,
+                                          segmThr, linkThr);
+            out = maskToBoxes(mask, kMinArea, kMinHeight, imgSize);
+        } else {
+            out = coordToBoxes(det1.ptr<float>(), det1.total(), kMinArea, kMinHeight,
+                               inputSize, imgSize);
+        }
+        if (maxRectsNum >= 0 && out.size() > maxRectsNum) {
+            std::sort(out.begin(), out.end(),
```
Wovchena, line 235: `std::nth_element()` should be enough

demos/text_detection_demo/cpp_gapi/main.cpp
```diff
@@ -0,0 +1,349 @@
...
+                config.getTRoutputInfo(FLAGS_tr_o_blb_nm, FLAGS_tr_pt_first, kPadSymbol,
+                                       FLAGS_start_index, trSymbolsSet);
+                config.configureTR(FLAGS_d_tr);
+            slog::info << "The Text Recognition model " << FLAGS_m_tr << " is loaded to " <<
+                FLAGS_d_tr << slog::endl;
+            }
+        }
+
+        std::vector<cv::RotatedRect> emptyBox =
+            { {cv::Point2f(0.0f, 0.0f), cv::Size2f(0.0f, 0.0f), 0.0f} };
```
Wovchena, line 140: Why not empty vector?

demos/text_detection_demo/cpp_gapi/main.cpp
```diff
@@ -0,0 +1,349 @@
...
+                outVector += cv::gout(outTexts);
+            }
+            outVector += cv::gout(outPts);
+            return outVector;
+        };
+
+        /** ---------------- The execution part ---------------- **/
+        PerformanceMetrics metrics;
+        std::chrono::steady_clock::time_point beginFrame = std::chrono::steady_clock::now();
+        while (true) {
```
Wovchena, line 241:
```suggestion
        while (true) {
            std::chrono::steady_clock::time_point beginFrame = std::chrono::steady_clock::now();
```
With that you can remove `beginFrame = std::chrono::steady_clock::now();` in the end

Wovchena (changes requested):

vladimir-dudnik:
Jenkins please retry a build

PR #3111, opened 2022-01-26
Update OV2.0 c++ configuration API

ivikhrev:
update to OV2.0 configuration API (although get_metric func is not implemented yet on OV side)


vladimir-dudnik:
will not build until 2022.1.550 OV build installed

vladimir-dudnik (approved):

eaidova:
> will not build until 2022.1.550 OV build installed

it is not quite true, python changes are not in package yet, they are not merged, so even update on 550 does not make this PR working.

@ivikhrev  is it possible separate python and c++ changes? C++ demo changes can be merged early, when we update package. it helps to unblock demo building validation on ov side again.

PR #3112, opened 2022-01-26
AC: migrate on configuration api 2.0

eaidova:
related to https://github.com/openvinotoolkit/openvino/pull/9927
configuration changes are backward compatible on tool side, no impact on validation if we merge before ov part will be expected


tools/accuracy_checker/openvino/tools/accuracy_checker/launcher/openvino_launcher.py
```diff
@@ -990,7 +996,7 @@ def release(self):
         if 'network' in self.__dict__:
             del self.network
         if 'infer_request' in self.__dict__:
-            del self.infer_request
+            del sel f.infer_request
```
eaidova, line 999:
```suggestion
            del self.infer_request
```

Anna-Mironova (approved):

PR #3133, opened 2022-01-31
Removed short suffix from COCO annotation files

Anna-Mironova:


eaidova (approved):

data/dataset_definitions.yml
```diff
@@ -12,7 +12,7 @@ datasets:
   - name: ms_coco_mask_rcnn_short_80_classes
     annotation_conversion:
       converter: mscoco_mask_rcnn
-      annotation_file: annotations/instances_val2017_short.json
+      annotation_file: annotations/instances_val2017.json
       has_background: True
       sort_annotations: True
     annotation: mscoco_mask_rcnn_short_80.pickle
```
eaidova, line 18: probably we need to remove short suffix from pickle name too for avoid users confusion

PR #3134, opened 2022-01-31
AC: fix resolving output name if provided tensor name

eaidova:


tools/accuracy_checker/openvino/tools/accuracy_checker/launcher/openvino_launcher.py
```diff
@@ -953,7 +959,7 @@ def get_infer_queue(self, log=True):
         if self.config.get('num_requests', 'AUTO') == 'AUTO':
             num_requests = self.auto_num_requests()
         else:
-            num_requests = self.num_requests or 0
+            num_requests = selfnum_requests or 0
```
eaidova, line 962:
```suggestion
            num_requests = self.num_requests or 0
```

Anna-Mironova (approved):

PR #3138, opened 2022-01-31
segmentation_demo.py: load default Pascal palette from file 

anzhella-pankratova:
The purpose is to not keep this palette in SegmentationVisualizer class itself, but load colors from file in palettes folder, representing this palette.
Also, rendering is updated in order to decrease the latency of rendering time (80ms->10ms)

### DIFF
--- a/demos/segmentation_demo/python/README.md
+++ b/demos/segmentation_demo/python/README.md
@@ -130,7 +130,7 @@ Running the application with the empty list of options yields the usage message 
 You can use the following command to do inference on CPU on images captured by a camera using a pre-trained network:
 
 ```sh
-    python3 segmentation_demo.py -d CPU -i 0 -m <path_to_model>/semantic-segmentation-adas-0001.xml
+    python3 segmentation_demo.py -d CPU -i 0 -at segmentation -m <path_to_model>/semantic-segmentation-adas-0001.xml
 ```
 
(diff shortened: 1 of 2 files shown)


vladimir-dudnik:
@anzhella-pankratova what is the purpose of this change? There will no default palette anymore and you'll have to provide one as a file (yet another mandatory param)?

anzhella-pankratova:
> @anzhella-pankratova what is the purpose of this change? There will no default palette anymore and you'll have to provide one as a file (yet another mandatory param)?

@vladimir-dudnik The behavior of the demo is not changed: if one does not provide any color palette, the demo will take colors from Pascal VOC dataset palette by default, the argument `--colors` is still optional.
The purpose is to not keep this palette in `SegmentationVisualizer` class itself, but load colors from file in `palettes` folder, representing this palette.

vladimir-dudnik (approved):

vladimir-dudnik:
@anzhella-pankratova should not we check if file not found by default path (for example demo copied out to another place)?

anzhella-pankratova:
> should not we check if file not found by default path (for example demo copied out to another place)?

In this case `get_palette_from_file()` will accept the non-existent file and raise `FileNotFoundError`, trying to open it. So I think the failure message by  the error is enough here

anzhella-pankratova:
Jenkins please retry a build

anzhella-pankratova:
Jenkins please retry a build

anzhella-pankratova:
Jenkins please retry a build

Wovchena (approved):

PR #3157, opened 2022-02-03
Python demos: unify core.read_model()

anzhella-pankratova:
The weights path (.bin file) can be omitted  in the python demos, since we expect this file to be located in the same folder as .xml file, so the reading should be simplified to `core.read_model(model_path)` in all demos.


demos/3d_segmentation_demo/python/3d_segmentation_demo.py
```diff
@@ -262,7 +262,7 @@ def main():
         raise AttributeError("Device {} do not support of 3D convolution. "
                              "Please use CPU, GPU or HETERO:*CPU*, HETERO:*GPU*")
 
-    # --------------------- 2. Read IR Generated by ModelOptimizer (.xml and .bin files) ---------------------
+    # --------------------- 2. Read IR Generated by ModelOptimizer (.xml file) ---------------------
```
Wovchena, line 265: I suggest remodeling such comments: there is `Reading model` print right below

anzhella-pankratova: Removed all comments

Wovchena (approved):

vladimir-dudnik (approved):

PR #3158, opened 2022-02-03
CONTRIBUTING.md: update demo args

Wovchena:


CONTRIBUTING.md
```diff
@@ -185,22 +185,19 @@ Deep Learning Inference Engine (IE) supports models in the Intermediate Represen
...
-Demos are required to support the following keys:
+Demos are required to support the following args:
 
- -  `-i "<input>"`: Required. An input to process. The input can usually be a single image, a folder of images or anything that OpenCV's `VideoCapture` can process.
- -  `-m "<path>"`: Required. Path to an .xml file with a trained model. If the demo uses several models at the same time, use other keys prefixed with `-m_`.
- -  `-d "<device>"`: Optional. Specifies a target device to infer on. CPU, GPU, HDDL or MYRIAD is acceptable. Default must be CPU. If the demo uses several models at the same time, use keys prefixed with `d_` (just like keys `m_*` above) to specify device for each model.
- -  `-no_show`: Optional. Do not visualize inference results.
+* `-h, --help`        show this help message and exit
+* `-m <MODEL FILE>`   path to an .xml file with a trained model. If the demo uses several models at the same time, use format `--m?*`.
+* `-i "<INPUT ID>"`   an input id to process. For vision tasks the input must be a single image, a folder of images, video file or camera id. For vision tasks the default value must be `0`
```
ieliz, line 196: Are these quotes on INPUT ID are necessary or they are just left from the previous version?

Wovchena: Removed

CONTRIBUTING.md
```diff
@@ -185,22 +185,19 @@ Deep Learning Inference Engine (IE) supports models in the Intermediate Represen
...
 
-Demos are required to support the following keys:
+Demos are required to support the following args:
 
- -  `-i "<input>"`: Required. An input to process. The input can usually be a single image, a folder of images or anything that OpenCV's `VideoCapture` can process.
- -  `-m "<path>"`: Required. Path to an .xml file with a trained model. If the demo uses several models at the same time, use other keys prefixed with `-m_`.
- -  `-d "<device>"`: Optional. Specifies a target device to infer on. CPU, GPU, HDDL or MYRIAD is acceptable. Default must be CPU. If the demo uses several models at the same time, use keys prefixed with `d_` (just like keys `m_*` above) to specify device for each model.
- -  `-no_show`: Optional. Do not visualize inference results.
+* `-h, --help`        show this help message and exit
+* `-m <MODEL FILE>`   path to an .xml file with a trained model. If the demo uses several models at the same time, use format `--m?*`.
```
anzhella-pankratova, line 195:
```suggestion
* `-m <MODEL FILE>`   path to an .xml file with a trained model. If the demo uses several models at the same time, use format `-m?*`.
```

Wovchena: `-m?*` is at least 2 symbol flag, so --

Wovchena: Though in fact we will support both

vladimir-dudnik:
comment on behalf of @andrew-zaytsev : still see Inference Engine mentioned - The following Inference Engine C++ API is used in the application:

CONTRIBUTING.md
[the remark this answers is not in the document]
Wovchena: Applied except for `specification of`: every arg specifies something

CONTRIBUTING.md
Wovchena: I don't think I see the point.
Your note
> better to have `-m` as a general rule

is satisficed by `-m <MODEL FILE>`   path to an .xml file with a trained model.`

The note
> comment in description, like "some demos might use extended syntax, like -m_det"

is implemented by `If the demo uses several models at the same time, use format --m?*`

vladimir-dudnik: point is that this "regular expression" like `?*` might be not obvious. While explaining that "might use extended syntax, like `-m_det` seems to be more clear

Wovchena: Updated. I also removed `_`. That allows to mimic `-nstreams` arg

anzhella-pankratova: The point of my suggestion was to left only one `-` because the documentation describes the flags for both c++ and python demos. There are no `--m*?` in c++ demos
According to our test, `--mdet` is not the common view of the flag. In almost demos we use `-m_det`, `-m_ag` etc.

Wovchena: I think we should change this style

anzhella-pankratova: Okay

vladimir-dudnik (approved):

CONTRIBUTING.md
```diff
@@ -185,22 +185,19 @@ Deep Learning Inference Engine (IE) supports models in the Intermediate Represen
...
- -  `-m "<path>"`: Required. Path to an .xml file with a trained model. If the demo uses several models at the same time, use other keys prefixed with `-m_`.
- -  `-d "<device>"`: Optional. Specifies a target device to infer on. CPU, GPU, HDDL or MYRIAD is acceptable. Default must be CPU. If the demo uses several models at the same time, use keys prefixed with `d_` (just like keys `m_*` above) to specify device for each model.
- -  `-no_show`: Optional. Do not visualize inference results.
+* `-h, --help`        show this help message and exit
+* `-m <MODEL FILE>`   path to an .xml file with a trained model. If the demo uses several models an extended syntax can be used, like `--mdet`
+* `-i <INPUT>`        input to process. For vision tasks the input might be a path to single image or video file, a path to folder of images, or numeric camera id. For vision tasks the default value must be `0`. For speech/audio tasks input is path to WAV file. For NLP tasks input might be a path to text file or just quoted sentence of text.
```
ivikhrev, line 196:
```suggestion
* `-i <INPUT>`        input to process. For vision tasks the input might be a path to single image or video file, a path to folder of images, or numeric camera id. For speech/audio tasks input is path to WAV file. For NLP tasks input might be a path to text file or just quoted sentence of text.
```
Why `'0'`? Currently default value just empty string https://github.com/openvinotoolkit/open_model_zoo/blob/3b8c2270fd2268a2aabf3c1c9d72592bfdafdae3/demos/common/cpp/utils/include/utils/default_flags.hpp#L10

Wovchena: That's my proposal :) For vision we can make a good guess just to show what a demo does

anzhella-pankratova (approved):

vladimir-dudnik (approved):

PR #3160, opened 2022-02-03
demos: merge requirements.txt

Wovchena:
The merge guarantees that Snyk scans all dependencies

### DIFF
--- a/demos/requirements.txt
+++ b/demos/requirements.txt
@@ -1,13 +1,18 @@
+inflect>=5.3.0
 librosa>=0.8.0
 matplotlib>=3.3.4
 motmetrics>=1.2.0
 nibabel>=3.2.1
 numpy>=1.16.6,<1.20; python_version <= "3.6"
 numpy>=1.16.6,<=1.21; python_version > "3.6"
+opencv-python==4.5.*
+openvino-dev
+ovmsclient
+pillow>=8.1.2
+pyyaml>=5.4.1
 scikit-learn~=0.24.1
 scipy~=1.5.4
 sympy>=1.8
-tokenizers~=0.10.1
 tensorboardX>=2.1
+tokenizers~=0.10.1
 tqdm>=4.54.1
-inflect>=5.3.0
--- a/demos/requirements.txt
+++ b/demos/requirements.txt
@@ -6,7 +6,7 @@ nibabel>=3.2.1
 numpy>=1.16.6,<1.20; python_version <= "3.6"
 numpy>=1.16.6,<=1.21; python_version > "3.6"
 opencv-python==4.5.*
-openvino-dev
+openvino
 ovmsclient
 pillow>=8.1.2
 pyyaml>=5.4.1
@@ -15,4 +15,5 @@ scipy~=1.5.4
 sympy>=1.8
 tensorboardX>=2.1
 tokenizers~=0.10.1
+tkinter
 tqdm>=4.54.1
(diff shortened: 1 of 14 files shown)


demos/requirements.txt
```diff
@@ -1,13 +1,18 @@
...
+openvino-dev
+ovmsclient
+pillow>=8.1.2
+pyyaml>=5.4.1
 scikit-learn~=0.24.1
 scipy~=1.5.4
 sympy>=1.8
-tokenizers~=0.10.1
 tensorboardX>=2.1
+tokenizers~=0.10.1
```
eaidova, line 17: probably it was not added explicitly, but there is also tkinter
https://github.com/openvinotoolkit/open_model_zoo/blob/master/demos/text_to_speech_demo/python/utils/gui.py#L2

demos/requirements.txt
```diff
@@ -1,13 +1,18 @@
+inflect>=5.3.0
 librosa>=0.8.0
 matplotlib>=3.3.4
 motmetrics>=1.2.0
 nibabel>=3.2.1
 numpy>=1.16.6,<1.20; python_version <= "3.6"
 numpy>=1.16.6,<=1.21; python_version > "3.6"
+opencv-python==4.5.*
+openvino-dev
```
eaidova, line 9: currently demos really depends on openvino-dev? If I right remember, it needs only if user wants to use downloader or mo, I can not to say that it is a strong requirement
only `openvino` package required for now

Wovchena: Updated

demos/requirements.txt
Wovchena: Added

eaidova:
@Wovchena as I can see, you include model api dependency to common list, but not remove file.
Potentially, these 2 dependencies lists can have differences in future.
What about use something like `-r common/python/requirements.txt` inside general demo requirements?

Wovchena:
Thanks!

eaidova (approved):

vladimir-dudnik (approved):
will not this bring difficulties in building appropriate environment for one who interested only in some demos?

Wovchena:
Who cares. Some demos didn't have requirements.txt even before.

vladimir-dudnik:
Those which did not have own requirements.txt probably were covered by general demos requirements.txt. But what is the root cause of this issue? Is that SNYK scan not able to scan requirements.txt in subfolders?

Wovchena:
Snyk requires to point to requirerements.txt explicitly. To avoid risks of forgetting to add new files for scanning we should prohibit adding new requirerements.txt

eaidova:
> will not this bring difficulties in building appropriate environment for one who interested only in some demos?

python packages installation not a big deal generally, even when user want to run one specific demo, it can change own preference and easily try to run another (as it is not required additional hard installation process). Demos code is not production ready solution, so I do not think that we need to take care a lot about compatibility with user environment. 
Beside that it will be easier track misalignment, duplication and conflicts inside (when I update numpy, I see that several demos had duplication with common requirements and misalignment in opencv version). Also we do not have procedure how to specify demo dependencies (some people modify common part e.g. tensorboardx and mot-metrics used only in one specific demo, but it included generally, others create own list and duplicate the same deps as in common (numpy, opencv and so on).

From user perspective it is easier to understand that you need install all in one, then additionally install something else if he want do something else

vladimir-dudnik:
ok, got this

PR #3164, opened 2022-02-04
Added converters for KITTI dataset

Anna-Mironova:


eaidova (approved):

tools/accuracy_checker/openvino/tools/accuracy_checker/annotation_converters/kitti_converter.py
```diff
@@ -0,0 +1,94 @@
...
+
+    def get_meta(self):
+        labels = read_txt(self.labels_file)
+        reversed_label_map = {}
+        label_map = {}
+        for idx, label_name in enumerate(labels, start=self.label_start):
+            label_map[idx] = label_name.lower()
+            reversed_label_map[label_name.lower()] = idx
+        meta = {'label_map': label_map}
+        return meta, reversed_label_map
```
eaidova, line 94: is it possible return only meta in this method?
get_meta used as part of api for getting dataset info without conversion, iy will be strange if somebody uses it and get not only meta but something else?

Anna-Mironova: Done

PR #3170, opened 2022-02-07
instance_segmentation_demo.py: use Python Model API

anzhella-pankratova:
Used `MaskRCNNModel` and `YolactModel` wrappers in the demo.  
Updated `Visualizer` class and move it `/common/python`, reuse it in text_spotting_demo.


Wovchena (changes requested):

demos/common/python/openvino/model_zoo/model_api/models/instance_segmentation.py
```diff
@@ -40,6 +42,10 @@ def parameters(cls):
                 default_value=0.5,
                 description='Probability threshold for detections filtering'
             ),
+            'labels': ListValue(description="List of class labels"),
+            'path_to_labels': StringValue(
+                description="Path to file with labels. Overrides the labels, if they sets via 'labels' parameter"
```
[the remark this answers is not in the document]
anzhella-pankratova: Updated

Wovchena (approved):

demos/instance_segmentation_demo/python/instance_segmentation_demo.py
```diff
@@ -18,205 +18,229 @@
...
-
-    if args.no_track:
-        tracker = None
-    else:
-        tracker = StaticIOUTracker()
-
-    if args.delay:
-        delay = args.delay
-    else:
-        delay = int(cap.get_type() in ('VIDEO', 'CAMERA'))
```
ivikhrev, line 134: Without` delay` it's hard to see the result on images, so I suggest leave it in code, without allowing modifying it from cmd:
https://github.com/openvinotoolkit/open_model_zoo/blob/0d23ffa744c6c6dabec6a3afeac3a023d11b946c/demos/classification_demo/python/classification_demo.py#L158

ivikhrev:
Also while I run the demo, I faced with error:
```
  File "C:\Users\ivikhrev\Documents\open_model_zoo\demos\instance_segmentation_demo\python\instance_segmentation_demo\tracker.py", line 40, in affinity
    intersection = np.count_nonzero(np.logical_and(history_mask, mask))
ValueError: operands could not be broadcast together with shapes (375,500) (333,500)
```
It seems tracker in demo doesn't work on images with different sizes, idk what could be the best solution here, but I suggest just add try-except in this [place](https://github.com/openvinotoolkit/open_model_zoo/blob/0d23ffa744c6c6dabec6a3afeac3a023d11b946c/demos/instance_segmentation_demo/python/instance_segmentation_demo/tracker.py#L40)
```
  try:
      intersection = np.count_nonzero(np.logical_and(history_mask, mask))
  except ValueError:
      intersection = 0
```

demos/instance_segmentation_demo/python/instance_segmentation_demo.py
anzhella-pankratova: Most of Python demos support folder with images as input, so the delay handling should be introduced in all demos, theoretically, since the images are changed quickly in this case. 
I would prefer to unify Python model API demos at the same time in order to align with classification_demo.py, and investigate this option in more details, if you agree with it

anzhella-pankratova:
@ivikhrev Thanks for the observation! 
The tracking assumes to work with video input with the static scene and the same resolution of frames. Folder with images might consists of not related to each other images, so in this option there are no objects to track, since they are different, and tracking should be disabled at all. 
I modified the tracker setting, it will be used only for video input.

demos/instance_segmentation_demo/python/instance_segmentation_demo.py
ivikhrev: ok

akorobeinikov (commented):
I think we need separate folder for `visualizers`, where we will be able to move all visualizers from demos. Also possible it make sense to do visualizers as part of modelAPI (it would be helpful for users). So maybe we should create a folder to start with and move `InstanceSegmentationVisualizer` there. What do you think about?

anzhella-pankratova:
> I think we need separate folder for `visualizers`, where we will be able to move all visualizers from demos. Also possible it make sense to do visualizers as part of modelAPI (it would be helpful for users). So maybe we should create a folder to start with and move `InstanceSegmentationVisualizer` there. What do you think about?

@akorobeinikov I agree with you, the separate folder would be useful. Now the visualizer is imported from demos/common/python/visualizers. What about its inclusion in Python Model API, we should discuss this option.

Currently I added the instance_segmentation.py file with `InstanceSegmentationVisualizer`. I think the folder might contain the visualizers in files named by the specific task. For example, object_detection.py will include a visualizer for predicted boxes, human_pose_estimation.py - for poses, classification.py - for labels etc.

akorobeinikov:
LGTM, only please update copyright date for all changed files

anzhella-pankratova:
> LGTM, only please update copyright date for all changed files

@akorobeinikov The year was updated

akorobeinikov (approved):

ivikhrev (approved):

vladimir-dudnik:
merge with know issues (non relevant to PR)

PR #3190, opened 2022-02-09
Add safe read

Wovchena:
Some VideoCapture backends continue owning the video buffer under cv::Mat. safe_copy forses to return a copy from read()
https://github.com/opencv/opencv/blob/46e1560678dba83d25d309d8fbce01c40f21b7be/modules/gapi/include/opencv2/gapi/streaming/cap.hpp#L72-L76
Unsuccessful read with throw exception, so no need to check cv::Mat returned at app level


demos/common/cpp/utils/src/images_capture.cpp
```diff
@@ -163,15 +167,26 @@ class VideoCapWrapper : public ImagesCapture {
...
+                    img = img.clone();
+                }
                 readerMetrics.update(startTime);
                 return img;
             }
             return cv::Mat{};
         }
         cv::Mat img;
-        if (!cap.read(img) && loop && cap.set(cv::CAP_PROP_POS_FRAMES, initialImageId)) {
+        bool success = cap.read(img);
```
anzhella-pankratova, line 179: There should also be the image cloning in case of `safe_read` after first frame reading, isn't it?

demos/gaze_estimation_demo/cpp/main.cpp
```diff
@@ -104,7 +104,7 @@ int main(int argc, char* argv[]) {
         std::string windowName = "Gaze estimation demo";
 
         std::unique_ptr<ImagesCapture> cap = openImagesCapture(
-            FLAGS_i, FLAGS_loop, 0, std::numeric_limits<size_t>::max(), stringToSize(FLAGS_res));
+            FLAGS_i, FLAGS_loop, EFFICIENT_READ, 0, std::numeric_limits<size_t>::max(), stringToSize(FLAGS_res));
```
anzhella-pankratova, line 107: EFFICIENT_READ is a default value for `safe_read`, no need to change the line

anzhella-pankratova (commented):
It seems `Can't read an image from the input` error was removed not in all demos, please verify it

demos/common/cpp/utils/src/images_capture.cpp
Wovchena: It's below: `img = img.clone();`

Wovchena:
> It seems `Can't read an image from the input` error was removed not in all demos, please verify it

Updated

demos/common/cpp/utils/src/images_capture.cpp
anzhella-pankratova: It's cloned if the condition is satisfied and we reach the end of the video

Wovchena: Updated

vladimir-dudnik:
non relevant issue with python object_detection_demo on person-vehicle-bike-detection-2003 model

vladimir-dudnik:
Jenkins please retry a build

demos/common/cpp/utils/include/utils/images_capture.h
```diff
@@ -9,6 +9,9 @@
 #include <opencv2/videoio.hpp>
 #include "utils/performance_metrics.hpp"
 
+constexpr static bool SAFE_READ = true;
+constexpr static bool EFFICIENT_READ = false;
```
ivikhrev, line 13: Would it better to use enum instead? This will allow to specify type for argument and rename  `safe_read ` arg to `read_type` for example, because now it's a bit clumsy:
```
bool safe_read=EFFICIENT_READ
```

Wovchena: Added

ivikhrev (approved):

demos/human_pose_estimation_demo/cpp/main.cpp
```diff
@@ -211,12 +211,9 @@ int main(int argc, char *argv[]) {
         }
 
         //------------------------------- Preparing Input ------------------------------------------------------
-        auto cap = openImagesCapture(FLAGS_i, FLAGS_loop);
+        auto cap = openImagesCapture(FLAGS_i, FLAGS_loop, FLAGS_nireq != 1);
```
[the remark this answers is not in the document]
Wovchena: Fixed

PR #3204, opened 2022-02-11
restore test case for detr-resnet50

vladimir-dudnik:
2022.1.579 verified to convert detr-resnet50


vladimir-dudnik:
Jenkins please retry a build

PR #3209, opened 2022-02-14
AC: Fixed outputs preparation in retinanet_tf2 adapter

Anna-Mironova:


tools/accuracy_checker/openvino/tools/accuracy_checker/adapters/retinanet.py
```diff
@@ -455,9 +459,10 @@ def unpack_labels(labels):
 
     def prepare_boxes_and_classes(self, raw, batch_id):
         boxes_outs, classes_outs = [], []
+        order = (1, 2, 0) if self.output_format == 'BHW' else (0, 1, 2)
```
eaidova, line 462: is it possible to bound input shape layout and output format without explicit parameter setting?
I mean e.g. if model input is NCHW, then BHW, else HWB

Anna-Mironova: Done

eaidova (approved):

PR #3210, opened 2022-02-14
Python demos: modify output result accessing in case of single output layer

anzhella-pankratova:
The usage of `output_tensor` as the dictionary key is safer and more convenient.


anzhella-pankratova:
Jenkins please retry a build

anzhella-pankratova:
Jenkins please retry a build

eaidova (approved):

PR #3215, opened 2022-02-14
text_spotting_demo: fix output layer name, restore the test

anzhella-pankratova:
In the latest IR version `text_features` name is used.


Wovchena (approved):
@VoronovaIntern, FYI

PR #3237, opened 2022-02-16
speech recognition dynamic version

akorobeinikov:


anzhella-pankratova (approved):

models/public/wav2vec2-base/accuracy-check.yml
```diff
@@ -2,7 +2,6 @@ models:
   - name: wav2vec2-base
     launchers:
       - framework: openvino
-        allow_reshape_input: true
```
eaidova, line 5: please keep AC config, it is still used in internal infra for validation models with static shape

demos/speech_recognition_wav2vec_demo/python/speech_recognition_wav2vec_demo.py
```diff
@@ -57,17 +57,16 @@ def __init__(self, core, model_path, input_shape, device, vocab_file):
         if len(model.inputs) != 1:
             raise RuntimeError('Wav2Vec must have one input')
         self.input_tensor_name = model.inputs[0].get_any_name()
-        model_input_shape = model.inputs[0].shape
+        model_input_shape = model.inputs[0].partial_shape
```
eaidova, line 60: as demo can work with custom models , I think we should also check that partial_shape is dynamic. In future model can be supported with different devices where dynamic shapes are not supported. Also for some scenarios it can be useful to compare demo results with dynamic and static shape, so possibly it can be useful if you allow switching between dynamic and static model using comand line parameter

models/public/wav2vec2-base/model.yml
```diff
@@ -107,7 +107,6 @@ conversion_to_onnx_args:
   - '--conversion-param=dynamic_axes={"inputs": {0: "batch_size", 1: "sequence_len"},
     "logits": {0: "batch_size", 1: "sequence_len"}}'
 model_optimizer_args:
-  - --input_shape=[1,30480]
```
eaidova, line 110: probably you need to update shape in model README.md too

models/public/wav2vec2-base/README.md
```diff
@@ -40,12 +42,13 @@ The converted model has the same parameters as the original model.
...
 #### Original model
 
-Per-token probabilities (after LogSoftmax) for every symbol in the alphabet, name - `logits`,  shape - `1, 95, 32`, output data format is `B, N, C`, where:
+Per-token probabilities (after LogSoftmax) for every symbol in the alphabet, name - `logits`,  shape - `B, N, 32`, output data format is `B, N, C`, where:
 
 - `B` - batch size
 - `N` - number of recognized tokens
 - `C` - alphabet size
 
+`B` and `N` dimensions can take different values, because model is dynamic. Alphabet size `C` is static and equal 32.
```
anzhella-pankratova, line 51:
```suggestion
`B` and `N` dimensions can take different values, because model is dynamic. Alphabet size `C` is static and equals 32.
```

demos/speech_recognition_wav2vec_demo/python/speech_recognition_wav2vec_demo.py
```diff
@@ -51,23 +53,25 @@ class Wav2Vec:
...
 
-    def __init__(self, core, model_path, input_shape, device, vocab_file):
+    def __init__(self, core, model_path, input_shape, device, vocab_file, dynamic_flag):
         log.info('Reading model {}'.format(model_path))
         model = core.read_model(model_path)
         if len(model.inputs) != 1:
             raise RuntimeError('Wav2Vec must have one input')
         self.input_tensor_name = model.inputs[0].get_any_name()
-        model_input_shape = model.inputs[0].shape
+        self.dynamic = model.is_dynamic() and dynamic_flag and device == 'CPU'
```
anzhella-pankratova, line 62: The check `device == 'CPU'` doesn't work for HETERO case, better to use check `device in set(parse_devices(device))` (if it is actually should be dynamic for HETERO case)

demos/speech_recognition_wav2vec_demo/python/speech_recognition_wav2vec_demo.py
```diff
@@ -40,6 +40,8 @@ def build_argparser():
                              "CPU, GPU, HDDL, MYRIAD or HETERO. "
                              "The demo will look for a suitable IE plugin for this device. Default value is CPU.")
     parser.add_argument('--vocab', help='Optional. Path to an .json file with encoding vocabulary.')
+    parser.add_argument('--dynamic_shape', default=False, action='store_true',
```
anzhella-pankratova, line 43:
```suggestion
    parser.add_argument('--dynamic_shape', action='store_true',
```

demos/speech_recognition_wav2vec_demo/python/speech_recognition_wav2vec_demo.py
```diff
@@ -40,6 +40,8 @@ def build_argparser():
...
+                        help='Optional. Using dynamic shapes for inputs and outputs of model.')
```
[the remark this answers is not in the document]
eaidova: agree, you have not a direct influence on the output, message looks confusing

demos/speech_recognition_wav2vec_demo/python/speech_recognition_wav2vec_demo.py
eaidova: I think we should not take care about which device used, if user says that he want run dynamic model and device does not support it, it get error from runtime, in this case, in future when other devices will support dynamic shapes you do not need modify demo for enabling for them

eaidova: BTW possibly we can add some info about dynamic models support into supported_devices

eaidova:
@anzhella-pankratova  @akorobeinikov  do you know why demo tests failed on instance segmentation demo with 
ZeroDevision error?

vladimir-dudnik:
Jenkins please retry a build

demos/speech_recognition_wav2vec_demo/python/speech_recognition_wav2vec_demo.py
```diff
@@ -51,23 +53,25 @@ class Wav2Vec:
...
 
-    def __init__(self, core, model_path, input_shape, device, vocab_file):
+    def __init__(self, core, model_path, input_shape, device, vocab_file, dynamic_flag):
         log.info('Reading model {}'.format(model_path))
         model = core.read_model(model_path)
         if len(model.inputs) != 1:
             raise RuntimeError('Wav2Vec must have one input')
         self.input_tensor_name = model.inputs[0].get_any_name()
-        model_input_shape = model.inputs[0].shape
+        self.dynamic = model.is_dynamic() and dynamic_flag
```
eaidova, line 62: I think user flag should have priority under model shape, if model is static, but dynamic flag you can specify reshape with dynamic shape (-1, -1)

eaidova (approved):

PR #3240, opened 2022-02-16
text_detection_demo: fix auto resize 

ivikhrev:
GPU plugin doesn't support auto resize feature. Make opencv resize as default.
p.s  `text-recognition-0015-encoder` model will fail with stack overflow error.

### DIFF
--- a/demos/text_detection_demo/cpp/src/cnn.cpp
+++ b/demos/text_detection_demo/cpp/src/cnn.cpp
@@ -13,9 +13,9 @@
 
 Cnn::Cnn(
     const std::string& modelPath, const std::string& modelType, const std::string& deviceName,
-    ov::Core& core, const cv::Size& new_input_resolution) :
+    ov::Core& core, const cv::Size& new_input_resolution, bool use_auto_resize) :
     m_modelPath(modelPath), m_modelType(modelType), m_deviceName(deviceName),
-    m_core(core), m_new_input_resolution(new_input_resolution),
+    m_core(core), m_new_input_resolution(new_input_resolution), use_auto_resize(use_auto_resize),
     m_channels(0), m_time_elapsed(0), m_ncalls(0)
 {
     slog::info << "Reading model: " << m_modelPath << slog::endl;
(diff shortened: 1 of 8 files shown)


demos/text_detection_demo/cpp/src/cnn.cpp
```diff
@@ -66,16 +66,19 @@ Cnn::Cnn(
 
     // we'd like to pass input image (NCWH, u8) to model input
```
vladimir-dudnik, line 67:
```suggestion
    // we'd like to pass input image (NHWC, u8) to model input
```

demos/text_detection_demo/cpp/src/cnn.cpp
```diff
@@ -64,18 +64,21 @@ Cnn::Cnn(
...
+    // we'd like to pass input image (NHWC, u8) to model input
     // and let OpenVINO do necessary conversions
+
     ppp.input().tensor()
         .set_element_type(ov::element::u8)
-        .set_layout({ "NHWC" })
-        .set_spatial_dynamic_shape();
+        .set_layout({ "NHWC" });
 
-    ppp.input().preprocess().convert_layout(input_layout);
```
vladimir-dudnik, line 74: why removed? will this work with IR v10?

vladimir-dudnik (approved):

demos/text_detection_demo/cpp/src/cnn.cpp
ivikhrev: layout conversion step is added when you set layout to tensor:
```
   ppp.input().tensor()
        .set_element_type(ov::element::u8)
        .set_layout({ "NHWC" });
```
We can see it in log:
```
[ INFO ] Preprocessor configuration:
[ INFO ] Input "Placeholder":
    User's input tensor: {1,768,1280,3}, [N,H,W,C], u8
    Model's expected tensor: {1,3,768,1280}, [N,C,H,W], f32
    Implicit pre-processing steps (2):
      convert type (f32): ({1,768,1280,3}, [N,H,W,C], u8) -> ({1,768,1280,3}, [N,H,W,C], f32)
      convert layout [N,C,H,W]: ({1,768,1280,3}, [N,H,W,C], f32) -> ({1,3,768,1280}, [N,C,H,W], f32)
```
no need to set it in preprocess.

Wovchena (approved):
You shouldn't set a default value for `auto_resize` in `Cnn` and the children because you never use that default value

demos/text_detection_demo/cpp/src/cnn.cpp
vladimir-dudnik: that was to make that obvious, but ok

vladimir-dudnik:
ignore not relevant issues

PR #2923, opened 2021-11-23
OV2.0 c++ smart classroom demo update

ieliz:
Still have problems with face recognition and face reidentification.

PR #3245, opened 2022-02-16
return size_t type to result in kuhn munkres

ivikhrev:
Undo changes done in https://github.com/openvinotoolkit/open_model_zoo/pull/3227 and https://github.com/openvinotoolkit/open_model_zoo/pull/2923. 
Changing `size_t` to `int` broke initial logic and causes next error:
`
[ ERROR ] OpenCV(4.5.2-openvino) C:\Users\ivikhrev\Documents\Work\openvino\temp\opencv_4.5.2\opencv\include\opencv2/core/mat.inl.hpp:872: error: (-215:Assertion failed) (unsigned)(i1 * DataType<_Tp>::channels) < (unsigned)(size.p[1] * channels()) in function 'cv::Mat::at' 
`
in some cases.

### DIFF
--- a/demos/common/cpp/utils/include/utils/kuhn_munkres.hpp
+++ b/demos/common/cpp/utils/include/utils/kuhn_munkres.hpp
@@ -31,7 +31,7 @@ public:
     /// \return Optimal column index for each row. -1 means that there is no
     /// column for row.
     ///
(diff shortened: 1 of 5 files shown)


vladimir-dudnik (approved):

vladimir-dudnik:
ignoring non relevant issues with INT8 instance-segmentation (IRs should be updated after CF)

PR #3246, opened 2022-02-16
Fix ModelYolo possible loss of data warning

Wovchena:
Cherry pick
6b126df45492c2316664163af6dcd921b2b58383

### DIFF
--- a/demos/common/cpp/models/include/models/detection_model_yolo.h
+++ b/demos/common/cpp/models/include/models/detection_model_yolo.h
@@ -27,14 +27,14 @@ protected:
     class Region {
     public:
         int num = 0;
-        int classes = 0;
+        size_t classes = 0;
         int coords = 0;
         std::vector<float> anchors;
-        int outputWidth = 0;
-        int outputHeight = 0;
+        size_t outputWidth = 0;
+        size_t outputHeight = 0;
 
         Region(const std::shared_ptr<ov::op::v0::RegionYolo>& regionYolo);
-        Region(int classes, int coords, const std::vector<float>& anchors, const std::vector<int64_t>& masks, int outputWidth, int outputHeight);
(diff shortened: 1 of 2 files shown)


demos/common/cpp/models/include/models/detection_model_yolo.h
```diff
@@ -27,14 +27,14 @@ class ModelYolo : public DetectionModel {
     class Region {
     public:
         int num = 0;
-        int classes = 0;
+        size_t classes = 0;
```
ivikhrev, line 30: It is still constructed from `int clasess` in 
https://github.com/openvinotoolkit/open_model_zoo/blob/b418da60e0c6aeb749b7acc9a9078b9c886a844b/demos/common/cpp/models/src/detection_model_yolo.cpp#L184-L188

demos/common/cpp/models/include/models/detection_model_yolo.h
```diff
@@ -27,14 +27,14 @@ class ModelYolo : public DetectionModel {
...
         int coords = 0;
         std::vector<float> anchors;
-        int outputWidth = 0;
-        int outputHeight = 0;
+        size_t outputWidth = 0;
```
ivikhrev, line 33: It's too
https://github.com/openvinotoolkit/open_model_zoo/blob/b418da60e0c6aeb749b7acc9a9078b9c886a844b/demos/common/cpp/models/src/detection_model_yolo.cpp#L356-L357

demos/common/cpp/models/include/models/detection_model_yolo.h
Wovchena: Updated

demos/common/cpp/models/include/models/detection_model_yolo.h
Wovchena: Updated

ivikhrev (approved):

vladimir-dudnik:
ignore non relevant issues

PR #3255, opened 2022-02-17
model_tools: add message about unsupported model precisions

jkamelin:
Fixes downloader message in case if specified precision is unsupported by the model (76778)

### DIFF
--- a/tools/model_tools/src/openvino/model_zoo/download_engine/downloader.py
+++ b/tools/model_tools/src/openvino/model_zoo/download_engine/downloader.py
@@ -213,6 +213,12 @@ class Downloader:
 
         reporter.print_group_heading('Downloading {}', model.name)
 
+        model_unsupported_precisions = self.requested_precisions - model.precisions
+        if model_unsupported_precisions:
+            reporter.print_section_heading('Model {} is not supported in {} precisions. '
+                + 'Skipping downloading for these precisions.', model.name, model_unsupported_precisions)
+            reporter.print()
+
         reporter.emit_event('model_download_begin', model=model.name, num_files=len(model.files))
 
         output = self.output_dir / model.subdirectory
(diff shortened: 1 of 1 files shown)


tools/model_tools/src/openvino/model_zoo/download_engine/downloader.py
```diff
@@ -213,6 +213,12 @@ def _download_model(self, reporter, session_factory, model, known_precisions: se
 
         reporter.print_group_heading('Downloading {}', model.name)
 
+        model_unsupported_precisions = self.requested_precisions - model.precisions
+        if model_unsupported_precisions:
+            reporter.print_section_heading('Model {} is not supported in {} precisions. '
+                + 'Skipping downloading for these precisions.', model.name, model_unsupported_precisions)
+            reporter.print()
```
Wovchena, line 220: I think the message should be an alternative to `reporter.print_group_heading('Downloading {}', model.name)`, not a follow up.

tools/model_tools/src/openvino/model_zoo/download_engine/downloader.py
```diff
@@ -213,6 +213,12 @@ def _download_model(self, reporter, session_factory, model, known_precisions: se
 
         reporter.print_group_heading('Downloading {}', model.name)
 
+        model_unsupported_precisions = self.requested_precisions - model.precisions
+        if model_unsupported_precisions:
```
Wovchena, line 217: If `self.requested_precisions & model.precisions` than you should't complain - at least something will be downloaded

Wovchena (changes requested):

tools/model_tools/src/openvino/model_zoo/download_engine/downloader.py
jkamelin: Done

Wovchena (approved):

PR #3259, opened 2022-02-17
Use : as delimiter for layout flag

ivikhrev:
https://github.com/openvinotoolkit/open_model_zoo/pull/3104#discussion_r802896207

### DIFF
--- a/demos/background_subtraction_demo/python/background_subtraction_demo.py
+++ b/demos/background_subtraction_demo/python/background_subtraction_demo.py
@@ -70,7 +70,7 @@ def build_argparser():
                       help='Optional. Background blur strength (by default with value 0 is not applied).')
     args.add_argument('--layout', type=str, default=None,
                       help='Optional. Model inputs layouts. '
-                           'Format "<layout>" or "<input1>:<layout1>,<input2>:<layout2>" in case of more than one input.'
+                           'Ex. NCHW or input0:NCHW,input1:NC in case of more than one input.'
                            'To define layout you should use only capital letters')
 
     infer_args = parser.add_argument_group('Inference options')
--- a/demos/background_subtraction_demo/python/background_subtraction_demo.py
+++ b/demos/background_subtraction_demo/python/background_subtraction_demo.py
@@ -70,8 +70,7 @@ def build_argparser():
                       help='Optional. Background blur strength (by default with value 0 is not applied).')
     args.add_argument('--layout', type=str, default=None,
                       help='Optional. Model inputs layouts. '
-                           'Ex. NCHW or input0:NCHW,input1:NC in case of more than one input.'
-                           'To define layout you should use only capital letters')
+                           'Ex. NCHW or input0:NCHW,input1:NC in case of more than one input.')
 
     infer_args = parser.add_argument_group('Inference options')
     infer_args.add_argument('-nireq', '--num_infer_requests', help='Optional. Number of infer requests.',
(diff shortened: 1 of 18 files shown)


vladimir-dudnik:
@ivikhrev would be good to add some details on new format to corresponding demo descriptions, which some challenging examples, like `--layout image_info:0:?c` if this is what you meant by this change

ivikhrev:
@vladimir-dudnik should layout flag support dynamic layout or layout with unknown dimensions? Because now it doesn't, and there are no any challenging examples, just straightforward `input1:NCHW,input2:NC`.  And I can replace `<layout>" or "<input1>:<layout1>,<input2>:<layout2>`  from help message with this one.

vladimir-dudnik:
@ivikhrev even if dynamic layout not supported (not sure, we already have [wav2vec](https://github.com/openvinotoolkit/open_model_zoo/blob/master/models/public/wav2vec2-base/model.yml) with dynamic shape) but at least an example like `--layout inage_info:0:nc` would be useful for those who will see this for the first time

demos/common/python/openvino/model_zoo/model_api/adapters/utils.py
```diff
@@ -59,17 +59,21 @@ def from_user_layouts(input_names: set, user_layouts: dict):
...
-        Parse layout parameter in format "input0[NCHW],input1[NC]" or "[NCHW]" (applied to all inputs)
+        Parse layout parameter in format "input0:NCHW,input1:NC" or "NCHW" (applied to all inputs)
         '''
         if not layout_regex:
             return None
         inputs_layouts = layout_regex.split(',')
         user_layouts = {}
         for layout in inputs_layouts:
-            if not re.fullmatch(r"[^\[\]]*\[[A-Z]+\]", layout):
+            if not re.fullmatch(r".*\:?[A-Z]+", layout):
```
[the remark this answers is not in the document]
ivikhrev: Done

demos/common/cpp/utils/src/args_helper.cpp
```diff
@@ -130,25 +130,23 @@ cv::Size stringToSize(const std::string& str) {
...
-        if (searchStr.empty() || searchStr.front() != ',') {
-            break;
-        }
-        searchStr = searchStr.substr(1);
-        startPos = searchStr.find_first_of('[');
+        searchStr.pop_back();
+        colonPos = searchStr.find_last_of(':');
     }
     if (!searchStr.empty()) {
         throw std::logic_error("Can't parse input layout string: " + layout_string);
```
Wovchena, line 152: I suggest `std::invalid_argument`.
Or `RuntimeError` for both languages

demos/background_subtraction_demo/python/background_subtraction_demo.py
```diff
@@ -70,7 +70,7 @@ def build_argparser():
                       help='Optional. Background blur strength (by default with value 0 is not applied).')
     args.add_argument('--layout', type=str, default=None,
                       help='Optional. Model inputs layouts. '
-                           'Format "[<layout>]" or "<input1>[<layout1>],<input2>[<layout2>]" in case of more than one input.'
+                           'Ex. NCHW or input0:NCHW,input1:NC in case of more than one input.'
                            'To define layout you should use only capital letters')
```
Wovchena, line 74: There is no such note for C++. `-d` and `--streams` require the same syntax, but they don't have such note either. I think `To define layout you should use only capital letters` should be removed

Wovchena (changes requested):

demos/common/cpp/utils/src/args_helper.cpp
ivikhrev: Changed to `std::invalid_argument`.

demos/background_subtraction_demo/python/background_subtraction_demo.py
ivikhrev: Removed.

Wovchena (approved):

akorobeinikov (approved):

PR #3289, opened 2022-02-24
Reduce test set for mask rcnn demo

ivikhrev:
`semantic-segmentation-adas` testset is used in segmentation demos also, so I decided to create separate smaller test set for this demo to avoid CI issues.


vladimir-dudnik (approved):

vladimir-dudnik:
Jenkins please retry a build

PR #3299, opened 2022-02-25
demos/cpp: remove IE, align lib location

Wovchena:


vladimir-dudnik (approved):

vladimir-dudnik:
currently we have three python modules which are built,
demos/common/python/monitors_extension
demos/human_pose_estimation_3d/python/pose_extractor
demos/speech_recognition_deepspeech_demo/python/ctcdecode-numpy

and in ctcdecode-numpy CmakeLists.txt there is comment, which show potential problem with building different python modules in one folder, because of __init__.py file
# The Python files have to be in the same directory as the native module,
# because a non-namespace package (i.e., a package with an __init__.py file)
# has to have all of its modules in the same directory.

the first two python modules does not have any files with non-unique names, but ctcdecode-numpy has __init__.py and if there will be need for other python module with __init__.py we have to use separate build folders for then and add these folders to PYTHONPATH

Wovchena:
The file is copied to the dedicated folder.

PR #3311, opened 2022-03-01
AC requirements: pin imagecodecs for py3.7 [release]

eaidova:


Anna-Mironova (approved):

eaidova:
fyi @slyubimt possibly we need update openvino-dev for avoid issues with installing for python3.7

PR #3328, opened 2022-03-03
Check intervals of labels array

alpkn:
Issue #80766
Prevent app crash when results are out of array bounds

### DIFF
--- a/demos/common/cpp/models/src/classification_model.cpp
+++ b/demos/common/cpp/models/src/classification_model.cpp
@@ -42,7 +42,11 @@ std::unique_ptr<ResultBase> ClassificationModel::postprocess(InferenceResult& in
 
     result->topLabels.reserve(scoresTensor.get_size());
     for (size_t i = 0; i < scoresTensor.get_size(); ++i) {
-        result->topLabels.emplace_back(indicesPtr[i], labels[indicesPtr[i]], scoresPtr[i]);
+        int ind = indicesPtr[i];
+        if (ind < 0 || ind >= labels.size()) {
+            throw std::runtime_error("Invalid index for the class label is found during postprocessing");
+        }
+        result->topLabels.emplace_back(ind, labels[ind], scoresPtr[i]);
     }
 
     return retVal;


alpkn:
> What index value is when this error occures? Maybe something wrong with that particular model (or topk operation for VPU)?

Index is random. Something is definitely wrong with run on VPU; will be further investigated

Wovchena (approved):

ivikhrev:
@alpkn we don't validate results of demos, so such fails could indicate on problem with models or plugins, so maybe we shouldn't change the behaviour? Otherwise demo will silently finishes and reports that everything is ok, while it isn't.
@Wovchena what do you think?

Wovchena:
Is the problem in topk?
If so @ivikhrev right. It's strange to expect and handle any value out of range.

alpkn:
> @alpkn we don't validate results of demos, so such fails could indicate on problem with models or plugins, so maybe we shouldn't change the behaviour? Otherwise demo will silently finishes and reports that everything is ok, while it isn't. 

Crash is not good indication of the improper behavior; assert/exception and graceful exit would be the proper one.
However such change will affect more codebase - could you think of it?
Meanwhile proper execution (destruction/etc) is needed for further investigation.

Wovchena:
The demo already performs out of range check on init:
https://github.com/openvinotoolkit/open_model_zoo/blob/8eef2b9ccb20cc7428e34ae753cd88b4c98f3ecb/demos/common/cpp/models/src/classification_model.cpp#L142

I don't think there exists a reasonable excuse which could be put to a comment explaining range check for an output which is never supposed to provide such values. The demo shouldn't be modified

+@vladimir-dudnik

alpkn:
> Is the problem in topk? If so @ivikhrev right. It's strange to expect and handle any value out of range.

The reason is yet to discover.

Invalid results from the network should be expected ones and the demo tool must be ready to handle them properly (i.e., don't crash at least)

ivikhrev:
> Invalid results from the network should be expected ones and the demo tool must be ready to handle them properly (i.e., don't crash at least)

Than just exception with corresponding message should be added like you said:

> Crash is not good indication of the improper behavior; assert/exception and graceful exit would be the proper one.

One exception isn't big changes 

> However such change will affect more codebase - could you think of it?

alpkn:
Have to weight options then: displaying improper value against application exit.

Usecase
* Invalid class value comes from the network. Invalid class is the class which does not have matching label.

Should demo 
* exit in such a case with appropriate message (caused by an exception)
* display the class as not having matching label (there still may be valid other classes)


@ivikhrev @Wovchena @vladimir-dudnik +someoneelse

vladimir-dudnik:
@alpkn I agree with @Wovchena and @ivikhrev demo should not try to handle unexpected values in the ways other than just throw exception with message that unrecoverable error met during execution.

alpkn:
> just throw exception with message that unrecoverable error met during execution.

ok, I see your choice.
I emphasize that if at least one invalid class is returned (that has no matching label) then the demo will not show any valid classes to the user at all.

vladimir-dudnik:
@alpkn you can't rely on validity of classes in case you get corrupted output from model, there is nothing app can do about this

alpkn:
> @alpkn you can't rely on validity of classes in case you get corrupted output from model, there is nothing app can do about this

The corruption of the data is out of the question. It is only about the design of the demo and how it works with the returned data which you choose here.

vladimir-dudnik:
@alpkn did not get your point. All this PR is because of corrupted output data, is not it? Is there any issue with demo when model output is valid?

alpkn:
The data which does not fill into the range of labels is not necessarily corrupted, it might be only invalid result of the network being evaluated.
And demo application must be ready to take such data and handle it properly. It's just as simple as that.

vladimir-dudnik:
@alpkn there is no way to handle invalid results properly, invalid results, simple by definitions, it is result which is not valid.

alpkn:
> @alpkn there is no way to handle invalid results properly, invalid results, simple by definitions, it is result which is not valid.

There are at least couple of ways to properly handle invalid results (and appcrash is not among them): https://github.com/openvinotoolkit/open_model_zoo/pull/3328#issuecomment-1057949809

vladimir-dudnik:
@alpkn I agree that app should not crash, in case of invalid output. When invalid output detected we should show error message and exit.

alpkn:
Switched the fix to the option
* exit in such a case with appropriate message (caused by an exception)

ivikhrev (approved):

PR #3358, opened 2022-03-11
Work around Layout == str issue

ngaloppo:
The Layout utility datatype was recently introduced. However, it seems
to be polymorphic depending on the way it was constructed: sometimes
it's just a string, sometimes it has a data member called "layout".

This is just a fix to make OVMSAdapter work with the most common use
case. There is a bigger fix needed to patch all code paths in
OVMSAdapter and OpenvinoAdapter.


vladimir-dudnik:
cc @nignatovsky mp4 codec on CI

anzhella-pankratova (approved):

vladimir-dudnik:
merge with known CI issue (MP4 codec)

PR #3361, opened 2022-03-14
AC: Added support audio in FLAC format

Anna-Mironova:


tools/accuracy_checker/openvino/tools/accuracy_checker/data_readers/audio_readers.py
```diff
@@ -183,3 +188,34 @@ def read_ascii_mat(fd):
...
+        if isinstance(sf, UnsupportedPackage):
+            sf.raise_error(self.__provider__)
+        super().configure()
+        self.dtype = self.get_value_from_config('dtype')
+
+    def read(self, data_id):
+        data_path = self.data_source / data_id if self.data_source is not None else data_id
+        data, samplerate = sf.read(data_path, dtype=self.dtype)
+        channels = 1
+        data = data.reshape(-1, channels).T
```
eaidova, line 217: why do we need this step?
audio can have more than 1 channel

eaidova (approved):

PR #3394, opened 2022-03-23
Python Model API: support OVMS adapter in openpose wrapper

anzhella-pankratova:
- Add "pooled_heatmaps" node to the model graph if the model_adapter is `OpenvinoAdapter`. In other cases, pooled_heatmaps are calculated using numpy

- Add `--adapter` key to human_pose_estimation_demo.py


openmodelzooci:

PR #3405, opened 2022-03-24
Performance scope: add demos with dynamism

anzhella-pankratova:
The PR extends performance scenario with demos supporting dynamic shapes. 

Also, the model precisions are written to .csv tables, because we might launch performance job with different precisions and it should be highlighted.

### DIFF
--- a/demos/tests/cases.py
+++ b/demos/tests/cases.py
@@ -15,6 +15,7 @@
 import collections
 import itertools
 import sys
+from copy import deepcopy
 
 from args import (
     DataDirectoryArg, DataDirectoryOrigFileNamesArg, DataPatternArg,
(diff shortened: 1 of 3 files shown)


anzhella-pankratova:
Jenkins please retry a build

vladimir-dudnik (approved):

demos/tests/cases.py
```diff
@@ -50,6 +52,8 @@ def models_lst_path(self, source_dir):
     def device_args(self, device_list):
         if len(self.device_keys) == 0:
             return {'CPU': []}
+        if self.supported_devices:
+            device_list = self.supported_devices
```
Wovchena, line 56: Should this be an intersection of `device_list` and `self.supported_devices`?

anzhella-pankratova: Updated

Wovchena (approved):

PR #3424, opened 2022-03-30
model_tools: update telemetry data

jkamelin:
Added version, type of exception, updated precision sending. Added `selection_mode`: `None` to count usage of tools without params.


vladimir-dudnik (approved):

tools/model_tools/src/openvino/model_zoo/_common.py
```diff
@@ -135,25 +135,30 @@ def get_package_path(python_executable, package_name):
...
 def get_version():
-    if VERSION_FILE and VERSION_FILE.is_file():
-        with VERSION_FILE.open('r') as version_file:
-            version = version_file.readline().rstrip('\n')
-        return f'commit {version}'
-    else:
-        return 'unknown'
+    try:
+        from openvino.runtime import get_version
+        version = get_version()
```
eaidova, line 140: you can get infinity cycle with overloading and mixing inner and outer scope, let's clarify
```suggestion
        from openvino.runtime import get_version as ov_get_version
        version = ov_get_version()
```

tools/model_tools/src/openvino/model_zoo/_common.py
```diff
@@ -135,25 +135,30 @@ def get_package_path(python_executable, package_name):
...
+        version_match = re.match(r"^([0-9]+).([0-9]+)*", version)
```
eaidova, line 141: it maybe not be very robust and informative if you provide ov version instead of your own (e.g. using development state of the tool from Github with already released ov package), can we combine ov version and your as common solution?

tools/model_tools/src/openvino/model_zoo/_common.py
jkamelin: Applied

tools/model_tools/src/openvino/model_zoo/_common.py
jkamelin: Yes, thanks, updated version

eaidova (approved):

PR #3459, opened 2022-04-08
G-API Cascade detection demo

TolyaTalamanov:


dmatveev:
@TolyaTalamanov can you  please remind the status of this -- do we still need one?

PR #3469, opened 2022-04-27
Update blob parsing to match master

aleckey:
An error is occuring when loading blobs due to iobuffer. This is not present on master branch.
_TypeError: a bytes-like object is required, not '_io.BufferedReader'_

This PR updates the release branch with the same code present on master at this line


aleckey:
Duplicate - not required

PR #3476, opened 2022-05-20
fix TOC reference for nanodet

kblaszczak-intel:
fix based on documentation build errors


dtrawins (approved):
Changes passed documentation tests

PR #3488, opened 2022-06-28
Fix the run_tests.py terminated with exception when timeout occurs.

yangwang201911:
Fix the run_tests.py terminated with exception when timeout occurs.
Signed-off-by: Wang, Yang <yang4.wang@intel.com>

### DIFF
--- a/demos/tests/run_tests.py
+++ b/demos/tests/run_tests.py
@@ -333,10 +333,9 @@ def main():
                             execution_time = timeit.default_timer() - start_time
                             demo.parse_output(output, test_case, device)
                         except (subprocess.CalledProcessError, subprocess.TimeoutExpired) as e:
-                            if type(e) == subprocess.TimeoutExpired:
+                            output = e.output
+                            if not isinstance(output, str):
                                 output = e.output.decode('utf-8')
-                            else:
-                                output = e.output
                             if isinstance(e, subprocess.CalledProcessError):
                                 exit_msg = f'Exit code: {e.returncode}\n'
                             elif isinstance(e, subprocess.TimeoutExpired):
--- a/demos/tests/run_tests.py
+++ b/demos/tests/run_tests.py
@@ -333,7 +333,10 @@ def main():
                             execution_time = timeit.default_timer() - start_time
                             demo.parse_output(output, test_case, device)
                         except (subprocess.CalledProcessError, subprocess.TimeoutExpired) as e:
-                            output = e.output
+                            if type(e) == subprocess.TimeoutExpired:
+                                output = e.output.decode('utf-8')
+                            else:
+                                output = e.output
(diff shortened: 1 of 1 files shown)


demos/tests/run_tests.py
```diff
@@ -333,7 +333,10 @@ def option_to_args(key, value):
                             execution_time = timeit.default_timer() - start_time
                             demo.parse_output(output, test_case, device)
                         except (subprocess.CalledProcessError, subprocess.TimeoutExpired) as e:
-                            output = e.output
+                            if type(e) == subprocess.TimeoutExpired:
```
Wovchena, line 336: I think it's handled by `elif isinstance(e, subprocess.TimeoutExpired):` below

Wovchena (changes requested):

demos/tests/run_tests.py
yangwang201911: Yes. The exception found serval days ago can't be reproduced now. So Close this PR now.

yangwang201911:
Can't reproduced now.

demos/tests/run_tests.py
yangwang201911: Failure appeared again as below caused by the timeout. I'm not sure if reopen this PR again? @Wovchena
image

demos/tests/run_tests.py
```diff
@@ -338,8 +338,7 @@ def option_to_args(key, value):
                                 exit_msg = f'Exit code: {e.returncode}\n'
                             elif isinstance(e, subprocess.TimeoutExpired):
                                 exit_msg = f'Command timed out after {e.timeout} seconds\n'
-                            output += exit_msg
-                            print(output)
+                            print('{}\n{}'.format(output, exit_msg))
```
Wovchena, line 341: `output` var is used to save log file below: `write_log(output, args.log_file)`. So you need this `output += exit_msg`.

If I understand correctly, `TimeoutExpired.output` is of bytes type for you. Maybe you can cast it to str before that `+=`

Please, add a comment to code with link to a bug https://bugs.python.org/issue43431 to explain why we have such cast.

Wovchena (changes requested):

demos/tests/run_tests.py
yangwang201911: > If I understand correctly, `TimeoutExpired.output` is of bytes type for you. Maybe you can cast it to str before that `+=`
> 
Yes, you are right. `output `var is bytes type under that host env when timeout happened. While don't know why the `output ` is Str type on my host when timeout happened.

Wovchena: Check the bug. It appears on python3.8+ may be you have python3.7 or 3.6 on your host

PR #3489, opened 2022-06-30
 Fix NaN issue for exp() in softmax function

goodsong81:
Fixing a issue regarding 86756.
This change would enhance numerical stability by preventing NaN output from softmax by shifting input range of exp() input.


goodsong81:
Not sure of target branch. Will set 'Ready for review' flag once the target (master or releases/2022/SCv1.1) decided.

eaidova (approved):

PR #3491, opened 2022-07-06
[NOTFORMERGE] Update nltk version in requirement to resolve SNYK scan

byungilm:
+ Clear SNYK-PYTHON-NLTK-1932014

Signed-off-by: Min, Byungil <byungil.min@intel.com>

### DIFF
--- a/tools/accuracy_checker/requirements.in
+++ b/tools/accuracy_checker/requirements.in
@@ -36,7 +36,7 @@ tokenizers>=0.10.1;python_version>="3.7"
 transformers>=4.5,<4.17;python_version<"3.7"
 transformers>=4.5;python_version>="3.7"
 
-nltk>=3.5
+nltk>=3.6.6
 
 # DNA sequence matching
 parasail>=1.2.4


eaidova:
@byungilm please stay files in ci folder as is or use special script https://github.com/openvinotoolkit/open_model_zoo/blob/master/ci/update-requirements.py

ci folder is not a part of openvino package, it  is only OMZ precommit validation infra dependencies and they should have pinned version, relaxation (e.g. >=) are not accepted

byungilm:
> @byungilm please stay files in ci folder as is or use special script for update https://github.com/openvinotoolkit/open_model_zoo/blob/master/ci/update-requirements.py
> 
> ci folder is not a part of openvino package, it is only OMZ precommit validation infra dependencies (should be excluded from snyk scans if I right remember) and they should have pinned version, relaxation (e.g. >=) are not accepted

Thanks for the info! I will remove the update from ci folder.
How about tools/accuracy_checker ? Does the tools have have this kind of limit?

eaidova:
> How about tools/accuracy_checker ? Does the 'tools' have this kind of limit?

For accuracy checker, the only recommendation that version should be aligned with other openvino components, so I believe it is fine

byungilm:
Not covered at this release. Closed.

PR #3495, opened 2022-07-13
Fixing headers and links in Datasets.

sgolebiewski-intel:
This PR addresses CVS-86927.

Hyperlinks are not rendered when used as a sole text in headers. 
Additionally, it appears that using nested markdown (backquotes within square brackets in hyperlinks) also is problematic

image

### DIFF
--- a/data/datasets.md
+++ b/data/datasets.md
@@ -11,18 +11,18 @@ More detailed information about using predefined configuration files you can fin
 
 ## ImageNet
 
-### How download dataset
+### Download dataset
 
-To download images from ImageNet, you need to have an account and agree to the Terms of Access. Follow the steps below:
-1. Go to the [ImageNet](http://www.image-net.org/) homepage
-2. If you have an account, click `Login`. Otherwise, click `Signup` in the right upper corner, provide your data, and wait for a confirmation email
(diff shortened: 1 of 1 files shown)


kblaszczak-intel:
Could we have links to the pages moved under the headers? Something like:

## ImageNET
See the [ImageNet home page](http://image-net.org)

This way we can be sure not to lose the link... I would prefer to unify the "how to download" sections to include "go to www" as step 1. But it may be a future improvement.

Also, we could improve language a bit while we're at it :) For example, "To download WIDER Face dataset:" seems enough as a call to action.

sgolebiewski-intel:
> Also, we could improve language a bit while we're at it :) For example, "To download WIDER Face dataset:" seems enough as a call to action.

I'm thinking about removing even such sentences, since step 1 reads: "Go to .... website" or "Download ... ". 

Of course that would not apply to such cases as ImageNet, where there's info about registering etc.

sgolebiewski-intel:
I went for "unifying 'how to download' sections to include 'go to www' as step1" where possible. So all the links to home pages/websites have been brought back.

kblaszczak-intel (approved):
Great! Looks nice.

AdamTumi (approved):

PR #3500, opened 2022-07-21
[Master]openvino.model_zoo.model_api.models.model.WrapperError: There is no model with name "blur_segmentation" in list

JihwanEom:
### Summary
- Resolve 84981, include similar issue for classification task.


demos/common/python/openvino/model_zoo/model_api/models/classification.py
```diff
@@ -23,11 +23,11 @@
 
 
 class Classification(ImageModel):
-    __model__ = 'Classification'
+    __model__ = 'ote_classification'
```
eaidova, line 26: I believe these changes break classification demo, which uses this model wrapper.
Why do you need 2 additional outputs?
Can we just relax check of outputs if prediction behavior is not changed, just allow ignoring unused for getting classification results outputs?

demos/common/python/openvino/model_zoo/model_api/models/blur.py
```diff
@@ -0,0 +1,74 @@
...
+    def parameters(cls):
+        parameters = super().parameters()
+        parameters.update({
+            'soft_threshold': NumericalValue(default_value=0.5, min=0.0, max=1.0),
+            'blur_strength': NumericalValue(value_type=int, default_value=1, min=0, max=25)
+        })
+
+        return parameters
+
+    def _check_io_number(self, number_of_inputs, number_of_outputs):
```
eaidova, line 41: why do you ignore number of inputs/outputs?
Does it mean that it can be any?

demos/common/python/openvino/model_zoo/model_api/models/classification.py
JihwanEom: Why do you need 2 additional outputs? 
=> by this error : `openvino.model_zoo.model_api.models.model.WrapperError: None: There is no model with name "ote_classification" in list: bert, bert-named-entity-recognition, bert-embedding, bert-question-answering, Robust-video-matting, Background-matting, centernet, DETR, CTPN, FaceBoxes, RetinaFace, RetinaFace-PyTorch, SSD, Ultra_LightWeight_Face_Detection, YOLO, YOLOV4, YOLOF, YOLOX, YOLOv3-ONNX, Deblurring, HPE-assosiative-embedding, MaskRCNN, Yolact, Segmentation, Salient_Object_Detection, MonoDepth, OpenPose ` 

Can we just relax check of outputs if prediction behavior is not changed, just allow ignoring unused for getting classification results outputs?
=> Yes, I'm OK with your suggestion. Thank you!

eaidova: Wrapper error explains only why you tried to rename `__model__`, but my question about this check:
`self._check_io_number(1, 3)`.
I'm not sure that renaming on OMZ side is good idea, because there is nothing changed from original classification wrapper, just small check which can be generalized to accept both 1 and 3 outputs if it is necessary.

demos/common/python/openvino/model_zoo/model_api/models/blur.py
```diff
@@ -0,0 +1,74 @@
...
+        pass
+
+    def _get_outputs(self):
+        layer_name = 'output'
```
eaidova, line 45: possibly need check that layer with this name exists

demos/common/python/openvino/model_zoo/model_api/models/blur.py
eaidova: hmm... this class inherited from our Segmentation model wrapper, where preprocessing step expects that model have 1 inputs. Probably you can just call basic method with parameters
(1, -1). -1 is specific value which speak that number of outputs should be ignored

demos/common/python/openvino/model_zoo/model_api/models/classification.py
```diff
@@ -23,11 +23,11 @@
...
 
     def __init__(self, model_adapter, configuration=None, preload=False):
         super().__init__(model_adapter, configuration, preload)
-        self._check_io_number(1, 1)
+        self._check_io_number(1, 3)
```
eaidova, line 30:
```suggestion
        self._check_io_number(1, (1, 3))
```

demos/common/python/openvino/model_zoo/model_api/models/classification.py
JihwanEom: The reason why I changed to get 3 outputs is from here : https://github.com/openvinotoolkit/model_preparation_algorithm/blob/develop/mpa/modules/models/classifiers/sam_classifier.py#L276, OTE classifier outputs (output_logit, feature_vector, saliency_map) when we export the model. 

For `ote_classification`, I agree with your suggestion is more reasonable for me, I'll review OTE side code to fit `ote_classification` -> `classification`

demos/common/python/openvino/model_zoo/model_api/models/blur.py
JihwanEom: Thank you for your comment! I'll also check this part and investigate if there is side effect for this :)

demos/common/python/openvino/model_zoo/model_api/models/classification.py
JihwanEom: Hi @eaidova, I think OTE need new template instead of renaming `__model__`. 
OTE is using OteClassification which inherits Classification class and overrides quiet many methods : https://github.com/openvinotoolkit/training_extensions/blob/develop/external/deep-object-reid/torchreid_tasks/model_wrappers/classification.py#L30

Can I create new file like `blur.py` for OTE classification? How do you think about this?

eaidova: @JihwanEom it is up to your, generally, we prefer to have all wrappers for one task is one file (ote_classification is still classification, is not it?), exception from this rule is case when model class represent specific topology or family (e.g. yolo)

demos/common/python/openvino/model_zoo/model_api/models/blur.py
JihwanEom: `self.outputs = {'output': Metadata(names={'output'}, shape=[1, 2, 544, 544], layout='', precision='f32', type='Softmax', meta={'axis': 1})}`

JihwanEom:
Hi @eaidova , we found some approach to resolve on OTE side. Can I close this PR?

eaidova:
> Hi @eaidova , we found some approach to resolve this issue on OTE side. Can I close this PR?

@JihwanEom sure, no problem :)

JihwanEom:
> > Hi @eaidova , we found some approach to resolve this issue on OTE side. Can I close this PR?
> 
> @JihwanEom sure, no problem :)

Thank you for kind and quick reply! I'll close this PR :)

PR #3513, opened 2022-08-03
demos/cpp: support OpenCV3

Wovchena:


Wovchena:
It is stated on wiki https://github.com/opencv/opencv/wiki/BuildOpenCV4OpenVINO

eaidova:
> It is stated on wiki https://github.com/opencv/opencv/wiki/BuildOpenCV4OpenVINO

but people who tries to run demos in usual way does not read wiki, so it is better to provide details explicitly

Wovchena:
The usual way is dropped. There is no OpenCV in OpenVINO now

eaidova:
> The usual way is dropped. There is no OpenCV in OpenVINO now

not understand, how it is connected? 
I just ask highlight what if user get standard or community opencv version and compile demos, in this case, it is not possible to run GAPI demos. Right now, it is not obvious not from demos documentation, not from wiki

eaidova (approved):

PR #3534, opened 2022-08-24
Bump networkx version to 2.8.6

mateusztabaka:


eaidova:
@mateusztabaka is it ready to review?
We are pinned networkx explicitly due to incompatibility with other openvino components (in our case the latest version was installed by scikit-image lib, we do not have direct dependency from networkx), so it will be nice if you make sure that we still have not pip-conflicts after this removal.

cc @akladiev

PR #3553, opened 2022-09-20
AC: support inference precision hint setting

eaidova:


tools/accuracy_checker/openvino/tools/accuracy_checker/config/config_reader.py
```diff
@@ -18,6 +18,7 @@
 import copy
 from pathlib import Path
 import os
+from tkinter.messagebox import NO
```
eaidova, line 21:
```suggestion
```

Wovchena (approved):

PR #3573, opened 2022-10-27
Bump torch version

p-wysocki:
Bump `torch` version to fix a bug in OpenVINO (see ticket).

Ticket: 95233


eaidova:
@p-wysocki on internal architectural review we decided relax pytorch version for omz in general way (upperbound is the latest released torch version, lowerbound is currently supported version), this action also requires some additional rework on model downloader side, so your changes are not applicable anymore (see https://github.com/openvinotoolkit/open_model_zoo/pull/3580)

PR #3574, opened 2022-10-28
demos: don't cast shape to str

Wovchena:
Merge after https://github.com/openvinotoolkit/openvino/pull/13613


eaidova:
Not sure that I understand idea, because the issue why this code was added in the impossibility of getting dim value from the partial shape in the dynamic case, so I believe it makes demos more dependent on shape realization and does not solve the initial problem for demos, where dynamic shapes supported

Wovchena:
I tested that [bert_named_entity_recognition_demo/python](https://github.com/openvinotoolkit/open_model_zoo/tree/master/demos/bert_named_entity_recognition_demo/python) works both with and without `--dynamic_shape` with this patch

Wovchena:
Done

eaidova (approved):

PR #3582, opened 2022-11-02
AC: fix add_extension()

Wovchena:
Ticket 94472


tools/accuracy_checker/openvino/tools/accuracy_checker/launcher/openvino_launcher.py
```diff
@@ -362,7 +362,7 @@ def _device_specific_configuration(self):
             if cpu_extensions:
                 selection_mode = self.config.get('_cpu_extensions_mode')
                 cpu_extensions = get_cpu_extension(cpu_extensions, selection_mode)
```
eaidova, line 364: probably this part should be rewritten too, because now it contains old logic for naming with CPU instructions (avx2 e.t.c.)

eaidova: I do not say that it should be fixed by you, just note for future development

eaidova (approved):

PR #3583, opened 2022-11-03
AC: update image processing adapter handle out precision

eaidova:


tools/accuracy_checker/openvino/tools/accuracy_checker/adapters/image_processing.py
```diff
@@ -94,8 +94,8 @@ def process(self, raw, identifiers, frame_meta):
 
     def _basic_postprocess(self, img):
         img = img.transpose((1, 2, 0)) if img.shape[-1] > 4 else img
-        img *= self.std
-        img += self.mean
+        img *= np.array(self.std).astype(img.dtype)
+        img += np.array(self.mean).astype(img.dtype)
```
Wovchena, line 98:
```suggestion
        img *= np.array(self.std, dtype=img.dtype)
        img += np.array(self.mean, dtype=img.dtype)
```

Wovchena (changes requested):

Wovchena (approved):

PR #3584, opened 2022-11-03
AC: Handle delayed model loading in get_model_file_type

jkamelin:
AC fails with `AttributeError: 'OpenVINOLauncher' object has no attribute '_model'` when running in delayed model loading mode, e.g. when restoring from stored predictions.

### DIFF
--- a/tools/accuracy_checker/openvino/tools/accuracy_checker/launcher/dlsdk_launcher.py
+++ b/tools/accuracy_checker/openvino/tools/accuracy_checker/launcher/dlsdk_launcher.py
@@ -931,10 +931,10 @@ class DLSDKLauncher(Launcher):
         self.disable_resize_to_input = preprocess.ie_processor.has_resize()
 
     def get_model_file_type(self):
-        if self._delayed_model_loading:
-            return None
-        else:
+        if hasattr(self, '_model'):
             return self._model.suffix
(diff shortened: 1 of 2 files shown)


eaidova (approved):

tools/accuracy_checker/openvino/tools/accuracy_checker/launcher/dlsdk_launcher.py
```diff
@@ -931,7 +931,10 @@ def _set_preprocess(self, preprocess):
         self.disable_resize_to_input = preprocess.ie_processor.has_resize()
 
     def get_model_file_type(self):
-        return self._model.suffix
+        if self._delayed_model_loading:
```
eaidova, line 934: what if initially model loading is postponed, but italready loaded at the moment when we call interface, probably we should check self._model itself for this case instead

eaidova: something like:
```suggestion
        if hasattr(self, '_model'):
```

jkamelin: Done

eaidova (approved):

PR #3585, opened 2022-11-03
Bump `scipy` to a higher version

p-wysocki:
There is a mismatch between `scipy` version in OpenVINO and OMZ, which causes pip conflicts in the CI pipelines. We would be thankful for a quick review. 

`scipy==1.5.4` has been released on November 5, 2020.

### DIFF
--- a/demos/common/python/requirements.txt
+++ b/demos/common/python/requirements.txt
@@ -1,3 +1,4 @@
 numpy>=1.16.6,<=1.23.1
 opencv-python
-scipy~=1.5.4
+scipy~=1.7.3; python_version == '3.7'
+scipy~=1.8.0; python_version >= '3.8'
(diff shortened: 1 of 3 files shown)


demos/common/python/requirements.txt
```diff
@@ -1,3 +1,4 @@
 numpy>=1.16.6,<=1.23.1
 opencv-python
-scipy~=1.5.4
+scipy~=1.7.3; python_version == '3.7'
```
eaidova, line 3: instead of pining specific version can we relax dependency, just >= 1.5.4?

p-wysocki: We had a discussion about it, would current dependency be alright? `scipy>=1.5.4,<1.9.0`

eaidova: yes, it is ok

eaidova (approved):

PR #3594, opened 2022-11-09
Update CI requirements to Python 3.8 (Ubuntu20 - OV 2022.2)

nignatovskiy:
- Updated update_requirements.py script (fixed paths to reqs sources files)
- Updated requirements-*.txt files.

### DIFF
--- a/ci/update-requirements.py
+++ b/ci/update-requirements.py
@@ -20,7 +20,7 @@ from pkginfo import Wheel
 # Package dependencies can vary depending on the Python version.
 # We thus have to run pip-compile with the lowest Python version that
 # the project supports.
-EXPECTED_PYTHON_VERSION = (3, 6)
+EXPECTED_PYTHON_VERSION = (3, 8)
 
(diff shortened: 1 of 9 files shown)


ci/update-requirements.py
```diff
@@ -20,7 +20,7 @@
 # Package dependencies can vary depending on the Python version.
 # We thus have to run pip-compile with the lowest Python version that
 # the project supports.
-EXPECTED_PYTHON_VERSION = (3, 6)
+EXPECTED_PYTHON_VERSION = (3, 8)
```
Wovchena, line 23: It should be python3.7

eaidova: I think it will be easier to use 3.8 as it is default for ubuntu 20, otherwise for setup ci environment we need manually install python3.7

Wovchena: ok

Wovchena (approved):

PR #3636, opened 2022-12-20
fix truth value of of numpy array

apram0d:
when --input_crop arg is given, it is throwing error with "truth value of an array with more than on element ambiguous". since already there is check for crop_size>0 I I guess .any() should be sufficient.

### DIFF
--- a/demos/face_recognition_demo/python/face_recognition_demo.py
+++ b/demos/face_recognition_demo/python/face_recognition_demo.py
@@ -228,7 +228,7 @@ def main():
             if frame_num == 0:
                 raise ValueError("Can't read an image from the input")
             break
-        if input_crop:
+        if input_crop it not None:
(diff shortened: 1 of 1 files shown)


demos/face_recognition_demo/python/face_recognition_demo.py
```diff
@@ -228,7 +228,7 @@ def main():
             if frame_num == 0:
                 raise ValueError("Can't read an image from the input")
             break
-        if input_crop:
+        if input_crop.any():
```
Wovchena, line 231:
```suggestion
        if input_crop is not None:
```
I don't think `input_crop` can be an empty array. But `input_crop` can be `None`

Wovchena (changes requested):

Wovchena (approved):

PR #3677, opened 2023-01-20
demos/tests: supress efficientdets

Wovchena:
Conversion fails with `ModuleNotFoundError: No module named 'keras.api'`


eaidova (approved):

aeozyalcin:
I am also running into `ModuleNotFoundError: No module named 'keras.api'` when trying to convert efficientdets model to IR. The changes on this PR just seem to supress the error message, but the conversion doesn't actually succeed. Is there a known fix for this?

Wovchena:
I don't have a fix

PR #3684, opened 2023-01-30
fix background matting representation

akorobeinikov:


tools/accuracy_checker/openvino/tools/accuracy_checker/representation/segmentation_representation.py
```diff
@@ -387,7 +387,7 @@ def __init__(self, identifier, path_to_mask, mask_to_gray=False, load_alpha=Fals
     def _load_mask(self):
         mask = super()._load_mask()
         if np.ndim(mask) == 3 and mask.shape[-1] == 4 and not self.load_alpha:
-            mask = mask[:, :, -1]
+            mask = mask[:, :, :-1]
```
eaidova, line 390: not, it is not correct.
when image has 4 channels, it 3 RGB + alpha. This code extracts alpha from image, your suggestion extracts RGB image instead

PR #3687, opened 2023-02-01
[AC]: background matting support per clip subfolders structure

eaidova:


Wovchena (approved):
There are still code style problems

PR #3696, opened 2023-02-08
[AC}: fix retinanet adapter

eaidova:


Wovchena (approved):
Does it just remove `s` from `nms_scores` name?

eaidova:
> Does it just remove `s` from `nms_scores` name?

yes, variable with the same name defined above for storing results for all nms results, in cycle we override it. it leads to invalid results

PR #3708, opened 2023-03-06
Expand image_processing_demo to work with EDSR like SR models

TolyaTalamanov:


demos/common/cpp/models/src/model_base.cpp
```diff
@@ -34,9 +34,6 @@ std::shared_ptr<ov::Model> ModelBase::prepareModel(ov::Core& core) {
     // -------------------------- Reading all outputs names and customizing I/O tensors (in inherited classes)
     prepareInputsOutputs(model);
 
-    /** Set batch size to 1 **/
-    ov::set_batch(model, 1);
```
TolyaTalamanov, line 38: @Wovchena Since `MO` can return IR with batch, should we move to separate `setBatch` method?

TolyaTalamanov:
@Wovchena Could you have a look, please?

demos/image_processing_demo/cpp/main.cpp
```diff
@@ -57,7 +57,8 @@ DEFINE_INPUT_FLAGS
 DEFINE_OUTPUT_FLAGS
 
 static const char help_message[] = "Print a usage message.";
-static const char at_message[] = "Required. Type of the model, either 'sr' for Super Resolution task, 'deblur' for "
+static const char at_message[] = "Required. Type of the model, either 'sr' for Super Resolution task, "
+                                 "'sr_channel_joint' for Super Resolution task that accepts and returns 1 channel image, 'deblur' for "
```
TolyaTalamanov, line 61: `task` -> `model` ???

Wovchena: Yes. Please change it. Also please update README Running section after that

demos/common/cpp/models/src/model_base.cpp
Wovchena: Alternative solution could be to modify `prepareInputsOutputs()` for every model wrapper. I don't see any reason to prefer one or another. Let's proceed with your suggestion to introduce `setBatch` method.

demos/image_processing_demo/cpp/main.cpp
TolyaTalamanov: Done

demos/common/cpp/models/src/model_base.cpp
TolyaTalamanov: Done

Wovchena (approved):
Is it urgent to merge? I'm not sure when the problem with model will be resolved

PR #3734, opened 2023-04-05
Fix model serialize

Wovchena:
OpenVINO started to use mmap allocator for IR reading. That prohibits serializing to the same file until the option to disable mmap is introduced.

Ticket 107646

### DIFF
--- a/tools/model_tools/src/openvino/model_zoo/omz_converter.py
+++ b/tools/model_tools/src/openvino/model_zoo/omz_converter.py
@@ -19,6 +19,7 @@ import os
 import shutil
 import string
 import sys
+import tempfile
 
 from openvino.runtime import Core, serialize
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


eaidova (approved):

tools/model_tools/src/openvino/model_zoo/omz_converter.py
```diff
@@ -130,31 +131,30 @@ def convert(reporter, model, output_dir, args, mo_props, requested_precisions):
...
+                    telemetry.send_event('md', 'converter_error',
+                        json.dumps({'error': 'mo-failed', 'model': model.name, 'precision': model_precision}))
+                    return False
 
-        reporter.print()
-
-        xml_path = str(output_dir / model.subdirectory / model_precision / model.name) + '.xml'
-        rt_model = Core().read_model(xml_path)
+            reporter.print()
+            rt_model = Core().read_model(os.path.join(mo_output_dir, model.name + '.xml'))
```
vurusovs, line 157: please, note that .bin file should be available until `ov::Model` is destructed (.bin file deletion will fail on Windows because `file is used by another process`). And looks like further logic also should be under `with` scope

vurusovs (approved):

ilyachur (approved):

PR #3739, opened 2023-04-12
Disable paddlepaddle

p-wysocki:
Paddlepaddle has to be disabled at all because even for lower Python versions its installation results in pip conflicts due to incompatible protobuf version


p-wysocki:
Please do not merge/review for now, I may have a better solution. I'm testing it right now.

PR #3743, opened 2023-04-18
[AC]: add yolov8 support

eaidova:


joelbudu:
This is brilliant thanks so much for adding this @eaidova . Is there any chance of having Yolov8 for instance segmentation integrated as well?

I've tested classification and that works out of the box.

PR #3750, opened 2023-04-21
[AC] fix model search issue

xczhai:


xczhai:
@eaidova Could you help review this PR if free? Thanks.

PR #3755, opened 2023-04-26
Fix incorrect rendering metric in case of processing 1 frame

smirnov-alexey:
This PR fixes the case where we get `Rendering:      nan(snan) ms` for 1 frame which is rather confusing. With properly updating render metric for leftover frames we get the correct number of frames in `getTotal()` function thus correct metric (`Rendering:      0.6 ms`).


smirnov-alexey:
@Wovchena please take a look

smirnov-alexey:
The CI is failing because of `PermissionError: [Errno 13] Permission denied: '/nfs/ov-share-01/data/volatile/omz_ci/md-cache/1/staging/...'`. How do I fix it?

Wovchena (approved):

PR #3757, opened 2023-04-27
Adjust inpainting_rect_masks from calibration_definitions.yml

pwolnows:
There was difference between dataset definition in open_model_zoo and [algo/annotation_converters_ext](https://gitlab-icv.toolbox.iotg.sclab.intel.com/algo/annotation_converters_ext/-/blob/masteracac#L3559)


pwolnows:
Is that OK that new dataset defines crop as preprocessing & postprocessing step where ac config contains resize ?
https://github.com/intel-innersource/frameworks.ai.openvino.accuracy-checker-configs/pull/16/files

PR #3762, opened 2023-05-17
Ea/efficientdet preconverted

eaidova:


Wovchena:
I would prefer to merge release branch into master. That way guarantees we don't leave any commit outside of master.

PR #3781, opened 2023-06-05
Restrict numpy for unsupported Python versions

p-wysocki:
## Details

- pip resolver is sporadically choosing numpy==1.25.0rc1 for unsupported Python versions, causing random CI fails
- the PR is currently a draft because of a **discussion in the ticket** - I recommend waiting to see if there are more reproductions before merging, but it's a subject to change and discuss

Ticket: 112557


Wovchena:
If your hypothesis turns out to be true
>This then could be a cause of a dependency of a dependency causing a sporadic issue

you should update numpy version only in a file that brings that dependency

PR #3788, opened 2023-06-14
EISW-83459 [MTL-P][Corp] OpenVINO Segmentation Demo reports incorrect output on VPUX

sivanov-work:
Fix for [EISW-83459](https://jira.devtools.intel.com/browse/EISW-83459)
[MTL-P][Corp] OpenVINO Segmentation Demo reports incorrect output on VPUX

It's back-merge of  https://github.com/openvinotoolkit/model_api/pull/63/files 
Plus, It introduces  layout guessing method based on the idea: C must be less then H and C must be less then W

### DIFF
--- a/demos/common/cpp/utils/include/utils/ocv_common.hpp
+++ b/demos/common/cpp/utils/include/utils/ocv_common.hpp
@@ -111,21 +111,78 @@ static inline void resize2tensor(const cv::Mat& mat, const ov::Tensor& tensor) {
     cv::resize(mat, cv::Mat{size, CV_8UC3, tensor.data()}, size);
 }
 
-static inline ov::Layout getLayoutFromShape(const ov::Shape& shape) {
+struct IntervalCondition {
+    using DimType = size_t;
+    using IndexType = size_t;
+    using ConditionChecker = std::function<bool(IndexType, const ov::Shape&)>;
+
+    template<class Cond>
+    constexpr IntervalCondition(IndexType i1, IndexType i2, Cond c) :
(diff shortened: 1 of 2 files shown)


sivanov-work:
@Wovchena Could you review it please?

demos/common/cpp/utils/include/utils/ocv_common.hpp
```diff
@@ -111,21 +111,78 @@ static inline void resize2tensor(const cv::Mat& mat, const ov::Tensor& tensor) {
...
+        if (gues_layout) {
+            bool guesResult = false;
+            ov::Layout guessedLayout;
+            std::tie(guesResult, guessedLayout) = makeGuesLayoutFrom4DShape(shape);
+            if (guesResult) {
+                return guessedLayout;
+            }
+        }
     }
+    throw std::runtime_error(std::string("Usupported " + std::to_string(shape.size()) + "D shape: ") +
```
Wovchena, line 184: Why do you cast `"Usupported " + std::to_string(shape.size()) + "D shape: "` to string explicitly?

Wovchena (approved):

demos/common/cpp/utils/include/utils/ocv_common.hpp
sivanov-work: as far I remember there are no global `operator+ (const char*, std::string)` but method `std::string::operator+(const char* OR std::string)` is. 
But I could be wrong, maybe it depends on c++ standard version

PR #3800, opened 2023-07-17
GAPI - classification benchmark porting

sivanov-work:
Rewriting classification benchmark using G-API


sivanov-work:
@TolyaTalamanov @Wovchena I wonder if you could review it

Wovchena (changes requested):

demos/classification_benchmark_demo/cpp_gapi/classification_benchmark_demo_gapi.hpp
```diff
@@ -0,0 +1,110 @@
...
+DEFINE_uint32(nireq, 1, num_inf_req_message);
+DEFINE_uint32(nt, 5, ntop_message);
+DEFINE_string(res, "1280x720", image_grid_resolution_message);
+DEFINE_bool(auto_resize, false, input_resizable_message);
+DEFINE_bool(no_show, false, no_show_message);
+DEFINE_uint32(time, std::numeric_limits<gflags::uint32>::max(), execution_time_message);
+DEFINE_string(u, "", utilization_monitors_message);
+DEFINE_bool(reverse_input_channels, false, reverse_input_channels_message);
+DEFINE_string(mean_values, "", mean_values_message);
+DEFINE_string(scale_values, "", scale_values_message);
```
Wovchena, line 74: These, `layout`, `auto_resize` are ignored in your implementation

Wovchena (changes requested):

demos/classification_benchmark_demo/cpp_gapi/main.cpp
```diff
@@ -0,0 +1,549 @@
...
+                               FLAGS_nt,
+                               isTestMode,
+                               !FLAGS_gt.empty(),
+                               presenter);
+            renderMetrics.update(renderingStart);
+            elapsedSeconds = std::chrono::steady_clock::now() - startTime;
+            if (!FLAGS_no_show) {
+                cv::imshow("classification_demo_gapi", gridMat.outImg);
+                //--- Processing keyboard events
+                int key = cv::waitKey(1);
```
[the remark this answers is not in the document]
sivanov-work: done

demos/classification_benchmark_demo/cpp_gapi/main.cpp
```diff
@@ -0,0 +1,549 @@
...
+    std::vector<cv::gapi::wip::onevpl::CfgParam> source_cfgs;
+    std::stringstream params_list(cfg_params);
+    std::string line;
+    while (std::getline(params_list, line, ',')) {
+        source_cfgs.push_back(createFromString(line));
+    }
+    return source_cfgs;
+}
+
+static cv::gapi::GKernelPackage getKernelPackage(const std::string& type) {
```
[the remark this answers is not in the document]
Wovchena: Well, that's up to you. I'd preferer not to have these options :) I just spotted a difference, that's why I used the word "probably".

demos/classification_benchmark_demo/cpp_gapi/main.cpp
```diff
@@ -0,0 +1,549 @@
...
+        try {
+            // fill starting from max confidence
+            for (auto conf_index_it = max_confidence_with_indices.rbegin();
+                 conf_index_it != max_confidence_with_indices.rend();
+                 ++conf_index_it) {
+                std::string str_label = in_labes.at(*conf_index_it->second);
+                out_scored_labels_to_append.emplace_back(conf_index_it->first, *conf_index_it->second,  std::move(str_label));
+            }
+        } catch (const std::out_of_range& ex) {
+            throw std::out_of_range(std::string("Provided labels file doesn't contain detected class index\nException: ") + ex.what());
```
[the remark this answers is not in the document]
Wovchena: You now have `size_t recalculated_label_id = *conf_index_it->second + labels_offset_correction;` in scope of this try. Assuming, you don't expect it to throw, you should move this line out of try catch

demos/classification_benchmark_demo/cpp_gapi/main.cpp
```diff
@@ -0,0 +1,549 @@
...
+                if (separatorIdx == std::string::npos) {
+                    throw std::runtime_error("The ground truth file has incorrect format.");
+                }
+                std::string imagePath = line.substr(0, separatorIdx);
+                size_t imagePathEndIdx = imagePath.rfind('/');
+                unsigned classIndex = static_cast<unsigned>(std::stoul(line.substr(separatorIdx + 1)));
+                if ((imagePathEndIdx != 1 || imagePath[0] != '.') && imagePathEndIdx != std::string::npos) {
+                    throw std::runtime_error("The ground truth file has incorrect format.");
+                }
+                // README
```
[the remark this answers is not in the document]
Wovchena: This is still applicable for demos/classification_benchmark_demo/cpp/main.cpp

demos/classification_benchmark_demo/cpp_gapi/main.cpp
```diff
@@ -0,0 +1,549 @@
...
+        bool isTestMode = true;
+        std::chrono::steady_clock::duration elapsedSeconds = std::chrono::steady_clock::duration(0);
+        std::chrono::seconds testDuration = std::chrono::seconds(3);
+        std::chrono::seconds fpsCalculationDuration = std::chrono::seconds(1);
+        auto startTime = std::chrono::steady_clock::now();
+        pipeline.start();
+        IndexScore::LabelsStorage top_k_scored_labels;
+        top_k_scored_labels.reserve(FLAGS_nt);
+        int64_t timestamp = 0;
+        while (keepRunning && elapsedSeconds < std::chrono::seconds(FLAGS_time) && pipeline.pull(cv::gout(output, infer_result, timestamp))) {
```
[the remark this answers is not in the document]
Wovchena: Is there a reason you suggest to introduce `-loop` instead of `-time` from cpp demo? I prefer to keep demos aligned, unless you provide other arguments

demos/classification_benchmark_demo/cpp_gapi/main.cpp
```diff
@@ -0,0 +1,549 @@
...
+
+    static IndexScore create_from_array(const float *out_blob_data_ptr, size_t out_blob_element_count,
+                                        size_t top_k_amount) {
+        IndexScore ret;
+        if (!out_blob_data_ptr) {
+            return IndexScore();
+        }
+        // find top K
+        size_t i = 0;
+        // fill & sort topK with first K elements of N array: O(K*Log(K))
```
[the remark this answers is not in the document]
Wovchena: Why do you need first k labels to be sorted?
About disadvantages:
The proposal is to have a vector of pairs<idx, score>, where pairs[i].second == out_blob_data_ptr[i]. it does require a copy of out_blob_data_ptr[], as you stated in 1. But it removes a need to introduce your own algorithm.

demos/classification_benchmark_demo/cpp_gapi/main.cpp
```diff
@@ -0,0 +1,549 @@
...
+                    if (predicted_class_id == classIndices.at(framesNum % classIndices.size())) {
+                        predictionResult = PredictionResult::Correct;
+                        correctPredictionsCount++;
+                    }
+                } else {
+                    predictionResult = PredictionResult::Unknown;
+                }
+            } catch (const std::out_of_range &ex) {
+                std::cerr << ex.what()
+                          << "\nPlease, make sure the model or the label file are correct";
```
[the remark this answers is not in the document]
Wovchena: Why did you keep try catch? My guess is to refine a error message. If so, why didn't you include all the needed info in the original exception which is cough and rethrown here?

demos/classification_benchmark_demo/cpp_gapi/main.cpp
```diff
@@ -0,0 +1,549 @@
...
+        } catch (const std::out_of_range& ex) {
+            throw std::out_of_range(std::string("Provided labels file doesn't contain detected class index\nException: ") + ex.what());
+        }
+    }
+
+    static IndexScore create_from_array(const float *out_blob_data_ptr, size_t out_blob_element_count,
+                                        size_t top_k_amount) {
+        IndexScore ret;
+        if (!out_blob_data_ptr) {
+            return IndexScore();
```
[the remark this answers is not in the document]
Wovchena: The question is not about `IndexScore`, but about `out_blob_data_ptr` being null. Why do you care? If G-API provides nullptr as `out_blob.ptr<float>()`, there's probably a bug in it, so you shouldn't pretend that it's fine

demos/classification_benchmark_demo/cpp_gapi/main.cpp
```diff
@@ -0,0 +1,549 @@
...
+        }
+        // find top K
+        size_t i = 0;
+        // fill & sort topK with first K elements of N array: O(K*Log(K))
+        while(i < std::min(top_k_amount, out_blob_element_count)) {
+            ret.max_element_indices.push_back(i);
+            ret.max_confidence_with_indices.emplace(out_blob_data_ptr[i], std::prev(ret.max_element_indices.end()));
+            i++;
```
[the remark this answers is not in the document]
Wovchena: I see
> Seems on micro-optimizations

But this is not about performance. This is about readability. My intuition is that having `while` suggests `i` changes in a way that can't be described with simple instruction, which doesn't hold here

demos/classification_benchmark_demo/cpp_gapi/main.cpp
```diff
@@ -0,0 +1,549 @@
...
+        }
+
+        // search K elements through remnant N-K array elements
+        // greater than the minimum element in the pivot topK
+        // O((N-K)*Log(K))
+        for (i = top_k_amount; i < out_blob_element_count && !ret.max_confidence_with_indices.empty(); i++) {
```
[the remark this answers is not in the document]
Wovchena: 1. If it's 0, why would someone call it in first place? I believe you should handle this as an error or ignore.
2. Agree
3. Will be discussed in https://github.com/openvinotoolkit/open_model_zoo/pull/3800#discussion_r1267996567
4. Let's hold this conversation until 3. is resolved, because it is essentially the same to me

demos/classification_benchmark_demo/cpp_gapi/main.cpp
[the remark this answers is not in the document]
sivanov-work: the upper level exception removed

demos/classification_benchmark_demo/cpp_gapi/main.cpp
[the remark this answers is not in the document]
sivanov-work: because `IndexScore` here doesn't know anything about G-API and high-level rules and might be used separately (after refactoring for example)  - its responsibility is build data structure from input.

So this check on empty array must be done in G-API kernel - I think I had missed that

demos/classification_benchmark_demo/cpp_gapi/main.cpp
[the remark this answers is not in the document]
sivanov-work: removed second assignment

demos/classification_benchmark_demo/cpp_gapi/main.cpp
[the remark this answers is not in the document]
sivanov-work: removed i assignment, and empty check. Added contract check in the beginning

demos/classification_benchmark_demo/cpp_gapi/main.cpp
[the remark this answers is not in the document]
sivanov-work: cannot assigned in `true` by default, as it declared in common gapi header.
So overridden in benchmark main.cpp

demos/classification_benchmark_demo/cpp_gapi/main.cpp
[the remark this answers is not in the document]
Wovchena: I'm ok with the current state. Now I'm just trying to understand:
1. How do you define high-level rules.
2. How do you define class having knowledge about something. And about G-API and high-level rules in particular. I would guess mentioning any G-API specific entity in scope of `IndexScore` makes it to know about G-API. But I can't apply the same idea to "rules".
3. > after refactoring for example

    Can you name an event that would trigger such refactoring?

demos/classification_benchmark_demo/cpp_gapi/main.cpp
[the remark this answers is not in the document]
Wovchena: The same is defined in `demos\background_subtraction_demo\cpp_gapi\main.cpp`. Please, factor it out to common, similar to `createFromString`. Feel free to create a new file if you believe it doesn't belong to `stream_source`

demos/classification_benchmark_demo/cpp_gapi/main.cpp
```diff
@@ -0,0 +1,549 @@
...
+                               frame_timestamp,  //from original image created timestamp
+                               accuracy,
+                               FLAGS_nt,
+                               isTestMode,
+                               !FLAGS_gt.empty(),
+                               presenter);
+            renderMetrics.update(renderingStart);
+            elapsedSeconds = std::chrono::steady_clock::now() - startTime;
+            if (!FLAGS_no_show) {
+                cv::imshow("classification_demo_gapi", gridMat.outImg);
```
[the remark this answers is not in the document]
Wovchena: I can see there's commit stating `fix label index`. Not sure if you finished work. At this moment the command above for classification_benchmark_demo_gapi still shows all labels red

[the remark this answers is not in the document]
sivanov-work: fixed

demos/classification_benchmark_demo/cpp_gapi/main.cpp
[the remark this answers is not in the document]
sivanov-work: done

demos/classification_benchmark_demo/cpp_gapi/main.cpp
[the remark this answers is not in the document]
sivanov-work: done

Wovchena (changes requested):
Please, publish kernel_package.hpp

demos/classification_benchmark_demo/cpp_gapi/classification_benchmark_demo_gapi.hpp
Wovchena: These three still exist: `reverse_input_channels`, `mean_values`, `scale_values`

Wovchena:
LGTM except for https://github.com/openvinotoolkit/open_model_zoo/pull/3800#discussion_r1265741982. Please ignore broken precommit builds

demos/background_subtraction_demo/cpp_gapi/main.cpp
```diff
@@ -84,37 +84,6 @@ static cv::gapi::GKernelPackage getKernelPackage(const std::string& type) {
     GAPI_Assert(false && "Unreachable code!");
 }
 
-cv::gapi::wip::onevpl::CfgParam createFromString(const std::string &line) {
```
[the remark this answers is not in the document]
sivanov-work: yes, here https://github.com/openvinotoolkit/open_model_zoo/pull/3800/files#diff-a44d06f960de66ff0634dcb013f3da9e7125e180cbfb3f59b089e6ca1f7af367R78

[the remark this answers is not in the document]
Wovchena: Yes, to demos/common/cpp_gapi/utils_gapi/include/utils_gapi/stream_source.hpp

Wovchena (approved):
Should I merge or wait for @TolyaTalamanov's approval?

demos/common/cpp_gapi/utils_gapi/src/stream_source.cpp
```diff
@@ -61,6 +61,8 @@ cv::MediaFrame::View MediaBGRAdapter::access(cv::MediaFrame::Access) {
 bool MediaCommonCapSrc::pull(cv::gapi::wip::Data& data) {
```
TolyaTalamanov, line 61: do we have source that returns `cv::Mat`?

demos/common/cpp_gapi/utils_gapi/src/stream_source.cpp
```diff
@@ -61,6 +61,8 @@ cv::MediaFrame::View MediaBGRAdapter::access(cv::MediaFrame::Access) {
...
     if (CommonCapSrc::pull(data)) {
         data = cv::MediaFrame::Create<MediaBGRAdapter>(cv::util::get<cv::Mat>(data));
+        std::chrono::steady_clock::time_point now = std::chrono::steady_clock::now();
+        data.meta[cv::gapi::streaming::meta_tag::timestamp] = int64_t{std::chrono::duration_cast<std::chrono::milliseconds>(now.time_since_epoch()).count()};
```
TolyaTalamanov, line 65: we use `mircoseconds` in our `GCaptureSource` I don't if it makes sense to do the same

demos/classification_benchmark_demo/cpp_gapi/src/custom_kernels.cpp
```diff
@@ -0,0 +1,134 @@
...
+        out_rect = cv::Rect{ center.x - sqside/2
+                             , center.y - sqside/2
+                             , sqside
+                             , sqside
+                            };
+    }
+};
+
+GAPI_OCV_KERNEL(OCVTopK, custom::TopK) {
+    static void run(const cv::MediaFrame &in, const cv::Mat &out_blob, uint32_t top_k_amount, IndexScore &out) {
```
TolyaTalamanov, line 111: Why do we need `cv::MediaFrame &in` there?

demos/classification_benchmark_demo/cpp_gapi/src/custom_kernels.cpp
```diff
@@ -0,0 +1,134 @@
...
+            out_scored_labels_to_append.emplace_back(conf_index_it->first,
+                                                     *conf_index_it->second,
+                                                     recalculated_label_id != 0 ? in_labes.at(recalculated_label_id) : "others");
+        } catch (const std::out_of_range& ex) {
+            throw std::out_of_range(std::string("Provided labels file doesn't contain classified label index\nException: ") + ex.what());
+        }
+    }
+}
+
+IndexScore IndexScore::create_from_array(const float *out_blob_data_ptr, size_t out_blob_element_count,
```
TolyaTalamanov, line 53: Wow, I expected to see it as small function that just use `sort` :)
We have only `1000` values there...

demos/classification_benchmark_demo/cpp_gapi/CMakeLists.txt
```diff
@@ -0,0 +1,18 @@
...
+
+file(GLOB_RECURSE SOURCES ${CMAKE_CURRENT_SOURCE_DIR}/*.cpp)
+file(GLOB_RECURSE HEADERS ${CMAKE_CURRENT_SOURCE_DIR}/*.hpp)
+
+add_demo(NAME classification_benchmark_demo_gapi
+    SOURCES ${SOURCES}
+    HEADERS ${HEADERS}
+    INCLUDE_DIRECTORIES "${CMAKE_CURRENT_SOURCE_DIR}/include"
+    DEPENDENCIES monitors utils_gapi models pipelines
+    OPENCV_VERSION_REQUIRED 4.5.2)
```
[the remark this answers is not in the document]
sivanov-work: I suppose it requires all demos version alignment and it'd better done in separate PR. @Wovchena  what do you think?

demos/classification_benchmark_demo/cpp_gapi/README.md
```diff
@@ -0,0 +1,183 @@
...
+
+An example of using the Model Converter:
+
+```sh
+omz_converter --list models.lst
+```
+
+### Supported Models
+
+* alexnet
```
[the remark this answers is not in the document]
sivanov-work: not tested. Do you have points that G-API cannot cover them automatically?
I mean - is there clear understanding that something wouldn't be supported?

demos/classification_benchmark_demo/cpp_gapi/README.md
```diff
@@ -0,0 +1,183 @@
...
+    -d "<device>"             Optional. Specify the target device to infer on (the list of available devices is shown below). Default value is CPU. The demo will look for a suitable plugin for device specified.
+    -nthreads "<integer>"     Optional. Specify count of threads.
+    -nstreams "<integer>"     Optional. Specify count of streams.
+    -nireq "<integer>"        Optional. Number of infer requests.
+    -nt "<integer>"           Optional. Number of top results. Default value is 5. Must be >= 1.
+    -res "<WxH>"              Optional. Set image grid resolution in format WxH. Default value is 1280x720.
+    -no_show                  Optional. Disable showing of processed images.
+    -time "<integer>"         Optional. Time in seconds to execute program. Default is -1 (infinite time).
+    -u                        Optional. List of monitors to show initially.
+    -use_onevpl                Optional. Use onevpl video decoding.
```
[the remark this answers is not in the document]
sivanov-work: ok, will remove any mentions about vpl at all

demos/classification_benchmark_demo/cpp_gapi/README.md
```diff
@@ -0,0 +1,183 @@
...
+    -onevpl_params             Optional. Parameters for onevpl video decoding. OneVPL source can be fine-grained by providing configuration parameters. Format: <prop name>:<value>,<prop name>:<value> Several important configuration parameters: 'mfxImplDescription.mfxDecoderDescription.decoder.CodecID' values: https://spec.oneapi.io/onevpl/2.7.0/API_ref/VPL_enums.html?highlight=mfx_codec_hevc#codecformatfourcc and 'mfxImplDescription.AccelerationMode' values: https://spec.oneapi.io/onevpl/2.7.0/API_ref/VPL_disp_api_enum.html?highlight=d3d11#mfxaccelerationmode(see `MFXSetConfigFilterProperty` by https://spec.oneapi.io/versions/latest/elements/oneVPL/source/index.html)
```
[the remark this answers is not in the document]
sivanov-work: ok

demos/classification_benchmark_demo/cpp_gapi/README.md
```diff
@@ -0,0 +1,183 @@
...
+    -onevpl_pool_size          OneVPL source applies this parameter as preallocated frames pool size. 0 leaves frames pool size default for your system. This parameter doesn't have a god default value. It must be adjusted for specific execution (video, model, system ...).
```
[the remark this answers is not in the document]
sivanov-work: will remove

demos/classification_benchmark_demo/cpp_gapi/README.md
```diff
@@ -0,0 +1,183 @@
...
+                      -labels <path_to_file_with_list_of_labels> \
+                      -gt <path_to_ground_truth_data_file> \
+                      -u CDM
+```
+
+## Demo Output
+
+The demo uses OpenCV to display the resulting image grid with classification results presented as a text above images. The demo reports:
+
+* **FPS**: average rate of video frame processing (frames per second).
```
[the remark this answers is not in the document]
sivanov-work: what is it?

demos/classification_benchmark_demo/cpp_gapi/main.cpp
```diff
@@ -0,0 +1,344 @@
...
+                                                               read_type::safe,
+                                                               0,
+                                                               std::numeric_limits<size_t>::max(),
+                                                               stringToSize(FLAGS_res));
+        const auto tmp = cap->read();
+        cv::Size frame_size = cv::Size{tmp.cols, tmp.rows};
+        cap.reset();
+        // NB: oneVPL source rounds up frame size by 16
+        // so size might be different from what ImagesCapture reads.
+        if (FLAGS_use_onevpl) {
```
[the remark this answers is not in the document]
sivanov-work: vpl removed

demos/classification_benchmark_demo/cpp_gapi/main.cpp
```diff
@@ -0,0 +1,344 @@
...
+            frame_size.width  = cv::alignSize(frame_size.width, 16);
+            frame_size.height = cv::alignSize(frame_size.height, 16);
+        }
+
+        cv::GComputation comp([&] {
+            cv::GFrame in;
+            cv::GOpaque<int64_t> outTs = cv::gapi::streaming::timestamp(in);
+
+            auto size = cv::gapi::streaming::size(in);
+            cv::GOpaque<cv::Rect> in_roi = custom::LocateROI::on(size);
```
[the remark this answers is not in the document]
sivanov-work: original demo uses centerCrop

demos/classification_benchmark_demo/cpp_gapi/main.cpp
```diff
@@ -0,0 +1,344 @@
...
+        cap.reset();
+        // NB: oneVPL source rounds up frame size by 16
+        // so size might be different from what ImagesCapture reads.
+        if (FLAGS_use_onevpl) {
+            frame_size.width  = cv::alignSize(frame_size.width, 16);
+            frame_size.height = cv::alignSize(frame_size.height, 16);
+        }
+
+        cv::GComputation comp([&] {
+            cv::GFrame in;
```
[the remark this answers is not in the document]
sivanov-work: it seems after abolishing vpl source we could use cv::Mat here. 
will change on Mat

demos/classification_benchmark_demo/cpp_gapi/main.cpp
```diff
@@ -0,0 +1,344 @@
...
+            cv::GOpaque<int64_t> outTs = cv::gapi::streaming::timestamp(in);
+
+            auto size = cv::gapi::streaming::size(in);
+            cv::GOpaque<cv::Rect> in_roi = custom::LocateROI::on(size);
+
+            auto blob = cv::gapi::infer<nets::Classification>(in_roi, in);
+            cv::GOpaque<IndexScore> index_score = custom::TopK::on(in, blob, FLAGS_nt);
+
+            cv::GMat bgr = cv::gapi::streaming::BGR(in);
```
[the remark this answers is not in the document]
sivanov-work: agree

demos/common/cpp_gapi/utils_gapi/src/stream_source.cpp
sivanov-work: yes, it's `CommonSrc` which will require add timestamp to that instead of MediaCap

demos/common/cpp_gapi/utils_gapi/src/stream_source.cpp
sivanov-work: ok

demos/classification_benchmark_demo/cpp_gapi/src/custom_kernels.cpp
sivanov-work: not sure, does graph can be constructed without source frame/mat linkage?

demos/classification_benchmark_demo/cpp_gapi/src/custom_kernels.cpp
sivanov-work: please check the previous comments from @Wovchena about the discussion of using `nth_element`. All of them applied in case of `std::sort` as well

demos/classification_benchmark_demo/cpp_gapi/CMakeLists.txt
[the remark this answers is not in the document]
TolyaTalamanov: Yes, let's have it later then

demos/classification_benchmark_demo/cpp_gapi/README.md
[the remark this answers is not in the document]
TolyaTalamanov: Well, classification models are pretty the same in terms of parsing output, just double check that your code working if there are `1001` class instead of `1000` just in case

demos/classification_benchmark_demo/cpp_gapi/README.md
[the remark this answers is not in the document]
TolyaTalamanov: Yes, I think it's not needed at this point

demos/common/cpp_gapi/utils_gapi/src/stream_source.cpp
TolyaTalamanov: You'd proceed with `cv::GMat` so far...

demos/classification_benchmark_demo/cpp_gapi/src/custom_kernels.cpp
TolyaTalamanov: You have source linkage (in infer at least) but there you don't need this source at all

demos/classification_benchmark_demo/cpp_gapi/README.md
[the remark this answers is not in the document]
TolyaTalamanov: I mean we have the `FPS` and `latency` per iteration for the entire graph.
But we don't have latency for `decoding`, `preprocessing` etc

demos/classification_benchmark_demo/cpp_gapi/classification_benchmark_demo_gapi.hpp
```diff
@@ -0,0 +1,77 @@
...
+#include <iostream>
+
+#include <gflags/gflags.h>
+
+static const char help_message[] = "Print a usage message.";
+static const char image_message[] = "Required. Path to a folder with images or path to an image file.";
+static const char model_message[] = "Required. Path to an .xml file with a trained model.";
+static const char labels_message[] = "Required. Path to .txt file with labels.";
+static const char kernel_package_message[] =
+    "Optional. G-API kernel package type: opencv, fluid (by default opencv is used).";
```
TolyaTalamanov, line 16: We don't use `fluid` at all there, since the graph is pretty straightforward: `in`->`infer`->`topK` there is no operations that could be run with `fluid` backend

demos/classification_benchmark_demo/cpp_gapi/classification_benchmark_demo_gapi.hpp
```diff
@@ -0,0 +1,77 @@
...
+static const char mean_values_message[] =
+    "Optional. Normalize input by subtracting the mean values per channel. Example: \"255.0 255.0 255.0\"";
+static const char scale_values_message[] = "Optional. Divide input by scale values per channel. Division is applied "
+                                           "after mean values subtraction. Example: \"255.0 255.0 255.0\"";
+
+DEFINE_bool(h, false, help_message);
+DEFINE_string(i, "", image_message);
+DEFINE_string(m, "", model_message);
+DEFINE_string(labels, "", labels_message);
+DEFINE_string(kernel_package, "opencv", kernel_package_message);
```
TolyaTalamanov, line 41: I'd remove since we don't use it there

demos/classification_benchmark_demo/cpp_gapi/include/custom_kernels.hpp
```diff
@@ -0,0 +1,59 @@
...
+//
+
+#pragma once
+
+#include <list>
+#include <map>
+#include <string>
+#include <vector>
+
+#include <cpp/ie_cnn_network.h>
```
TolyaTalamanov, line 12: I don't think you need this header there

demos/classification_benchmark_demo/cpp_gapi/include/custom_kernels.hpp
```diff
@@ -0,0 +1,59 @@
...
+    IndicesStorage max_element_indices;
+    size_t classification_indices_scale;
+};
+
+using GRect = cv::GOpaque<cv::Rect>;
+using GSize = cv::GOpaque<cv::Size>;
+using GIndexScore = cv::GOpaque<IndexScore>;
+
+namespace custom {
+G_API_OP(LocateROI, <GRect(GSize)>, "classification_benchmark.custom.locate-roi") {
```
TolyaTalamanov, line 45: Why do we need to locate roi? I thought the inference is supposed to work on the full frame

demos/classification_benchmark_demo/cpp_gapi/main.cpp
```diff
@@ -0,0 +1,314 @@
+// Copyright (C) 2022-2023 Intel Corporation
```
TolyaTalamanov, line 1: just `2023` ?

demos/classification_benchmark_demo/cpp_gapi/main.cpp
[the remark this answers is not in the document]
TolyaTalamanov: Make sense, could you rename this to `CentralCrop` please?
I don't know why it's called like this there: https://github.com/opencv/opencv/blob/4.x/modules/gapi/samples/infer_single_roi.cpp#L85-L108

demos/classification_benchmark_demo/cpp_gapi/main.cpp
```diff
@@ -0,0 +1,314 @@
...
+        const auto net =
+            cv::gapi::ie::Params<nets::Classification>{
+                FLAGS_m,  // path to topology IR
+                fileNameNoExt(FLAGS_m) + ".bin",  // path to weights
+                FLAGS_d  // device specifier
+            }.cfgNumRequests(config.maxAsyncRequests)
+             .pluginConfig(config.getLegacyConfig());
+        // clang-format on
+
+        auto kernels = cv::gapi::combine(custom::kernels(), util::getKernelPackage(FLAGS_kernel_package));
```
TolyaTalamanov, line 158: Probably just `cv::gapi::kernels<OCVLocateROI>()`

demos/classification_benchmark_demo/cpp_gapi/src/custom_kernels.cpp
```diff
@@ -0,0 +1,134 @@
+// Copyright (C) 2022-2023 Intel Corporation
```
TolyaTalamanov, line 1: `2023`

demos/classification_benchmark_demo/cpp_gapi/src/custom_kernels.cpp
```diff
@@ -0,0 +1,134 @@
...
+
+#include "custom_kernels.hpp"
+
+#include <algorithm>
+#include <map>
+#include <memory>
+#include <stdexcept>
+#include <utility>
+
+#include <ie_core.hpp>
```
TolyaTalamanov, line 13: I think you don't need `ie` headers there

demos/classification_benchmark_demo/cpp_gapi/src/custom_kernels.cpp
```diff
@@ -0,0 +1,134 @@
...
+#include <ie_data.h>
+#include <ie_layouts.h>
+#include <opencv2/gapi/core.hpp>
```
TolyaTalamanov, line 16: and most of `opencv` headers as well

TolyaTalamanov:
@AsyaPronina Could you have a look, please?

demos/classification_benchmark_demo/cpp_gapi/CMakeLists.txt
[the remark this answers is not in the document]
Wovchena: > I suppose it requires all demos version alignment and it'd better done in separate PR. @Wovchena what do you think?

I agree

demos/classification_benchmark_demo/cpp_gapi/README.md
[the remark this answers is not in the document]
sivanov-work: It was the reason why gapi benchmark gives different output initially. The fix if here https://github.com/openvinotoolkit/open_model_zoo/pull/3800/files#diff-e122ede2725aacb2bad4130ba9f3c7b9855157b24d61440bcd7a4985f3d2713dR29

demos/classification_benchmark_demo/cpp_gapi/README.md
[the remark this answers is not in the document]
sivanov-work: let's remove them

demos/classification_benchmark_demo/cpp_gapi/classification_benchmark_demo_gapi.hpp
sivanov-work: let's remove it

demos/classification_benchmark_demo/cpp_gapi/classification_benchmark_demo_gapi.hpp
sivanov-work: ok

demos/classification_benchmark_demo/cpp_gapi/include/custom_kernels.hpp
sivanov-work: ok

demos/classification_benchmark_demo/cpp_gapi/main.cpp
sivanov-work: ok

demos/classification_benchmark_demo/cpp_gapi/src/custom_kernels.cpp
sivanov-work: ok

demos/classification_benchmark_demo/cpp_gapi/src/custom_kernels.cpp
sivanov-work: ok

demos/classification_benchmark_demo/cpp_gapi/src/custom_kernels.cpp
sivanov-work: if comments were lost: 
sort as well as other std algorithms  require additional memory allocations for both vertices & indices either; copy vertices + fill (by iota) indices; + custom modifications for keep stable indices in parallel. On every frame
So it makes no sense to bring up such a premature pessimization if  you has better solution already

sivanov-work:
> @AsyaPronina Could you have a look, please?

I hope it worth all such effort

demos/classification_benchmark_demo/cpp_gapi/CMakeLists.txt
```diff
@@ -0,0 +1,18 @@
+# Copyright (C) 2022-2023 Intel Corporation
```
TolyaTalamanov, line 1: `2023`

demos/classification_benchmark_demo/cpp_gapi/README.md
```diff
@@ -0,0 +1,177 @@
...
+
+The demo starts in "Testing mode" with fixed grid size. After calculating the average FPS result, it will switch to normal mode and grid will be readjusted depending on model performance. Bigger grid means higher performance. You can repeat testing by pressing "Space" or "R" button.
+
+When "ground truth" data is applied, the color coding for the text, drawn above each image, shows whether the classification was correct: green means correct class prediction, red means wrong.
+
+You can stop the demo by pressing "Esc" or "Q" button. After that, the average metrics values will be printed to the console.
+
+> **NOTE**: By default, Open Model Zoo demos expect input with BGR channels order. If you trained your model to work with RGB order, you need to manually rearrange the default channels order in the demo application or reconvert your model using the Model Optimizer tool with the `--reverse_input_channels` argument specified. For more information about the argument, refer to **When to Reverse Input Channels** section of [Embedding Preprocessing Computation](@ref openvino_docs_MO_DG_Additional_Optimization_Use_Cases).
```
TolyaTalamanov, line 17: Do we support `--reverse_input_channels` flag properly?

demos/classification_benchmark_demo/cpp_gapi/classification_benchmark_demo_gapi.hpp
```diff
@@ -0,0 +1,73 @@
...
+static const char num_streams_message[] = "Optional. Specify count of streams.";
+static const char num_inf_req_message[] = "Optional. Number of infer requests.";
+static const char image_grid_resolution_message[] = "Optional. Set image grid resolution in format WxH. "
+                                                    "Default value is 1280x720.";
+static const char ntop_message[] = "Optional. Number of top results. Default value is 5. Must be >= 1.";
+static const char no_show_message[] = "Optional. Disable showing of processed images.";
+static const char execution_time_message[] = "Optional. Time in seconds to execute program. "
+                                             "Default is -1 (infinite time).";
+static const char utilization_monitors_message[] = "Optional. List of monitors to show initially.";
+static const char reverse_input_channels_message[] = "Optional. Switch the input channels order from BGR to RGB.";
```
TolyaTalamanov, line 29: It's not used, let's remove if we don't plan to handle this case so far

demos/classification_benchmark_demo/cpp_gapi/main.cpp
```diff
@@ -0,0 +1,313 @@
...
+#include <utils/images_capture.h>
+#include <utils/ocv_common.hpp>
+#include <utils/performance_metrics.hpp>
+#include <utils/slog.hpp>
+#include <utils_gapi/kernel_package.hpp>
+#include <utils_gapi/stream_source.hpp>
+
+#include "classification_benchmark_demo_gapi.hpp"
+#include "custom_kernels.hpp"
+#include <models/classification_model.h>
```
TolyaTalamanov, line 53: this one definitely not used

demos/classification_benchmark_demo/cpp_gapi/main.cpp
```diff
@@ -0,0 +1,313 @@
...
+#include <fstream>
+#include <iomanip>
+#include <limits>
+#include <memory>
+#include <stdexcept>
+#include <string>
+#include <utility>
+#include <vector>
+
+#include <gflags/gflags.h>
```
TolyaTalamanov, line 19: all of these headers are used?

TolyaTalamanov (approved):
Double check if all headers included are really needed :)

Great, LGTM 👍

demos/classification_benchmark_demo/cpp_gapi/README.md
Wovchena: That's model optimizer's flag, not demo's one

TolyaTalamanov: https://github.com/openvinotoolkit/open_model_zoo/blob/master/demos/classification_benchmark_demo/cpp/main.cpp#L233

Wovchena: Your point is valid overall, I just wanted to point out that the piece of text which you started this conversation under shouldn't be modified because it relates to Model Optimizer, not the demo

demos/classification_benchmark_demo/cpp_gapi/main.cpp
```diff
@@ -0,0 +1,313 @@
...
+        std::chrono::seconds fpsCalculationDuration = std::chrono::seconds(1);
+        auto startTime = std::chrono::steady_clock::now();
+        pipeline.start();
+        IndexScore::LabelsStorage top_k_scored_labels;
+        top_k_scored_labels.reserve(FLAGS_nt);
+        int64_t timestamp = 0;
+        size_t total_produced_image_count = 0;
+        while (keepRunning && elapsedSeconds < std::chrono::seconds(FLAGS_time) && pipeline.pull(cv::gout(output, infer_result, timestamp))) {
+            std::chrono::microseconds dur(timestamp);
+            std::chrono::time_point<std::chrono::steady_clock> frame_timestamp(dur);
```
TolyaTalamanov, line 205: Looks like this is the latency of the source itself

TolyaTalamanov: To get the latency for pipeline iteration you need to measure time between `timestamp` that returned from the pipeline and `timestamp` after the `pull` is completed.

demos/classification_benchmark_demo/cpp_gapi/main.cpp
```diff
@@ -0,0 +1,313 @@
...
+        const auto net =
+            cv::gapi::ie::Params<nets::Classification>{
+                FLAGS_m,  // path to topology IR
+                fileNameNoExt(FLAGS_m) + ".bin",  // path to weights
+                FLAGS_d  // device specifier
+            }.cfgNumRequests(config.maxAsyncRequests)
+             .pluginConfig(config.getLegacyConfig());
+        // clang-format on
+
+        auto pipeline = comp.compileStreaming(cv::compile_args(custom::kernels(), cv::gapi::networks(net)));
```
TolyaTalamanov, line 158: Could you also add `cv::gapi::streaming::queue_capacity{1}`

demos/common/cpp_gapi/utils_gapi/src/kernel_package.cpp
```diff
@@ -0,0 +1,18 @@
+// Copyright (C) 2021-2023 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+//
+
+#include "utils_gapi/kernel_package.hpp"
+
+namespace util {
+cv::gapi::GKernelPackage getKernelPackage(const std::string& type) {
```
TolyaTalamanov, line 8: This is not needed if we're going to use `cpu` backend only

demos/classification_benchmark_demo/cpp_gapi/main.cpp
```diff
@@ -0,0 +1,313 @@
...
+            if (classIndex >= labels.size()) {
+                throw std::runtime_error("Class index " + std::to_string(classIndex) +
+                                         " is outside the range supported by the model with max index: " +
+                                         std::to_string(labels.size()));
+            }
+        }
+
+        cv::GComputation comp([&] {
+            cv::GMat in;
+            cv::GOpaque<int64_t> out_ts = cv::gapi::streaming::timestamp(in);
```
TolyaTalamanov, line 133: This is not needed if we're going only expose `pull` latency (`output latency`)

demos/classification_benchmark_demo/cpp_gapi/main.cpp
sivanov-work: this magic happens inside of `gridMat`
```
gridMat.textUpdate(metrics,
                               frame_timestamp,  <---- here
```
The logic of measure pipeline latency is quite similar to `classic benchmark demo`
1) mark source timestamp before invoke inferRequest
2) pass it through gridMat after pull

demos/common/cpp_gapi/utils_gapi/src/kernel_package.cpp
sivanov-work: are we going?

by the way - it's `utility` common place, which is usable by other gapi demo

PR #3847, opened 2023-08-21
demos: fix some warnings

Wovchena:


TolyaTalamanov (approved):

AsyaPronina (approved):

ngaloppo:
I believe that this broke the Windows build. `SharedMatAllocator` does not inherit from `ov::Allocator`, which breaks the `Tensor` constructor in ocv_common.hpp:

```
C:\Users\alikhola\Downloads\open_model_zoo\demos\common\cpp\utils\include\utils/ocv_common.hpp(99,98): error C2440: '<function-style-cast>': cannot convert from 'initializ
er list' to 'ov::Tensor' 
C:\Users\alikhola\Downloads\open_model_zoo\demos\common\cpp\utils\include\utils/ocv_common.hpp(99,98): message : No constructor could take the source type, or constructor
overload resolution was ambiguous 
```

Wovchena:
The compiler is supposed to use https://github.com/openvinotoolkit/openvino/blob/47fe50ca35305d0f24e4e34c2f84d571c4730694/src/core/include/openvino/runtime/allocator.hpp#L173 constructor, no inheritance required. The solution works for many people. Not just for me and CI. Maybe your openvino is outdated? It's supposed to be 2023.0

ngaloppo:
Thanks. This issue was reported to me, I do not have a Windows environment to test this. Indeed, it's possible that they were compiling with an older version of OpenVINO. Either that or their version of MSVC compiler is not supporting this C++ feature.

PR #3866, opened 2023-10-07
demos: add llm

Wovchena:
Ticket 121865

### DIFF
--- a/demos/CMakeLists.txt
+++ b/demos/CMakeLists.txt
@@ -143,6 +143,9 @@ find_package(OpenVINO REQUIRED COMPONENTS Runtime)
 
 add_subdirectory(thirdparty/gflags)
 add_subdirectory(thirdparty/llama.cpp)
+if(COMPILER_IS_GCC_LIKE)  # Disable the warnings from llama.cpp
+    target_compile_options(llama PRIVATE -Wno-implicit-fallthrough -Wno-unused-function)
+endif()
 add_subdirectory(common/cpp)
 
 find_package(OpenCV QUIET COMPONENTS gapi)
--- a/demos/CMakeLists.txt
+++ b/demos/CMakeLists.txt
@@ -145,6 +145,7 @@ add_subdirectory(thirdparty/gflags)
 add_subdirectory(thirdparty/llama.cpp)
 if(COMPILER_IS_GCC_LIKE)  # Disable the warnings from llama.cpp
     target_compile_options(llama PRIVATE -Wno-implicit-fallthrough -Wno-unused-function)
+    target_compile_options(common PRIVATE -Wno-unused-function)
 endif()
 add_subdirectory(common/cpp)
 
--- a/demos/CMakeLists.txt
+++ b/demos/CMakeLists.txt
@@ -142,6 +142,7 @@ find_package(OpenCV REQUIRED COMPONENTS core highgui videoio imgproc imgcodecs)
 find_package(OpenVINO REQUIRED COMPONENTS Runtime)
 
 add_subdirectory(thirdparty/gflags)
+add_subdirectory(thirdparty/llama.cpp)
 add_subdirectory(common/cpp)
 
 find_package(OpenCV QUIET COMPONENTS gapi)
--- a/demos/CMakeLists.txt
+++ b/demos/CMakeLists.txt
@@ -146,6 +146,9 @@ add_subdirectory(thirdparty/llama.cpp)
 if(COMPILER_IS_GCC_LIKE)  # Disable the warnings from llama.cpp
     target_compile_options(llama PRIVATE -Wno-implicit-fallthrough -Wno-unused-function)
     target_compile_options(common PRIVATE -Wno-unused-function)
+elseif(MSVC)
+    target_compile_options(llama PRIVATE /wd4146 /wd4305 /wd4715)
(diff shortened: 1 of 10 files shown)


demos/llm_demo/cpp/README.md
```diff
@@ -0,0 +1,41 @@
+# LLM C++ Demo
+
+This application showcases inference of s large language model (LLM). Unlike most of other demos, this application doesn't have a rich set of command line arguments to encourage the reader to explore and modify the source code.
+
+## How it works
+
+The demo loads a model (`.xml` and `.bin`) to OpenVINO™ and a provided vocab (`.gguf`) to use for tokinezation. The model is reshaped to batch 1 and variable prompt size. A prompt is tokenized and passed to the model. The model greedily generates token by token until the special end of sentence token is obtained. The predicted tokens are converted to chars and printed in a streaming fashion.
```
yury-gorbachev, line 7: *tokenization*

demos/llm_demo/cpp/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+   ```sh
+   git lfs install
+   python -m pip install --extra-index-url https://download.pytorch.org/whl/cpu onnx git+https://github.com/huggingface/optimum-intel.git
+   ```
+
+2. Download and convert the model
+
+   ```sh
+   git clone https://huggingface.co/openlm-research/open_llama_3b_v2
```
yury-gorbachev, line 27: Would it be easier to use optimum-cli instead? we recommend this now I think.

demos/llm_demo/cpp/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+   `python demos/thirdparty/llama.cpp/convert.py open_llama_3b_v2/ --vocab-only --outfile open_llama_3b_v2/vocab.gguf`
+
+## Running
+
+Usage: `llm_demo <model_path> <vocab_path> "<prompt>"`
+
+Example: `llm_demo openvino_model.xml vocab.gguf "Why is the Sun yellow?"`
+
+Follow https://github.com/ggerganov/llama.cpp/discussions/366#discussioncomment-5384744 to enable non ASCII characters for Windows cmd.
```
yury-gorbachev, line 41: I am not very excited that we are sending people outside to solve problems, entire purpose of this sample is to show how to do things e2e. Entire fix is not very complicated to describe.

yury-gorbachev:
@slyalin , @apaniukov , @ilya-lavrenov - see this PR. It is basically dependent on llama.cpp due to tokenization and this is not great to be honest., could we cover that differently with ov-tokenizers perhaps?

yury-gorbachev:
In general, I do not like this sample as it depends on llama.cpp and specific functions like llama_* there. It does not make sense to me - it is very narrow in terms of what it shows and I would argue that it shows more llama.cpp than openvino.

yury-gorbachev (changes requested):

demos/llm_demo/cpp/README.md
Wovchena: Fixed

demos/llm_demo/cpp/README.md
Wovchena: Moved

apaniukov:
> @slyalin , @apaniukov , @ilya-lavrenov - see this PR. It is basically dependent on llama.cpp due to tokenization and this is not great to be honest., could we cover that differently with ov-tokenizers perhaps?

Yes, there is a Python full pipeline example that uses llama tokenizer, which can be used here, after tokenzier conversion: https://github.com/apaniukov/openvino_contrib/tree/tokenizer-fix-decode/modules/custom_operations/user_ie_extensions/tokenizer/python#text-generation-pipeline

```bash
python demos/thirdparty/llama.cpp/convert.py open_llama_3b_v2/ --vocab-only --outfile open_llama_3b_v2/vocab.gguf
```
 We could provide similar `convert_tokenizer.py` cli script, to get tokenizer/detokenizer.xml from HF hub or disk.

Wovchena:
Moved to https://github.com/openvinotoolkit/openvino.genai/pull/1

demos/CMakeLists.txt
```diff
@@ -142,6 +142,14 @@ find_package(OpenCV REQUIRED COMPONENTS core highgui videoio imgproc imgcodecs)
 find_package(OpenVINO REQUIRED COMPONENTS Runtime)
 
 add_subdirectory(thirdparty/gflags)
+add_subdirectory(thirdparty/llama.cpp)
```
slyalin, line 145: We discussed offline and decided to try https://github.com/openvinotoolkit/openvino_contrib/pull/687 instead of `llama.cpp`. The steps are:
- replace llama.cpp dependency by that tokenizer branch from contrib in the submodules and cmake
- then use a short snippet in Python to convert the tokenizer in HF to OpenVINO IR using methods implemented in that branch
- replace the code of tokenization/detokenization in the sample by OV inference of tokenization/detokenization parts produced in the previous step.

@apaniukov, please comment on the functionality of the tokenizers: is it enough to be used here?
@Wovchena, could you cooperate with @apaniukov to understand how to connect and use that branch with tokenizers?

apaniukov:
@slyalin yes, the functionality is enough, already give instruction here: https://github.com/openvinotoolkit/openvino.genai/pull/2#issuecomment-1764890516

PR #3868, opened 2023-10-09
Add support for IFRNet

pwolnows:
To be validated with related change for accuracy checker config


prabin693 (approved):

pwolnows:
Validation passed with https://ci-accuracy-icv.iotg.sclab.intel.com/job/OMZ-Validation/job/try/job/ubuntu20/22650/
OVVP results http://ovvp.intel.com/acc/snapshot?md5id=2394d1214cac05be95ce18ee2f15ef60

skuros (approved):

tools/accuracy_checker/openvino/tools/accuracy_checker/annotation_converters/vimeo90k.py
```diff
@@ -50,3 +50,30 @@ def convert(self, check_content=False, progress_callback=None, progress_interval
             annotations.append(SuperResolutionAnnotation(
                 MultiFramesInputIdentifier(list(range(len(input_data))), input_data), target))
         return ConverterReturn(annotations, None, None)
+
+class Vimeo90K_IFRNetDataset(BaseFormatConverter):
```
eaidova, line 54: I do not think that it is good idea to name dataset by model name as it does not contains any model specific logic. can it be named in more generic way? (potentially it can be reused for other models in future)

tools/accuracy_checker/openvino/tools/accuracy_checker/metrics/image_quality_assessment.py
```diff
@@ -478,3 +478,30 @@ def _create_loss(self):
                 sl._modules[module_id] = feats[int(module_id)] # pylint: disable=W0212
             setattr(loss.net, 'slice{}'.format(slice_id), sl)
         return loss
+
+
+class PeakSignalToNoiseRatioIFRNet(BaseRegressionMetric):
```
eaidova, line 483: The same comment for naming. What the difference with original psnr metric?
Can it be just configured for getting desired behaviour by parameters instead of introducing new metric?

tools/accuracy_checker/openvino/tools/accuracy_checker/annotation_converters/vimeo90k.py
pwolnows: Name has been changed to Vimeo90KIntermediateFrameInterpolationDataset

tools/accuracy_checker/openvino/tools/accuracy_checker/metrics/image_quality_assessment.py
pwolnows: Reused PSRN with added 'unweighted_average' parameter to calculate psnr as for grayscale image.

PR #3883, opened 2023-12-19
Correct name of yolo-v3-tiny-tf model in accuracy-check.yml

pwolnows:


openmodelzooci:

PR #3884, opened 2023-12-20
Enable use of local copy of model weights

pwolnows:
Enable use of local copy of model weights to avoid SSL certificate issues on Windows builds


openmodelzooci:

PR #3902, opened 2024-01-26
Correct selection of metrics with sub_evaluation disabled

pwolnows:
Added fix and tests to check correct metrics selection.


openmodelzooci:

eaidova (approved):

PR #3904, opened 2024-01-29
Remove all deprecated models

Wovchena:


data/dataset_classes/imagenet_2015.txt
```diff
@@ -1,1000 +0,0 @@
-n02119789 kit fox
```
eaidova, line 1: I think it can be stayed for opportunity to run somebody converted models outside of omz

eaidova (approved):

data/dataset_classes/imagenet_2015.txt
Wovchena: Reverted

PR #3916, opened 2024-02-16
Adjusts to new namespaces

Wovchena:
Ticket 110661


tools/model_tools/src/omz_tools/omz_converter.py
```diff
@@ -20,7 +20,7 @@
 import string
 import sys
 
-from openvino.runtime import Core, serialize
+from openvino import Core, serialize
```
eaidova, line 23: I think it is also will be good to use save_model instead of serialize as it is deprecated

eaidova (approved):

andrei-kochin (approved):

PR #3921, opened 2024-02-27
Support URLNet in accuracy checker

pwolnows:
The current implementation supports URLNet with 2 inputs. 
There is no tokenization of input_1 data, instead input_1 is fed with tokenized data from input_file defined in URLNet dataset.
Tokenization of input_1 requires TADK Tokenizer that is not supported on all validation platforms.


sstrehlk (approved):

tools/accuracy_checker/accuracy_checker/metrics/classification.py
```diff
@@ -671,3 +671,56 @@ def reset(self):
     @classmethod
     def get_common_meta(cls):
         return {'target': 'higher-worse'}
+
+
+class URLNetAccuracy(Metric):
+    __provider__ = 'urlnet_score'
```
eaidova, line 677: could you please provide more clear metric name, this approach is not related to specific model

pwolnows: renamed to class to UrlCassificationScore and __provider__ to url_classification_score

eaidova (approved):

PR #3936, opened 2024-04-05
fix setting config for async pipeline

eaidova:
CVS-136393

PR #3951, opened 2024-04-24
Fix plugin bug in object detection demo

junwenwu:
In the Python demo, the argument -device can only take AUTO because of the previous bug inside of the async_pipeline.py script. Updated the script so it can handle the device configuration correctly.


openmodelzooci:

Wovchena:
Wasn't it fixed by https://github.com/openvinotoolkit/open_model_zoo/pull/3936?

junwenwu:
#3936 introduced a bug. config is initialized as an empty dict and then 'device' is used as a key to reference the empty dictionary and causes error. This is trying to fix this bug.

Wovchena (approved):
Thank you

PR #3959, opened 2024-06-20
Update kenlm in accuracy setup.py

pwolnows:
Update the kenlm package revision to resolve errors during building wheels for kenlm on python 3.10


openmodelzooci:

eaidova (approved):

PR #3974, opened 2024-09-10
Exclude a specific tensorflow version which does not support Python 3.8 on MacOS

p-wysocki:
Details in CVS-151865

Related PR: https://github.com/openvinotoolkit/openvino/pull/26507


p-wysocki:
@eaidova could you please take a look?

PR #3975, opened 2024-09-16
Fix kaldi subset creation convert.py

pwolnows:


openmodelzooci:

PR #3985, opened 2024-10-25
Add custom evaluator for LDM Super Resolution

pwolnows:
Refactored Stable Diffusion custom evaluator to reuse some code from base classes

### DIFF
--- a/tools/accuracy_checker/accuracy_checker/evaluators/custom_evaluators/ldm_super_resolution_evaluator.py
+++ b/tools/accuracy_checker/accuracy_checker/evaluators/custom_evaluators/ldm_super_resolution_evaluator.py
@@ -1,0 +1,265 @@
+"""
+Copyright (c) 2024 Intel Corporation
+
+Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


tools/accuracy_checker/accuracy_checker/evaluators/custom_evaluators/ldm_super_resolution_evaluator.py
```diff
@@ -0,0 +1,265 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+
+import inspect
+from typing import Union, List, Optional
+import numpy as np
+import PIL
+from diffusers.utils import torch_utils
```
eaidova, line 21: that is probably need to be added under try-except with diffusers package import as diffusers may be not installed yet

pwolnows: Fixed

sstrehlk (approved):

PR #3997, opened 2025-02-11
Set weights_only to false for pytorch convert

mkruszyx:
Fix for failing cpp_gapi-demos workflow with PyTorch 2.6

> Cannot use ``weights_only=True`` with files saved in the legacy .tar format. In PyTorch 2.6, we changed the default value of the `weights_only` argument in `torch.load` from `False` to `True`. Re-running `torch.load` with `weights_only` set to `False` will likely succeed, but it can result in arbitrary code execution. Do it only if you got the file from a trusted source.
> 
> FAILED:
> resnet-18-pytorch
> regnetx-3.2gf
> resnet-50-pytorch
> resnet-34-pytorch

### DIFF
--- a/models/public/regnetx-3.2gf/model.py
+++ b/models/public/regnetx-3.2gf/model.py
@@ -12,11 +12,18 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import pycls.core.checkpoint
+import torch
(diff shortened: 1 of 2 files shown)


Wovchena:
Jenkins ok to test

PR #4010, opened 2025-04-25
Support native pytroch models with transformers class

pwolnows:


Wovchena:
Jenkins Accuracy Checker failed with
> accuracy_checker/launcher/launcher.py:266: [R0201(no-self-use), Launcher.fit_to_input] Method could be a function

PR #4039, opened 2026-03-31
Changed "face_recognition_demo" - fixing the device argument to accept AUTO: GPU,CPU style inputs for flexibility

InfoSage05:
## Summary

  - I have removed `choices=DEVICE_KINDS` restriction from `-d_fd`, `-d_lm`, and `-d_reid`
    arguments in `face_recognition_demo.py`
  - The old code only accepted bare device names (`CPU`, `GPU`, `AUTO`, `HETERO`) and
    rejected valid OpenVINO compound strings like `AUTO:GPU,CPU` or `HETERO:GPU,CPU`
    with an argparse error so I have added them in the argument. 
  - I have also updated help text for all three arguments to document the correct compound format

  ## Root Cause

  OpenVINO supports compound device strings as a standard way to configure
  the AUTO and HETERO plugins. For example:

  - `AUTO:GPU,CPU` — automatically selects the best device, and it prefers GPU
    and falls back to CPU
  - `HETERO:GPU,CPU` — splits model layers across GPU and CPU

  These strings are not standalone device names and they carry a priority list
  after the colon. However, the demo used Python's `argparse` with
  `choices=DEVICE_KINDS` where:

 ```python
  DEVICE_KINDS = ['CPU', 'GPU', 'NPU', 'AUTO', 'HETERO'] 

  argparse choices= performs exact string matching against the list.
  So when a user passed -d_fd AUTO:GPU,CPU, argparse compared AUTO:GPU,CPU
  against each entry in the list, found no exact match, and immediately raised
  an error:

  error: argument -d_fd: invalid choice: 'AUTO:GPU,CPU'
  (choose from CPU, GPU, NPU, AUTO, HETERO)

  This means the AUTO plugin could never be used to its full potential from
  the command line  and users were silently limited to just the AUTO feature with no device
  priority control.

  Solution

  Removed the choices=DEVICE_KINDS restriction from the three device
  arguments -d_fd, -d_lm, and -d_reid. The arguments now accept any
  string, which allows all valid OpenVINO device specifiers including compound
  ones. The help text for each argument was also updated to explicitly document
  the AUTO:<dev1>,<dev2> and HETERO:<dev1>,<dev2> formats so users know
  what is supported.  
```

## How to Test

  ```bash
  python face_recognition_demo.py -i <input> -m_fd <model> -m_lm <model> \
    -m_reid <model> -fg <gallery> -d_fd AUTO:GPU,CPU

### DIFF
--- a/demos/face_recognition_demo/python/face_recognition_demo.py
+++ b/demos/face_recognition_demo/python/face_recognition_demo.py
@@ -210,6 +210,10 @@ def center_crop(frame, crop_size):
 def main():
     args = build_argparser().parse_args()
 
+    for arg_name, device in [('-d_fd', args.d_fd), ('-d_lm', args.d_lm), ('-d_reid', args.d_reid)]:
+        if device == 'HETERO':
+            raise ValueError('{} requires sub-devices, e.g. HETERO:GPU,CPU'.format(arg_name))
+
     cap = open_images_capture(args.input, args.loop)
     frame_processor = FrameProcessor(args)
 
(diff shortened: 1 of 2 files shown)


workflow-lab:

PR #4041, opened 2026-03-31
face_detector: detect NCHW layout after model compilation

InfoSage05:
## Summary 
I have moved NCHW layout detection from model construction to after compilation, ensuring the layout matches the actual compiled model selected by AUTO plugin.                 
                                                                                                                                                                                                                                                                    

## Root Cause             
The `nchw_layout` flag was detected in `__init__` using the uncompiled model's input shape. When AUTO plugin is used, the device is not yet known at that point. Different devices (CPU vs GPU) may have different input layout requirements.                                                  After `deploy()` compiles with the chosen device, the actual layout might differ, which is causing incorrect preprocessing.                                                                                                                                                                                                                                                                                                           
  
## Solution
Overriding the `deploy()` function in the `FaceDetector` file to detect `nchw_layout` from the `compiled_model.inputs[0].shape` AFTER `compile_model()` completes. This ensures the layout detection uses the device-specific shape and not the defined one.

  ## How to Test
  ```bash
  # Should work correctly with AUTO on systems with mixed CPU/GPU
  python face_recognition_demo.py -d_fd AUTO ...

### DIFF
--- a/demos/face_recognition_demo/python/face_detector.py
+++ b/demos/face_recognition_demo/python/face_detector.py
@@ -14,12 +14,13 @@
  limitations under the License.
 """
 
+import logging as log
 import numpy as np
 
 from ie_module import Module
 from utils import resize_input
 
-from openvino import PartialShape
+from openvino import PartialShape, AsyncInferQueue
 
 
 class FaceDetector(Module):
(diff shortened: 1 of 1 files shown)


workflow-lab:

